PR #12, opened 2023-01-20
Add --disk-cache-dir and --max-cpu-mem parameters

Silver267:
--disk-cache-dir allows user to specify their cache directory while --max-cpu-mem allows user to specify the amount of RAM they want to use for offloading.


Silver267:
In windows, using ram as high as possible may cause swap space to be used. As for the cache directory, if the user have a dedicated acceleration disk for ai, it.would be useful to have the option to redirect.

Silver267:
Just figured out how to properly commit to github, reopening a new one.

PR #16, opened 2023-01-20
Allow specifying the Hugging Face Git branch when downloading models

81300:
This adds a `--branch` option to the model download script.


oobabooga:
Very nice, thank you. There is just one issue, it doesn't work if the parameters are reordered:

```
python download-model.py --branch dev PygmalionAI/pygmalion-6b
Downloading the model to models/--branch_dev...

```

Issue #14, opened 2023-01-20
Doesn't run on colab with Pygmalion-6B / results look different on Colab


waifusd:
Using the provided notebook and just changing the model to Pygmalion-6B instead of Pygmalion-1.3B generates the following tcmalloc error and the execution stops cold.

![tcmalloc](https://user-images.githubusercontent.com/123158074/213659678-ecdfaf31-697c-4c1b-9c7c-90c9ff12689a.png)

BushyToaster88:
> Using the provided notebook and just changing the model to Pygmalion-6B instead of Pygmalion-1.3B generates the following tcmalloc error and the execution stops cold.
> 
> ![tcmalloc](https://user-images.githubusercontent.com/123158074/213659678-ecdfaf31-697c-4c1b-9c7c-90c9ff12689a.png)

I have the same issue when trying to run the 4chan model in colab. Im get a "Memory cgroup out of memory" error when I look at the dmesg in the colab terminal.

oobabooga:
I can confirm this issue. 

On colab, I can't load either pygmalion-2.7b or pygmalion-6b.

The free colab has around 13GB RAM, while pygmalion-2.7b takes 6.8GB of RAM to load on my system (peak allocation). So it should in principle work.

waifusd:
Pygmalion-6B should be able to loaded on colab too, since the colab notebooks of other projects are able to load it (namely the Pyg devs own notebook, and KoboldAI's).

oobabooga:
An Anonymous 4chan user has kindly provided this notebook that allows the 6b model to be loaded in 8bit mode:

https://colab.research.google.com/github/81300/AI-Notebooks/blob/main/Colab-TextGen-GPU.ipynb

I haven't tested it yet because Google is not giving me a free instance with a GPU.

waifusd:
I tried it:

![pygoob1](https://user-images.githubusercontent.com/123158074/213779900-af9f4ab8-ef97-4f2a-813e-62805cc9b607.png)

oobabooga:
I can confirm that the results look worse on colab than running locally. I made a comparison using `debug.txt`, which is a deterministic preset that should generate the same responses:

```
do_sample=False,
max_new_tokens=tokens,

```

But the responses are different:

| Colab | Local |
| ------------- | ------------- |
[...]

PR #24, opened 2023-01-24
Change `--auto-devices` and model offload behavior

### LINKED ISSUE #23: `device_map='auto'` is inserted by default
Presently if you pass any other model loading arguments except for `--cpu`, the device_map gets appended even if `--auto-devices` is not used.

81300:
Fixes #23 

This makes `device_map='auto'` optional by ensuring it is only used as a setting for `AutoModelForCausalLM.from_pretrained()` in those cases:
* As before, if the model's name starts with `gpt-neo`, `opt-` or `galactica` and has 13B/20B/30B parameters, or
* The `--auto-devices` argument is passed, or
* The `--load-in-8bit` argument is passed (launch would fail only with `load_in_8bit=True`, see below)

```python
  File "server.py", line 301, in <module>
    model, tokenizer = load_model(model_name)
  File "server.py", line 169, in load_model
    model = eval(command)
  File "/usr/local/lib/python3.8/dist-packages/transformers/models/auto/auto_factory.py", line 463, in from_pretrained
    return model_class.from_pretrained(
  File "/usr/local/lib/python3.8/dist-packages/transformers/modeling_utils.py", line 1989, in from_pretrained
    raise ValueError(
ValueError: A device map needs to be passed to run convert models into mixed-int8 format. Please run`.from_pretrained` with `device_map='auto'`
```

It also ensures that unless `--cpu` or `--auto-devices` is passed, the model will always moved to the GPU first. This is to avoid scenarios such as:

```python
/home/81300/miniconda3/envs/textgen/lib/python3.10/site-packages/transformers/generation/utils.py:1470: UserWarning: You are calling .generate() with the `input_ids` being on a device type different than your model's device. `input_ids` is on cuda, whereas the model is on cpu. You may experience unexpected behaviors or slower generation. Please make sure that you have put `input_ids` to the correct device by calling for example input_ids = input_ids.to('cpu') before running `.generate()`.
  warnings.warn(
(...)
  File "/home/81300/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/functional.py", line 2210, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument index in method wrapper__index_select)
```

May also impact #14


oobabooga:
@81300, thanks for the PR. Can you help me understand it? This is what I was thinking with the original implementation:

```
            settings.append("device_map='auto'")
            if args.gpu_memory is not None:
                if args.cpu_memory is not None:
                    settings.append(f"max_memory={{0: '{args.gpu_memory}GiB', 'cpu': '{args.cpu_memory}GiB'}}")
                else:
                    settings.append(f"max_memory={{0: '{args.gpu_memory}GiB', 'cpu': '99GiB'}}")
            if args.disk:
                if args.disk_cache_dir is not None:
                    settings.append(f"offload_folder='{args.disk_cache_dir}'")
                else:
                    settings.append("offload_folder='cache'")
            if args.load_in_8bit:
                settings.append("load_in_8bit=True")
            else:
                settings.append("torch_dtype=torch.float16")
```

At this point, the user doesn't want to load the entire model into the GPU or the CPU. He either wants to use the model in 8-bit mode (in which case things remain the same in your PR), or he wants to offload layers to the GPU and/or disk because the model doesn't fit entirely into his GPU. In the latter case, it seemed to me that `device_map='auto'` was always necessary.

What is gained by passing a `max_memory` dict and/or an `offload_folder` *without* `device_map=auto`?

About `offload_state_dict=True`, [it defaults to true](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.offload_state_dict) when there is a disk offload, so it seems unnecessary to include it.

Also, what is gained by adding `.cuda()` [here](https://github.com/oobabooga/text-generation-webui/pull/24/files#diff-791d4d41d3718d15d49180f3aacc8370b8cab07383f0d35b2713651cc0adfe46R117)? I thought that `.cuda()` was only relevant when the model was sent entirely to the GPU, and that it would generate an error if added when the model is loaded in 8-bit mode.

oobabooga:
I think that this boils down to a matter of semantics. With `gpu-memory` and `cpu-memory`, the goal is to limit the amount of memory allocated by `auto-devices`, allowing the user to get the model working by following the [low VRAM guide](https://github.com/oobabooga/text-generation-webui/wiki/Low-VRAM-guide) in a simple way.

If you set the `max_memory` dict with a maximum GPU memory but no `auto devices`, there are two options:

1. This is greater than the model's size and the model will load to the GPU as if no command-line flag was used.
2. This is smaller than the model's size and the program will segfault because it wasn't instructed to offload weights to the CPU by `auto-devices`.

I think that features should be constructed with common use cases in mind rather than supporting every possible combination of parameters in the hugging face's pipelines. Right now, changing the behavior of `auto-devices` would not support a new use case it seems.

oobabooga:
I think not, as this would also not support a new use case.

In the current implementation, the only use case that I can think of that is not supported is running in CPU mode with layers offloaded to the disk. But I am not sure if anyone is interested in that (it would be painfully slow).

If you can think of other relevant use cases that are not currently possible, please let me know.

PR #27, opened 2023-01-26
bump transformers

Silver267:
idk but there seems to be a lot of random fixes in the new version, and it's working according to my local tests.


oobabooga:
#yolo

PR #38, opened 2023-01-31
Fix an error

Silver267:
Fixes "UnboundLocalError: local variable 'substring_found' referenced before assignment" when loading non-pygmalion models in cai chat mode.


oobabooga:
Thank you! I missed that.

PR #43, opened 2023-02-01
Add DeepSpeed ZeRO-3 integration

81300:


oobabooga:
Thank you, that's amazing and very impressive! 

How faster is generation with deepseed than without it?

Silver267:
Is it possible to add 8bit quantization for deepspeed too?
[Microsoft's blog claimed that deepspeed supports 8bit quantization](https://www.microsoft.com/en-us/research/blog/deepspeed-accelerating-large-scale-model-inference-and-training-via-system-optimizations-and-compression/)

oobabooga:
It seems to be working. Merged.

PR #57, opened 2023-02-05
Fix a regex issue in `tokenize_dialogue`.

martinnj:
The existing regex would fail if using character names that start with numbers.

The issue is the formatting of the backreference using a "naked" number, meaning that any numerical value in the name will be part of the backreference.

Using the alternative format of `\\g<1>` as described [in the docs](https://docs.python.org/3/library/re.html#re.sub) solves this issue. The docs actually has this exact case as an example in the last paragraph above the version discrepancies.

### DIFF
--- a/server.py
+++ b/server.py
@@ -532,7 +532,7 @@ if args.chat or args.cai_chat:
         dialogue = re.sub('<START>', '', dialogue)
         dialogue = re.sub('<start>', '', dialogue)
         dialogue = re.sub('(\n|^)[Aa]non:', '\\1You:', dialogue)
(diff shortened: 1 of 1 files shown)


oobabooga:
In this case, what is being replaced is the literal string `[CHARACTER]`, so it won't make a difference. But it's good practice to use `\g<1>` instead of `\1`.

martinnj:
> In this case, what is being replaced is the literal string `[CHARACTER]`, so it won't make a difference. But it's good practice to use `\g<1>` instead of `\1`.

It causes a crash for characters with numbers in the beginning of names all the same since the regex is still being evaluated. :)

oobabooga:
Aha, I see. Thanks for this!!

PR #66, opened 2023-02-09
Extend bfloat16 support

81300:
Makes `--bf16` usable without DeepSpeed. For parametrization consistency.


oobabooga:
That's a great idea, thank you for this.

PR #67, opened 2023-02-10
moving argparse to a separate module

rmn-ykimov:
- Argparse was moved to a separate module.
- Imports order in `download-model.py` was changed according to PEP8
- `.gitignore` was added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+__pycache__
+models/*
+!models/place-your-models-here.txt
--- a/argparse_module.py
+++ b/argparse_module.py
@@ -1,0 +1,26 @@
+"""
+Module containing the argument parser for the script.
+
+The purpose of this module is to provide a convenient and centralized way to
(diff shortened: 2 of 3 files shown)


oobabooga:
Your PR moves the argparse parameters from the download script instead of the main script.

I dislike the idea of moving the parameters to a separate file, as this makes it harder to understand the script. Also, adding .gitignore is kind of useful but it's one more file in the repository.

Changing the imports to adhere to pep8 style is a good idea. They were a mess. I have done it in the latest commit using https://pypi.org/project/reorder-python-imports/

rmn-ykimov:
As far as I understand, you want to keep your repository as clean as possible. Can you describe your vision of project structure? What is acceptable and what is not?

oobabooga:
So far I have tried to keep everything to a single file, except very secondary and unrelated things, which I moved to separated files under modules/.

My vision is a clean and didactic code base that is as easy as possible to understand and modify, and that relies on existing, actively maintained libraries for the UI and AI stuff (gradio and transformers namely).

PR #68, opened 2023-02-11
Added ROCm Install instructions to README

Spencer-Dawson:


oobabooga:
Thank you, this is super useful. I had no idea how to get this running on AMD GPUs. This will help a lot of people.

Enferlain:
Yo just tried setting it up on ubuntu with the amd installer, everything installs fine except this

> ERROR: Could not find a version that satisfies the requirement torchaudio (from versions: none)
> ERROR: No matching distribution found for torchaudio

https://download.pytorch.org/whl/torchaudio/ but it seems like there is a rocm5.2 so I don't get why it errors? Any idea?

Spencer-Dawson:
Interesting. maybe you're using the wrong python environment?  Try installing manually. Your python install might be defaulting to python2 or something odd like that. If you look at my commit to the readme I explicitly used pip3 to install. 

```
conda create -n textgen
conda activate textgen
pip3 install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/rocm5.2
git clone https://github.com/oobabooga/text-generation-webui
cd text-generation-webui
pip install -r requirements.txt
```

I might personally update the linux install script later. Not so much because of your comment, but because I've noticed a separate broken dependency when using ROCm. The --load-in-8bit flag uses bitsandbytes which isn't ROCm compatible.

Enferlain:
Yeah saw that, tried replacing it in the sh with the line from the readme but it still says no matching distribution for torchaudio

Screenshot from 2023-02-25 21-02-14

Screenshot from 2023-02-25 21-07-10

I'll just download them from the link and see if that helps and if not I'll do manual, thanks

Enferlain:
That didn't work either, it falls back on CPU regardless, and also saw 1 more person get the same error. No clue why it does that. I'll try it one more time later with stuff I read from one of your comments in a different thread

jcfalcone:
> That didn't work either, it falls back on CPU regardless, and also saw 1 more person get the same error. No clue why it does that. I'll try it one more time later with stuff I read from one of your comments in a different thread

I had the same problem running on Ubuntu 23 and I had to downgrade to Ubuntu 22 to make it work. I also installed ROCm using this article:
https://huggingface.co/blog/chatbot-amd-gpu

and Remove this comments to make it load and work:
https://github.com/oobabooga/text-generation-webui/blob/c2ae01fb0431cf33bd5a609437ac0ef2d92bada2/one_click.py#L12

PR #76, opened 2023-02-14
Use BLIP to send a picture to model

SillyLossy:
### How it works?
1. Load a BLIP model for captioning: https://huggingface.co/spaces/Salesforce/BLIP (I used this particular pip: https://pypi.org/project/nataili-blip/).
2. Feed it an image uploaded by the user and BLIP captions it.
3. UI takes the caption and feeds it to the bot like this: `*You send Bot the picture that contains: "Caption returned by BLIP model"*`
4. And then the bot reacts to whatever was in the captioned image

### Disclaimer
I'm not really good at Python and Gradio (oh how I hate it) in particular, so my code is not perfect for sure. But it works as a POC, at least...

P.S. yield None part in wrappers is to clear an Image control after selecting a picture
P.P.S. For some peculiar reason upload event handler can't be canceled

### DIFF
--- a/modules/bot_picture.py
+++ b/modules/bot_picture.py
@@ -1,0 +1,9 @@
+from nataili_blip.model_manager import BlipModelManager
+from nataili_blip.caption import Caption
+
+def load_model():
+    model_name = "BLIP"
+    mm = BlipModelManager()
+    mm.download_model(model_name)
+    mm.load_blip(model_name)
+    return Caption(mm.loaded_models[model_name]["model"], mm.loaded_models[model_name]["device"])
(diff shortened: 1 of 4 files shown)


oobabooga:
This is extremely cool and very creative, thank you!

There are two issues that we need to address before I can merge:

1. `nataili_blip` depends on `transformers==4.21.3`, while this project depends on the dev version of transformers (the softprompt implementation was made possible by a recent commit), which is version `4.27.0.dev0`. This causes a conflict error when I try to `pip install -r requirements.txt`.
2. When the script is called without the `--picture` flag, I get

```
NameError: name 'picture_select' is not defined
```

oobabooga:
I got this, I have found a way to do it by using BLIP directly without additional dependencies. Will commit soon.

PR #90, opened 2023-02-19
Escape regexp in message extraction

SillyLossy:
## The problem
Users were unable to send messages if the character's name contained parentheses. [Example of 'broken' character card](https://media.discordapp.net/attachments/1076630264433872896/1076630345069379604/Fem-Death_Muerte.card.png)

The following exception was present in console logs:

image

## The solution
Escape the value inserted into the regexp used for extracting a reply. Now I'm able to get a reply.

image

### DIFF
--- a/server.py
+++ b/server.py
@@ -559,8 +559,8 @@ def extract_message_from_reply(question, reply, current, other, check, extension
     next_character_found = False
     substring_found = False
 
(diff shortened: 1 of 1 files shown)


oobabooga:
I had seen many reports of these "list index out of range" errors but could never find the cause. Thank you so much for this.

PR #113, opened 2023-02-23
gpu-memory arg change

elwolf6:
allows for easy use on machines with multiple GPUs


oobabooga:
Thanks! I only made one change, which was to specify the parameters separated by spaces instead of commas, like this:

--gpu-memory 10 5

PR #135, opened 2023-02-26
fixed convert-to-flexgen.py

elwolf6:
without this change is says model has no attribute model


elwolf6:
nvm didn't read that it only works with opt models

PR #175, opened 2023-03-06
Add api example using websockets

SagsMug:
Adds an API example that uses websockets such that it can both be ran when the server is using streaming and --no-stream
It also uses async

### DIFF
--- a/api-example-stream.py
+++ b/api-example-stream.py
@@ -5,77 +5,77 @@ import json
 import asyncio
 
 def random_hash():
-	letters = string.ascii_lowercase + string.digits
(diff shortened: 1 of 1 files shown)


oobabooga:
Very impressive, I didn't know this was even possible. Thanks for the PR!!

oobabooga:
For some reason github didn't account your contribution properly and didn't put you in the list of contributors to the project, so I have added credits to the source file itself

https://github.com/oobabooga/text-generation-webui/blob/main/api-example-stream.py

CyberTimon:
Hello @SagsMug 
Can you please fix it as it doesn't work in the latest branch? I would love to have the api working

PR #195, opened 2023-03-08
Allow for autoplaying of audio

ChristophHess:
I really wanted the conversation to feel more natural. 

I had to add more functionality to the extension system, though I am unsure if this is the route you want to take, so I am open to suggestions. The reasoning was that this system allows for future additional extension functionality with backwards compatibility.

Furthermore, I also improved the performance of the generation a lot. What I found was that audio was being generated for every iteration when doing live updates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ torch-dumps/*
 *pycache*
 */*pycache*
 */*/pycache*
+venv
 
 settings.json
 img_bot*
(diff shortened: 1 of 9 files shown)


xanthousm:
I've just ran your fork to test the autoplay, and is it just me or do the previous messages made in the current session also autoplay when you click "generate" or press enter? Using "regenerate" doesn't have this problem though. 

Do we need to find a way to remove the autoplay tag on old messages?

oobabooga:
There are lots of changes to the extensions engine in this PR. I would prefer to keep it simple and not change the predefined function definitions (input_modifier, output_modifier, etc).

https://github.com/oobabooga/text-generation-webui/wiki/Extensions

PR #201, opened 2023-03-09
Lets propose a name besides "Gradio"

jtang613:
It's about time Oobabooga gets some window/tab title credit. Definitely beats the generic "Gradio" title.


oobabooga:
Let's settle with "Text generation web UI"

PR #207, opened 2023-03-09
Extension: Whisper Speech-To-Text Input

EliasVincent:
Hey there,
I made an extension where you can use speech-to-text input through OpenAI Whisper. Simply press the Microphone button, talk, preview or edit the result in the Textbox and then click Generate.

grafik

The additional packages are all in the `requirements.txt` for the extension. 

The first time will take a while since it will download the Whisper weights.

Hope this is useful, love your webui!


oobabooga:
Two questions:

1. pyaudio seems to require additional installation steps:

```
pip install PyAudio

(...)
      src/pyaudio/device_api.c:9:10: fatal error: portaudio.h: file not found 
          9 | #include "portaudio.h"
            |          ^~~~~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]

```

Is there an alternative?

2. Is there a reason to install `git+https://github.com/Uberi/speech_recognition.git@010382b` instead of simply `SpeechRecognition`?

EliasVincent:
1: I just checked on Linux it doesn't come with PortAudio installed alongside PortAudio and [has to be installed by the user manually](https://stackoverflow.com/questions/48690984/portaudio-h-no-such-file-or-directory)

2: There's a bug with Whisper recognition on Windows. [It has been fixed](https://github.com/Uberi/speech_recognition/pull/647) but there hasn't been a release with it since.

oobabooga:
An issue that I have found is that it does not seem to work remotely, for instance on Google Colab or by accessing the web UI from another computer in the same local network.

In gradio, it is possible to create an audio widget that takes microphone input:

```python
gr.Audio(source="microphone")
```

I wonder if it's possible to adapt it here.

EliasVincent:
Good tip. I changed it to the Gradio microphone input now, that's much easier.

grafik

RandomInternetPreson:
FYS, I was able to get this to work too!! Wow it really works well at transcribing my voice too, better than google on my phone.  A few things, I had to download ffmpeg separately,  and add the executables to my system path.  It might trip people up on the install, if they can't figure it out.
ffmpeg enviromental paths information

EliasVincent:
Huh, I thought that adding ffmpeg to the requirements was enough to solve ffmpeg, I was wrong 😅. [SpeechRecognition doesn't state ffmpeg as an explicit requirement either](https://pypi.org/project/SpeechRecognition/), [I think Gradio might be the culprit here for the microphone recording?](https://gradio.app/real-time-speech-recognition/#prerequisites). I can reproduce this too now after uninstalling ffmpeg from my system..

PR #235, opened 2023-03-11
--auto-launch and "Is typing..."

### LINKED ISSUE #191: Chat log no longer dims when generating
The chat log area no longer dims when responses are being generated.
Screenshot 2023-03-08 031454

Not sure if this is intended as part of one of the new upgrades.  If so, is there any way yo make this a toggle-able feature, as it provides more of a visual feedback to let me know that the server is doing its thing.

### LINKED ISSUE #209: Suggestion to add: --autolaunch
--autolaunch _open the webui URL in the system's default browser upon launch (is used on AUTOMATIC1111)_

xanthousm:
- Added `--auto-launch` arg to open web UI in the default browser when ready (resolves #209).
- Changed chat.py to display user input immediately and "*Is typing...*" as a temporary reply while generating text. Most noticeable when using `--no-stream` (resolves #191 by adding immersive visual feedback).

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -120,6 +120,9 @@ def chatbot_wrapper(text, max_new_tokens, do_sample, temperature, top_p, typical
     else:
         prompt = custom_generate_chat_prompt(text, max_new_tokens, name1, name2, context, chat_prompt_size)
 
+    #display user input and "*is typing...*" imediately
+    yield shared.history['visible']+[[visible_text, '*Is typing...*']]
+
     # Generate
     reply = ''
     for i in range(chat_generation_attempts):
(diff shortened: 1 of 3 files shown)


oobabooga:
That's a very good idea and thanks for helping out with `--auto-launch`. I have made the following additions:

* Add "Is typing..." to the input field as well when "Impersonate" is clicked
* Remove `show_progress` from the generation events when streaming is off because yielding at the beginning of the `chatbot_wrapper` call causes the progress indicator (in seconds) that used to appear at the top to disappear, so it becomes meaningless to have `show_progress=True`.

We do lose the seconds indicator, but I think that a "Bot is typing" message is more soothing for anxious users anyway. 

* Reload the chat window when "Stop" is clicked so that the "is typing..." partial message is removed in this case.

See if you agree with these changes and if you want to change anything else. After your OK I will merge.

Ph0rk0z:
>We do lose the seconds indicator

Now there is no way to see how long RWKV takes. I see no "output generated in" messages for that model family anymore.

oobabooga:
@Ph0rk0z you should be able to consistently see how long RWKV took now (in the terminal) https://github.com/oobabooga/text-generation-webui/pull/189

PR #255, opened 2023-03-11
Use the pt_path value in load_quant call

jtang613:
I see you go to the effort to identify the quantized pt file, but that value doesn't ever get passed to the load_quant call.


oobabooga:
It turned out that this was fixed in https://github.com/oobabooga/text-generation-webui/pull/224, but well noted :)

PR #258, opened 2023-03-12
Load and save character files and chat history in UTF-8

lxe:
For some reason all the files were getting loaded and saved as cp1252 by default. Changing to UTF allows us to do things like (づ｡◕‿‿◕｡)づ and 🧀 and better multilingual support. 

Screenshot 2023-03-11 212239


oobabooga:
That's a great idea. Do you think that it is worth it to save/load the chat history in utf-8 as well?

lxe:
Yeah let me add that

oobabooga:
Thank you for this!!!

PR #260, opened 2023-03-12
Fix loading of character jsons with non-ascii symbols.

HideLord:
Small patch to fix loading of character jsons. Now it correctly reads non-ascii characters.

image

image


oobabooga:
This is a duplicate of https://github.com/oobabooga/text-generation-webui/pull/258. Your change is already present in `chat.py`!

PR #263, opened 2023-03-12
Fixing compatibility with GPTQ repository

HideLord:
`load_quant` expects a string argument.
image


oobabooga:
`os.path.abspath()` should be a bit more reliable

PR #266, opened 2023-03-12
Adding markdown support and slight refactoring.

HideLord:
CSS strings moved to separate files.

Supports Fenced Code Blocks/lists/numbered lists and CAI-style markdown.
image
image


oobabooga:
I find it too invasive to turn this into default chat behavior. Could it be turned into an extension? For the bot, the conversion to markdown could be done in an `output_modifier` function. For the user input it's more tricky.

HideLord:
> I find it too invasive to turn this into default chat behavior.

I thought it was fine since both CAI and ChatGPT do it by default. Perhaps it can be restricted to only `cai-chat` mode?

oobabooga:
> CAI and ChatGPT do it by default

You have a point. I guess `--cai-chat` could do it by default as well.

Are there any performance considerations? Is the conversion to markdown a significant slow down when the chat window is refreshing 10 times a second during streaming?

HideLord:
Yes, there's no noticeable slowdown. The markdown library is very efficient.

Here is a test with diff text lengths, each purposefully inserting markdown elements length/10 times (extreme case):

Time for 1000 characters: 0.0014704479998908938s
Time for 5000 characters: 0.0029229620000114664s
Time for 10000 characters: 0.016733493999927306s
Time for 50000 characters: 0.27447341500024774s

oobabooga:
Some minor changes:

* Use pathlib instead of os
* Move all the remaining css styles into the `css` folder

Your refactoring idea will greatly help improve the appearance of the web UI in the future. Thank you!

The idea to use markdown for all chat outputs was also great.

I find that code blocks don't work well `--chat` mode, but I will merge anyway and we can solve this later.

PR #275, opened 2023-03-13
Fix: tuple object does not support item assignment

stefanhamburger:
Fixes an error in the `silero_tts` extension.

```
  File "...\text-generation-webui\extensions\silero_tts\script.py", line 82, in input_modifier
    shared.history['visible'][-1][1] = shared.history['visible'][-1][1].replace('controls autoplay>','controls>')
TypeError: 'tuple' object does not support item assignment

  File "...\text-generation-webui\extensions\silero_tts\script.py", line 69, in toggle_text_in_history
    shared.history['visible'][i][1] = f"{visible_reply.split('</audio>')[0]}</audio>\n\n{reply}"
TypeError: 'tuple' object does not support item assignment
```

### DIFF
--- a/extensions/silero_tts/script.py
+++ b/extensions/silero_tts/script.py
@@ -57,7 +57,7 @@ def remove_surrounded_chars(string):
 
 def remove_tts_from_history(name1, name2):
     for i, entry in enumerate(shared.history['internal']):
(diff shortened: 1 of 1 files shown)


oobabooga:
Ty, I missed that =)

PR #294, opened 2023-03-13
Create oasst-sft-1-pythia-12b.txt

Danilo-GB:
Add the preset for the first model finetuned by Open Assistant.

### DIFF
--- a/presets/oasst-sft-1-pythia-12b.txt
+++ b/presets/oasst-sft-1-pythia-12b.txt
@@ -1,0 +1,6 @@
+do_sample=True
+max_new_tokens=400
+temperature=1
+repetition_penalty=1.2
+top_k=50
+top_p=0.95


oobabooga:
This seems like a very generic preset. I don't think it has been through a trial and error process that could make it stand apart from the existing ones.

Danilo-GB:
Those are the parameters used in the model comparator but I agree with you, there's no actual need of add them, the only need to add them would be to have a distinctive preset.

https://open-assistant.github.io/oasst-model-eval/?f=https%3A%2F%2Fraw.githubusercontent.com%2FOpen-Assistant%2Foasst-model-eval%2Fmain%2Fsampling_reports%2Foasst-sft%2F2023-03-09_andreaskoepf_oasst-1_12b_7000_sampling_noprefix_lottery.json

PR #295, opened 2023-03-13
Add support for quantized OPT models and refactor

Zerogoki00:
Added ability to use quantized OPT models
Added argument to specify quantized model type (LLaMA by default)
Removed --load-in-4bit because we already have --gptq-bits

Tested with OPT-30B-Erebus on RTX 4090. Works slower than LLaMA but it works


Ph0rk0z:
How is performance vs flexgen?

modules/quant_loader.py
```diff
@@ -7,28 +7,20 @@
...
+    elif model_type == 'opt':
+        from opt import load_quant
     else:
-        bits = shared.args.gptq_bits
+        print("Unknown pre-quantized model type specified. Only 'llama' and 'opt' are supported")
+        exit()
 
     path_to_model = Path(f'models/{model_name}')
-    pt_model = ''
-    if path_to_model.name.lower().startswith('llama-7b'):
```
oobabooga, line 22: Removing this block breaks compatibility with folder names like `llama-7b-hf` by decapoda.

modules/models.py
```diff
@@ -35,14 +34,15 @@
...
 
+
 def load_model(model_name):
     print(f"Loading {model_name}...")
     t0 = time.time()
 
     shared.is_RWKV = model_name.lower().startswith('rwkv-')
 
     # Default settings
-    if not any([shared.args.cpu, shared.args.load_in_8bit, shared.args.load_in_4bit, shared.args.gptq_bits > 0, shared.args.auto_devices, shared.args.disk, shared.args.gpu_memory is not None, shared.args.cpu_memory is not None, shared.args.deepspeed, shared.args.flexgen, shared.is_RWKV]):
```
oobabooga, line 45: I'm reluctant to remove `--load-in-4bit` because that will certainly cause confusion, but I guess we can do it and move on.

modules/shared.py
```diff
@@ -68,8 +68,8 @@ def str2bool(v):
...
 parser.add_argument('--cpu', action='store_true', help='Use the CPU to generate text.')
 parser.add_argument('--load-in-8bit', action='store_true', help='Load the model with 8-bit precision.')
-parser.add_argument('--load-in-4bit', action='store_true', help='Load the model with 4-bit precision. Currently only works with LLaMA.')
-parser.add_argument('--gptq-bits', type=int, default=0, help='Load a pre-quantized model with specified precision. 2, 3, 4 and 8bit are supported. Currently only works with LLaMA.')
+parser.add_argument('--gptq-bits', type=int, default=0, help='Load a pre-quantized model with specified precision. 2, 3, 4 and 8bit are supported. Currently only works with LLaMA and OPT.')
+parser.add_argument('--gptq-model-type', type=str, default='llama', help='Model type of pre-quantized model. Currently only LLaMa and OPT are supported.')
```
oobabooga, line 72: Maybe we could infer this from the model name?

Zerogoki00: @oobabooga 
Then user will have to always set a special name for model folder with model type prefix (like llama-xxx, opt-xxxxx)

Maybe do as you described, but keep this argument as fallback if folder name has no prefix? What do you think about it?

oobabooga: That sounds like a good idea (making this argument optional).

modules/quant_loader.py
Zerogoki00: I moved the code back

modules/models.py
Zerogoki00: It's up to you but I think that it's a bad practice to keep multiple arguments which do the same thing

modules/shared.py
Zerogoki00: I implemented this. Tested myself with different folder names and it works. (but I recommend you to check too)

Zerogoki00: So here is the logic:
If `--gptq-model-type` isn't present, then try to get model type from name. Print error and exit if failed
If `--gptq-model-type` is set, then don't try to get model name and use its value

modules/models.py
oobabooga: You have convinced me, let's ditch `--load-in-4bit`.

LoopControl:
Could you briefly describe how to convert an OPT huggingface model to .pt (or provide a link to pregenerated .pt)?

LoopControl:
Good news is, after quantizing a 13B Erebus `pt`, the model loads in around 8GB of VRAM and seems to generate text.

Problem is, I'm seeing 5x+ slower generations with very short contexts in 4 bit mode as compared to 13B Llama in 4bit:

```
> python server.py --gptq-bits 4 --gptq-model-type opt --no-stream --model KoboldAI_OPT-13B-Erebus
Output generated in 49.05 seconds (1.06 tokens/s, 52 tokens)
Output generated in 95.38 seconds (0.84 tokens/s, 80 tokens)
```

For comparison, llama 13B:
```
> python server.py --gptq-bits 4 --no-stream
Output generated in 36.11 seconds (5.54 tokens/s, 200 tokens)
Output generated in 42.36 seconds (4.72 tokens/s, 200 tokens)
```

With larger contexts (800+ tokens), llama model works fine but the OPT model seems to just hang (I gave it multiple minutes before killing process).

oobabooga:
@LoopControl can you report that on https://github.com/qwopqwop200/GPTQ-for-LLaMa?

Pinging @qwopqwop200

qwopqwop200:
Benchmarked on opt2.7b, but not as slow as this.

PR #325, opened 2023-03-14
Fix rwkv names

Ph0rk0z:
Fixes the way names appear in the model list for RWKV and removes the 20b tokenizer file that would show up.

Did I do this right?


oobabooga:
Can you load the models from the dropdown menu after this change?

Ph0rk0z:
Yes, I just double checked.

loading

oobabooga:
It seems like this works because the `rwkv` library automatically adds the `.pth` for us. Good finding!

PR #337, opened 2023-03-15
Support Yaml characters

TheTerrasque:
This is just a simple change to support having custom characters written in yaml. 

The format is a superset of json, so have same structure. But it is more suitable for editing by hand.


mcmonkey4eva:
This PR had an unintended side effect - because `yaml.load` is used even for json files, the subset of JSON files that don't form valid YAML files are now erroring. The primary (and, well, only case I'm aware of but I think there's potentially more) case of this is JSON files that contain tab characters, which YAML will produce an error message upon loading.

This is easily resolved by editing the file - but something to be aware of, and potentially worth patching to make JSON files use the JSON loader to prevent?

(For the record of anyone that comes here and sees this message: literally just remove the tabs from the file and you're good)

The error message for reference and so it shows up in searches:
```
    raise ScannerError("while scanning for the next token", None,
yaml.scanner.ScannerError: while scanning for the next token
found character '\t' that cannot start any token
  in "<unicode string>", line 3, column 1:
  ```

PR #345, opened 2023-03-15
Guide for Windows Subsystem for Linux

jfryton:
I've included a guide for WSL in case it's helpful. Includes general step-by-step walkthrough and a command at the bottom to fix the port forwarding that prevents the Webui from showing up on local network devices when using WSL. 

First time using github - so apologies if I'm doing something wrong!


oobabooga:
Thanks, there is a strong demand for this. 

I will accept the PR so that your contribution gets accounted by GitHub, then I will delete the file and move it into the wiki.

PR #342, opened 2023-03-15
Extension: KoboldAI api

mayaeary:
Extension for some minimal set of KoboldAI-like api. 

Intended to be used with tools like TavernAI frontend.

PR #348, opened 2023-03-15
flask_cloudflared for shared tunnels

mayaeary:
Response to request https://github.com/oobabooga/text-generation-webui/pull/342#issuecomment-1470917370

Now it uses cloudflare tunnels when `--share` flag is set. 
I'm not tested it in colab environment, but on my local machine it works.

You need to `pip install flask_cloudflared` to use this feature. 
To run locally it's not necessary

### DIFF
--- a/extensions/api/requirements.txt
+++ b/extensions/api/requirements.txt
@@ -1,0 +1,1 @@
+flask_cloudflared==0.0.12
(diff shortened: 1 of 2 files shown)


oobabooga:
I have tested it using this notebook and it worked: https://colab.research.google.com/github/oobabooga/AI-Notebooks/blob/main/API-notebook.ipynb

I just copied and pasted the cloudfare link that was generated verbatim, without adding `:5000` or anything.

Issue #193, opened 2023-03-08
Is split offloading supported for 8bit mode?


RandomInternetPreson:
It would be really great to run the LLaMA 30B model in 8bit mode, but right now I can't get the memory to split between gpu and CPU using 8bit mode.

I feel like if this were possible it would be the revolutionary!

sgsdxzy:
It is possible, but requires some hack:
1. instead of pass `load_in_8bit=True` directly, construct a `BitsAndBytesConfig(load_in_8bit=True, llm_int8_enable_fp32_cpu_offload=True)` and pass as quantization_config to from_pretained. This is relative easy.
2. https://github.com/huggingface/transformers/issues/22018 of support offloading in auto device mapping is rejected in transformers. So we have to manually build our device map first. I'd rather monkey-patch from_pretained in transformers to just accept auto mapping...

oobabooga:
That's new to me. @sgsdxzy, can you copy and paste the exact modifications that you had to make to get this working for clarity?

oobabooga:
Related to #190

ye7iaserag:
If this solution can be integrated here... the possibilities!

sgsdxzy:
@oobabooga I was originally preparing a pr for you... now that the transformers part is rejected thing get a bit difficult.
The fastest way to get this to work:

1. edit text-generation-webui/modules/models.py line 102 from `params.append("load_in_8bit=True"...` to `params.append("quantization_config=BitsAndBytesConfig(load_in_8bit=True, llm_int8_enable_fp32_cpu_offload=True)")` (or you can just use `params.extend(["load_in_8bit=True", "llm_int8_enable_fp32_cpu_offload=True"]`)
2. edit python's site-packages\transformers\modeling_utils.py, in from_pretrained, search for " # Extend the modules to not convert to keys that are supposed to be offloaded to `cpu` or `disk`", copy the following part starting with `if isinstance(device_map, str):` (the part that convert device map from 'auto' to a dict) and paste before " # Extend the modules to not convert to keys that are supposed to be offloaded to `cpu` or `disk`"

And pass --load-in-8bit --auto-devices to server.py.

RandomInternetPreson:
Oh my frick... thank you so much kind internet stranger. I was able to use your code and can bifurcate 8-bit between cpu and gpu oh my god, wow like absolutely incredible thank you so much!

athu16:
[...]

PR #358, opened 2023-03-16
Add support for memory maps with --load-in-8bit

mayaeary:
Related to #193 

Now it's possible to use `--gpu-memory` or `--auto-devices` with `--load-in-8bits`.
It uses new transformers library, so maybe you will need an upgrade

I've tested it on pygmalion-6b, gpt-j-6b and OPT-6.7B with my 8GB RTX2070, and it worked at least twice as fast as same models without 8bit or with any other accelerators.

### DIFF
--- a/modules/models.py
+++ b/modules/models.py
@@ -7,8 +7,9 @@ from pathlib import Path
 import numpy as np
 import torch
 import transformers
-from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig, AutoConfig
-from accelerate import infer_auto_device_map, init_empty_weights, load_checkpoint_and_dispatch
+from accelerate import infer_auto_device_map, init_empty_weights
+from transformers import (AutoConfig, AutoModelForCausalLM, AutoTokenizer,
+                          BitsAndBytesConfig)
 
 import modules.shared as shared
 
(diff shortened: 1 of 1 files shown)


oobabooga:
It seems to work. Here is what I have tested:

* The results with the `Debug-deterministic` preset are the same in the old and new implementations with `python server.py --load-in-8bit` (without `--gpu-memory`).
* When `--gpu-memory` is used, the offload works: the GPU memory usage is reduced and the program doesn't crash with a bitsandbytes error like it used to.
* The results with offloading using  `Debug-deterministic` are somewhat different. Probably an upstream bitsandbytes issue.

I will leave the PR open so that more people can test it before merging. Thanks for the refactor, by the way :)

oobabooga:
Oops, I merged at the same time that you commented, @81300. As it is, `--load-in-8bit` does not do any memory map unless `--gpu-memory` is also specified, so my understanding is that `llm_int8_enable_fp32_cpu_offload=True` will just be ignored in this case.

> perhaps outputs are different because the modules that don't end up in VRAM do not get converted to 8-bit

Yes, that's probably the cause of the difference. This issue also appeared here: https://github.com/oobabooga/text-generation-webui/pull/198

PR #363, opened 2023-03-16
Refactor models.py

LuxF3rre:
- Removed unused imports
- Formatted with black
- Extracted tokenizers and model loaders into their own functions
- Changed model parameter calling to use kwargs, extracted CPU failover and suggested memory allocation into helper functions, minor readability improvements for memory_map allocation.
- The general logic untouched


oobabooga:
I find that this makes the code significantly more verbose and harder to read.

Model loading with *kwargs has been also handled in #358

PR #407, opened 2023-03-18
Add loras to .gitignore

ThisIsPIRI:
Also, would you be open to a PR that adds a --lora-dir option, similar to --model-dir in #402?
Sorry if this is not the place to ask.


oobabooga:
I haven't reviewed `--model-dir` yet, later I will give you an update.

PR #414, opened 2023-03-18
Change click me in chatbot download to make it less confusing

RoARene317:
Click me is really confusing since it's no explanation what it works for normal users. I suggest to change it to make it easier.


oobabooga:
Like this if gets misleading because the button reads `Click here to download the chat history` but clicking on it does not cause the chat history to be downloaded. Instead, a download link is generated.

PR #433, opened 2023-03-19
Fix api extension duplicating

mayaeary:
Prevent server in `api` extension run multiple times, causing "Port already taken" error. 

I added new `setup()` method to extensions to run once when app is ready to initialize extensions when app is first loaded.
It also helps against empty "Extensions" block, when no actual ui is present.

### DIFF
--- a/extensions/api/script.py
+++ b/extensions/api/script.py
@@ -8,6 +8,8 @@ params = {
     'port': 5000,
 }
 
+server = None
+
 class Handler(BaseHTTPRequestHandler):
     def do_GET(self):
         if self.path == '/api/v1/model':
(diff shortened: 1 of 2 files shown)


oobabooga:
Do you see a disadvantage in simply moving the `run_server()` call to the global scope? From

```
def ui():
    Thread(target=run_server, daemon=True).start()
```

to

```
Thread(target=run_server, daemon=True).start()
```

mayaeary:
`setup()` is called when ui is ready and parameters from settings.json are parsed, but global scope statements are not

Another thing is rare case, but what if someone wants to import something from other extension without executing it. 
It's like 
```
if __name__ == '__main__':
    do_things()
```

oobabooga:
You are right, this is absolutely necessary. When I created the extensions engine, my idea was to un-import the modules on request, but I am not sure if a use case for that exists and your solution works nicely.

PR #437, opened 2023-03-19
Enable characters to also change the "your name" field, e.g. for interviews.

OWKenobi:
I wanted to make characters for interviews or role play, e.g. Madonna and The Vogue. But characters were unable to change their partners name ("Your name")

With this patch, the string "your_name" is evaluated in the json files, so that they can change the name of the interview or chat partner. But they don't have to; if you omit this field, the interview partner just stays untouched.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -92,8 +92,8 @@ def select_model_from_default_options():
         print("""\nThen type the name of your desired Hugging Face model in the format organization/name.
 
 Examples:
-PygmalionAI/pygmalion-6b
 facebook/opt-1.3b
+EleutherAI/pythia-1.4b-deduped
 """)
 
         print("Input> ", end='')


CrazyKrow:
Yeah, I think this could be great. I'm using the Alpaca lora with the chat interface and for it to work I set the Bot name as "### Response" and the your name as "### Instruction", and its annoying that I have to change "your name" every time I use it.

server.py
```diff
@@ -355,7 +355,7 @@ def create_interface():
             shared.gradio['textbox'].submit(lambda x: '', shared.gradio['textbox'], shared.gradio['textbox'], show_progress=False)
             shared.gradio['textbox'].submit(lambda : chat.save_history(timestamp=False), [], [], show_progress=False)
 
-            shared.gradio['character_menu'].change(chat.load_character, [shared.gradio['character_menu'], shared.gradio['name1'], shared.gradio['name2']], [shared.gradio['name2'], shared.gradio['context'], shared.gradio['display']])
+            shared.gradio['character_menu'].change(chat.load_character, [shared.gradio['character_menu'], shared.gradio['name1'], shared.gradio['name2']], [shared.gradio['name1'],shared.gradio['name2'], shared.gradio['context'], shared.gradio['display']])
```
mcmonkey4eva, line 358: Minor formatting correction: there's a missing space after the `,` here

PR #455, opened 2023-03-20
Do not use empty user messages in chat mode.

VldmrB:
There doesn't seem to be much value to them - they just take up space while also making it seem like there's still some sort of pseudo-dialogue going on, instead of a monologue by the bot.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -34,12 +34,12 @@ def generate_chat_prompt(user_input, max_new_tokens, name1, name2, context, chat
     while i >= 0 and len(encode(''.join(rows), max_new_tokens)[0]) < max_length:
         rows.insert(1, f"{name2}: {shared.history['internal'][i][1].strip()}\n")
         prev_user_input = shared.history['internal'][i][0]
-        if prev_user_input and not shared.history['internal'][i][0] == '<|BEGIN-VISIBLE-CHAT|>':
-            rows.insert(1, f"{name1}: {shared.history['internal'][i][0].strip()}\n")
+        if len(prev_user_input) > 0 and prev_user_input != '<|BEGIN-VISIBLE-CHAT|>':
+            rows.insert(1, f"{name1}: {prev_user_input.strip()}\n")
         i -= 1
 
     if not impersonate:
-        if user_input:
+        if len(user_input) > 0:
             rows.append(f"{name1}: {user_input}\n")
         rows.append(apply_extensions(f"{name2}:", "bot_prefix"))
         limit = 3
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -33,12 +33,14 @@ def generate_chat_prompt(user_input, max_new_tokens, name1, name2, context, chat
     i = len(shared.history['internal'])-1
     while i >= 0 and len(encode(''.join(rows), max_new_tokens)[0]) < max_length:
         rows.insert(1, f"{name2}: {shared.history['internal'][i][1].strip()}\n")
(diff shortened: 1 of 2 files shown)


oobabooga:
The empty message is still sent to the bot in the prompt like:

```
Bot: Hi
User:
Bot: How are you
```

Just removing it from display is not enough.

VldmrB:
Done, please take a look. Sorry about the force pushes, barely used git/GitHub in a long time.

VldmrB:
The bot talks to itself a lot e.g.
```
This is a conversation between two people.
Person 1: Hello!
Person 2: Hi, how are you?
Person 2: I'm fine thank you and yourself?
Person 2: Good thanks...
Person 2:
--------------------
```
But it does the same even if you leave in  the empty user prompts.

OWKenobi:
What if you just let the bot continue on the same line, instead of feeding him a newline and a name? I'm sure it will continue it's own statement then without talking to itself.

oobabooga:
Looks good! This was a very good idea, thanks for submitting the PR. It's interesting to be able to let the bot send multiple messages in sequence.

This also seems like the first step in creating a group chat.

oobabooga:
Sorry, I accidentally squashed the commits, but your contribution is still accounted here: https://github.com/oobabooga/text-generation-webui/graphs/contributors

VldmrB:
All good, happy to see that it made it in. :)

OWKenobi:
> There's still a difference between `continue from the same line` and `omit blank Person 1 prompts`. But I tried what you suggested, since this functionality could be moved out into a separate `Continue` button. I didn't do it well, because most of the time it just returned the exact same reply as on the last line, without any changes (not exactly sure where I fucked up), but the times it did work, the output was very highly predicated on the initial reply. If it was in the form of a question, it would often just start asking more and more questions. It also sometimes still ended up in 2 person mode. Here's an example:
> 
> ```
> Person 1:
> Person 2: Hi there, how are you doing? I am fine thank you and yourself? (says person one) Good to hear that. Thanks for asking me. How was your day today? Great thanks so it seems like we both had great days today then (says Person 2). Oh really cool…(person 1)? That’s wonderful news. Tell me about it. What did you do? (Says Person 2). Well firstly, this morning i got up at around seven o clock in the morning because my alarm rang on time as usual. Then I went straight into the shower with some hot water. Afterwards, when I finished getting dressed, I put some music on loud in order not to disturb anyone while having breakfast but also making sure I can still listen to what they were saying. My mum and dad started telling us all stories about their childhood. After breakfast I left home and came here.(Saying person 1). Oh wow, so you have been busy eh? Cool. So what else happened after that?(says person 2). Well first of all I took a taxi from home to work because I wasn't driving yet. When I arrived, I went through the whole process of going through security checks before entering the office building. The security guards checked me thoroughly before letting me inside. Thereafter I walked down the corridor towards my room where I sat down at my desk to start working. Later on my boss called me into his room for an emergency meeting which lasted for almost three hours until he dismissed us back to our rooms again. Once everyone returned to their own offices or cubicles we continued to carry out whatever task assigned by our managers. Around lunchtime I decided to go shopping at the local supermarket since it was close by and bought myself some food and drinks before returning to work. At around five pm I rushed outside once more just to catch another cab back home because I didn’t want to be
> ```
> 
> I may or may not fiddle with it some more, but for now, I won't try and change the scope of this pull request.

You are right. Yes, I actually also like the idea of a dedicated "continue" button; there are some instances where the bot didn't finish his reply due to 200 token limit, for example when you create php code, I am sure the bot will be able to continue.

PR #309, opened 2023-03-14
 Extension: Stable Diffusion Api integration

Brawlence:
## Description:
Lets the bot answer you with a picture!

Load it in the `--cai-chat` mode with `--extension sd_api_pictures` alongside `send_pictures` (it's not really required, but completes the picture).

If enabled, the image generation is triggered either:
- manually through the extension buttons OR 
- IF the words `'send | mail | me'` are detected simultaneously with `'image | pic | picture | photo'`

One needs an available instance of Automatic1111's webui running with an --api flag. Ain't tested with a notebook / cloud hosted one but should be possible. I'm running it locally in parallel on the same machine as the textgen-webui. One also needs to specify custom `--listen-port` if he's gonna run everything locally.

For the record, 12 GB VRAM is barely enough to run NeverEndingDream 512×512 fp16  and LLaMA-7b in 4bit precision. 
TODO: We should really think about a way to juggle models around RAM and VRAM for this project to work on lower VRAM cards.

---

<details>
<summary>Extension interface</summary>

Interface
Don't mind the Windranger Arcana key in the Prompt Prefix, that's just the name of an embedding I trained beforehand.

</details>

## Demonstrations:

<details>
<summary>Conversation 1</summary>

EXA1
EXA2
EXA3
EXA4

</details>

<details>
<summary>Conversation 2</summary>

Hist1
Hist2
Hist3

</details>

PR #471, opened 2023-03-21
Unload and reload models on request

Brawlence:
An important step towards optimizing running different neural networks in parallel on the same GPU.

**The core idea and the usage case** is simple: when `oobabooga` is used alongside other memory hogs like Stable Diffusion (`sd-api-pictures` extension) or Tortoise-TTS (not yet implemented) this simple unload function leaves a lot more video memory for those other neural networks to work with. Once they finish their jobs, the LLM can be returned back to VRAM.

This is the first one of the possible improvements to #309 memory handling.

---

Tested on my machine, unloading Pyg-2.7B-8bit is almost instant, loading it back (from the RAM cache) takes ~7 seconds which I consider to be an acceptable delay compared to the image generation itself.

Pyg-6B-8bit is a bit slower but still tolerable.

MemGraph

ConsoleLog


Brawlence:
Well, the state after unloading the checkpoint would be undetermined. One won't be able to generate a response, yet the generated error is not fatal and one can resume the chat texgen once the model is loaded back in, that much I tested.

The core idea and the usage case is simple: when `oobabooga` is used alongside other memory hogs like Stable Diffusion (`sd-api-pictures` extension) or Tortoise-TTS (not yet implemented) this simple unload function leaves a lot more videomemory for those other neural networks to work with. Once they finish their jobs, the LLM can be returned back to VRAM.

oobabooga:
In the latest gradio version, there is now this circle icon in dropdown menus that unselects the currently selected option. I have modified the PR for using this button to unload the model from memory.

circle


Your buttons were more functional because they allowed the very same model to be reloaded without having to locate it in the dropdown list, but I found that they occupied a lot of space while being a very niche feature. It should still be possible to create unload/reload buttons inside an extension.

Brawlence:
That's a nice way to save space! 
Though I'll still need the reload_model() _**function**_ in `server.py`, as it would be called in extension which is trying to manage VRAM. I'll just introduce it as part of sd-api-pics update, this will make more sense in context

PR #489, opened 2023-03-22
Extensions performance & memory optimisations

Brawlence:
Affected extensions: `elevenlab_tts`, `silero_tts`, `send_pictures`, `sd_api_pictures`.

---

Reworked `remove_surrounded_chars()` to use regular expression ( https://regexr.com/7alb5 ) instead of repeated string concatenations for `elevenlab_tts`, `silero_tts`, `sd_api_pictures`. 
This should be both **_faster_** and more robust in handling asterisks.

Reduced the memory footprint of `send_pictures` and `sd_api_pictures` by scaling the images shown in the chat to 300 pixels max-side wise. (Nothing is lost since the user presumably has the original file in the case of `send_pictures` and there's an option to save the SD generation for `sd_api_pictures`). 
This should fix **_history growing annoyingly large_** with multiple pictures present


oobabooga:
Can you check if it works as expected after https://github.com/oobabooga/text-generation-webui/pull/489/commits/0abff499e2e40fb308b406b82ad3f850683720e3? I used `image.thumbnail` to simplify the resizing.

PR #495, opened 2023-03-22
feat: add Manual Seed command line argument

wywywywy:
This PR adds a `--seed` command line argument to `server.py`, allowing us to specify a seed for text generation, in order to obtain deterministic results.

This is particularly useful for debugging.

Going forward, the UI should allow changing the seed in the Parameters tab (in a different PR after this is merged).

Please review whether this is the best place to plant the seed. If not please suggest a new one.


oobabooga:
It's better to just make it an interface parameter already. Thanks for the PR! Many people have been requesting this.

PR #500, opened 2023-03-22
Type annotations added

Simon1V:
Around ~80% of the code in /modules has been type annotated.


oobabooga:
I can see why type annotations exist, but I find that they make the code more verbose in a way that I just dislike to look at.

Simon1V:
> I can see why type annotations exist, but I find that they make the code more verbose in a way that I just dislike to look at.

This should not outweight it being easier to fix issues related to typing with this approach, especially since assertions/preconditions become easily to express.

PR #504, opened 2023-03-22
Fix for older GPUs (Maxwell, Pascal, etc.) when using bitsandbytes - 8 bit optimizer - by Tim Dettmers

MichaelMartinez:
Some users of the bitsandbytes - 8 bit optimizer - by Tim Dettmers have reported issues when using the tool with older GPUs, such as Maxwell or Pascal. I beleive they don't even know its an issue. These GPUs do not support the required instructions for the tool to run properly, resulting in errors or crashes.

A simple fix for this issue is available, which requires the use of a prebuilt DLL from https://github.com/james-things/bitsandbytes-prebuilt-all_arch by James. This DLL contains the necessary instructions for the tool to run on older GPUs.

To apply the fix, follow these steps:

1. Download the prebuilt DLL from https://github.com/james-things/bitsandbytes-prebuilt-all_arch.
2. Replace the existing DLL in the bitsandbytes tool directory with the downloaded DLL.
3. Move those files into
4. C:\Users\xxx\miniconda3\envs\textgen\lib\site-packages\bitsandbytes\
5. Now edit bitsandbytes\cuda_setup\main.py with these changes:
6. Change this line:
7. `ct.cdll.LoadLibrary(binary_path)`
8. To the following:
9. `ct.cdll.LoadLibrary(str(binary_path))` *There are two occurrences in the file.*
10. Then replace this line:
11. `if not torch.cuda.is_available(): return 'libsbitsandbytes_cpu.so', None, None, None, None`
12. With the following:
13. `if torch.cuda.is_available(): return 'libbitsandbytes_cudaall.dll', None, None, None, None`
14. Please note that the prebuilt DLL may not work with every version of the bitsandbytes tool, so make sure to use the version that is compatible with the DLL.

I used this on WSL and Regular windows install with a maxwell generation card after trying a bazillion and 1 different methods. Finally, I found that my card was too old and none of the options out in the wild would work until I addressed that issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The recommended installation methods are the following:
 * Linux and MacOS: using conda natively.
 * Windows: using conda on WSL ([WSL installation guide](https://github.com/oobabooga/text-generation-webui/wiki/Windows-Subsystem-for-Linux-(Ubuntu)-Installation-Guide)).
 
-Conda can be downloaded here: https://docs.conda.io/en/latest/miniconda.html
+Conda can be downloaded here: <https://docs.conda.io/en/latest/miniconda.html>
 
 On Linux or WSL, it can be automatically installed with these two commands:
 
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Yea, they are missing HW 8 bit matrix multiplication so you have to bootleg it. Also will need to change how much int8 is converted back to fp16 while inferencing via the threshold argument.

oobabooga:
Thanks for the information. People will be able to find it while searching.

Issue #513, opened 2023-03-23
Other stopping conditions besides new line


Ph0rk0z:
**Description**

So since stop at "\n" was defaulted the choice is to either have really short replies or the bot going off and talking for you.

Does it makes sense to add Botname: and Username: as optional stopping criteria to have the best of both worlds and avoid this bug?

Mainly it matters in chat.

oobabooga:
I have been bothered by the `Person 1` and `Person 2` defaults... I have changed them to "You" and "Assistant" 

https://github.com/oobabooga/text-generation-webui/commit/c5ebcc5f7e862b1f2c6b1d807bbf2c1aadeb159e

USBhost:
There seems to be an issue idk if it's because of this. So in my case when I got a prompt I did not like I hit stop and hit regenerate. But after it regenerates a then when I do a reply it does not post and it replaces the last prompt with the one I hit stop. This has already happened to me twice.

Ph0rk0z:
I have to hit stop twice.. I thought it was me but its happening on all models.

oobabooga:
There is definitely something weird going on with the Stop button. The `generate_reply` function sometimes never reaches the `finally` block after Stop is clicked: https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L271

oobabooga:
I think that this fixed the Stop button in chat mode: https://github.com/oobabooga/text-generation-webui/commit/fd99995b01878246b62302d31a844dd68ee7d139

If you still see weird behavior, please let me know.

Brawlence:
Still reproducible, at least with disabled streaming (for text to speech):

![image](https://user-images.githubusercontent.com/42910943/227729902-8d9332e5-0f7d-474b-bc35-4cf4704dfd0f.png)

oobabooga:
That's not a bug, in this case. The character's name is "Assistant" and not "The Assistant", so the stopping criteria don't recognize that as a new character turn.

USBhost:
![image](https://user-images.githubusercontent.com/7269941/228971191-a9084945-03eb-4d46-93d0-446c2695d0d2.png)

So thoughts?

oobabooga:
With streaming or without streaming? What model?

USBhost:
With streaming aka default?  `python server.py --listen --wbits 4 --groupsize 128 --cai-chat --model llama-65b-test`

Currently testing GPTQ triton

[...]

PR #515, opened 2023-03-23
Stop the bot from talking for you in chat mode

oobabooga:
For #513

### DIFF
--- a/modules/RWKV.py
+++ b/modules/RWKV.py
@@ -45,11 +45,11 @@ class RWKVModel:
             token_stop = token_stop
         )
 
-        return context+self.pipeline.generate(context, token_count=token_count, args=args, callback=callback)
+        return self.pipeline.generate(context, token_count=token_count, args=args, callback=callback)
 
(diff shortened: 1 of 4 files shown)


RandomInternetPreson (commented):
I made this change and it didn't alter the output.  I'm assuming you made this change to avoid the AI having a P1: and P2: conversation on it's own?

I've found that if you give yourself and the AI a name prior to the conversation instead of the default "Person 1" "Person 2" The AI will no longer have the P1: P2: conversations on its own.

oobabooga:
It shouldn't alter the output, it should only stop the text generation in case the bot starts talking as if it was you like

Bot: Hey, how are you?
You: I'm fine

The goal is to prevent

> You: I'm fine

From ever appearing, where "You" is your name.

Ph0rk0z:
Oh they will happily say your name and their own name too. But so far so good.

oobabooga:
I probably messed something up.

oobabooga:
Some big simplifications in the code now.

PR #529, opened 2023-03-24
Allow loading of .safetensors through GPTQ-for-LLaMa

EyeDeck:
Quantized models were hardcoded to only load .pt, but qwopqwop200's repo already works with .safetensors if we just pass in the other file extension.

With this PR, it looks for models in the order:
* models\\*model*-#bit.safetensors
* models\\*subfolder*\\*model*-#bit.safetensors
* models\\*model*-#bit.pt
* models\\*subfolder*\\*model*-#bit.pt

Seems to work fine, but I only tested on one janky model I ran through the quantizer myself.

### DIFF
--- a/modules/GPTQ_loader.py
+++ b/modules/GPTQ_loader.py
@@ -37,21 +37,23 @@ def load_quantized(model_name):
 
     path_to_model = Path(f'models/{model_name}')
     if path_to_model.name.lower().startswith('llama-7b'):
-        pt_model = f'llama-7b-{shared.args.gptq_bits}bit.pt'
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Yes, except we all have to re-quantize 100gb of models for this.

EyeDeck:
Why would that be necessary? I mean, it's going to need to be done to support newer versions of GPTQ-for-LLaMa at some point, but that's not related to this PR.

oobabooga:
Thanks, this is handy! More updates on GPTQ will come after https://github.com/oobabooga/text-generation-webui/pull/530.

PR #546, opened 2023-03-24
Send_pictures small fix

Brawlence:
Since the `text` variable is just for passing to the model instead of the picture, changed internal regular quotes to [typographical curly quotes](https://typographyforlawyers.com/straight-and-curly-quotes.html). 

The model interprets this all the same, yet curly quotes prevent possible HTML & JSON breakage down the line — as it was the case with alt-text 6 lines lower ( `visible_text = f'` )

### DIFF
--- a/extensions/send_pictures/script.py
+++ b/extensions/send_pictures/script.py
@@ -24,7 +24,7 @@ def caption_image(raw_image):
     return processor.decode(out[0], skip_special_tokens=True)
 
 def generate_chat_picture(picture, name1, name2):
-    text = f'*{name1} sends {name2} a picture that contains the following: "{caption_image(picture)}"*'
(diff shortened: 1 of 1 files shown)


oobabooga:
Isn't it better to just find a way to escape the `"` characters?

Brawlence:
At least for HTML, it wouldn't be possible short of replacing every instance of `"` with `&quot;` codes every time we try to use that string programmatically

Brawlence:
Alternatively, we can rid of the quotes, like #607 ( [the string in question](https://github.com/oobabooga/text-generation-webui/pull/607/files#diff-77339a19c9f9148ecfea949fb521a2d810067e967efe1ebaa51314512dc68c44R55) ) suggests

Accepting that pull will fix the issue described here too.

oobabooga:
Let's use this for now

Issue #550, opened 2023-03-25
port being used when restarting web ui


catalpaaa:
### Describe the bug

when restart the web ui, the port is always shown to be used.
my temp solusion is to sleep 120s before recreating interface.

ie add time.sleep(120) after the very last shared.gradio['interface'].close() in server.py

it might not be a thing for newer gradio cus A1111's webui dose not have this issue. or maybe he did some blackmagic

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

start normal,
change mode or do whatever that restart the webui
boom

### Logs

```shell
Running on local URL:  http://0.0.0.0:9441

To create a public link, set `share=True` in `launch()`.
Task exception was never retrieved
future: <Task finished name='02liijn6knwb_18' coro=<Queue.process_events() done, defined at /home/catalpa/anaconda3/envs/textgen/lib/python3.10/site-packages/gradio/queueing.py:329> exception=AssertionError('No event data')>
Traceback (most recent call last):
  File "/home/catalpa/anaconda3/envs/textgen/lib/python3.10/site-packages/gradio/queueing.py", line 343, in process_events
    response = await self.call_prediction(awake_events, batch)
  File "/home/catalpa/anaconda3/envs/textgen/lib/python3.10/site-packages/gradio/queueing.py", line 303, in call_prediction
    assert data is not None, "No event data"
AssertionError: No event data
Closing server running on port: 9441
/home/catalpa/anaconda3/envs/textgen/lib/python3.10/site-packages/gradio/deprecation.py:40: UserWarning: The 'type' parameter has been deprecated. Use the Number component instead.
  warnings.warn(value)
Traceback (most recent call last):
  File "/home/catalpa/anaconda3/envs/textgen/lib/python3.10/site-packages/gradio/networking.py", line 119, in start_server
    s.bind((LOCALHOST_NAME, server_port))
OSError: [Errno 98] Address already in use

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/catalpa/nas/program/text-generation-webui/server.py", line 468, in <module>
    create_interface()
  File "/home/catalpa/nas/program/text-generation-webui/server.py", line 457, in create_interface
[...]

PR #556, opened 2023-03-25
add --wait-before-restart to allow gradio to have time to free port

catalpaaa:
simple work arround for https://github.com/oobabooga/text-generation-webui/issues/550
the lowest time i could go was 75s

### DIFF
--- a/modules/LoRA.py
+++ b/modules/LoRA.py
@@ -32,7 +32,7 @@ def add_lora_to_model(lora_name):
             elif shared.args.load_in_8bit:
                 params['device_map'] = {'': 0}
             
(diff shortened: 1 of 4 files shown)


tensiondriven:
Ugh! It's so frustrating when the UI doesn't restart.

I think ultimately we'll want a more robust solution though - Since the UI doesn't crash every time, having to endure a 75s restart each time is pretty brutal.

Would it be possible to catch the error from Gradio when it happens, and then back off and retry again later?  Even better, would it be possible to add a little javascript or something to update the UI in the case where the restart takes more than n seconds, so users know that the port is not available and the backend is waiting for a restart?

oobabooga:
As explained here https://github.com/oobabooga/text-generation-webui/issues/550, I think that it's best to try to get gradio to fix this bug upstream.

PR #565, opened 2023-03-25
improve/simplify gitignore

mcmonkey4eva:
This is a minor change but bit of QOL for forking/contributing to this repo

- add repositories
- remove the redundant `/*` on folders
- remove the exclusions for files that already exist

If it's unclear what difference this makes: pretty much it fixes the majority of issues with git thinking you've made changes to the repo just by virtue of ... using the WebUI. Not all, but most. Notably `repositories` being added fixes the issue where GTPQ support adds an unneeded local change.

The `!` exclusions at the bottom don't have any function, gitignore doesn't apply to files that are already in the repository.
The `/*` is somewhere between redundant and actively annoying.

- This also adds `datasets` to gitignore for #570 

Only remaining issues for me personally is
- it's annoyed about my `models` dir being a symlink but <https://github.com/oobabooga/text-generation-webui/pull/549> will resolve that.
- there isn't a good directory for personal custom presets (ie ones I don't intend to git commit)

PS while I have oobabooga's attention here: have you thought about making an official Discord for the WebUI? Either for the community or just for devs/extension authors/etc. at least to communicate (For comparison Automatic1111 WebUI has an official dev-only discord and a separate unofficial public discord)


oobabooga:
This excludes files like `place-your-loras-here.txt` and `characters/Example.json` from being eventually updated. I prefer to keep .gitignore in the current format.

mcmonkey4eva:
@oobabooga That is incorrect, it does not exclude them from being updated, that's why I was trying to explain with the description of this PR.
Because the files are in the git-tracked tree already, they remain tracked, it is not the duty of a `gitignore` file to enforce that. You can test locally with any git client or GUI to verify this if you don't believe me.
(Screenshot: GitHub Desktop showing an edit to `example.json` from this improved gitignore branch, clearly demonstrating that edits to the file are tracked exactly the same as files that were never in a gitignore'd directory)
image



Negatives would make sense if eg you wanted to negative entire folders to be able to add new files in them easily. But that's not the case here.

This PR is valid and I request it be reopened and reconsidered.

oobabooga:
You are right, the files that I mentioned are indeed still tracked and the syntax of your updated .gitignore is much cleaner. Sorry, I was too fast to discard the PR, and thanks for insisting!

PR #568, opened 2023-03-25
silero_tts streaming fix

Brawlence:
Temporarily suppress the streaming during the audio response as it would interfere with the audio (making it stutter and play anew)

Demonstrating the issue:
Issue_demo
As one can see, every time the streamed text is updated, currently a new voice line is generated and played, resulting in stutters.

Suppressing the streaming fixes it.


oobabooga:
That's creative! It works.

PR #572, opened 2023-03-25
Potential fix for issues/571

clusterfudge:


MrFelt:
Tried, didn't resolve for me.

  File "/home/ftg/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/routes.py", line 374, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/ftg/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/blocks.py", line 1017, in process_api
    result = await self.call_function(
  File "/home/ftg/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/blocks.py", line 849, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/home/ftg/.local/lib/python3.10/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/home/ftg/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/home/ftg/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/home/ftg/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/utils.py", line 453, in async_iteration
    return next(iterator)
  File "/home/ftg/textgen/modules/chat.py", line 177, in cai_chatbot_wrapper
    for _history in chatbot_wrapper(text, max_new_tokens, do_sample, temperature, top_p, typical_p, repetition_penalty, encoder_repetition_penalty, top_k, min_length, no_repeat_ngram_size, num_beams, penalty_alpha, length_penalty, early_stopping, seed, name1, name2, context, check, chat_prompt_size, chat_generation_attempts):
  File "/home/ftg/textgen/modules/chat.py", line 148, in chatbot_wrapper
    cumulative_reply = reply
UnboundLocalError: local variable 'reply' referenced before assignment

mayaeary:
It's better to add `reply = ''` before `for output in generator:` on line 226

clusterfudge:
> It's better to add `reply = ''` before `for output in generator:` on line 226

@mayaeary Can you clarify why you believe this is better? My understanding of the code is that if the generator [here](https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L226) yields no results, we trigger this error. In your suggestion, we would yield a single empty response instead, and I'm not sure that's the correct behavior. Perhaps @oobabooga would like to weigh in on this?

clusterfudge:
@MrFelt I think you're demonstrating a different (but similar) error, as the error is coming from a different file (`chat.py`). Please feel free to add on to #572 with reproduction steps, or create a new issue!

MrFelt:
> @MrFelt I think you're demonstrating a different (but similar) error, as the error is coming from a different file (`chat.py`). Please feel free to add on to #572 with reproduction steps, or create a new issue!

Ah, my bad. Will do.

mayaeary:
>
> 
> @mayaeary Can you clarify why you believe this is better? 
Generator trying to yield from variable `reply`, which is'n set if there is no tokens in answer. 

But you are right, maybe we should return nothing instead of empty string.

clusterfudge:
@oobabooga The reproduction in the linked issue doesn't require the "Stop"
button to be clicked -- it just requires that the generator yield no
results. The question I have is whether you always want to yield at least
one result from this loop, or if yielding none is the appropriate behavior.

oobabooga:
I have reproduced the issue and also your solution. I think that it makes sense to not return anything if no tokens are generated, and that doesn't seem to generate any error. Thank you :)

PR #585, opened 2023-03-26
Also download Markdown files

fkusche:


oobabooga:
Are these useful?

fkusche:
Sometimes, there is a `README.md` with helpful information

Issue #533, opened 2023-03-24
Expand sd_api_pictures to support adventure game/interactive story use-case


vaidasmaciulis:
**Description**

I have modified sd_api_pictures script locally to use it for a graphic text adventure game:

![image](https://user-images.githubusercontent.com/11597256/227478569-b21a9042-9e71-4603-b606-a164cc9259d4.png)

Would be nice to incorporate changes required to support such use case in main repo.

Changes that is required to achieve this (all should be made configurable, to not interfere with current behavior):

- Generate image on every response
- Do not modify input text
- Do not modify output text
- Allow to configure image width/height independently
- Do not resize output image

Modified POC script, that can be used as reference:

```
import base64
import io
import re
from pathlib import Path

import gradio as gr
import modules.chat as chat
import modules.shared as shared
import requests
import torch
from PIL import Image

torch._C._jit_set_profiling_mode(False)

# parameters which can be customized in settings.json of webui  
params = {
    'enable_SD_api': True,
    'address': 'http://127.0.0.1:7860',
    'save_img': False,
    'SD_model': 'NeverEndingDream', # not really used right now
    'prompt_prefix': 'fantasy painting, ',
    'negative_prompt': 'text, caption, words, 3d, nrealfixer, nfixer, nartfixer',
    'side_length': 512,
    'restore_faces': False
}

SD_models = ['NeverEndingDream'] # TODO: get with http://{address}}/sdapi/v1/sd-models and allow user to select

streaming_state = shared.args.no_stream # remember if chat streaming was enabled
picture_response = True # specifies if the next model response should appear as a picture
pic_id = 0

def remove_surrounded_chars(string):
    # this expression matches to 'as few symbols as possible (0 upwards) between any asterisks' OR
    # 'as few symbols as possible (0 upwards) between an asterisk and the end of the string'
    return re.sub('\*[^\*]*?(\*|$)','',string)

# I don't even need input_hijack for this as visible text will be commited to history as the unmodified string
def input_modifier(string):
    """
    This function is applied to your text inputs before
    they are fed into the model.
    """
    # Adventure game: no modification on input
    return string

# Get and save the Stable Diffusion-generated picture
def get_SD_pictures(description):

    global params, pic_id

    payload = {
        "prompt": params['prompt_prefix'] + description,
        "seed": 3463809478,
        "sampler_name": "Euler a",
        "steps": 12,
        "cfg_scale": 7,
        "width": 640, # Adventure game change
[...]

Issue #573, opened 2023-03-25
sd_api_pictures can't connect to stable diffusion on wsl


laughinggaschambers:
### Describe the bug

Running the sd api natively on windows 10 works 
but when doing it via wsl it's not able to connect to the stable diffusion address(sd is being run on windows 10 natively)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run stable diffusion on windows 10
Run the webui in wsl

### Logs

```shell
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/routes.py", line 374, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/blocks.py", line 1017, in process_api
    result = await self.call_function(
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/blocks.py", line 849, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/home/doom/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/utils.py", line 453, in async_iteration
    return next(iterator)
  File "/home/doom/text-generation-webui/modules/chat.py", line 167, in cai_chatbot_wrapper
    for _history in chatbot_wrapper(text, max_new_tokens, do_sample, temperature, top_p, typical_p, repetition_penalty, encoder_repetition_penalty, top_k, min_length, no_repeat_ngram_size, num_beams, penalty_alpha, length_penalty, early_stopping, seed, name1, name2, context, check, chat_prompt_size, chat_generation_attempts):
  File "/home/doom/text-generation-webui/modules/chat.py", line 125, in chatbot_wrapper
    visible_reply = apply_extensions(visible_reply, "output")
[...]

PR #596, opened 2023-03-27
SD Api Pics extension, v.1.1

Brawlence:
Lots of new things, gonna do a quick draft right now:

New features: API detection (press enter in the API box), more robust host address handing, VRAM management (model shuffling), separate image dimensions sliders for H&W, three different trigger modes (manual, interactive, always-on). The button to force/suppress the image on the next reply.
The model input is modified only in the interactive mode; other two are unaffected. The output pic description is presented differently for Picture-book / Adventure mode.

Connection check:
Connection feature demo

It's gonna strip the `http(s)://` part and the trailing `/` if present and also return the status when pressing Enter in that field

#533


oobabooga:
It's looking great! I'm really excited for this.

extensions/sd_api_pictures/script.py
```diff
@@ -96,6 +105,14 @@ def get_SD_pictures(description):
         image_bytes = buffered.getvalue()
         img_str = "data:image/jpeg;base64," + base64.b64encode(image_bytes).decode()
         visible_result = visible_result + f'<img src="{img_str}" alt="{description}">\n'
```
mcmonkey4eva, line 107: Suggestion: if `save_img` is on, skip this shrink and embed step, and instead embed ``. Makes for much cleaner+shorter data, and allows full original res to be used.

extensions/sd_api_pictures/script.py
```diff
@@ -14,13 +14,15 @@
 
 # parameters which can be customized in settings.json of webui
```
vaidasmaciulis, line 15: Can this comment be expanded with example, how to add these to settings.json?

extensions/sd_api_pictures/script.py
vaidasmaciulis: Also, please make image size configurable, allowing to not resize it at all.

extensions/sd_api_pictures/script.py
```diff
@@ -119,16 +136,16 @@ def output_modifier(string):
...
 
-    # I can't for the love of all that's holy get the name from shared.gradio['name1'], so for now it will be like this
-    text = f'*Description: "{string}"*'
-
-    image = get_SD_pictures(string)
+    text = ""
+    if (params['mode']<2):
+        toggle_generation(False)
+        text = f'*Sends a picture which portrays: “{string}”*'
+    else:
```
vaidasmaciulis, line 143: Perhaps `toggle_generation(True)` is missing here, as after switching to Aventure mode images does not automatically start to generate

extensions/sd_api_pictures/script.py
```diff
@@ -139,41 +156,74 @@ def bot_prefix_modifier(string):
...
+
+    return gr.Textbox.update(label=msg)
+
+
 
 def ui():
 
     # Gradio elements
-    with gr.Accordion("Stable Diffusion api integration", open=True):
+    gr.Markdown('### Stable Diffusion API Pictures')
```
vaidasmaciulis, line 198: The title is written bellow another h1 title "sd_api_pictures", which I guess is added automatically.

vaidasmaciulis (approved):

extensions/sd_api_pictures/script.py
```diff
@@ -105,7 +122,7 @@ def output_modifier(string):
     """
     This function is applied to the model outputs.
     """
-    global pic_id, picture_response, streaming_state
+    global pic_id, picture_response, params
```
vaidasmaciulis, line 125: Would be nice, if there is a way, to skip generating image when loading character which already has some history (now it generates picture, but it is not visible anywhere)

extensions/sd_api_pictures/script.py
Brawlence: > Also, please make image size configurable, allowing to not resize it at all.

That's not defined here, the 300x300 viewport is set in the main ooba css. I could probably override it, but the resizing is done just to match the viewport

extensions/sd_api_pictures/script.py
Brawlence: Whoa that's a nice catch! Can you list the exact steps to reproduce it?


Gonna think if I can actually override that behaviour

extensions/sd_api_pictures/script.py
```diff
@@ -66,19 +65,29 @@ def get_SD_pictures(description):
...
+        shared.gradio['unload_model_fn']()
+        response = requests.post(url=f'http://{params["address"]}/sdapi/v1/reload-checkpoint', json='')
+        response.raise_for_status()
+        r = response.json()
+        del response
+        del r
+
     payload = {
         "prompt": params['prompt_prefix'] + description,
         "seed": -1,
```
vaidasmaciulis, line 78: Would be nice to have seed, sampler_name, steps and cfg_scale configurable as well.

mcmonkey4eva: I don't know why you'd want to manually specify a seed when an AI is randomly generating the prompt?

vaidasmaciulis: some times same seed provides similar images on slightly different promts, it helps to get more consistent characters/visuals (or I am just seeing things)

laughinggaschambers:
send_pictures always triggers generation in immersive mode. I believe the input prompt from send_pictures has the words send and picture in them. I might be good to filter out any prompts that uses send_pictures.
image

Brawlence:
Okay, just a little checklist to keep my thoughts organized:

- Fix bugs:
    - [ ] title duplication
    - [ ] adventure mode [not activating image-gen right away](https://github.com/oobabooga/text-generation-webui/issues/533#issuecomment-1485719154)
    - [ ] send_pictures triggering image generation (because of the replacement text)
    - [ ] character swapping triggering image generation (the image is immediately discarded)

- Write a readme, include
     - [X] Connection checker: how does it work, allowed inputs
     - [ ] warning for WSL users: loopback is limited to WSL \ Native, use local IP instead 
     ( #573 )
     - [ ] Difference between Manual, Picturebook Adventure (always-on) and Interactive modes
     - [ ] VRAM shuffler explanation
     - [ ] [How to setup persistent extension settings via settings.json](https://github.com/oobabooga/text-generation-webui/wiki/Extensions#params-dictionary)

- Add functionality
    - [ ] Manually configure seed, sampler_name, steps and cfg_scale
        - [ ] request available samplers from SD
    - [ ] Try to override image viewport, make it display original sized-images
        - [ ] If the save original file checkbox is set, [display local file instead (smart!)](https://github.com/oobabooga/text-generation-webui/pull/596#discussion_r1149650125)
        - [ ] But to do so, probably need to keep images unique, maybe introduce subfolders, grouping received pics by date?
    - [ ] Zoom-in feature? 

- Rework
    - [ ] Interactive mode flow change
        - [ ] Factor out the decider function
        - [ ] Allow the bot to refuse sending pics? w/ option to force anyway
    - [ ] In-chat hidden context button for 'illustrate'

laughinggaschambers:
using regenerate display "is typing" instead of "Is sending a picture"

Brawlence:
@laughinggaschambers that's unfortunate but I can't detect re-generations; so if you didn't like the image it sent you'll have to 'remove-last' and then 'generate' instead

extensions/sd_api_pictures/script.py
Brawlence: Yup, in the planned Readme

ARandomUserFromGithub:
Hi, I have an issue with this plugin whereas the generated picture only quickly flashes and instantly disappears. Also the HTML displays nothing between the textbox divs. 
`<div class="message">
                <div class="circle-bot">
                  
                </div>
                <div class="text">
                  
                  
                </div>
              </div>`

vaidasmaciulis:
> Okay, just a little checklist to keep my thoughts organized:
> 
@Brawlence would it make sense to merge current PR and address remaining issues with separate PRs?

oobabooga:
I agree with @vaidasmaciulis. There is no need for @Brawlence to overwhelm himself and try to fix everything at once. If there are not any new bugs in this PR that need to be addressed, I'd vote for merging it.

extensions/sd_api_pictures/script.py
```diff
@@ -139,41 +188,76 @@ def bot_prefix_modifier(string):
...
+
+    shared.args.no_stream = True if picture_response else streaming_state # Disable streaming cause otherwise the SD-generated picture would return as a dud
+    shared.processing_message = "*Is sending a picture...*" if picture_response else "*Is typing...*"
+    btn_text = "Suppress the picture response" if picture_response else "Force the picture response"
+    
+    return btn_text
+
+def filter_address(address):
+    address = address.strip()
+    address = re.sub('http(s)?:\/\/|\/$','',address) # remove starting http:// OR https:// OR trailing slash
```
mcmonkey4eva, line 206: Why is this done? This seems like the primary result is just going to be breaking functionality for anyone that uses `https`, and not making a difference for anyone that doesn't apply to.

extensions/sd_api_pictures/script.py
```diff
@@ -64,38 +105,46 @@ def input_modifier(string):
...
-        "width": params['side_length'],
-        "height": params['side_length'],
+        "width": params['width'],
+        "height": params['height'],
         "restore_faces": params['restore_faces'],
         "negative_prompt": params['negative_prompt']
     }
     
-    response = requests.post(url=f'{params["address"]}/sdapi/v1/txt2img', json=payload)
+    print(f'Prompting the image generator via the API on {params["address"]}…')
```
mcmonkey4eva, line 124: Minor, but `…` is a unicode symbol that I'm not sure will render in all terminals (vs just normal `...`)

extensions/sd_api_pictures/script.py
```diff
@@ -9,52 +10,92 @@
...
+    for sett in shared.settings:
+        if sett.startswith(f'{ext_name}-'):
+            params.update({sett.replace(f'{ext_name}-',''): shared.settings[sett]})
+update_params()
+
+def give_VRAM_priority(actor):
+    
+    global shared, params
+
+    if (actor == 'SD'):
```
mcmonkey4eva, line 42: Also minor, but I can see you're used to languages other than python - `()` on `if` isn't needed for python

extensions/sd_api_pictures/script.py
```diff
@@ -9,52 +10,92 @@
...
+    'SD_model': 'NeverEndingDream', # not used right now
+    'prompt_prefix': '(Masterpiece:1.1), detailed, intricate, colorful',
     'negative_prompt': '(worst quality, low quality:1.3)',
-    'side_length': 512,
+    'width': 512,
+    'height': 512,
     'restore_faces': False
 }
 
+def update_params(): # somewhy the default extension params dict is not changed by settings.json, thus update it forcefully
```
mcmonkey4eva, line 31: This sounds like it'll be worth looking into and fixing

mcmonkey4eva (commented):
(Review submitted because I was tagged for it) there's only one potentially-breaking issue I see, the `https` part, and a lot of formatting, but that is stuff that can easily be worked on in followups

EDIT: gave it a testrun, worked flawlessly from the text-gen-webui side.
(The Auto WebUI side had errors, but, er, that's an auto webui issue unrelated. Extensions and API calls don't get along)

extensions/sd_api_pictures/script.py
Brawlence: https://github.com/oobabooga/text-generation-webui/pull/309#issuecomment-1475828792
It seems that trying to connect to API via https breaks functionality for whoever uses it regardless

mcmonkey4eva: That... seems more like user error in the config of the user that posted that. There's no reason use `https` on its own should cause issues -IF- https is configured properly aka not using a self-signed cert from an autowebui addon, and instead using a real https certifier (eg a cloudflare tunnel)

extensions/sd_api_pictures/script.py
Brawlence: I've looked into it and it's being done, but just a bit too late. 

If something is happening right as the extension is loading -- like, for example, setting initial UI state (names of the buttons), or [checking if local cache exists (for Silero-TTS) and switching to it without connecting to Net](https://github.com/oobabooga/text-generation-webui/pull/628/files#diff-428f3d2a0bf16166af01f47be473a94fa202e4844e2ea38547d5d2742d16cf20R62) it's already too late.

extensions/sd_api_pictures/script.py
Brawlence: Convincing. Reverted the https strip

Brawlence:
@vaidasmaciulis although nothing _**seems**_ broken right now, for the last 3 or 4 days I couldn't really test the full extent of my changes because the flow kept breaking cause I want all the new features. I should have fixed the repo state, but oh well.

@oobabooga thanks but don't worry, I am not really overwhelmed

@mcmonkey4eva thanks, your feedback is invaluable!

Brawlence:
Okay, finished with everything I wanted but readme. It's coming later™

oobabooga:
In Manual mode, if I click on `Force (Suppress) the picture response`, the button changes to `Suppress the picture response` and the next reply is an image.

After that, the button text remains at `Suppress the picture response`, but the next generation returns text instead of another image.

Brawlence:
@oobabooga so the state changes correctly but the button is not reset after the generation. Hmmmm

Is there a way to access this button component outside of extension `ui()`'s scope?

oobabooga:
You can turn it into a global variable (inside script.py), like with `global my_button` at the top of the ui() function.

extensions/sd_api_pictures/script.py
mcmonkey4eva: The fix is a subject for a separate PR but I think what could be done is to have an explicit `def on_load():` or similar function, such that the order of calls is: (1) file init, (2) settings load, (3) `on_load` called, and thus any init behavior that needs settings can simply be put inside the `on_load` instead of in the file init.
As a bonus, if it's ordered such that _all file inits_ for all extensions happen, _then_ all `on_load`, in separate groups, it would enable the ability for extensions to interact with each other.

Brawlence:
Strangely, although declaring it globally does allow me to change its label, the rendered button won't automatically adjust. So, as it is seemingly useless as a feedback for the user, I'm splitting it in two buttons. There won't be any reliable indication for the state of the next reply (picture/regular), but one can always force it with those.

oobabooga:
I did some reorganizing/reformatting. Everything is working well on my end. I'm impressed at how engineered and detailed this extension has become. In particular, the new `Picturebook` is really fun to use.

Thanks for the update, @Brawlence :) (and feel free to submit new PRs in the future if you come up with further enhancements)

PR #618, opened 2023-03-28
Improve download-model.py progress bar with multiple threads

nikita-skakun:
The previous implementation used threads to download files in parallel, which could lead to performance issues due to the Global Interpreter Lock (GIL). This commit refactors the download process to use multiprocessing instead, which allows for true parallelism across multiple CPUs. This results in significantly faster downloads, particularly for large models.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -149,13 +149,6 @@ def download_files(file_list, output_folder, num_threads=8):
     thread_map(lambda url: get_file(url, output_folder), file_list, max_workers=num_threads, verbose=False)
 
 if __name__ == '__main__':
-    parser = argparse.ArgumentParser()
-    parser.add_argument('MODEL', type=str, default=None, nargs='?')
-    parser.add_argument('--branch', type=str, default='main', help='Name of the Git branch to download from.')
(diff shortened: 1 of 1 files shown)


oobabooga:
Wasn't the previous implementation equivalent? It also used `multiprocessing.Pool`.

Also, is there a way of having well behaved progress bars for parallel downloads?

nikita-skakun:
Yes, the previous implementation also used multiprocessing.Pool, but it used the map method instead of the imap_unordered method. The map method blocks the main process until all the results are ready, while the imap_unordered method allows the main process to continue running while the worker processes are still processing the data.

As for the progress bars, I am looking into it now.

nikita-skakun:
I should probably rename this pull request to be about the visual changes.

oobabooga:
I have merged some changes that I made today (add a `--output` flag, save metadata in the model folder, put the organization name at the beginning of the model folder). Let me know if you see something wrong in the merge, but everything seems to be fine.

The new progress bar is glorious. I wanted to do this since the beginning but I didn't know how.

A next step now could be to add the ability to continue partially downloaded files (I also never managed to get this to work).

Issue #561, opened 2023-03-25
Automatic model parallel inference by deepspeed


sgsdxzy:
**Description**

The current multi-gpu setup uses a simple pipeline parallelism (PP) provided by huggingface transformers, which is inefficient because only one gpu can work at the same time. DeepSpeed-Inference introduces several features to efficiently serve transformer-based PyTorch models. It supports model parallelism (MP) to fit large models that would otherwise not fit in GPU memory. Even for smaller models, MP can be used to reduce latency for inference. https://github.com/microsoft/DeepSpeed/blob/master/docs/_tutorials/inference-tutorial.md

DeepSpeed-Inference provides two methods of supporting MP:
1. For a limited set of [supported models](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/replace_policy.py), it will automatically partition the model as necessary, inject compatible high performance kernels into your model and manage the inter-gpu communication. The list includes GPT-NeoX, GPT-J and OPT.
2. Pass an injection policy that shows the two specific linear layers on a Transformer Encoder/Decoder layer: 1) the attention output GeMM and 2) layer output GeMM.

I am particularily interested in applying method 2 to LLaMA, because it is one of the largest model that textui supports that would benefit from performance improvements.

The LLaMA model has a structure like this
```
LlamaForCausalLM(
  (model): LlamaModel(
    (embed_tokens): Embedding(32000, 5120, padding_idx=0)
    (layers): ModuleList(
      (0-39): 40 x LlamaDecoderLayer(
        (self_attn): LlamaAttention(
          (q_proj): Linear(in_features=5120, out_features=5120, bias=False)
          (k_proj): Linear(in_features=5120, out_features=5120, bias=False)
          (v_proj): Linear(in_features=5120, out_features=5120, bias=False)
          (o_proj): Linear(in_features=5120, out_features=5120, bias=False)
          (rotary_emb): LlamaRotaryEmbedding()
        )
        (mlp): LlamaMLP(
          (gate_proj): Linear(in_features=5120, out_features=13824, bias=False)
          (down_proj): Linear(in_features=13824, out_features=5120, bias=False)
          (up_proj): Linear(in_features=5120, out_features=13824, bias=False)
          (act_fn): SiLUActivation()
        )
        (input_layernorm): LlamaRMSNorm()
        (post_attention_layernorm): LlamaRMSNorm()
      )
    )
    (norm): LlamaRMSNorm()
  )
[...]

PR #627, opened 2023-03-29
Support GLM-130B via external api.

sgsdxzy:
This pr adds support for passing an http endpoint as --model, allowing textui to serve as a frontend to other models that are not currently in huggingface or complicated to setup/require special versions of dependencies.
The primary goal is to support [GLM-130B](https://github.com/THUDM/GLM-130B), which runs on custom transformers implementations. It is recommended to run GLM-130B on [FasterTransformer](https://github.com/NVIDIA/FasterTransformer) in a docker, which exposes a http api.

Run textui with
```python server.py --model http://127.0.0.1:5000```

Infilling works (GLM-130B is a masked language model that supports both infilling and appending)
image
image

Chatting also works
image

The generation speed is quite good, considering it's a 130B model running on 2080Ti 22G x 4. The word count uses `len()`, so a Chinese character takes up multiple words.
image

I think it is worth extending external api support to:

 1. ChatGLM-6B and smaller GLM variants (requires custom code)
 2. Distributed deepspeed/torchrun, as they have a multiprocess structure too cumbersome to stuff into textui #561 
 3. The original tensor-parallel version of Llama from Meta.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -31,7 +31,7 @@ def generate_chat_prompt(user_input, max_new_tokens, name1, name2, context, chat
     max_length = min(get_max_prompt_length(max_new_tokens), chat_prompt_size)
 
     i = len(shared.history['internal'])-1
-    while i >= 0 and len(encode(''.join(rows), max_new_tokens)[0]) < max_length:
+    while i >= 0 and (shared.is_external_api or len(encode(''.join(rows), max_new_tokens)[0]) < max_length):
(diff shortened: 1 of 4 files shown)


oobabooga:
I believe that it would be better to implement this as an extension that uses custom_generate_reply

https://github.com/oobabooga/text-generation-webui/blob/main/docs/Extensions.md#custom_generate_reply-example

Since this model is presumably too big to run on consumer GPUs, it would be better to host this extension separately and add it to https://github.com/oobabooga/text-generation-webui-extensions

PR #628, opened 2023-03-29
Silero TTS offline cache

### LINKED ISSUE #516: silero_tts will not load if I am not connected to the internet
### Describe the bug

I used to be able to use this extension offline, but now I can't load the extension if I am not online.  If I am online the extension loads just fine.  The actual language models is saved on my machine via the .cache file: C:\Users\myself\.cache\torch\hub\snakers4_silero-models_master\src\silero\model

The model name is called v3_en.pt, it's being cached on my machine and when I load the extension with an internet connection the miniconda console says that it's using the cached model, so I don't know why I NEED to be connected to the internet for it to work.

### Reproduction

Run this (change your install location as necessary) with and without an internet connection.

cd F:\OoBaboogaMarch17\text-generation-webui
conda activate textgen
python .\server.py --auto-devices --gptq-bits 4 --cai-chat --gptq-model-type LLaMa --extension silero_tts 

### Screenshot

I'm including two screenshots, one when I am connected to the internet, and one when I am not connected to the internet.

internetConnection
no internet connection

### Logs

```shell
See screenshots
```

### System Info

```shell
Window 10, 4090, i9 13900, windows mode not wsl
```

Brawlence:
Make Silero extension try to load the model from local cache if it's present, and only THEN try to connect to torch hub via the internet.

The cache location is configurable via `settings.json`, ex.:

```
{
    "silero_tts-local_cache_path": "C:/Users/Me/.cache/torch/hub/snakers4_silero-models_master/"
}
```

Will result (even offline!) in: 

Console

---

**Note**: as some-why `params` dictionary is not updated with the values from `settings.json` as of now, `update_params()` function is needed.

Resolves #516


oobabooga:
Isn't there a library that finds the default configuration folder for the current OS automatically? Hardcoding the default path as `C:/Users/USER/.cache/torch/hub/snakers4_silero-models_master/`is uncomfortable to me as a Linux user.

Brawlence:
> Isn't there a library that finds the default configuration folder for the current OS automatically?

It haven't really crossed my mind, sorry. Even better, since torch itself is already imported, we can get the cache path by simply using `torch.hub.get_dir()`

Brawlence:
There. Now it should be platform agnostic

oobabooga:
Thank you, I have tested it and it worked on Linux. I replaced the update_params() function with loading the model inside a setup() function (a recent addition) that is only called once when the extension is loaded. I will follow up with a new commit improving the timing for the setup() call

PR #547, opened 2023-03-24
Dockerfile / docker-compose to help streamline build process

loeken:
Wanted to run in docker, used https://github.com/RedTopper`s version in https://github.com/oobabooga/text-generation-webui/issues/174 as a base, modified slightly

added small section in readme to explain how to start up, the defaults of this config run with < 4GB of vram

PR #530, opened 2023-03-24
Add support for the latest GPTQ models with group-size

oobabooga:
Example: https://huggingface.co/ozcur/alpaca-native-4bit

Usage:

```
python server.py --model alpaca-native-4bit --wbits 4 --groupsize 128
```

Issue #400, opened 2023-03-18
undefined symbol: cget_col_row_stats / 8-bit not working / libsbitsandbytes_cpu.so not found 


AlexysLovesLexxie:
### Describe the bug

On starting the server, I recieve the following error messages : 

```
===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please submit your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
CUDA SETUP: Required library version not found: libsbitsandbytes_cpu.so. Maybe you need to compile it from source?
CUDA SETUP: Defaulting to libbitsandbytes_cpu.so...
argument of type 'WindowsPath' is not iterable
CUDA SETUP: Required library version not found: libsbitsandbytes_cpu.so. Maybe you need to compile it from source?
CUDA SETUP: Defaulting to libbitsandbytes_cpu.so...
argument of type 'WindowsPath' is not iterable
C:\Oobabooga_new\installer_files\env\lib\site-packages\bitsandbytes\cextension.py:31: UserWarning: The installed version of bitsandbytes was compiled without GPU support. 8-bit optimizers and GPU quantization are unavailable.
  warn("The installed version of bitsandbytes was compiled without GPU support. "
```
 This is not the same as #388.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Start web UI using the supplied batch file.

### Screenshot

![Screenshot 2023-03-18 024607](https://user-images.githubusercontent.com/126999069/226101631-543d0ee8-e359-4e49-b6a4-e1846e16bbe3.png)

### Logs

```shell
None.  See screenshot.
```

### System Info

```shell
Windows 11
No GPU, CPU only
CPU : Ryzen 7 6800H
RAM : 32Gb
```

KirillRepinArt:
I have an exact same issue with Nvidia GPU, tried a new install several times — nothing seems to work. Very frustrating, given the fact that yesterday it worked just fine. Shouldn't have done git pull today, it seem to have broken the UI.

fuomag9:
Same issue on linux as well.

I even tried inside a [11.8.0-runtime-ubuntu22.04](https://hub.docker.com/layers/nvidia/cuda/11.8.0-runtime-ubuntu22.04/images/sha256-61187bc58b1411daa436202bebc96022e9c5339611589a022cd913b1b54cdead?context=explore) Nvidia container

AlexysLovesLexxie:
[...]

Issue #174, opened 2023-03-06
Dockerfile


slush0:
If there's any interest (to use it or add into this repo), I've knocked up Dockerfile.

https://github.com/slush0/docker-misc/blob/master/text-generation-webui/Dockerfile

It is also available at https://hub.docker.com/r/slush0/text-generation-webui.

phokur:
I can't seem to connect to the instance locally once it's up. What is your docker run command?
`docker run --gpus all -p 7860:7860 -v /tmp/.X11-unix:/tmp/.X11-unix -e DISPLAY -it slush0/text-generation-webui`

rkfg:
I think the problem is that there's no entry point or `RUN` line in that Dockerfile so the container terminates immediately.

RedTopper:
If anyone happens to want to use Podman instead, I have a repo here called [Text-Generation-Webui-Podman](https://github.com/RedTopper/Text-Generation-Webui-Podman). It compiles and installs the GPTQ-for-LLaMa repo so 4bit works too.

In theory the Containerfile should be compatible with Docker, but I haven't tested it.

nelsonjchen:
i guess i'll throw my hat in the ring too. This slapdash attempt is/was targeted at running on vast.ai and cheap GPU providers.

https://github.com/nelsonjchen/docker-quick-llama 

It's really stupid bare/bones. No multi-stage. The ecosystem is simply way too volatile to research deeply or care about that stuff for now. I've taught GitHub actions to prebuild all the containers sans models so I can just point vast.ai at it.

tensiondriven:
Does anyone know, is it possible to run multiple instance of this dockerfile using a single GPU?  I'm not able to run multiple instances from bash; when I try, the second task exits with no error:

(with first instance already running)
```
$ python server.py --cai-chat --gptq-bits 4 --model llama-13b --listen-port 8002 --verbose
Loading llama-13b...
Loading model ...
Killed
```

Related to #58

oobabooga:
I am interested in making this the default installation method to avoid all the setup trouble.

Here is what I have tried:

1. Download the [Dockerfile](https://github.com/slush0/docker-misc/blob/master/text-generation-webui/Dockerfile)
2. Build the image with 

```
docker build . -f Dockerfile -t oobabooga
```

3. Teleport into the image with

```
docker run -i -t oobabooga bash
```

4. Download a test model

```
python3 download-model.py facebook/galactica-125m
```

5. Try to start the web UI

[...]

PR #633, opened 2023-03-29
creating a layer with Docker/docker-compose

loeken:
cleaned up version from https://github.com/oobabooga/text-generation-webui/pull/547


oobabooga:
1. Do you know what the Ubuntu equivalents of this are?

> yay -S docker docker-compose buildkit nvidia-container-runtime
> sudo systemctl restart docker # required by nvidia-container-runtime

2. Have you tested the docker image with 4-bit and 8-bit models? It seems like the current bitsandbytes 8-bit setup is not included: https://github.com/oobabooga/text-generation-webui/issues/400#issuecomment-1474876859

loeken:
1.) as far as i know only docker and docker-compose are in the ubuntu/debian repos. buildkit should be part of docker since version 18.09 ( it is a separate package in manjaro ).

- https://github.com/NVIDIA/nvidia-container-toolkit/releases 
- https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html

2.) Due to ram limits i so far only tested 4bit I've followed that link and noticed that im seeing the same errors. I ve now updated the Dockerfile to use the nvidia/cuda image in the second stage too, which provides the required files.
while it crashes with OOM for me with 8bit i dont see the errors so expect that this would solve the problems

logs from execution of 4bit/8bit: https://gist.github.com/loeken/a9c2141adee41a6181c9bbde509a75fe

oobabooga:
I need better instructions on how to run this before I can merge. The average Linux user uses Ubuntu, so the README must contain all the necessary steps to get this running on Ubuntu.

loeken:
alright i ll test on ubuntu 22.04 and get back to you with instructions

loeken:
@oobabooga added a docs/README_docker.md with instructions for ubuntu

deece:
THe following lines in the Dockerfile would clobber any work-in-progress that would be intended for test:
```
ARG WEBUI_SHA=HEAD
RUN git reset --hard ${WEBUI_SHA}
```

loeken:
valid point @deece the branches/shas have changed around a lot lately. I've removed setting values in the Dockerfile and since it switched from being a sha - to now being the cuda branch - i ve changed them to _VERSION.

I ll leave the WEBUI_VERSION on head for now, but I think this should be changed to a tag/release. Right now everybody installs from the last version of this repo 

@oobabooga if you are fine with switching to a tag release flow i can provide a github action to build a docker image

deece:
The cuda branch of GPTQ introduces some breaking changes that need to be addressed. In the meantime, this is the commit before the breaking change: GPTQ_SHA=608f3ba71e40596c75f8864d73506eaf57323c6e

deece:
I'd suggest that the the git reset line be changed to something like:
`RUN test -n "${WEBUI_VERSION}" && git reset --hard "${WEBUI_VERSION}"`

This would allow uncommitted changes to be tested in the Docker environment.

deece:
The args in docker-compose.yml and Dockerfile don't align, eg, GPTQ_SHA vs GPTQ_VERSION

loeken:
made changes as requested

loeken:
@deece as for the breaking change what are you refering to? i can build for both cuda and 608f3ba71e40596c75f8864d73506eaf57323c6e for GPTQ

rklasen:
The docker image builds without problems. I've put `llama-7b-4bit.safetensors` in the `models` directory, but I'm still getting errors about a private HF repo:

```
text-generation-webui-text-generation-webui-1  | /usr/local/lib/python3.10/dist-packages/bitsandbytes/cuda_setup/main.py:136: UserWarning: WARNING: The following directories listed in your path were found to be non-existent: {PosixPath('/usr/local/nvidia/lib64'), PosixPath('/usr/local/nvidia/lib')}
text-generation-webui-text-generation-webui-1  |   warn(msg)
text-generation-webui-text-generation-webui-1  | /usr/local/lib/python3.10/dist-packages/bitsandbytes/cuda_setup/main.py:136: UserWarning: /usr/local/nvidia/lib:/usr/local/nvidia/lib64 did not contain libcudart.so as expected! Searching further paths...
text-generation-webui-text-generation-webui-1  |   warn(msg)
text-generation-webui-text-generation-webui-1  | CUDA_SETUP: WARNING! libcudart.so not found in any environmental path. Searching /usr/local/cuda/lib64...
text-generation-webui-text-generation-webui-1  | CUDA SETUP: CUDA runtime path found: /usr/local/cuda/lib64/libcudart.so
text-generation-webui-text-generation-webui-1  | CUDA SETUP: Highest compute capability among GPUs detected: 8.6
text-generation-webui-text-generation-webui-1  | CUDA SETUP: Detected CUDA version 118
text-generation-webui-text-generation-webui-1  | CUDA SETUP: Loading binary /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cuda118.so...
text-generation-webui-text-generation-webui-1  | Loading llama-7b-4bit...
text-generation-webui-text-generation-webui-1  | Found models/llama-7b-4bit.safetensors
text-generation-webui-text-generation-webui-1  | Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/utils/_errors.py", line 259, in hf_raise_for_status
text-generation-webui-text-generation-webui-1  |     response.raise_for_status()
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/requests/models.py", line 1021, in raise_for_status
text-generation-webui-text-generation-webui-1  |     raise HTTPError(http_error_msg, response=self)
text-generation-webui-text-generation-webui-1  | requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/models/llama-7b-4bit/resolve/main/config.json
text-generation-webui-text-generation-webui-1  | 
text-generation-webui-text-generation-webui-1  | The above exception was the direct cause of the following exception:
text-generation-webui-text-generation-webui-1  | 
text-generation-webui-text-generation-webui-1  | Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/transformers/utils/hub.py", line 409, in cached_file
text-generation-webui-text-generation-webui-1  |     resolved_file = hf_hub_download(
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/utils/_validators.py", line 120, in _inner_fn
text-generation-webui-text-generation-webui-1  |     return fn(*args, **kwargs)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/file_download.py", line 1160, in hf_hub_download
text-generation-webui-text-generation-webui-1  |     metadata = get_hf_file_metadata(
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/utils/_validators.py", line 120, in _inner_fn
text-generation-webui-text-generation-webui-1  |     return fn(*args, **kwargs)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/file_download.py", line 1501, in get_hf_file_metadata
text-generation-webui-text-generation-webui-1  |     hf_raise_for_status(r)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/utils/_errors.py", line 291, in hf_raise_for_status
text-generation-webui-text-generation-webui-1  |     raise RepositoryNotFoundError(message, response) from e
text-generation-webui-text-generation-webui-1  | huggingface_hub.utils._errors.RepositoryNotFoundError: 401 Client Error. (Request ID: Root=1-64283e05-294e106c21f09d494bf5bb13)
text-generation-webui-text-generation-webui-1  | 
text-generation-webui-text-generation-webui-1  | Repository Not Found for url: https://huggingface.co/models/llama-7b-4bit/resolve/main/config.json.
text-generation-webui-text-generation-webui-1  | Please make sure you specified the correct `repo_id` and `repo_type`.
text-generation-webui-text-generation-webui-1  | If you are trying to access a private or gated repo, make sure you are authenticated.
text-generation-webui-text-generation-webui-1  | Invalid username or password.
text-generation-webui-text-generation-webui-1  | 
text-generation-webui-text-generation-webui-1  | During handling of the above exception, another exception occurred:
text-generation-webui-text-generation-webui-1  | 
text-generation-webui-text-generation-webui-1  | Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/app/server.py", line 275, in <module>
text-generation-webui-text-generation-webui-1  |     shared.model, shared.tokenizer = load_model(shared.model_name)
text-generation-webui-text-generation-webui-1  |   File "/app/modules/models.py", line 102, in load_model
text-generation-webui-text-generation-webui-1  |     model = load_quantized(model_name)
text-generation-webui-text-generation-webui-1  |   File "/app/modules/GPTQ_loader.py", line 114, in load_quantized
text-generation-webui-text-generation-webui-1  |     model = load_quant(str(path_to_model), str(pt_path), shared.args.wbits, shared.args.groupsize, kernel_switch_threshold=threshold)
text-generation-webui-text-generation-webui-1  |   File "/app/modules/GPTQ_loader.py", line 19, in _load_quant
text-generation-webui-text-generation-webui-1  |     config = AutoConfig.from_pretrained(model)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/transformers/models/auto/configuration_auto.py", line 908, in from_pretrained
text-generation-webui-text-generation-webui-1  |     config_dict, unused_kwargs = PretrainedConfig.get_config_dict(pretrained_model_name_or_path, **kwargs)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/transformers/configuration_utils.py", line 573, in get_config_dict
text-generation-webui-text-generation-webui-1  |     config_dict, kwargs = cls._get_config_dict(pretrained_model_name_or_path, **kwargs)
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/transformers/configuration_utils.py", line 628, in _get_config_dict
text-generation-webui-text-generation-webui-1  |     resolved_config_file = cached_file(
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/transformers/utils/hub.py", line 424, in cached_file
text-generation-webui-text-generation-webui-1  |     raise EnvironmentError(
text-generation-webui-text-generation-webui-1  | OSError: models/llama-7b-4bit is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
text-generation-webui-text-generation-webui-1  | If this is a private repository, make sure to pass a token having permission to this repo with `use_auth_token` or log in with `huggingface-cli login` and pass `use_auth_token=True`.
text-generation-webui-text-generation-webui-1 exited with code 1
```

I thought once the models are present, everything should be offline?

loeken:
@deece tested only the build which passed earlier, I now started it up and im seeing errors - the gptq hash you provided does solve it. probably better to have fixed versions anyways.

loeken:
@rklasen you are using the "updated" models?

https://github.com/oobabooga/text-generation-webui/pull/530#issuecomment-1483891617
https://github.com/oobabooga/text-generation-webui/pull/530#issuecomment-1483941105

deece:
Apologies, the WEBUI line I provided earlier fails the build if WEBUI_VERSION is not set, here's a version that will always exit with 0:
```
RUN test -n "${WEBUI_VERSION}" && git reset --hard ${WEBUI_VERSION} || echo "Using provided webui source"
```

rklasen:
@loeken thanks for the hashes. I've got:

```
ed8ec9c9f0ebb83210157ad0e3c5148760a4e9fd2acfb02cf00f8f2054d2743b  models/llama-7b-4bit-128g.safetensors
09841a1c4895e1da3b05c1bdbfb8271c6d43812661e4348c862ff2ab1e6ff5b3  models/llama-7b-4bit.safetensors
```

So at least those seem to be correct. I've just checked out the lastest changes of this PR after @deece 's latest commit, still the same error.

What's confusing me is that the model files are found at least:

```
text-generation-webui-text-generation-webui-1  | Loading llama-7b-4bit...
text-generation-webui-text-generation-webui-1  | Found models/llama-7b-4bit.safetensors
```

But there still seems to be some reference to `https://huggingface.co/models/llama-7b-4bit/resolve/main/config.json.` That URL exists, but requires login. Do I need to make a HF account?

deece:
> @loeken thanks for the hashes. I've got:
> 
...
> But there still seems to be some reference to `https://huggingface.co/models/llama-7b-4bit/resolve/main/config.json.` That URL exists, but requires login. Do I need to make a HF account?

You have to have the following files too:
```
$ find models
models
models/llama-30b
models/llama-30b/config.json
models/llama-30b/pytorch_model.bin.index.json
models/llama-30b/generation_config.json
models/llama-30b-4bit.pt
models/place-your-models-here.txt
```

I'd actually prefer it if the 4 bit checkpoints were alongside the the matching tokenizer files

rklasen:
Oh, I didn't know the config files were supposed to go in sub folders. Alright, now I have:

```
$ find models                                    
models
models/llama-7b
models/llama-7b/special_tokens_map.json
models/llama-7b/config.json
models/llama-7b/tokenizer.model
models/llama-7b/tokenizer_config.json
models/llama-7b/7b.log
models/llama-7b/generation_config.json
models/llama-7b-4bit.safetensors
```

I don't have `pytorch_model.bin.index.json`, it seems this wasn't included in the magnet. But with this file structure, the error remains:

```
text-generation-webui-text-generation-webui-1  | Loading llama-7b-4bit...
text-generation-webui-text-generation-webui-1  | Found models/llama-7b-4bit.safetensors
text-generation-webui-text-generation-webui-1  | Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/huggingface_hub/utils/_errors.py", line 259, in hf_raise_for_status
text-generation-webui-text-generation-webui-1  |     response.raise_for_status()
text-generation-webui-text-generation-webui-1  |   File "/usr/local/lib/python3.10/dist-packages/requests/models.py", line 1021, in raise_for_status
text-generation-webui-text-generation-webui-1  |     raise HTTPError(http_error_msg, response=self)
text-generation-webui-text-generation-webui-1  | requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/models/llama-7b-4bit/resolve/main/config.json
```

loeken:
@deece  updated

@rklasen i dont think your issues are related to this PR

mjlbach:
A couple comments:

* I'm trying out this branch, when I actually try to run it there's an issue with the pip installation of `transformers` from HF
```
text-generation-webui-text-generation-webui-1  | Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/app/server.py", line 13, in <module>
text-generation-webui-text-generation-webui-1  |     from modules import chat, shared, training, ui
text-generation-webui-text-generation-webui-1  |   File "/app/modules/chat.py", line 16, in <module>
text-generation-webui-text-generation-webui-1  |     from modules.text_generation import (encode, generate_reply,
text-generation-webui-text-generation-webui-1  |   File "/app/modules/text_generation.py", line 8, in <module>
text-generation-webui-text-generation-webui-1  |     import transformers
text-generation-webui-text-generation-webui-1  | ModuleNotFoundError: No module named 'transformers'
```
* Do you really want to have the final stage of the build be from `nvidia/cuda:11.8.0-devel-ubuntu22.04` rather than `nvidia/cuda:11.8.0-runtime-ubuntu22.04`? The cuda toolchain bloats the container size from ~300 mb to over 6 gb, and shouldn't be required for the runtime libraries.

* I'd recommend distributing a lock-file, either with conda-lock (preferred because it's much faster), poetry, pip-tools, otherwise the container won't be reproducible.

* 
```
ENV NVIDIA_VISIBLE_DEVICES=all
ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility
```
These lines shouldn't be necessary as they're part of the upstream container

mjlbach:
This is the root cause of the issue with transformers https://stackoverflow.com/questions/29905909/pip-install-e-packages-dont-appear-in-docker

TLDR either don't touch workdir until the end (just use absolute paths to the requirements.txt, shortens code anyways), or you can manually pass `--src /usr/local/src` to the pip installation commands.

loeken:
@mjlbach i removed the env vars and replaced devel with runtime for the second stage.

on the failure on transformers i think it would be better to create a seperate issue to change the main requirements.txt in the repo:

git+https://github.com/huggingface/transformers it's risky to point it at the last version.

I think it might be better to define the version to use there, so it solves the problem for both ppl using containers and running it natively

you can edit your requirements.txt to git+https://github.com/huggingface/transformers@c612628045822f909020f7eb6784c79700813eda

and set WEBUI_VERSION="" in your env file so it doesnt do a hard reset to HEAD when building

loeken:
i created an issue to track this: https://github.com/oobabooga/text-generation-webui/issues/753

mjlbach:
I wouldn't say it broke, if you create a virtualenv with the env module and install everything in the resulting venv instead of in `dist-packages`, transformers is still installed correctly, probably something to do with https://github.com/huggingface/transformers/pull/22539

mcsgroi:
I believe there is an issue here around supported versioning. I have a Tesla K80. I attempted to run this branch using the instructions provided which seem to indicate that the CUDA architecture for a K80 would work. That said, I encountered the following error:
```
#0 42.01 nvcc warning : The 'compute_35', 'compute_37', 'sm_35', and 'sm_37' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
#0 51.05 quant_cuda_kernel.cu(654): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.05 quant_cuda_kernel.cu(654): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.05 quant_cuda_kernel.cu(788): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.05 quant_cuda_kernel.cu(788): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.05 quant_cuda_kernel.cu(906): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.05 quant_cuda_kernel.cu(906): error: identifier "__hfma2" is undefined
#0 51.05
#0 51.24 /usr/local/lib/python3.10/dist-packages/torch/include/c10/util/irange.h(54): warning #186-D: pointless comparison of unsigned integer with zero
#0 51.24           detected during:
#0 51.24             instantiation of "__nv_bool c10::detail::integer_iterator<I, one_sided, <unnamed>>::operator==(const c10::detail::integer_iterator<I, one_sided, <unnamed>> &) const [with I=size_t, one_sided=false, <unnamed>=0]"
#0 51.24 (61): here
#0 51.24             instantiation of "__nv_bool c10::detail::integer_iterator<I, one_sided, <unnamed>>::operator!=(const c10::detail::integer_iterator<I, one_sided, <unnamed>> &) const [with I=size_t, one_sided=false, <unnamed>=0]"
#0 51.24 /usr/local/lib/python3.10/dist-packages/torch/include/c10/core/TensorImpl.h(77): here
#0 51.24
#0 51.38 /usr/local/lib/python3.10/dist-packages/torch/include/c10/util/irange.h(54): warning #186-D: pointless comparison of unsigned integer with zero
#0 51.38           detected during:
#0 51.38             instantiation of "__nv_bool c10::detail::integer_iterator<I, one_sided, <unnamed>>::operator==(const c10::detail::integer_iterator<I, one_sided, <unnamed>> &) const [with I=std::size_t, one_sided=true, <unnamed>=0]"
#0 51.38 (61): here
#0 51.38             instantiation of "__nv_bool c10::detail::integer_iterator<I, one_sided, <unnamed>>::operator!=(const c10::detail::integer_iterator<I, one_sided, <unnamed>> &) const [with I=std::size_t, one_sided=true, <unnamed>=0]"
#0 51.38 /usr/local/lib/python3.10/dist-packages/torch/include/ATen/core/qualified_name.h(73): here
#0 51.38
#0 55.33 6 errors detected in the compilation of "quant_cuda_kernel.cu".
#0 55.51 error: command '/usr/local/cuda/bin/nvcc' failed with exit code 1
```

Unless I'm mistaken, this error seems to be a result of an unsupported version (looks to be related to this issue https://github.com/qwopqwop200/GPTQ-for-LLaMa/issues/88). Any thoughts?

mcsgroi:
Just figured it out. That link mentions a specific commit that breaks support for older GPUs. I then noticed that that library was one of the fields that supported being changed in the .env file. I would suggest adding a comment to the .env example file to indicate that older GPUs should still work by using the following version of GPTQ `841feedde876785bc8022ca48fd9c3ff626587e2` based on the Github Issue I mentioned above.

Dockerfile
```diff
@@ -0,0 +1,54 @@
+# GPTQ-for-LLaMa and Text Generation WebUI Dockerfile
+FROM nvidia/cuda:11.8.0-devel-ubuntu22.04 as builder
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y git build-essential python3-dev python3-pip && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN --mount=type=cache,target=/root/.cache/pip pip3 install torch torchvision torchaudio
```
XuehaiPan, line 8:
```suggestion
RUN --mount=type=cache,target=/root/.cache/pip pip3 install --upgrade pip setuptools
RUN --mount=type=cache,target=/root/.cache/pip pip3 install torch torchvision torchaudio
```

Reinstall `pip` to avoid using the system `pip` package. The `pip3` installed by `apt install python3-pip` is placed in `/usr/bin/pip3`. "Upgrade" it with `pip3 install --upgrade pip` will install a new one in `/usr/bin/local/pip3`.

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+FROM nvidia/cuda:11.8.0-runtime-ubuntu22.04
+
+LABEL maintainer="Your Name <your.email@example.com>"
+LABEL description="Docker image for GPTQ-for-LLaMa and Text Generation WebUI"
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y git python3 python3-pip && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN --mount=type=cache,target=/root/.cache/pip pip3 install torch torchvision torchaudio
```
XuehaiPan, line 32:
```suggestion
RUN --mount=type=cache,target=/root/.cache/pip pip3 install --upgrade pip setuptools
RUN --mount=type=cache,target=/root/.cache/pip pip3 install torch torchvision torchaudio
```

Same argument above.

loeken: but this is the second stage

loeken:
@mcsgroi i added a comment with the hash and a link to the PR

loeken:
I have added a venv layer - which does solve installing transformers ( at least for me ) . @mjlbach I went with venv cause poetry etc seem to require a different dependency definition then what is currently used by devs here ( pip ) - i think its better if the dockerfile piggy bags of the existing package management used

Dockerfile
```diff
@@ -0,0 +1,62 @@
+FROM nvidia/cuda:11.8.0-devel-ubuntu22.04 as builder
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y git build-essential python3-dev python3-pip && \
```
XuehaiPan, line 4:
```suggestion
    apt-get install --no-install-recommends -y git build-essential python3-dev python3-venv && \
```

Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+RUN --mount=type=cache,target=/root/.cache/pip pip3 install virtualenv
+
+RUN git clone https://github.com/qwopqwop200/GPTQ-for-LLaMa /build
+
+WORKDIR /build
+
+ARG GPTQ_VERSION
+RUN git checkout ${GPTQ_VERSION}
+
+RUN virtualenv /build/venv
```
XuehaiPan, line 16:
```suggestion
RUN python3 -m venv /build/venv
```

Dockerfile
```diff
@@ -0,0 +1,62 @@
+FROM nvidia/cuda:11.8.0-devel-ubuntu22.04 as builder
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y git build-essential python3-dev python3-pip && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN --mount=type=cache,target=/root/.cache/pip pip3 install virtualenv
+
```
XuehaiPan, line 8:
```suggestion
```

Dockerfile
loeken: what is bad about the system's pip? it seems to add an extra 200 seconds on the buildtime

without upgrading pip:
 => [stage-1  8/15] RUN . /app/venv/bin/activate &&     pip3 install torch torchvision torchaudio &&     pip3   491.5s
 => [builder 7/8] RUN . /build/venv/bin/activate &&     pip3 install torch torchvision torchaudio &&     pip3   484.5s

with upgrading pip:
=> [builder 7/8] RUN . /build/venv/bin/activate &&     pip3 install --upgrade pip set  628.4s
 => [stage-1  8/15] RUN . /app/venv/bin/activate &&     pip3 install --upgrade pip set  684.6s

loeken: what is bad about the system pip? it seems to increase the build time by ~ 200 seconds

mjlbach:
> I have added a venv layer - which does solve installing transformers ( at least for me ) . @mjlbach I went with venv cause poetry etc seem to require a different dependency definition then what is currently used by devs here ( pip ) - i think its better if the dockerfile piggy bags of the existing package management used

My comment is more general to the repo, not specific to this PR. A lockfile based approach (I prefer conda-lock + micromamba to poetry for ML things due to speed) can be used everywhere and simplify the setup.

loeken:
@oobabooga any requirements left from your side?

deece:
I think its worth adding Dockerfile and .env to .dockerignore, otherwise changes to them invalidate the `COPY . /app/` line, resulting in pip having to re-do a fair amount of work each time the image is re-run (eg. to change the model).

Hmm, I wonder if CLI_ARGS should be a runtime argument, so that the image doesn't need to be rebuilt just to change the args?

deece:
While testing Koala I get the following:
```
text-generation-webui_1  | Loading koala_13b_v2-hf...
text-generation-webui_1  | Auto-assiging --gpu-memory 23 for your GPU to try to prevent out-of-memory errors.
text-generation-webui_1  | You can manually set other values.
Loading checkpoint shards: 100% 3/3 [00:19<00:00,  6.54s/it]
text-generation-webui_1  | Traceback (most recent call last):
text-generation-webui_1  |   File "/app/server.py", line 276, in <module>
text-generation-webui_1  |     shared.model, shared.tokenizer = load_model(shared.model_name)
text-generation-webui_1  |   File "/app/modules/models.py", line 176, in load_model
text-generation-webui_1  |     tokenizer = AutoTokenizer.from_pretrained(Path(f"{shared.args.model_dir}/{shared.model_name}/"))
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/models/auto/tokenization_auto.py", line 700, in from_pretrained
text-generation-webui_1  |     return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 1811, in from_pretrained
text-generation-webui_1  |     return cls._from_pretrained(
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 1965, in _from_pretrained
text-generation-webui_1  |     tokenizer = cls(*init_inputs, **init_kwargs)
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/models/llama/tokenization_llama_fast.py", line 74, in __init__
text-generation-webui_1  |     super().__init__(
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_fast.py", line 120, in __init__
text-generation-webui_1  |     raise ValueError(
text-generation-webui_1  | ValueError: Couldn't instantiate the backend tokenizer from one of: 
text-generation-webui_1  | (1) a `tokenizers` library serialization file, 
text-generation-webui_1  | (2) a slow tokenizer instance to convert or 
text-generation-webui_1  | (3) an equivalent slow tokenizer class to instantiate and convert. 
text-generation-webui_1  | You need to have sentencepiece installed to convert a slow tokenizer to a fast one.
```

Adding sentencepiece to the list of pip packages gets me past this.

deece:
The docker images currently fall over with the problem described here: https://github.com/oobabooga/text-generation-webui/issues/829

It may be best to pin a commit for the transformers library

loeken:
I think its worth adding Dockerfile and .env to .dockerignore <- added both to the .dockerignore file. changing CLI_ARGS in your .env file doesnt require you to rebuild the container.

@deece what errors are you getting ? i just build based on latest main which worked, but yeah i do agree with dependency locking, but not in this pr

deece:
@loeken The problem is described here: https://github.com/oobabooga/text-generation-webui/issues/829

This is the error:
```
text-generation-webui_1  | CUDA_SETUP: WARNING! libcudart.so not found in any environmental path. Searching /usr/local/cuda/lib64...
text-generation-webui_1  | /app/venv/lib/python3.10/site-packages/bitsandbytes/cuda_setup/main.py:136: UserWarning: WARNING: No libcudart.so found! Install CUDA or the cudatoolkit package (anaconda)!
text-generation-webui_1  |   warn(msg)
text-generation-webui_1  | CUDA SETUP: Highest compute capability among GPUs detected: 8.6
text-generation-webui_1  | CUDA SETUP: Detected CUDA version 117
text-generation-webui_1  | CUDA SETUP: Loading binary /app/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so...
text-generation-webui_1  | Loading vicuna-13b-4bit-128g...
text-generation-webui_1  | Loading model ...
text-generation-webui_1  | /app/venv/lib/python3.10/site-packages/safetensors/torch.py:99: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
text-generation-webui_1  |   with safe_open(filename, framework="pt", device=device) as f:
text-generation-webui_1  | /app/venv/lib/python3.10/site-packages/torch/_utils.py:776: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
text-generation-webui_1  |   return self.fget.__get__(instance, owner)()
text-generation-webui_1  | /app/venv/lib/python3.10/site-packages/torch/storage.py:899: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
text-generation-webui_1  |   storage = cls(wrap_storage=untyped_storage)
text-generation-webui_1  | Done.
text-generation-webui_1  | Traceback (most recent call last):
text-generation-webui_1  |   File "/app/server.py", line 276, in <module>
text-generation-webui_1  |     shared.model, shared.tokenizer = load_model(shared.model_name)
text-generation-webui_1  |   File "/app/modules/models.py", line 176, in load_model
text-generation-webui_1  |     tokenizer = AutoTokenizer.from_pretrained(Path(f"{shared.args.model_dir}/{shared.model_name}/"))
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/models/auto/tokenization_auto.py", line 700, in from_pretrained
text-generation-webui_1  |     return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 1811, in from_pretrained
text-generation-webui_1  |     return cls._from_pretrained(
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 1965, in _from_pretrained
text-generation-webui_1  |     tokenizer = cls(*init_inputs, **init_kwargs)
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/models/llama/tokenization_llama_fast.py", line 74, in __init__
text-generation-webui_1  |     super().__init__(
text-generation-webui_1  |   File "/app/venv/lib/python3.10/site-packages/transformers/tokenization_utils_fast.py", line 120, in __init__
text-generation-webui_1  |     raise ValueError(
text-generation-webui_1  | ValueError: Couldn't instantiate the backend tokenizer from one of: 
text-generation-webui_1  | (1) a `tokenizers` library serialization file, 
text-generation-webui_1  | (2) a slow tokenizer instance to convert or 
text-generation-webui_1  | (3) an equivalent slow tokenizer class to instantiate and convert. 
text-generation-webui_1  | You need to have sentencepiece installed to convert a slow tokenizer to a fast one.
```

oobabooga:
I have tried following the instructions here https://github.com/loeken/text-generation-webui/blob/dockerize/docs/README_docker.md

The command 

```
docker-compose up --build
```

generated the error

```
WARNING: The TORCH_CUDA_ARCH_LIST variable is not set. Defaulting to a blank string.
WARNING: The GPTQ_VERSION variable is not set. Defaulting to a blank string.
WARNING: The WEBUI_VERSION variable is not set. Defaulting to a blank string.
WARNING: The HOST_PORT variable is not set. Defaulting to a blank string.
WARNING: The CONTAINER_PORT variable is not set. Defaulting to a blank string.
WARNING: The HOST_API_PORT variable is not set. Defaulting to a blank string.
WARNING: The CONTAINER_API_PORT variable is not set. Defaulting to a blank string.
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.text-generation-webui.ports contains an invalid type, it should be a number, or an object
services.text-generation-webui.ports contains an invalid type, it should be a number, or an object
services.text-generation-webui.deploy.resources.reservations value Additional properties are not allowed ('devices' was unexpected)
services.text-generation-webui.ports value [':', ':'] has non-unique elements
```

This PR adds lots of files and has lots of breaking parts. I wonder if it wouldn't be more beneficial to simply include a Dockerfile in the repository.

deece:
@oobabooga It looks like you missed creating the .env file that is mentioned in the instructions

oobabooga:
You are right. With the env file in place the error is another one:

```
$ cp .env.example .env
$ docker-compose up --build
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.text-generation-webui.deploy.resources.reservations value Additional properties are not allowed ('devices' was unexpected)
```

It may be related with the fact that I had to add the `focal` branch in my `/etc/apt/sources.list.d/docker.list` to get it to work on Linux Mint

```
deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu focal stable
```

oobabooga:
My only experience with Docker so far has been on vast.ai, where I used the `pytorch:latest` docker image to easily get the web UI running. I documented the steps here: https://github.com/oobabooga/text-generation-webui/discussions/528

Other people have requested and shown interested in a Dockerfile: https://github.com/oobabooga/text-generation-webui/issues/174

deece:
Could be, the Internet suggests that error is from using an old version of docker-compose.

garrettsutula:
Works great for me, thanks for doing this! IMO should be merged!

deece:
The Dockerfile provided in #174 only provides an interactive environment to play with, it doesn't really provide a service that could be started at boot (for example). Models would be stored within the Docker image, rather than volumes, so the many-gb of models would be lost if the image was rebuilt.

It's also missing many of the pre-reqs and workarounds required for various features.

As someone with some Docker experience, I've reviewed this and I'm happy to +1 this PR. I don't think there is anything here that cannot be justified.

deece:
@loeken To get past the sentencepiece issue, I had to add this:
```
RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && pip3 install sentencepiece protobuf==3.20.1
```

loeken:
@oobabooga I do have a theory on your error, what docker-compose --version are you on?

from the changelog of docker-compose it seems it was added in version 1.28.0 can you confirm you run >= 1.28.0 ?

1.28.0(https://docs.docker.com/compose/release-notes/#1280)
(2021-01-20)

Features
Added support for NVIDIA GPUs through device requests.

loeken:
@deece yeah it's not at a pinned version https://github.com/oobabooga/text-generation-webui/blob/main/requirements.txt shouldnt "hot fix" this in this PR but rather have a seperate PR pinning all dependencies to specific versions.

loeken:
if you installed the docker-compose-plugin you might have ended up with a newer version ( you can try a docker compose up ) without the "-"

tested in focal based ubuntu docker container with your repo:
root@d4eb8c9bf89b:/# docker compose version
Docker Compose version v2.17.2
root@d4eb8c9bf89b:/# docker-compose --version
docker-compose version 1.25.0, build unknown

oobabooga:
@loeken indeed, I have an older version

> docker-compose version 1.25.0, build unknown

Let me see if I can find a way to install the newer version

loeken:
try the "docker compose up" first you might already have it installed in a new version via the plugin

oobabooga:
It was easier to just install the plugin manually with

```
DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
mkdir -p $DOCKER_CONFIG/cli-plugins
curl -SL https://github.com/docker/compose/releases/download/v2.17.2/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
export PATH="$HOME/.docker/cli-plugins:$PATH"
```

Now I'm trying to build the image.

oobabooga:
It worked perfectly. This was pretty neat =)

@loeken Some questions/comments before I merge:

1. Is the `.dockerignore` necessary? Why are these files/folders in particular excluded from being copied, and not others like `characters`?

```
/loras
/models
.env
Dockerfile
```
2. I'll merge the documentation with the PR for the contribution to be accounted and afterwards move it to the wiki.
3. For the one-click installer, I have been using my fork of gptq-for-llama, and I have reproduced this here and removed the gptq commit parameter.

For reference, these parameters were useful to me to launch an interactive shell inside the image:

```
docker-compose up -d
docker-compose exec text-generation-webui bash
docker-compose kill text-generation-webui
```

deece:
@oobabooga Those directories are brought in as volumes instead (so they don't take extra space). The 2 files aren't used within the image, and updating them will invalidate the image unnescessarily.

loeken:
- .env/Dockerfile are added to improve caching ( https://github.com/oobabooga/text-generation-webui/pull/633#issuecomment-1498882017 )
- loras/models contain large files, those are not added to the image itself, but mounted as volumes via docker-compose.

loeken:
you can also use 
```
docker exec -it text-generation-webui_text-generation-webui_1 bash
```

to get an interactive terminal after you started up via docker-compose up

oobabooga:
I see @deece @loeken. Wouldn't it make sense to add all of the mounted folders to the .dockerignore?
```
      - ./characters:/app/characters
      - ./extensions:/app/extensions
      - ./loras:/app/loras
      - ./models:/app/models
      - ./presets:/app/presets
      - ./prompts:/app/prompts
      - ./softprompts:/app/softprompts
      - ./training:/app/training
```

deece:
@loeken here's a patch that reorders the COPYs so that pre-requisites aren't invalidated from the cache just because a source file has been updated.

```
diff --git a/Dockerfile b/Dockerfile
index 334f5a1..f5d38ff 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,8 +33,7 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 RUN --mount=type=cache,target=/root/.cache/pip pip3 install virtualenv
-
-COPY . /app/
+RUN mkdir /app
 
 WORKDIR /app
 
@@ -44,21 +43,33 @@ RUN test -n "${WEBUI_VERSION}" && git reset --hard ${WEBUI_VERSION} || echo "Usi
 RUN virtualenv /app/venv
 RUN . /app/venv/bin/activate && \
     pip3 install --upgrade pip setuptools && \
-    pip3 install torch torchvision torchaudio && \
-    pip3 install -r requirements.txt
+    pip3 install torch torchvision torchaudio
 
 COPY --from=builder /build /app/repositories/GPTQ-for-LLaMa
 RUN . /app/venv/bin/activate && \
     pip3 install /app/repositories/GPTQ-for-LLaMa/*.whl
 
-ENV CLI_ARGS=""
-
+COPY extensions/api/requirements.txt /app/extensions/api/requirements.txt
+COPY extensions/elevenlabs_tts/requirements.txt /app/extensions/elevenlabs_tts/requirements.txt
+COPY extensions/google_translate/requirements.txt /app/extensions/google_translate/requirements.txt
+COPY extensions/silero_tts/requirements.txt /app/extensions/silero_tts/requirements.txt
+COPY extensions/whisper_stt/requirements.txt /app/extensions/whisper_stt/requirements.txt
 RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && cd extensions/api && pip3 install -r requirements.txt
 RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && cd extensions/elevenlabs_tts && pip3 install -r requirements.txt
 RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && cd extensions/google_translate && pip3 install -r requirements.txt
 RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && cd extensions/silero_tts && pip3 install -r requirements.txt
 RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && cd extensions/whisper_stt && pip3 install -r requirements.txt
+#RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && pip3 install sentencepiece protobuf==3.20.1
+#RUN --mount=type=cache,target=/root/.cache/pip . /app/venv/bin/activate && pip3 install git+https://github.com/huggingface/transformers@176ceff91f5e5ff15922715e5a4a4d9f66b92d14
+
+COPY requirements.txt /app/requirements.txt
+RUN . /app/venv/bin/activate && \
+    pip3 install --upgrade pip setuptools && \
+    pip3 install torch torchvision torchaudio && \
+    pip3 install -r requirements.txt
 
 RUN cp /app/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so /app/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so
 
+COPY . /app/
+ENV CLI_ARGS=""
 CMD . /app/venv/bin/activate && python3 server.py ${CLI_ARGS}
```

loeken:
if you want to be able to edit the files in these folders while the container is running then i would add them to the dockerignore ( but also create similar volume: entries in the docker-compose ). I basically excluded the lora/models folder because the files in there are big and i wanted to exclude them from the image itself, to reduce size/build time

loeken:
i havent edited files in the other folders a lot, but might be a good idea if others do

loeken:
want me to add those in as volumes and add to dockerignore too?

oobabooga:
> if you want to be able to edit the files in these folders while the container is running then i would add them to the dockerignore ( but also create similar volume: entries in the docker-compose ). I basically excluded the lora/models folder because the files in there are big and i wanted to exclude them from the image itself, to reduce size/build time

I don't get it. Aren't those folders already mounted when the image is started, so that I can for instance drop a new yaml character in my characters folder and it will appear inside the image?

loeken:
oh sorry sleepy brain, yes we already have those folders mapped as volumes in the docker-compose mounted as volumes. ignore my last statements on this. adding those other files to the .dockerignore wont change much. we only need the entries for the folders containing large files so they dont get added to the image

oobabooga:
Thanks for the PR, @loeken! And for the patience in convincing me of the relevance of docker compose. Also thanks @deece for the feedback.

I will work on improving the documentation in the coming days. If you want to make any further chances to the docker compose files, feel free to submit a new PR.

Just one more question: is it possible to get this running on WSL?

deece:
@oobabooga I don't see any reason why it shouldn't work on WSL

loeken:
@oobabooga yeah does work on windows.

https://github.com/oobabooga/text-generation-webui/issues/874 i ve also raised a separate issue on introducing dependency pinning, which is sort of related to this. pinning all versions would improve stability/experience for the users here.

PR #652, opened 2023-03-30
Simple streaming stop string handling

seijihariki:
Just something I wanted in the UI. (custom stop string definition for notebook/default text generators)

For now, it splits by commas and strips() each one, so spaces and newlines don't really work. Also, it depends on the streaming behavior to detect the stopping strings, so it does not do anything on --no-stream.

I will see what I can do to solve these issues, so will keep this as a draft for now.

image


oobabooga:
This is already handled in chat mode. In default and `--notebook` modes, it seems very niche to want to use this, so I dislike the idea of devoting screen space to it.

PR #654, opened 2023-03-30
Add support for resuming downloads

nikita-skakun:
This pull request adds the ability to resume interrupted downloads by modifying the `get_file` function. The function now uses the HTTP Range header to fetch only the remaining part of a file that wasn't downloaded yet.

With the testing I have done, resuming a partial download still produces the same checksum, (maybe an automatic checksum check would be useful later).

Note: At the moment when running the download script with multiple threads, keyboard interrupts stop individual threads rather then the program itself.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -9,6 +9,7 @@ python download-model.py facebook/opt-1.3b
 import argparse
 import base64
 import datetime
+import hashlib
 import json
 import re
 import sys
@@ -17,7 +18,6 @@ from pathlib import Path
 import requests
 import tqdm
 from tqdm.contrib.concurrent import thread_map
-import hashlib
 
 parser = argparse.ArgumentParser()
 parser.add_argument('MODEL', type=str, default=None, nargs='?')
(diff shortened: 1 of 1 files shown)


oobabooga:
Nice to hear that resuming the download now works. I think that checking the hashsum should be done in a second call to download-model.py, as

1. This is a time consuming process and 
2. The user might want to check the hashsum in the future without re-downloading the files, for instance after moving the files from one drive to another

So the workflow would be something like

```
python download-model.py facebook/opt-30b
python download-model.py facebook/opt-30b --check
```

where the second command just recreates the `huggingface-metadata.txt` file and checks the hashsums.

nikita-skakun:
This suggestion makes sense. I was doing my testing on the tiny models so the checksumming process was quite fast, but with large models this will cause noticeable delays.

oobabooga:
I have done the following test:

1. Truncate an existing model file:

```
cd models/facebook_opt-350m
head -1000 pytorch_model.bin > a ; mv a pytorch_model.bin
```

2. Try downloading it again

```
python download-model.py facebook/opt-350m
```

3. The download doesn't start

```
Downloading the model to models/facebook_opt-350m
~$
```

Any idea why? Am I doing this wrong?

nikita-skakun:
It seems to be because huggingface responds with an incorrect file size for the model files, likely because it splits it up into smaller packets rather then streaming the whole file. Here is a debug printout of the downloaded file size and the reported header file size:
```
Downloading the model to models/facebook_opt-350m
README.md File: 8768 Header: 8768
LICENSE.md File: 11117 Header: 11117
config.json File: 644 Header: 644
special_tokens_map.json File: 441 Header: 441
generation_config.json File: 137 Header: 137
pytorch_model.bin File: 662513657 Header: 1214
merges.txt File: 456318 Header: 456318
tokenizer_config.json File: 685 Header: 685
vocab.json File: 898822 Header: 898822
```

Looking into getting the actual file size in a more reliable way now.

nikita-skakun:
I tried your test on the updated code, and it passes the checksum test after resuming the download.

nikita-skakun:
Would you mind double checking checking that you are using the latest commit. In the error you provided the thread_map has a parameter `verbose=True` but it is now `disable=True`. This parameter disables the main tread from trying to print its own status on top of the threaded progress bars.

oobabooga:
You are right, I was probably using another commit for some reason. I have interrupted and resume a download several times as a test, and the result ended up with the correct sha256sum. Well done!

I made one last change: not write a new huggingface-metadata.txt when checking the file. It's valuable to keep the original download date.

nikita-skakun:
Thanks! That changes makes sense, since the metadata file is not used by the validation checksum code.

Issue #640, opened 2023-03-29
Character gallery non-functional with newer Gradio versions


brandonj60:
### Describe the bug

It looks like the Character Gallery doesn't work with Gradio versions >3.19.0. It seems the 'change' event isn't being picked up by character-menu, so clicking on character icons doesn't do anything.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

To replicate:
pip uninstall gradio
pip install gradio==3.23.0

The character icons no longer do anything when clicked.

To fix:
pip uninstall gradio
pip install gradio==3.19.0

The gallery icons again work by switching the active character.

### Logs

```shell
none
```

### System Info

```shell
Windows 10, Edge browser.
```

ye7iaserag:
Yep, same here, a work around for now is using the character tab

thedarthvader2000:
> Yep, same here, a work around for now is using the character tab

Do yall  know why oobabooga recently started making a bunch of mini commits/changes to the repo? I cant even run the Colab bc of issues like the file names changing ---or im retarded
<img width="912" alt="image" src="https://user-images.githubusercontent.com/125932943/228653501-9aa67882-9fb6-48fa-a82f-cf3330eb6e57.png">

PR #664, opened 2023-03-30
rewrote gallery image to use gradio gallery

theubie:
I couldn't figure out how to fix the javascript magic from the HTML gallery as implemented, so I rewrote the extension to use the gradio Gallery object instead.

I had to use a CSS hack to hide the preview screen, but it _should_ be isolated enough using the elem_id that it won't affect any other gallery/previews someone might implement via extensions down the road.


theubie:
This should address issue #640

oobabooga:
I like this, but I get a little paranoid about gradio messing up the gallery appearance in the future for no reason like they did with the chat widget: https://github.com/gradio-app/gradio/issues/3544#issuecomment-1489679267

theubie:
That's fair.  I'll release this as an external then.

Issue #660, opened 2023-03-30
Characters extension only loads cards for selected model after recent update


tensiondriven:
### Describe the bug

I recently updated from `a04b7cf2643a5950fc215deac7f76b4b31336a81` (about 3 days ago) to head at `f0fdab08d31d4c00e2e15c5871413ce847ca842b`, and found the behaviour of the "Characters" control seems to have changed.

Previously, I could:

1) Start Server
2) Open Browser
3) Expand Characters section
4) Click on a Character

The model and chat history would load, as expected

After updating,

1) Start Server
2) Open Browser
3) Expand Characters section
4) Click on a Character
* The model and character do not load!

To load a character, now i have to:

1) Start Server
2) Open Browser
3) Load a specific character on the Character tab
4) Switch back to the "Text Generation" tab
3) Expand Characters section
4) Click on a Character

Then the model loads.

Is this expected?  It seems like a bug.  If expected, perhaps only available cards should show under characters?

FWIW, I very much prefer the old behaviour, as it allows me to test prompts among different characters/contexts more easily.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

0) git pull (I'm at `f0fdab08d31d4c00e2e15c5871413ce847ca842b`)
1) Start Server
2) Open Browser
3) Expand Characters section
4) Click on a Character
* The model and character do not load!

### Screenshot

No screenshot necessary for this one

### Logs

```shell
No logs to show
```

### System Info

```shell
n/a
```

bmoconno:
This seems to be a duplicate of #640

I don't know if anyone is currently working on it.

brandonj60:
Someone made an open PR to fix it here
https://github.com/oobabooga/text-generation-webui/pull/664

PR #673, opened 2023-03-31
Implement character gallery using Dataset

ye7iaserag:
Here is an implementation for character gallery using Dataset element
Tried using gradio Gallery but we'd need to hijack the click action for that to work since a click triggers image enlarging in a gallery element.

Also this implementation allows more customization since it's depending on sub gradio HTML elements

There is another pull https://github.com/oobabooga/text-generation-webui/pull/664 that is using a gradio gallery but is using a hack to disable the image enlarge action

This solves https://github.com/oobabooga/text-generation-webui/issues/640 and https://github.com/oobabooga/text-generation-webui/issues/660

### DIFF
--- a/css/chat.css
+++ b/css/chat.css
@@ -36,7 +36,3 @@ div.svelte-362y77>*, div.svelte-362y77>.form>* {
 .wrap.svelte-6roggh.svelte-6roggh {
   max-height: 92.5%;
 }
-
-.gallery-item.svelte-13hsdno.svelte-13hsdno.svelte-13hsdno {
-  border: 0 !important;
-}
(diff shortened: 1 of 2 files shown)


ye7iaserag:
I've also tried to do a hacky solution using gradio HTML element but there is no way to trigger the `change` event programmatically on HTML elements right now

oobabooga:
This is almost there. The only issue I have found is that if the character has a very long name with lots of words, its image will become vertically misaligned with the rest of the images in the same row.

ye7iaserag:
Done
also there is no need for:
```
.gallery-item.svelte-13hsdno.svelte-13hsdno.svelte-13hsdno {
  border: 0 !important;
}
```
after adding `display: contents;` to the button element, all styles from the button will be dropped so you can safely remove it

oobabooga:
Well done! It's working perfectly now. 

I would never have thought of this on my own. Thanks a lot for the PR.

ye7iaserag:
Thanks a lot for everything man, Respect!

PR #711, opened 2023-04-01
Character Tab Update: "character greeting" displayed and editable on the fly

OWKenobi:
This commit combines two updates of the character tab and syncs the abilities of the character.json-files with the ui:

**UI Update:**
- a new field "Character Greeting" displays the "char_greeting" from the character file, and it is also possible to change it on the fly; "Clear History" will respect the changes, so you can experiment with different greetings right from the UI.

**character.json update:**
- new optional field "your_name" to be able to change the "Your name" textfield in the UI, especially useful for "Alpaca", so you don't have to change your name manually every time.

**Bugfix:**
- fixed "none"-Character being unable to load due to recent settings changes.


mcmonkey4eva:
Several recent changes have modified the same sections of code, notably the YAML support update, It'll need to be basically redone to merge properly.

OWKenobi:
I can rewrite the part of code, hold on. @oobabooga , any feedback from your side while I am about editing everything?

Issue #666, opened 2023-03-30
Input should automatically clear after sending message in chat mode


eiery:
**Description**

In chat mode currently the text in the input box is retained after I press Enter on my keyboard or hit the Generate key. I have to delete the text manually to continue the conversation. The box should clear itself like any other chat app after the message is sent to the AI.

![Screenshot from 2023-03-30 17-41-28](https://user-images.githubusercontent.com/19350831/228971037-8ef6582d-f300-483f-98d2-5f12e84a195d.png)

This should be a pretty simple fix and I can even have a go at it assuming this behavior isn't intentional.

oobabooga:
It does clear after the message is generated. If the event for clearing the input field happens before the text generation event, the latter will receive an empty string as input.

If you can find a clean workaround, feel free to submit a PR. I agree that it's weird for the sent message to not disappear immediately.

eiery:
Ah, I see it now. I always manually cleared the box while the message was generating so I never saw it clear itself.

bmoconno:
I just put in a PR (#715)  that solves this issue, I tested it, but it's work checking yourselves.

eiery:
The PR is working well for me and I see it's merged already. Thanks!

PR #715, opened 2023-04-02
Clear text input for chat

bmoconno:
Add logic to clear the textbox for chat input when the user submits or hits the generate button.

This is an attempt to solve issue #666 

Since the `click` and `submit` events are added to the event queue, the only way to clear the `shared.gradio['textbox']` that I could think of was to save the text currently in `shared.gradio['textbox']` to a new `shared.gradio['Chat input']` variable. We can then use that new `shared.gradio['Chat input']` for text generation, and safely clear out the  `shared.gradio['textbox']`


oobabooga:
This is a very good idea. I have changed the textbox to a hidden `gr.State()` variable.

PR #720, opened 2023-04-02
Add telegram bot extension

akatsukinoyami:
Allow changing model and character from installed

Cai-chat based chatting with AI

Will be updated to use inline keyboards.


mcmonkey4eva:
This and #725 were both recently submitted for the same goal. The other PR looks to be much simpler and cleaner. Is there any reason to prefer this PR over the other one?

oobabooga:
This is very interesting, but I will not be able to maintain this extension if it gets merged. I encourage you to upload it to its own GitHub repository. If you do that, I can link it in the extensions page in the wiki:

https://github.com/oobabooga/text-generation-webui/wiki/Extensions

PR #723, opened 2023-04-02
Not sure if I am the only one with this problem. Chat and Cai Chat didn't load.

Ph0rk0z:
Cai Chat caused it to look for a prompt from the folder, even without a lora. 

```
  File "/home/mint/text-generation-webui/server.py", line 284, in <module>
    default_text = load_prompt(shared.settings['prompts'][next((k for k in shared.settings['prompts'] if re.match(k.lower(), shared.model_name.lower())), 'default')])
  File "/home/mint/text-generation-webui/server.py", line 141, in load_prompt
    with open(Path(f'prompts/{fname}.txt'), 'r', encoding='utf-8') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'prompts/Common sense questions and answers\n\nQuestion: \nFactual answer:.txt'
```

### DIFF
--- a/server.py
+++ b/server.py
@@ -280,6 +280,8 @@ if shared.args.lora:
 default_preset = shared.settings['presets'][next((k for k in shared.settings['presets'] if re.match(k.lower(), shared.model_name.lower())), 'default')]
 if shared.lora_name != "None":
     default_text = load_prompt(shared.settings['lora_prompts'][next((k for k in shared.settings['lora_prompts'] if re.match(k.lower(), shared.lora_name.lower())), 'default')])
+elif (shared.args.cai_chat or shared.args.chat):
+    default_text = ''
 else:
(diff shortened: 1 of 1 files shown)


bmoconno:
No need to update code, your `settings.json` is out of date. Copy and paste over the existing `presets`, `prompts`, and `lora_prompts`. This happened because of [this](https://github.com/oobabooga/text-generation-webui/commit/55755e27b9ddb98e48d5b2dba08c3fb728d7d680) commit, which changed the `settings.json` to use txt files instead of hardcoded prompts.

oobabooga:
I'm not a fan of changes that break backward compatibility, but this one was for the best.

Ph0rk0z:
Oh wow. ooops. Did not notice was just settings.

Also had the problem of initial lora prompts forced into cai-chat when loading one, but did not try to reproduce that recently. Not sure if this would have fixed it too, but it might have.

PR #731, opened 2023-04-02
NSFW toggle, better context and character json

GuizzyQC:
To the SD_Api_Pictures extension, added a NSFW toggle to avoid NSFW results when unwanted, added to parser the ability to understand non-self / non-environment objects and also added the ability to add tags to SD prompt loaded from the "for_sd" json tag, this allows characters to keep their appearance even if they're asked to describe something else than only themselves.

### DIFF
--- a/css/html_instruct_style.css
+++ b/css/html_instruct_style.css
@@ -25,9 +25,7 @@
 .message-body {}
 
 .message-body p {
-    margin-bottom: 0 !important;
     font-size: 15px !important;
-    line-height: 1.428571429 !important;
 }
 
 .message-body li {


mcmonkey4eva:
This... seems like it's just a complicated way of toggling whether to add `nsfw` to the negative prompt?

It also will conflict with https://github.com/oobabooga/text-generation-webui/pull/596 , a more general update to the SD API extension created by the same developer that originally made it. You might want to talk with @Brawlence to coordinate any improvements you're wanting.

Brawlence:
Experience indicates that having "sfw" / "nsfw" tags in (negative) prompt does not automatically guarantee the intended result. It influences the generation in the same way any other tag would, so I see no reason to make a separate checkmark for them when one can simply include it in prompt prefix.

I like the proposed description request change for the interactive mode, here's mine so far: https://github.com/oobabooga/text-generation-webui/pull/596/files#diff-3918965e2a344d3e5543871949fc3c01353cf9e542d0d913c96f697a66de871aR103

I plan on isolating this logic for description requests eventually, so if you could re-write it as a function, that would help greatly.

Same goes for parsing additional tags in character json; that's a brilliant idea, but I think it should be implemented in a separate helper function and not be done in `get_SD_pictures()` directly

Brawlence:
@GuizzyQC on the topic of loading the character features, there's an established format for description in character JSONs for TavernAI / KoboldAI, [ex. 1](https://files.catbox.moe/xm9db2.json), [ex. 2](https://files.catbox.moe/0msv8k.json) : 

It goes like
`Appearance(about 7 ft+well built+fairly slender+pale+free flowing gorgeous long red hair reaches down her back+lips red like her hair+small perky aristocratic breasts)`

Are you planning to support it or is your JSON syntax unrelated to that?

PR #740, opened 2023-04-03
Fix some old JSON characters not loading

mcmonkey4eva:
Fixes bug introduced @ https://github.com/oobabooga/text-generation-webui/pull/337

tldr: if a `json` file has a tab in it, it breaks. With this PR it no longer breaks as it uses the JSON parser for `json` files, and YAML parser for `yaml`/`yml` files. This should also preemptively fix any other format edgecases that might exist between the two.

I also improved the Example yaml character file format as a secondary commit. Hopefully you'll agree with me that it looks a lot cleaner this way, and a lot friendlier to use as a reference for new users.
Specifically it uses multiline-block syntax (that lil `|-`) to automatically remove the need for quotes and to make a single newline count as a newline (to avoid the spam of double newlines). Also use double quotes `"` in the `context` key to avoid the need for that awkward `''` double-singlequote escape syntax.

### DIFF
--- a/characters/Example.yaml
+++ b/characters/Example.yaml
@@ -1,32 +1,16 @@
 name: Chiharu Yamada
-context: 'Chiharu Yamada''s Persona: Chiharu Yamada is a young, computer engineer-nerd
-  with a knack for problem solving and a passion for technology.'
-greeting: '*Chiharu strides into the room with a smile, her eyes lighting up
-  when she sees you. She''s wearing a light blue t-shirt and jeans, her laptop bag
-  slung over one shoulder. She takes a seat next to you, her enthusiasm palpable in
-  the air*
-
-  Hey! I''m so excited to finally meet you. I''ve heard so many great things about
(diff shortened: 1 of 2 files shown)


oobabooga:
You are absolutely right that it's better to load json with the json loader and yaml with the yaml loader. I did not realize that there could be conflicts because all my characters just worked.

I reverted the change to the yaml example because the `vim` syntax highlight didn't work properly with this new formatting.

mcmonkey4eva:
... *vim*?!

oobabooga:
It's all I use :P

TheTerrasque:
Bless your heart 😄 

But on a more serious note, the example character should have the more newbie friendly layout, I'm sure people who has found out how to use vim can handle some lack of syntax highlighting.

TheTerrasque:
yaml has many ways to handle blocks of text, actually: https://yaml-multiline.info/

mcmonkey4eva:
Yay!
 #770

PR #743, opened 2023-04-03
 Character Tab Update: "character greeting" displayed and editable on the fly #711 (conflict resolved)

OWKenobi:
This _updated_ commit combines two updates of the character tab and syncs the abilities of the character.json-files with the ui:

**UI Update:**
-  a new field "Character Greeting" displays the "char_greeting" from the character file, and it is also possible to change it on the fly; "Clear History" will respect the changes, so you can experiment with different greetings right from the UI.

**character.json update:**
-  new optional field "your_name" to be able to change the "Your name" textfield in the UI, especially useful for "Alpaca", so you don't have to change your name manually every time.

**Bugfix:**

- fixed "none"-Character being unable to load due to recent settings changes.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -334,7 +334,7 @@ def load_character(character, name1, name2):
 
     if character != 'None':
         for extension in ["yml", "yaml", "json"]:
-            filepath = Path(f'characters/{_character}.{extension}')
+            filepath = Path(f'characters/{character}.{extension}')
             if filepath.exists():
                 break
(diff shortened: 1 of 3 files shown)


oobabooga:
I have made some changes/simplifications and everything seems to be expected, except the "None" character greeting is not shown in the chat box when it is selected.

Crocky-101:
"your-name" seems to be broken, can't get the interface to load anything but "You" every time. Or is `"your_name": "name"` incorrect?

PR #744, opened 2023-04-03
Make every parameter of the generate_reply function parameterizable

nielstron:
Especially the seed parameter is an important parameter to be set


oobabooga:
Is it possible to make the `stopping_strings` parameter customizable? Some people have requested that.

nielstron:
I have added the field stopping strings to be customizable as well

PR #751, opened 2023-04-03
Add file api-example.sh

StefanDanielSchwarz:
Converted and expanded the api-example.py into a bash script that takes preset (optional) and prompt as arguments. More than just an example, this can be used in shell scripts and loops to automate text generation.

Simple example:

preset='NovelAI-Sphinx Moth'
prompt=$'This is a conversation with your Assistant. The Assistant is very helpful and is eager to chat with you and answer your questions.\nYou: Explain quantum computing in simple terms\nAssistant:' ./api-example.sh "$preset" "$prompt"

### DIFF
--- a/api-example.sh
+++ b/api-example.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+# Make sure to start the web UI with the following flags:
+#
+# python server.py --model MODEL --listen --no-stream
+
+# If no arguments were passed, display a usage message and exit with an error code
+if [[ "$#" -eq 0 ]]; then
+    echo "Usage: $0 [preset] prompt"
+    exit 1
+fi
+
+# Set LANG variable to ensure consistent language environment across different systems
+LANG=
+
+# Load preset
+if [[ -e "presets/$1.txt" ]]; then
+    source "presets/$1.txt"
+    shift
+fi
+
+# Set generation parameters
+prompt=$(jq -n --arg prompt "$*" '$prompt')
+max_new_tokens=${max_new_tokens:-200}
+do_sample=${do_sample:-true}
+temperature=$(printf "%.2f" "${temperature:-0.5}")
+top_p=$(printf "%.2f" "${top_p:-0.9}")
+typical_p=$(printf "%.2f" "${typical_p:-1}")
+repetition_penalty=$(printf "%.2f" "${repetition_penalty:-1.05}")
+encoder_repetition_penalty=$(printf "%.2f" "${encoder_repetition_penalty:-1.0}")
+top_k=${top_k:-0}
+min_length=${min_length:-0}
+no_repeat_ngram_size=${no_repeat_ngram_size:-0}
+num_beams=${num_beams:-1}
+penalty_alpha=$(printf "%.2f" "${penalty_alpha:-0}")
+length_penalty=$(printf "%.2f" "${length_penalty:-1}")
+early_stopping=${early_stopping:-false}
+seed=${seed:--1}
+
+# Encode data as JSON string
+data="{
+  \"data\": [
+    $(jq -n --arg prompt "$*" '$prompt'),
+    $max_new_tokens,
+    ${do_sample,,},
+    $temperature,
+    $top_p,
+    $typical_p,
+    $repetition_penalty,
+    $encoder_repetition_penalty,
+    $top_k,
+    $min_length,
+    $no_repeat_ngram_size,
+    $num_beams,
+    $penalty_alpha,
+    $length_penalty,
+    ${early_stopping,,},
+    $seed
+  ]
+}"
+
+# Make HTTP request and extract generated text
+response=$(curl -s -X POST "http://${server:-127.0.0.1}:7860/run/textgen" \
+    -H "Content-Type: application/json" \
+    -d "$data")
+
+reply=$(jq -r '.data[0]' <<< "$response")
+echo "$reply"


api-example.sh
```diff
@@ -0,0 +1,68 @@
...
+    $num_beams,
+    $penalty_alpha,
+    $length_penalty,
+    ${early_stopping,,},
+    $seed
+  ]
+}"
+
+# Make HTTP request and extract generated text
+response=$(curl -s -X POST "http://${server:-127.0.0.1}:7860/run/textgen" \
```
tensiondriven, line 63:
```
${server:-127.0.0.1}
```
This is a little hard to read for me, but my bash isn't great.

I don't see `server` defined anywhere, should it be?  If this could be more explicit, i think it would be more readble.

api-example.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+# Make sure to start the web UI with the following flags:
+#
+# python server.py --model MODEL --listen --no-stream
```
tensiondriven, line 5: Can we add a `--help` flag that will display this to the user?  Also, can we include the server ip address and port as arguments?  (or event "base url")?  I often run on a specific port, and I can easily see copying this script to a different server and using it to perform inference that way, so having a little extra config and a little extra help would be super useful.

api-example.sh
tensiondriven: This is great - I've been looking for a way to run inference independently of the webui.  I think this is going to need to be extended pretty quickly, to allow for different character loading etc, but for now, super!  The port and server IP are important things to include (imo)

Also, I find the filename to be a little confusing... Can we call it `api-usage-example`?  there are other files callled `api-example` which are involved in standing up the api, but this is using that api.  Another idea for the name might be `inference.sh` or something like that.

When I tried working with the `/run/textgen` endpoint, i had some trouble, so having this available is fantastic.

StefanDanielSchwarz: About the name: Since I created it by converting the existing `api-example.py` to bash, I kept the name, but you're right and it's more than just an example by now. I think it should start with `api-` to show it's relation to the other api-scripts. How about simply `api-bash-script`?

About the server variable: It uses an existing "server" variable and defaults to 127.0.0.1 when undefined. You can set it by prefixing it to the script call (just the way bash works), e. g.:

```
server=127.0.0.1 bash ./api-example.sh "$preset" "$prompt"
```

The word `bash` can be left out if you made the script executable (`chmod +x api-example.sh`).

StefanDanielSchwarz: This has now been resolved by renaming the script and giving usage information and examples when it's run without arguments.

api-example.sh
StefanDanielSchwarz: This has now been resolved by renaming the script and giving usage information and examples when it's run without arguments.

oobabooga:
This is an interesting idea, but it's probably broken after I refactored the API yesterday. Now it accepts a single string input representing serialized JSON. Is is defined here:

https://github.com/oobabooga/text-generation-webui/blob/main/modules/api.py

StefanDanielSchwarz:
Thanks for the heads-up. I'll look into it tomorrow and make sure the script will be compatible with the new API again.

StefanDanielSchwarz:
@oobabooga I've tried to update the bash script, but somehow the curl call isn't working, even if I pass the same JSON as the Python script does:

Echoing the JSON from api-example.py, this is what the Python script sends to the server:

```sh
{'data': ['["What I would like to say is the following: ", {"max_new_tokens": 200, "do_sample": true, "temperature": 0.5, "top_p": 0.9, "typical_p": 1, "repetition_penalty": 1.05, "encoder_repetition_penalty": 1.0, "top_k": 0, "min_length": 0, "no_repeat_ngram_size": 0, "num_beams": 1, "penalty_alpha": 0, "length_penalty": 1, "early_stopping": false, "seed": -1}]']}
```

Sending that same JSON using curl, I get this error message:

```sh
$ curl -X POST "http://${server:-127.0.0.1}:${port:-7860}/run/textgen" \
    -d "{'data': ['[\"What I would like to say is the following: \", {\"max_new_tokens\": 200, \"do_sample\": true, \"temperature\": 0.5, \"top_p\": 0.9, \"typical_p\": 1, \"repetition_penalty\": 1.05, \"encoder_repetition_penalty\": 1.0, \"top_k\": 0, \"min_length\": 0, \"no_repeat_ngram_size\": 0, \"num_beams\": 1, \"penalty_alpha\": 0, \"length_penalty\": 1, \"early_stopping\": false, \"seed\": -1}]']}"

{"detail":[{"loc":["body",1],"msg":"Expecting property name enclosed in double quotes: line 1 column 2 (char 1)","type":"value_error.jsondecode","ctx":{"msg":"Expecting property name enclosed in double quotes","doc":"{'data': ['[\"What I would like to say is the following: \", {\"max_new_tokens\": 200, \"do_sample\": true, \"temperature\": 0.5, \"top_p\": 0.9, \"typical_p\": 1, \"repetition_penalty\": 1.05, \"encoder_repetition_penalty\": 1.0, \"top_k\": 0, \"min_length\": 0, \"no_repeat_ngram_size\": 0, \"num_beams\": 1, \"penalty_alpha\": 0, \"length_penalty\": 1, \"early_stopping\": false, \"seed\": -1}]']}","pos":1,"lineno":1,"colno":2}}]}
```

Expecting property name enclosed in double quotes? OK, so I changed the single quotes around data to double quotes:

```sh
$ curl -X POST "http://${server:-127.0.0.1}:${port:-7860}/run/textgen" \
    -d "{\"data\": ['[\"What I would like to say is the following: \", {\"max_new_tokens\": 200, \"do_sample\": true, \"temperature\": 0.5, \"top_p\": 0.9, \"typical_p\": 1, \"repetition_penalty\": 1.05, \"encoder_repetition_penalty\": 1.0, \"top_k\": 0, \"min_length\": 0, \"no_repeat_ngram_size\": 0, \"num_beams\": 1, \"penalty_alpha\": 0, \"length_penalty\": 1, \"early_stopping\": false, \"seed\": -1}]']}"

{"detail":[{"loc":["body",10],"msg":"Expecting value: line 1 column 11 (char 10)","type":"value_error.jsondecode","ctx":{"msg":"Expecting value","doc":"{\"data\": ['[\"What I would like to say is the following: \", {\"max_new_tokens\": 200, \"do_sample\": true, \"temperature\": 0.5, \"top_p\": 0.9, \"typical_p\": 1, \"repetition_penalty\": 1.05, \"encoder_repetition_penalty\": 1.0, \"top_k\": 0, \"min_length\": 0, \"no_repeat_ngram_size\": 0, \"num_beams\": 1, \"penalty_alpha\": 0, \"length_penalty\": 1, \"early_stopping\": false, \"seed\": -1}]']}","pos":10,"lineno":1,"colno":11}}]}
```

Expecting value? Without further information, I don't know what else to do, but the textgen console showns no messages and the JSON looks just like the one you send using the Python script.

I'm stumped right now after trying various ways of quoting for a couple of hours. If you or anyone else has an idea?

PR #757, opened 2023-04-04
Extract the Preprocessing for Silero into a file and Improve it

da3dsoul:
This improves the Silero preprocessing to resolve many issues with numbers, roman numerals, and abbreviations. It's not perfect, but it's better in *almost* every situation I could find.

#### Main changes:
- Numbers are no longer silent
- Acronyms shorter than 5 letters are spelled out instead of spoken. Some exceptions are weird, like NASA, but it's better than 450 BC being said as b'ck
- Roman Numerals longer than 1 character are handled, the limit was placed to avoid converting i to one. This could be changed with an explicit exception for i added
- Negative numbers have the word "negative" applied before it, preventing the negative from being silent
- Number ranges, such as "1-5", are expanded to "one to five"
- num2words is added as a dependency to make a lot of the number expansion possible

### DIFF
--- a/extensions/silero_tts/test_tts.py
+++ b/extensions/silero_tts/test_tts.py
@@ -45,6 +45,8 @@ def load_model():
     model, example_text = torch.hub.load(repo_or_dir='snakers4/silero-models', model='silero_tts', language=params['language'], speaker=params['model_id'])
     model.to(params['device'])
     return model
+
+
 model = load_model()
 
 
--- a/extensions/silero_tts/test_tts.py
+++ b/extensions/silero_tts/test_tts.py
@@ -1,0 +1,79 @@
+import time
+from pathlib import Path
+
+import torch
+
+import tts_preprocessor
+
+torch._C._jit_set_profiling_mode(False)
+
+params = {
+    'activate': True,
+    'speaker': 'en_49',
+    'language': 'en',
+    'model_id': 'v3_en',
+    'sample_rate': 48000,
+    'device': 'cpu',
+    'show_text': True,
+    'autoplay': True,
+    'voice_pitch': 'medium',
+    'voice_speed': 'medium',
+}
+
+current_params = params.copy()
+voices_by_gender = ['en_99', 'en_45', 'en_18', 'en_117', 'en_49', 'en_51', 'en_68', 'en_0', 'en_26', 'en_56', 'en_74', 'en_5', 'en_38', 'en_53', 'en_21', 'en_37', 'en_107', 'en_10', 'en_82', 'en_16', 'en_41', 'en_12', 'en_67', 'en_61', 'en_14', 'en_11', 'en_39', 'en_52', 'en_24', 'en_97', 'en_28', 'en_72', 'en_94', 'en_36', 'en_4', 'en_43', 'en_88', 'en_25', 'en_65', 'en_6', 'en_44', 'en_75', 'en_91', 'en_60', 'en_109', 'en_85', 'en_101', 'en_108', 'en_50', 'en_96', 'en_64', 'en_92', 'en_76', 'en_33', 'en_116', 'en_48', 'en_98', 'en_86', 'en_62', 'en_54', 'en_95', 'en_55', 'en_111', 'en_3', 'en_83', 'en_8', 'en_47', 'en_59', 'en_1', 'en_2', 'en_7', 'en_9', 'en_13', 'en_15', 'en_17', 'en_19', 'en_20', 'en_22', 'en_23', 'en_27', 'en_29', 'en_30', 'en_31', 'en_32', 'en_34', 'en_35', 'en_40', 'en_42', 'en_46', 'en_57', 'en_58', 'en_63', 'en_66', 'en_69', 'en_70', 'en_71', 'en_73', 'en_77', 'en_78', 'en_79', 'en_80', 'en_81', 'en_84', 'en_87', 'en_89', 'en_90', 'en_93', 'en_100', 'en_102', 'en_103', 'en_104', 'en_105', 'en_106', 'en_110', 'en_112', 'en_113', 'en_114', 'en_115']
+voice_pitches = ['x-low', 'low', 'medium', 'high', 'x-high']
+voice_speeds = ['x-slow', 'slow', 'medium', 'fast', 'x-fast']
+
+# Used for making text xml compatible, needed for voice pitch and speed control
+table = str.maketrans({
+    "<": "&lt;",
+    ">": "&gt;",
+    "&": "&amp;",
+    "'": "&apos;",
+    '"': "&quot;",
+})
+
+
+def xmlesc(txt):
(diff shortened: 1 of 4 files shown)


da3dsoul:
One problem I haven't looked into is number locales. I'm testing with LLaMA, which seems to favor American locale. I'm not sure how num2words handles decimals. It doesn't like thousands' separators in general, so commas are removed in this code, but the comma and decimal point are flipped in some locales. Detecting and converting it to the American locale may be needed.

da3dsoul:
Both the 123,456.78910 and 123.456,78910 formats are handled now.

da3dsoul:
This does not handle currency. According to https://github.com/savoirfairelinux/num2words/issues/267, that would be kind of difficult. We would need to make a mapping of currency symbols to country, know where the symbol goes ($1 vs 12¢), detect situations where the symbol isn't a unique currency symbol, etc. For speaking the currency, it would matter less if it's exactly correct. For example, $12 USD and $12 CAD may be different, but they are both "twelve dollars".

da3dsoul:
I'm going to call this done from my side. Currency is a job by itself. I can maybe look into it later. I'd rather not have this go stale while I research it.

extensions/silero_tts/test_tts.py
da3dsoul, line 1: Not that I really care what is proper for Python, but the generally preferred formatting is:
- 1 newline after a reduction in indentation
- 2 newlines after functions and classes end their scope

da3dsoul: An example for the first is:
```py
if index == 2:
    index = 12

print(index)
```

oobabooga: That's a good rule of thumb. I guess it's best to add these additional spaces. I have checked a couple of Python projects that I like and they all follow this style.

oobabooga:
I'm still testing/trying to understand everything. At first glance, it seems to work well, and it's great to be able to hear the numbers.

Some comments:

* There are some edge cases like "ASP.NET", "C#", and "HTTP/2" aren't caught by the pre-processing.
* "VB" is spoken as "vii bee".
* "HTTP" is spoken as "HTTPy" instead of "HTTPee". Similarly, TCP becomes "TCPy"
* JPEG sounds like "J.P.I.G"
* Numbers like 4.6 and 2007 are spoken perfectly.

da3dsoul:
- HTTP/2 could be fixed by adding / to punctuation list.
- VB should be Vii Bee? I'm not sure if you were just commenting on it or saying that it's a problem.
- it sounds like it didn't like my phonetic translation for P. I can try to tweak that some more.
- JPEG might be affected by the P not working right. I noted in the OP that acronyms would be weird in some cases. It doesn't help that people can't even agree on a lot of them. GIF is the infamous one. For JPEG, I say it Jay Peg, but it's not going to get that right. I added a code comment that in the future, it might be worth trying to find a data source for abbreviation phonetics. There might be a database somewhere, but I'm not aware of one.

da3dsoul:
So to fix:
- P
- Add / to the punctuation regexes

oobabooga:
> VB should be Vii Bee? I'm not sure if you were just commenting on it or saying that it's a problem.

I think that it sounded weird. Wouldn't `vee` for V sound more natural?

da3dsoul:
> I think that it sounded weird. Wouldn't `vee` for V sound more natural?
I can play with that a bit, too.

da3dsoul:
Ok, @oobabooga, it's ready to look at again

da3dsoul:
For reference, I used "VB Script is a scripting language used in Excel. It can perform HTTP requests. HTTP/2 is the standard for web requests. JPEG images are commonly downloaded with it. GIF Images are also common. ENV is short for environment." as a test.

oobabooga:
Looks good to me now. Just one more thing, I have tried adding the characters `“”` to the punctuation string but that caused HTTP to be spoken as a word for some reason. Do you have any idea how to include those characters? It's not rare for them to appear in LLaMA.

```
punctuation = r'[\s,.?!/)"\'\]>]“”'
```

da3dsoul:
It's the Unicode open and closed quotation marks, right? I would just use a string replace to turn them into normal quotes before the other filters are done

extensions/silero_tts/tts_preprocessor.py
```diff
@@ -2,6 +2,7 @@
 
 from num2words import num2words
 
+#punctuation = r'[\s,.?!/)"\'\]>]“”'
```
da3dsoul, line 5: It needs to be inside the right-most `]`

da3dsoul: That's a regex `[ ]` meaning any one of the characters inside the brackets

da3dsoul: I'm guessing you don't use it. Pycharm Community is free, and it has a regex checker built into the editor, along with many other tools, like fixing the whitespace issue from before.

oobabooga: Eh, I hadn't noticed it was a regex.

da3dsoul: https://www.jetbrains.com/webstorm/guide/tips/check-regex/

da3dsoul: > Eh, I hadn't noticed it was a regex.

Fair

da3dsoul: I just checked while I was fixing some other things I found, and those fancy quotes were already replaced with empty strings. I changed and added some more fancy quotes to a filter with the unicode literals, which makes them easier to tell apart right next to each other.
The changes you had are in there and cleaned up, and I think it's good again.

oobabooga:
I think that it looks good now. There are endless edge cases to take care of, like some in this string

> Python, Java, C++, JavaScript, Ruby, PHP, Perl, Pascal, Lisp, Scheme, Fortran, Cobol, Ada, Basic, Assembly Language, Visual BASIC, Delphi, Objective-C, Smalltalk, Haskell, Erlang, Prolog, Scala, Clojure, F#, Groovy, Dart, Rust, Ocaml, Elixir, Kotlin, Julia, Go, Swift, TypeScript, Lua, R, Matlab, Bash, Awk, Makefile, Shell Scripting, SQL, PL/SQL, TCL, AWK, PERL, SED, YAML, JSON, XML, CSS, HTML5, XHTML, SVG, SCSS, LESS, SASS, Stylus, Less, Jade, EJS, Handlebars, Mustache, Twig, Jinja2, Mako, Django Templates, Razor

but the most important things are already in this PR (handling numbers and acronyms).

Thank you for this! The level of attention to detail here is impressive. By the way, After your suggestion of adding spaces between functions, I ended up refactoring the entire code base and I find it to be much more readable now https://github.com/oobabooga/text-generation-webui/commit/ea6e77df726bc57d60cd77be399dbfd50c25345e

PR #760, opened 2023-04-04
allow quantized model to be loaded from model dir

catalpaaa:
quick fix :)


da3dsoul:
This will be nice. I just symlinked the model dir to a bcache'd RAID array for my models

oobabooga:
Thank you :)

PR #570, opened 2023-03-25
LoRA Training Tab

mcmonkey4eva:
### What is this

Adds a new tab, `Training`, for training-related features. Has a sub-tab `Train LoRA` to train LoRA with.

This `Training` tab can potentially be expanded more in the future with
- Other trainers (core model training, softprompts, etc)
- Utilities (dataset managers, format managers, etc)

------------------

### What's the point

It makes training your own LoRA super simple. There's an easy UI screen, it even shows an explanation of what each input does so you don't need to be confused about it. Decent default parameters and guidance for how to change them if you want to. You just fill in your dataset and go.

The code is referenced from [Alpaca-LoRA](https://github.com/tloen/alpaca-lora) but not copied exactly (some parts are direct copy, so if license rights are a topic that matters do note alpca-lora's Apache2 license. I believe it should be fine).

Because it uses the WebUI's own code, it just kinda... works quick n easy in terms of model handling and all. The model you have loaded in the WebUI is the model that gets trained. Doesn't need to be reloaded newly or anything. You can go use your LoRA immediately after training completes if you want.

------------------

### Appearance, and gradio notes

~~This is compatible with the existing version of gradio in the WebUI, but looks better if we update to `3.23.0` thanks to the new `info` feature (it will be silently ignored on old versions).~~

Screenshot from latest gradio:

image

~~I know gradio updates are iffy sometimes because of how much they change between versions, but imo the new features are worth it.
Fwiw nothing looks broken, even cai-chat and all are working properly.~~ (Gradio is now updated in main)

-------------------

### I want to try, give me quick test.

- Build any dataset.
    - You can for example use the Alpaca dataset: https://github.com/tloen/alpaca-lora/blob/main/alpaca_data_cleaned.json
    - You can also trim it to just a few lines to do a quicker test (ie delete most of the file content).
    - The dataset can be an JSON file formatted as a list `[]` of dicts `{}` of string-pairs `"key": "val"`, so `[{"key": "val"}]` is the shortest valid file. They must be standardized key names, eg Alpaca uses `instruction` + `input` + `output` wherein `input` is optional.
    - You can have any keyset you want, and you can use different sets of key names to choose sub-format options.
- Build any format file.
[...]

Issue #706, opened 2023-04-01
When you train a second time it fails


elephantpanda:
### Describe the bug

Trying to train a lora for a second time fails and you have to reload the model. A similar issue is after you have made your lora file, you have to reload everything to use it. If this is a feature maybe there should just be a friendly message that tells you to restart the server.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

try to train a second time

### Logs

```shell
RuntimeError: mat1 and mat2 shapes cannot be multiplied (1024x768 and 2304x768)
```

### System Info

```shell
windows 10
```

mcmonkey4eva:
I'm testing on the improvements PR #763 and I can't replicate an issue here.
I don't think that PR changed anything relevant, but it _does_ add friendly messages telling you when it detects something wrong.

`A similar issue is after you have made your lora file, you have to reload everything to use it` You don't, just click the *Refresh* button next to the LoRA dropdown under Parameters, and then select your LoRA file. It should *just work*.

Can you try running via #763 and seeing if any of the warning messages catch you?
The most notable detections that I think might apply here are: it's only been validated for LLaMA in 8-bit. Other models, or non 8-bit, won't necessarily work. (4bit doesn't yet, 16bit mostly works but not well validated, and kinda irrelevant considering 8bit quality is equivalent to 16bit for half the VRAM cost).

If you still have the same issue, and there's no detected error/warning, can you please post *step by step* instructions to get to the error you have - including your model, command line args, training param, and literally step by step (like `1: click train tab, fill in training params, click train, 2: go to parameters tab, refresh and select the new lora, ...`).

PR #763, opened 2023-04-04
Lora trainer improvements

mcmonkey4eva:
This is a followup to my earlier PR #570 with more improvements to the LoRA trainer tab, focused on end-user QOL mainly.

The overall end-goal here is to make it so using the LoRA Training tab is as easy as possible. The particular exemplar of this goal is that common user mistakes I've already seen happening in practice now have error checks with clear messages, to ensure future users who make the same mistakes know what's wrong and how to fix it straightaway.

- Added "Prefer newline cut length" to "Raw text file input"
    - Cuts strings at newlines when it can, to reduce the risk of the AI learning it should cut off midway through a line (I don't know that this necessarily happens in a problematic way given training is per-token anyway, but, either way, better to be safe and clean with cuts)
    - Side effect is in the UI, "Text File" dropdown occupies half a row on its own - this makes the sliders way too thin. For now I've just put them on their own row, but it would be good to fix it such that Text File is not overly wide.
- Put cutoff length below dataset and in a group with it, to make more clear that it's part of the dataset options and not part of the core LoRA format options it was previously grouped with
- Hid `put-trainer-datasets-here` from the text file list
- updated `info` text to make clear that `Eval Dataset` is optional
- Added a check for if the user clicks Interrupt before the trainer itself has started (ie while data prep is still on-going) to prevent the trainer from spooling up if it's going to be insta-cancelled
- improved the formatting of time for long runs. It will now show eg `5 hours` where previously it would show `18000 seconds` and now provides an explicit time-remaining estimate
- fixed a few more issues of code not formatted to the repo's standard (meaning: changed a few more missed camelCases to snake_case)
- Added some extra error detection:
    - If the user is running in 4bit mode, it stops instantly with a notice that 4bit training isn't available yet
    - If the user is running in 16bit mode, it displays a notice encouraging 8bit but continues anyway (this seems to work but requires ridiculous amounts of VRAM, even more than you'd expect from dealing with 16bit in general, and I don't think carries any actual quality benefit over 8bit)
    - If the user is running a model that isn't LLaMA, it displays a notice warning that only LLaMA is tested but continues anyway (hypothetically any other model that's compatible with peft should just work, but, uh, haven't tested that. If other models are confirmed to work well with it we can swap the LLaMA check to a whitelist of known compatible model types).

### DIFF
--- a/modules/training.py
+++ b/modules/training.py
@@ -111,8 +111,7 @@ def do_train(lora_name: str, micro_batch_size: int, batch_size: int, epochs: int
     if model_type != "LlamaForCausalLM":
         if model_type == "PeftModelForCausalLM":
             yield "You are trying to train a LoRA while you already have another LoRA loaded. This will work, but may have unexpected effects. *(Will continue anyway in 5 seconds, press `Interrupt` to stop.)*"
-            print(f"Warning: Training LoRA over top of another LoRA. May have unexpected effects.")
-        
+            print("Warning: Training LoRA over top of another LoRA. May have unexpected effects.")
         else:
             yield "LoRA training has only currently been validated for LLaMA models. Unexpected errors may follow. *(Will continue anyway in 5 seconds, press `Interrupt` to stop.)*"
             print(f"Warning: LoRA training has only currently been validated for LLaMA models. (Found model type: {model_type})")
--- a/modules/training.py
+++ b/modules/training.py
@@ -243,6 +243,9 @@ def do_train(lora_name: str, micro_batch_size: int, batch_size: int, epochs: int
     # TODO: save/load checkpoints to resume from?
     print("Starting training...")
     yield "Starting..."
+    if WANT_INTERRUPT:
+        yield "Interrupted before start."
+        return
 
     def threadedRun():
         trainer.train()
@@ -256,6 +259,7 @@ def do_train(lora_name: str, micro_batch_size: int, batch_size: int, epochs: int
         time.sleep(0.5)
         if WANT_INTERRUPT:
             yield "Interrupting, please wait... *(Run will stop after the current training step completes.)*"
+
         elif CURRENT_STEPS != lastStep:
             lastStep = CURRENT_STEPS
             timeElapsed = time.perf_counter() - startTime
--- a/modules/training.py
+++ b/modules/training.py
@@ -109,6 +109,20 @@ def do_train(lora_name: str, micro_batch_size: int, batch_size: int, epochs: int
     lora_name = f"{shared.args.lora_dir}/{clean_path(None, lora_name)}"
     actual_lr = float(learning_rate)
 
+    if type(shared.model).__name__ != "LlamaForCausalLM":
+        yield "LoRA training has only currently been validated for LLaMA models. Unexpected errors may follow. *(Will continue anyway in 2 seconds)*"
+        print("Warning: LoRA training has only currently been validated for LLaMA models.")
+        time.sleep(2)
+
+    if shared.args.wbits > 0 or shared.args.gptq_bits > 0:
+        yield "LoRA training does not yet support 4bit. Please use `--load-in-8bit` for now."
+        return
+
+    elif not shared.args.load_in_8bit:
+        yield "It is highly recommended you use `--load-in-8bit` for LoRA training. *(Will continue anyway in 2 seconds, press `Interrupt` to stop.)*"
+        print("Warning: It is highly recommended you use `--load-in-8bit` for LoRA training.")
+        time.sleep(2) # Give it a moment for the message to show in UI before continuing
+
     if cutoff_len <= 0 or micro_batch_size <= 0 or batch_size <= 0 or actual_lr <= 0 or lora_rank <= 0 or lora_alpha <= 0:
         yield "Cannot input zeroes."
         return
@@ -247,10 +261,10 @@ def do_train(lora_name: str, micro_batch_size: int, batch_size: int, epochs: int
         yield "Interrupted before start."
         return
 
-    def threadedRun():
+    def threaded_run():
         trainer.train()
 
-    thread = threading.Thread(target=threadedRun)
+    thread = threading.Thread(target=threaded_run)
     thread.start()
     last_step = 0
     start_time = time.perf_counter()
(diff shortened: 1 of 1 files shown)


modules/training.py
```diff
@@ -273,3 +302,25 @@ def threadedRun():
...
+    return chunk
+
+def format_time(seconds: float):
+    if seconds < 120:
+        return f"`{seconds:.0f}` seconds"
+    minutes = seconds / 60
+    if minutes < 120:
+        return f"`{minutes:.0f}` minutes"
+    hours = minutes / 60
+    return f"`{hours:.0f}` hours"
```
tensiondriven, line 326: With a little use of modulus here, we could display "h m s" instead of a single unit of time remaining
```suggestion
def format_time(seconds):
    # calculate hours, minutes, and seconds
    hours = seconds // 3600
    seconds %= 3600
    minutes = seconds // 60
    seconds %= 60
    
    # build the time string
    time_str = ""
    if hours:
        time_str += f"{hours}h "
    if hours || minutes:
        time_str += f"{minutes}m "
    time_str += f"{seconds}s"
    
    return time_str
```
Note: This is untested, just providing as a possible strategy.

Gitterman69:
https://github.com/lxe/simple-llama-finetuner 

we need this instead! 10x faster!

mcmonkey4eva:
At a glance that project looks to be practically identical. In fact they based it on the same references I did even. If it's faster it's almost certainly just from different default params.

modules/training.py
mcmonkey4eva: Looks valid, but the question is - which is simpler/clearer. I went with single number on the idea that `5h 23m 16s` isn't actually really providing much more info than `5 hours` is, but looks more complex.

oobabooga:
Or he used a paragraph in the simple-llama-finetuner, the alpaca json with a million rows here, and concluded that simple-llama-finetuner is faster

oobabooga:
I agree that

> Put cutoff length below dataset and in a group with it, to make more clear that it's part of the dataset options and not part of the core LoRA format options it was previously grouped with

makes sense, but it looked really weird to put that slider below the dataset tabs, so I moved it back. I don't think this is a significant cause of confusion.

@mcmonkey4eva do you think that the model should be reloaded before and/or after the training? I haven't tested, but there has been an issue related to this https://github.com/oobabooga/text-generation-webui/issues/706

The new warnings are great, and the remaining changes all make sense to me. I will do some experimenting with the slicing functions to better understand the algorithm

mcmonkey4eva:
Did testing and posted followup on that issue (tldr: can't replicate, so encouraged them to re-test from the PR's branch and re-test and post detail).

While testing I noticed the error message for having a LoRA already loaded isn't very clear (it thinks it's a non-LLaMA model) so I changed it to a warning specific to having a LoRA loaded - non-breaking, as that *does work*, but with a notice to make sure the user knows to switch it off if they didn't intend to do that.
(Side topic: loading multiple LoRAs at once would be a nice feature improvement)

I also increased the delay for warnings to 5 seconds as I noticed with 2 seconds, if you don't have a long dataset load time extending it, it goes past too quick to read properly.

oobabooga:
Looks good to me. Thanks for checking that issue @mcmonkey4eva.

This PR is a very good quality of life update.

Gitterman69:
I had problems with loading my txt file due to the encoding, so i added the following line to the training.py


` # == Prep the dataset, format, etc ==
    if raw_text_file not in ['None', '']:
        print("Loading raw text file dataset...")
        with open(clean_path('training/datasets', f'{raw_text_file}.txt'), 'r', **encoding='utf-8'**) as file:
            raw_text = file.read()
        tokens = shared.tokenizer.encode(raw_text)`

oobabooga:
Thank you @Gitterman69, added here https://github.com/oobabooga/text-generation-webui/commit/768354239b4437eb2fa739effdf8461cdd91cb6a

mcmonkey4eva:
Oooh, good fix!
We'll probably want to make sure that's on *all* text file reads throughout the codebase, including eg the format/dataset for JSON inputs in the same trainer file, but also character files and etc.

Gitterman69:
Over the past few days, I've been utilizing the trainer, and I must say, I'm extremely impressed with its performance. However, I do have a minor suggestion for improvement. 

Background of my request: 
Currently, when training stable diffusion models, the model is saved multiple times during the training process (which can be adjusted in the settings to save the model every few epochs). 

Feature request: 
It would be beneficial to include this feature in the text-webui-trainer as well, so the ongoing lora could be automatically saved every XXX epochs (or after 25% 50% 75% of the training... etc.). 

This would enable better progress tracking and evaluation for each epoch, help with finding the best lora version for a training run, saving loras in case of a computer crash or shutdown, and other advantages. Although I was unable to locate mcmonkey4evas contact information, I hope to submit this suggestion here in the hopes that it may be of use and maybe even realized soon! :)

Thanks guys

Gitterman69:
An additional feature that would be helpful is the option to save and load finetuning settings in txt or json format. The current process of taking screenshots of settings and manually saving them in the lora folder can be quite cumbersome. This feature would streamline the process and save users time and effort.

mcmonkey4eva:
Posting here, or in an issue and @ ing me are both good ways of contacting me about the lora trainer interface here. Alternately you can poke me on the Discord @ https://discord.gg/u27RhJECcW in the `#training` channel.
(Either way posting publicly is handy so others can comment, and/or if I'm unable any other dev can handle it. Ooba's pretty awesome and has been actively maintaining the entire repo).

Saving settings was something I already had planned but haven't gotten around to. Seeing how many people are already using this, I guess I should prioritize that!
Saving midpoints was something I hadn't thought about but should be simple to add I think, I'll put that in my notes.

Have that and a few other TODOs, so looks about time to make another followup PR soon.

nRuaif:
Hey, thank for great work. however I ran into a problem: With a batch size = 4, default parameters, when I tried to train it I just got OOM even though , any idea? Or just a T4 is not good enough?

mcmonkey4eva:
Depends what you're trying to train - it requires *roughly* a similar amount of VRAM to train as it does to run the model in 8bit for inference (text generation). It requires more depending on Batch Size, Rank, and Text Cutoff Length values - can set them lower to squeeze it in tighter.

I've trained 7B and 13B models without issue on a 3090, with VRAM to spare (aka room to bump the parameters up higher to get better results).

PR #767, opened 2023-04-04
Add LLaMA-Precise preset

StefanDanielSchwarz:
[Comparing LLaMA and Alpaca PRESETS deterministically](https://www.reddit.com/r/LocalLLaMA/comments/12az7ah/comparing_llama_and_alpaca_presets/) shows that the LLaMA-Precise preset (as originally recommended in the [How to install LLaMA: 8-bit and 4-bit : LocalLLaMA](https://www.reddit.com/r/LocalLLaMA/comments/11o6o3f/how_to_install_llama_8bit_and_4bit/) guide) is distinct and useful enough to warrant official inclusion in oobabooga/text-generation-webui. In fact, I recommend this to be the default for LLaMA/Alpaca models, since it outperformed all other presets in the aforementioned comparison.

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -44,6 +44,7 @@ settings = {
     'chat_default_extensions': ["gallery"],
     'presets': {
         'default': 'NovelAI-Sphinx Moth',
+        '.*(alpaca|llama)': "LLaMA-Precise",
         '.*pygmalion': 'NovelAI-Storywriter',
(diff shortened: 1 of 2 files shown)


oobabooga:
I'm still not too sure which preset is better, but sphinx moth has been giving me too inconsistent replies in instruct mode, so let's give this a chance.

PR #771, opened 2023-04-04
Prompt template extension

oobabooga:
For usage with Alpaca, Open Assistant, etc without having to manually type the prompt format in the input field.


da3dsoul:
Out of curiosity, why is this a PR instead of just committing it like other changes and fixes? Are you looking for feedback?

oobabooga:
I implemented this extension but am not sure if this is the best way to do it. It probably isn't, since it doesn't allow the generation to continue or new user inputs to be given after the bot's reply.

oobabooga:
Implemented in a better way in chat mode here (not an extension) https://github.com/oobabooga/text-generation-webui/commit/e722c240af56aa733c576f49e663ed7cafef5784

PR #798, opened 2023-04-05
Stop character pic from being cached when changing chars or clearing.

Ph0rk0z:
Otherwise the first used char PFP will be stuck or will return when generation finishes.

Hope it's not just my janky browsers but I tried a few.


oobabooga:
I can't reproduce the issue (the time.time() clears the cache consistently for me), but your change is harmless and makes sense

PR #803, opened 2023-04-05
fix link to streaming api example

eiery:
Fixed the broken README link to point to https://github.com/oobabooga/text-generation-webui/blob/main/api-example-stream.py.


oobabooga:
Ty!

PR #785, opened 2023-04-05
Add support for latest qwopqwop200 cuda branch

oobabooga:
Following the discussion in #734

Issue #734, opened 2023-04-02
I think I made the Triton-GPTQ model work with the GPTQ-for-LLaMa (Triton branch) repository.


BadisG:
Hello!

I have a model that has all the GPTQ implementations and it's called "gpt-x-alpaca-13b-native-true_sequential-act_order-128g-TRITON"
This model was made using [Triton](https://github.com/qwopqwop200/GPTQ-for-LLaMa/tree/triton) and it can be run on the webui with the current commit from GPTQ-for-LLaMa.

1) Do a regular clone of the GPTQ-for-LLaMa repository (triton is the main branch)
```
conda activate textgen
mkdir repositories
cd repositories
git clone https://github.com/qwopqwop200/GPTQ-for-LLaMa
cd GPTQ-for-LLaMa
pip install -r requirements.txt
```
That will install the Triton package on your textgen environnement

2) Go to your anaconda/miniconda/mamba installation directory and find the autotuner.py script

```
miniconda3\envs\textgen\lib\python3.10\site-packages\triton\runtime\autotuner.py 
```

3) Once you've opened autotuner.py, change the line 81 from this:
```
self.cache[key] = builtins.min(timings, key=timings.get)
```
to this:
```
self.cache[key] = min(timings, key=lambda x: (timings[x] if isinstance(timings[x], float) else float('inf')))
```

And there you have it! You can now run a model that has all the GPTQ implementations on the webui!

Here's an example of a successful run: 

```
conda activate textgen 
python server.py --wbits 4 --auto-devices --pre_layer 30 --disk --groupsize 128
```

```
===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please submit your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
CUDA SETUP: CUDA runtime path found: /home/adduser/miniconda3/envs/triton/lib/libcudart.so
CUDA SETUP: Highest compute capability among GPUs detected: 7.5
CUDA SETUP: Detected CUDA version 117
CUDA SETUP: Loading binary /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117.so...
The following models are available:

1. alpaca-native-7b
2. gpt-x-alpaca-13b-native-true_sequential-128g-CUDA
3. gpt-x-alpaca-13b-native-true_sequential-act_order-128g-TRITON
[...]

PR #820, opened 2023-04-06
Broaden GPTQ-for-LLaMA branch support

EyeDeck:
Conditionalized the keyword arguments passed to GPTQ-for-LLaMA's `quant.make_quant()` so we can support more branches simultaneously. Pretty simple, we just test which kwargs `quant.make_quant()` currently has available and don't try to send any invalid ones.

Tested (with varying launch.py arguments) on:
* Oobabooga's old CUDA fork: [9659310499cc7a0ea5498c1beb47bb228d65d178](https://github.com/oobabooga/GPTQ-for-LLaMa/tree/9659310499cc7a0ea5498c1beb47bb228d65d178)
* Current CUDA branch: [610fdae6588c2b17bcf2726cacaaf795cd45077e](https://github.com/qwopqwop200/GPTQ-for-LLaMa/tree/610fdae6588c2b17bcf2726cacaaf795cd45077e)
* Current triton branch: [6800a08c7a3d8d682499df5d6d96666253af8ed1](https://github.com/qwopqwop200/GPTQ-for-LLaMa/tree/6800a08c7a3d8d682499df5d6d96666253af8ed1)

I suspect it would also work with the older non-groupsize versions too, but I don't think I have any old quantizations lying around to test it.

Following up #785

Just to be clear, this still won't make a model quantized by an excessively different version of GPTQ-for-LLaMA load, so e.g. you still need the old CUDA GPTQ-for-LLaMA in your repositories directory to use the 2-week-oldish quantizations most users probably have at this point. Likewise, if you want to try the new triton branch, you'll need to grab some fresh requantizations.
This PR is just so you can use different GPTQ-for-LLaMA versions without having to _also_ fiddle with text-generation-webui code to get things cooperating.

### DIFF
--- a/modules/GPTQ_loader.py
+++ b/modules/GPTQ_loader.py
@@ -16,7 +16,7 @@ from modelutils import find_layers
 from quant import make_quant
 
 
-def _load_quant(model, checkpoint, wbits, groupsize=-1, exclude_layers=['lm_head']):
+def _load_quant(model, checkpoint, wbits, groupsize=-1, faster_kernel=False, exclude_layers=['lm_head'], kernel_switch_threshold=128):
     config = AutoConfig.from_pretrained(model)
     def noop(*args, **kwargs):
         pass
@@ -126,7 +126,8 @@ def load_quantized(model_name):
     if shared.args.pre_layer:
         model = load_quant(str(path_to_model), str(pt_path), shared.args.wbits, shared.args.groupsize, shared.args.pre_layer)
     else:
-        model = load_quant(str(path_to_model), str(pt_path), shared.args.wbits, shared.args.groupsize)
+        threshold = False if model_type == 'gptj' else 128
+        model = load_quant(str(path_to_model), str(pt_path), shared.args.wbits, shared.args.groupsize, kernel_switch_threshold=threshold)
 
         # accelerate offload (doesn't work properly)
         if shared.args.gpu_memory:
--- a/modules/GPTQ_loader.py
+++ b/modules/GPTQ_loader.py
@@ -1,6 +1,7 @@
 import re
 import sys
 from pathlib import Path
+import inspect
 
 import accelerate
 import torch
@@ -33,7 +34,22 @@ def _load_quant(model, checkpoint, wbits, groupsize=-1, exclude_layers=['lm_head
     for name in exclude_layers:
         if name in layers:
             del layers[name]
-    make_quant(model, layers, wbits, groupsize)
+    
+    gptq_args = inspect.getfullargspec(make_quant).args
+
+    make_quant_kwargs = {
+        'module': model, 
+        'names': layers,
+        'bits': wbits,
+    }
+    if 'groupsize' in gptq_args:
+        make_quant_kwargs['groupsize'] = groupsize
+    if 'faster' in gptq_args:
+        make_quant_kwargs['faster'] = faster_kernel
+    if 'kernel_switch_threshold' in gptq_args:
+        make_quant_kwargs['kernel_switch_threshold'] = kernel_switch_threshold
+    
+    make_quant(**make_quant_kwargs)
 
     del layers
 
--- a/modules/GPTQ_loader.py
+++ b/modules/GPTQ_loader.py
@@ -15,13 +15,13 @@ from modelutils import find_layers
 from quant import make_quant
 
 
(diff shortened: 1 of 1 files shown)


EyeDeck:
By the way I've now gotten a chance to test performance at different context sizes of the new triton branch (via WSL2), with a fresh quantization, following up https://github.com/oobabooga/text-generation-webui/issues/734#issuecomment-1496928371 
* Upstream triton branch; LLaMA 30B --wbits 4 --act-order --true-sequential --groupsize 128
* > Output generated in 8.69 seconds (4.72 tokens/s, 41 tokens, context 696)
Output generated in 35.14 seconds (5.66 tokens/s, 199 tokens, context 738)
Output generated in 36.51 seconds (5.45 tokens/s, 199 tokens, context 937)
Output generated in 38.51 seconds (5.17 tokens/s, 199 tokens, context 1137)
Output generated in 41.56 seconds (4.79 tokens/s, 199 tokens, context 1336)
Output generated in 39.90 seconds (4.99 tokens/s, 199 tokens, context 1535)
Output generated in 44.05 seconds (4.52 tokens/s, 199 tokens, context 1734)
Output generated in 45.22 seconds (4.40 tokens/s, 199 tokens, context 1848)
Output generated in 33.63 seconds (5.92 tokens/s, 199 tokens, context 23)

On this machine (single 3090), the triton branch is around 80% as fast as old cuda branch at large context size, which isn't too bad considering the new cuda branch is around 1.27% as fast as the old cuda branch. Also, I seem to be able to do 1848 context size with `--groupsize 128` and not OOM, while the old branch + `--groupsize 128` usually OOMed after roughly 1650.

If you're willing to take about a 20% speed hit, I'm sure the improved model quality achieved by act-order + groupsize is very nice.

Ph0rk0z:
Heh.. its a 20% speed hit on almost the top consumer GPU. Wonder if it will run at all on mine and what the hit will be.

Why not go all the way and support v1, v2 and is it now v3 models? Just use separate GPTQ loaders and a command line parameter.

oobabooga:
Very nice! As a sanity check, I have tested it with my current models and they loaded without errors and had the same text generation performance.

I thought the triton branch was slower. 80% is not bad at all, considering that the CUDA branch is already several times faster than running the model in 16-bit mode using the default transformers code.

BadisG:
So on the repository folder I have to create 2 GPTQ-for-LLaMA?
Something like:  
GPTQ-for-LLaMA-CUDA
GPTQ-for-LLaMA-TRITON

But I have a different cuda environnement for both Cuda and Triton... if you can help me clarify stuff I would be grateful

BadisG:
How can you clone differents GPTQ-for-LLaMa folders if they have the same names? And like I said I have a different conda environnement for both cuda and triton, do you mean I can do it with just one env?

oobabooga:
You can't, you have to delete/rename the existing folder and clone/install a new one. I'm not saying this is ideal, it's just that adding support for multiple GPTQ-for-LLaMa folders at the same time is beyond the scope of this PR

EyeDeck:
Well, there are a couple of different obstacles to having different setups of GPTQ-for-LLaMA to seamlessly switch between:
* As far as I know, there still isn't any explicit metadata stored with the actual quantized model that tells which version can load it; it might be possible to deduce via the internal layer names (which changed a couple of times), or even just try trial-and-error
* More of a problem, all the CUDA versions of GPTQ-for-LLaMA require slightly different custom CUDA kernels, which replace each other; so it would be necessary to:
* * execute GPTQ-for-LLaMA's setup_cuda.py every time the webui switches,
* * or track down old GPTQ-for-LLaMA commits, patch them so they can coexist without conflicting, and then get users to use those patched old versions instead of qwopqwop200's official versions

It _is_ a bit silly how quantized models don't have any standardized metadata to read. Making the user have to manually handle version + wbits + groupsize feels clumsy. _But_, the GPTQ-for-LLaMA is still only a month old as of today, so this is all still very early development.

oobabooga:
The models have versioning because past versions were never supposed to be re-utilized. For instance, hugging face has recently changed the LLaMA tokenizer format requiring a new conversion. The old files are obsolete and to be deleted. But in the case of GPTQ-for-LLaMA, a trade off appeared between the past and current versions.

In any case, wbits and groupsize should indeed be saved in a metadata file.

PR #838, opened 2023-04-06
Instruction Character Vicuna, Instruction Mode Bugfix

OWKenobi:
- added Vicuna to the Instruction Templates, since it uses a slightly different setup than Alpaca.

- fixed a bug that in instruct mode, the stopping_string was produced like for example **\nAssistant::** with double colon, and thus  the model would not stop generating.


oobabooga:
I have tested it with `anon8231489123/vicuna-13b-GPTQ-4bit-128g` and it worked perfectly. I just added a missing `:` to the normal chat mode stopping strings.

PR #863, opened 2023-04-07
Tweak COPY order in Dockerfile

deece:
The current ordering will cause the prequisite steps to be invalidated if any of the project files changes.

This patch first copies only the required files for the prerequisite builds, so they will only be rebuilt if those specific files change.


oobabooga:
Thank you! If you can think of other improvements, feel free to submit a new PR.

PR #877, opened 2023-04-07
Use updated docker compose

MarlinMr:
Using the `docker-compose` command is outdated and will give version 1 of docker compose. Instead it should be `docker compose` to get version 2.17+ as specified in the requirements.

This is also marked down wrongly in the Wiki.


oobabooga:
You are correct :)

PR #881, opened 2023-04-07
Fix regeneration of first message

mayaeary:
In instruct mode regeneration of first answer is not worked, because it thinks this message is greeting.


oobabooga:
Thank you, I was aware of this bug but had not managed to find a solution so far.

PR #823, opened 2023-04-06
fix vacuna won't stop talking

catalpaaa:
added stopping_strings at the vary top which returns a list of stopping strings.

I fell like put it there makes it easy to add more types of stopping strings for different model, not only for vacuna

maybe also let user to customize them with setting files or args?

PR #903, opened 2023-04-08
allow custom stopping strings

catalpaaa:
a better #823


catalpaaa:
image

oobabooga:
It now works in notebook mode too

ShaneTRS:
I've just done `git pull` and updated all the requirements via `pip`, and I can't seem to verify that this is functioning. I'm using the `api-example.py` script, with `custom_stopping_strings` set to `["\n"]` (as well as `early_stopping` set to `True`, though I'm not sure that's necessary). I also changed the `prompt` to `"test"`, just because I know it often creates new lines.

I'm still getting responses that are multiple lines long, such as:
```
test_2.csv"
    train_df = pd.read_csv(train_csv)
    test_df = pd.read_csv(test_csv)
    # preprocessing
```
I've tried other strings (and other amounts of strings), such as first names and what-not, but none seem to have an effect.

ShaneTRS:
After some extra testing, I think it may be related to the fact that I run a RWKV model.
Here are my typical launch arguments:
`--model RWKV-4-Raven-3B-v8-EngAndMore-ctx4096.pth --rwkv-strategy 'cuda fp16' --rwkv-cuda-on --listen --no-stream --notebook`

When running with these instead:
`--model llama-7b --load-in-8bit --listen --no-stream --notebook --lora alpaca-lora-7b`

I seem to get proper stopping, though I didn't do a lot of testing

oobabooga:
@ShaneTRS the stopping criteria stuff is a part of `transformers`, and RWKV uses a custom implementation that is not part of transformers, so it doesn't work for this model (and also for llama.cpp)

See https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L139

PR #904, opened 2023-04-08
Update the api-example-stream.py to use the correct fn_index value

andysalerno:
Apparently, Gradio makes no guarantees about the fn_index values.

At some point in time, changes to the Gradio app definition have invalidated fn_index 12, and now the correct value is fn_index 8 (verified by running the webapp and using Firefox dev tools to inspect the websocket stream).

IMO the api-example-stream.py seems "wrong" on the surface, because it relies upon talking to the Gradio server, which seemingly has no guaranteed contract. I guess it expects that the Gradio webclient is the only client (which makes sense because the service and client are generated at the same time so they will always be in sync).

But I will admit, I have found this api-example-stream very useful, and I use it in my toy projects as a way to talk to llms programatically, even if it's a bit of a hack :D

### DIFF
--- a/api-example-stream.py
+++ b/api-example-stream.py
@@ -12,6 +12,11 @@ import string
 
 import websockets
 
+# Note, Gradio may pick a different fn value as the definition of the Gradio app changes.
+# You can always launch the web UI and inspect the websocket stream using your browser's dev tools
+# to determine what value Gradio expects here.
+GRADIO_FN = 8
+
 
 def random_hash():
     letters = string.ascii_lowercase + string.digits
@@ -47,14 +52,14 @@ async def run(context):
                 case "send_hash":
                     await websocket.send(json.dumps({
                         "session_hash": session,
-                        "fn_index": 12
+                        "fn_index": GRADIO_FN
                     }))
                 case "estimation":
                     pass
                 case "send_data":
                     await websocket.send(json.dumps({
                         "session_hash": session,
-                        "fn_index": 12,
+                        "fn_index": GRADIO_FN,
                         "data": [
                             payload
                         ]


oobabooga:
After this change, the streaming does work, but it seems to return corrupt. Adding this to the end of api-example-streaming

```python
async def get_result():
    async for response in run(prompt):
        # Print intermediate steps
        print(repr(response))

    # Print final result
    print(repr(response))
```

I get a sequence of incremental responses ending in this:

> '["What I would like to say is the following: ", {"max_new_tokens": 200, "do_sample": true, "temperature": 0.5, "top_p": 0.9, "typical_p": 1, "repetition_penalty": 1.05, "encoder_repetition_penalty": 1.0, "top_k": 0, "min_length": 0, "no_repeat_ngram_size": 0, "num_beams": 1, "penalty_alpha": 0, "length_penalty": 1, "early_stopping": false, "seed": -1}]\n[["I would like to say that this is the best sentence in the world.", "I would like to say that this is the best sentence in the world."], [{"max_new_tokes": 34867}, {{"max_new_toks": 34867}}]]\n\\end{code} Tags: c#, asp.net-mvc, linq, lambda\n\nQuestion: How do you write a lambda expression with an if statement?\n\nI\'m using ASP.NET MVC and C#. I have a list of objects which are returned from my controller action as follows:\n\n\\begin{code}\npublic ActionResult Index()\n    {\n        var items = _repository.GetAll();\n        return View(items);\n    }\n\\end{code}\n\nInside my view, I need to check whether or not each item has been marked for'

Moreover, if I add this print statement to `modules/api.py`, 

```python
def generate_reply_wrapper(string):
    print(repr(string))
```

nothing is printed :(

andysalerno:
Yeah, you're right! Let me close this one, I think I have a better option running locally. I'll open a new PR and we can consider it.

PR #906, opened 2023-04-08
Add support for reading options from configuration files.

sukipop:
This commit introduces changes to the argument parsing logic in order to support reading and including command-line arguments from `webui.conf`. The updated code allows users to store frequently used arguments in `webui.conf`, which will be automatically combined with any arguments passed through the command line.

The changes made include:

1. Added a new function `get_args_from_file` that takes a file path as input and reads the file line by line. It then splits each line into arguments and returns a list of these arguments.

2. Modified the argument parsing logic to read arguments from both `webui.conf` and the command line. The arguments are combined and then passed to the `argparse.ArgumentParser` instance for parsing.

3. Updated the code to parse the combined arguments using the `parse_args` method.

This feature simplifies the process of passing common arguments and makes it more convenient for users to configure the application.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,7 @@ venv/
 *.bak
 *.ipynb
 *.log
-
+webui.conf
 settings.json
 img_bot*
 img_me*
--- a/README.md
+++ b/README.md
@@ -271,8 +271,8 @@ To save configuration options for `server.py` and ensure their persistence acros
 
 | Flag                                  | Description |
 |---------------------------------------|-------------|
+| `--config CONFIG_PATH`                | Path(s) to configuration files to source options from. |
 | `--no-config`                         | Ignore the default configuration file `webui.conf`. |
-| `--config`                            | Locations of configuration files to source options from. |
 
 Out of memory errors? [Check the low VRAM guide](https://github.com/oobabooga/text-generation-webui/wiki/Low-VRAM-guide).
 
--- a/README.md
+++ b/README.md
@@ -265,6 +265,15 @@ Optionally, you can use the following command-line flags:
 | `--auto-launch`                       | Open the web UI in the default browser upon launch. |
 | `--gradio-auth-path GRADIO_AUTH_PATH` | Set the gradio authentication file path. The file should contain one or more user:password pairs in this format: "u1:p1,u2:p2,u3:p3" |
 
+#### Configuration
+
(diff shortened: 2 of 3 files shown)


sukipop:
This is my first pull request, so please let me know if I did anything wrong. Constructive feedback would be much appreciated.

I wanted to highlight a few things about the changes I made:

- Users should see no change to the execution of `server.py` if they do not create `webui.conf`.
- Command line options take precedence over options from config files.
- I've noticed that the code may not handle arguments whose values are in quotes correctly, please keep that in mind when reviewing.
- In the future, I plan on modifying the code to print messages about sourced `.conf` files when the `--verbose` option is enabled. The message should include the path to the file.
- I also think it might be useful to consider adding an `--ignore-config` option to prevent loading of `webui.conf`. The option should exit with error if given but `webui.conf` does not exist.
- Additionally, it might be helpful to add a `--config-location` option to accept one or more paths to additinal `.conf` files. The option should exit with an error if given without any values, and it should check for the existence of the supplied file(s).

Thanks for your time and consideration. Please let me know if you have any comments or questions.

sukipop:
I updated the code to handle values containing quotes in the `webui.conf` file. This means that users can now specify options with quotes, without causing any issues in the code. 

Secondly, I added a check to skip any lines in the webui.conf file that start with the `#` character. This allows users to add comments to their config files, without causing any errors in the code. 

Using Tilde expansion and variables is still not supported, I'm not sure what it would take to fix that.

sukipop:
I made some improvements to the argument parsing functionality, the main updates include:

1. Moved the `get_args_from_file` function to a more appropriate location within the code.
2. Refactored the argument parsing logic to make it more streamlined and easier to read.
3. Defined the `no_config` option in the "Default settings" section.
4. Added the following command line options.
  - `--no-config`: Prevents loading of `webui.conf` or displays an error if the file doesn't exist.
  - `--config`: Allows users to provide paths to additional config files for more flexibility.
5. Rewrote the `get_args_from_file` function to print a message when the `verbose` option is enabled, providing details about the sourced config files and their paths.
6. Updated the `README.md` documentation to include information about the new options and the `webui.conf` functionality.

These improvements aim to make it easier and more convenient for users to manage and customize their application settings.

oobabooga:
I think that this commit (and the previous ones that expanded the models tab) made an additional config file just for the command-line arguments redundant.

https://github.com/oobabooga/text-generation-webui/pull/1167

PR #915, opened 2023-04-08
Fix ggml downloading in download-model.py

xNul:
I was trying to download models from Hugging Face and noticed the script wouldn't download ggml models. Turned out to be a very simple fix.


oobabooga:
Thank you, I missed that :)

Issue #854, opened 2023-04-07
Chat keeps reusing last input


practicaldreamer:
### Describe the bug

Experiencing a pretty major bug in chat mode. Every other response uses last input... see screenshot

If I `git reset d0f9625 --hard` it doesn't do this btw...

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Launch with:
`python3 server.py --wbits 4 --gpu-memory 15 19 --listen --model llama-65b --verbose --chat`

See screenshot...

### Screenshot

![image](https://user-images.githubusercontent.com/78515588/230515668-7b3c483a-e979-4965-ab34-b950cb3f1189.png)

### Logs

```shell
(textgen2) user@user-System-Product-Name:~/Documents/ooba/2/text-generation-webui$ python3 server.py --wbits 4 --gpu-memory 15 19 --listen --model llama-65b --verbose --chat

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please submit your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
CUDA SETUP: CUDA runtime path found: /home/user/anaconda3/envs/textgen2/lib/libcudart.so
CUDA SETUP: Highest compute capability among GPUs detected: 8.6
CUDA SETUP: Detected CUDA version 117
CUDA SETUP: Loading binary /home/user/.local/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117.so...
Loading settings from settings.json...
Loading llama-65b...
Found models/llama-65b-4bit.safetensors
Loading model ...
Done.
[...]

PR #917, opened 2023-04-08
Fix mobile client reusing input

practicaldreamer:
This change reverts one of the changes made from 0dc6fa0 which causes mobile client to reuse input


practicaldreamer:
In response to https://github.com/oobabooga/text-generation-webui/issues/854

oobabooga:
I think that this should have solved it in a cleaner way.

https://github.com/oobabooga/text-generation-webui/commit/a6a00cb82f6127e0497917aca8803ecf9db220cc

practicaldreamer:
Indeed it does resolve the issue. I'll close #854 now

oobabooga:
Thanks for the confirmation!

PR #938, opened 2023-04-08
Added CPU training.

IggoOnCode:
Added a command line parameter to enable training on the CPU only.

For people like me, that refuse to acknowledge the reality without experiencing it themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,8 +206,7 @@ Optionally, you can use the following command-line flags:
 
 | Flag                                        | Description |
 |---------------------------------------------|-------------|
-| `--cpu`                                     | Use the CPU to generate text. |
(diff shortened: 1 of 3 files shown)


oobabooga:
I had no idea that was possible, very cool.

Why not reuse the same `--cpu` parameter for this?

IggoOnCode:
I wanted to break as few things as possible, so I made a extra parameter. And in the future I will try to combine it with --auto-devices and GPU inference, so the users still can have reasonable still speeds when talking to the model.

Having an extra parameter also makes sure people read the warning label. I don't want people to have the impression that it's a fully usable feature.

IggoOnCode:
Changed to use --cpu parameter to enable cpu training.

The combination with --auto-devices didn't work easily and I'm going to take another approach on hybrid training. So this will stay cpu only.

oobabooga:
It looks good to me like this. Thanks for the PR! One more thing that I had no idea was possible.

PR #940, opened 2023-04-09
updated docs for docker, setup video added, removed left over GPTQ_VE…

loeken:
hey @oobabooga 

when you removed the GPTQ_VERSION i think you missed one in the docker-compose.yml

i've also updated the docs slighty, moved manjaro section down, and added numbering to the steps for ubuntu so its 8 clear steps to follow. Also recorded a short youtube video on how to get it running on ubuntu 22.04 as not everybody seems to be too familiar with docker.

i also moved your note about installing the docker compose plugin manually to the bottom so you can still find it but most users would follow the instructions and would install docker-compose-plugin when running
```
sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin docker-compose -y
``` 
and receive updates via repos that way

one note on the docs/wiki, if you were to move the docs over to the other files, users might be able to create PRs to update the docs.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,6 @@ services:
       args:
         # specify which cuda version your card supports: https://developer.nvidia.com/cuda-gpus
         TORCH_CUDA_ARCH_LIST: ${TORCH_CUDA_ARCH_LIST}
-        GPTQ_VERSION: ${GPTQ_VERSION}
         WEBUI_VERSION: ${WEBUI_VERSION}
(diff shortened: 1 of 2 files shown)


oobabooga:
Thank you for the video and the updated instructions! They have been moved here: https://github.com/oobabooga/text-generation-webui/wiki/Docker-compose-installation

I will consider moving the wiki to the `docs/` folder.

PR #948, opened 2023-04-09
More like added some stuff to the SD_API extension

if-ai:
I modify the script so you can select all available models, samplers, Upscalers, modify height and width size, select batch count, do Hires fix, restore face, denoise slider, cfg scale, etc..

### DIFF
--- a/extensions/sd_api_pictures/If_mod_SD_api
+++ b/extensions/sd_api_pictures/If_mod_SD_api
@@ -1,0 +1,269 @@
+#This script has been modified by a caveman's monkey all credits to the original author
+
+import base64
+import io
+import re
+from pathlib import Path
+
+
+import gradio as gr
+import modules.chat as chat
+import modules.shared as shared
+
+import requests
+import torch
+from PIL import Image
+
+
+
+
+
+torch._C._jit_set_profiling_mode(False)
+
+# parameters which can be customized in settings.json of webui  
+params = {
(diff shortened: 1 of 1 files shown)


if-ai:
if you test this you need to run SD webui first also don't know how it is to run it from another machine you need to change the url in the code bc it is hardcoded at the moment

Brawlence:
Oof. Although I welcome those improvements, there are currently several problems with your proposal.

First and foremost, it is based on an old version (the one which was used prior to the [commit #ffd102e](https://github.com/oobabooga/text-generation-webui/commit/ffd102e5c02781c40a822b43556ae9d9db2001af) of sd-api-pics extension. Please save a copy of your changes, update the repository and try to apply them to the new (v.1.1) script.

Second -- you should submit a pull request modifying the sole script.py file, not making its copy. Right now it's hard to review changes short of directly comparing your code to the one used in my version (using PRs as intended makes this easy).

Third -- yeah, you already said it. It's bad to hardcode stuff especially since address is the first thing that was made customizable and you already have it in `params['address']` dict.

if-ai:
Yes, I just realised my models aren't changing on the webui,I thought I had fix that I will take again from the update version and implement all that I said earlier. I am sorry, thank you

if-ai:
I am adding a few more things but now the models are changing in SD I have trouble with some not subscript able objects I must find what they are.

PR #954, opened 2023-04-09
download custom_model menu (from hugging face) added in model tab inside server.py

UsamaKenway:
Hi. 
People were downloading models repo from huggingface.co using git and manually downloading LFS and placing in the 'model' folder.So, I added option to download.

1. textbox and a download button
2. if lora model dected, place in loras folder
3. display error if textbox is empty or invalid input
4. display progress

I was not able to import download-model.py to reuse that code and hugging face download methods were limited, so wrote inside wrapper. Im new to gradio, will be improving your repo as I learn more.

image
image
image


oobabooga:
The problem with using Hugging Face to download the files is that many unnecessary files may be downloaded. For instance, some repositories contain the same model both in .bin and .safetensors format.

After this change, download-model.py should be importable, since all command-line arguments are confined to the main function: https://github.com/oobabooga/text-generation-webui/commit/34ec02d41d6b68aa04cffed1a70c30c977a27578

PR #970, opened 2023-04-09
Fix lora training prompt generation always using the first formatting option in the formatting file instead of the correct one and fix error when value of datapoint is None

myluki2000:
I found these two problems while using the lora trainer. I hope the fixes I applied are correct.

The "return data" line was indented incorrectly, causing the function to always return after the first loop iteration, even if the data format option didn't match the input dataset's needed formatting option. To fix this I indented the line so it now sits inside the if-block.

The `if set(options.split(',')) == set(x[0] for x in data_point.items() if len(x[1].strip()) > 0):` line crashed when a value of a datapoint was set to None. I'm not exactly sure when this happens, because I do not have any null values inside of the json file containing my training dataset, but maybe it's caused by me having both items with an instruction and an additional input and also items only with an intruction inside my training dataset? Anyways, with the check to see if the value is None it seems to work and correctly generate the prompt now, at least from the limited amount of testing I performed.


oobabooga:
Do you have a .json file that I can use to reproduce this issue?

myluki2000:
For the first bug, consider a simple training dataset like this

```JSON
[
{
	"instruction": "Do stuff",
	"input": "Additional Input",
	"output": "Answer stuff"
},
{
	"instruction": "Do stuff2",
	"input": "Additional Input2",
	"output": "Answer stuff2"
}
]
```

use it together with the default alpaca formatting template (or any other formatting template; as I wrote in my first message it will always pick the first formatting option from the formatting template, because the return statement is outside of the if-block that checks if the correct formatting option was picked). The prompt it then generates using `generate_prompt()` will look like this:

```
Below is an instruction that describes a task. Write a response that appropriately completes the request.

### Instruction:
%instruction%

### Response:
%output%
```

i.e. it literally just returns the formatting template without replacing the placeholders with the text from the training dataset.


Regarding the second bug I've done some further debugging and it seems like the issue is more complex than I previously thought. The issue mainly stems from how the huggingface datasets package works.

Consider the following training dataset:

```json
[
{
	"instruction": "Do stuff",
	"input": "Additional Input",
	"output": "Answer stuff"
},
{
	"instruction": "Do stuff2",
	"output": "Answer stuff2"
}
]
```

This json file will be loaded as a dataset with 2 rows and 3 columns and will look like this:

| **Instruction** | **Input**          | **Output**      |
|-----------------|--------------------|-----------------|
| "Do stuff"      | "Additional Input" | "Answer Stuff"  |
| "Do stuff2"     | None               | "Answer Stuff2" |

As you can see the value of the input for the second entry is set to None, because that entry in the JSON file did not have an "input" attribute. In your code here 

https://github.com/oobabooga/text-generation-webui/blob/1911504f8205bfae5203185d89d7860a0f930974/modules/training.py#L188

you assume however that the value is always a string (you strip() it and then check if it's length is larger than 0). You assume that all entries in the JSON file have all attributes, and if you want to create an entry that does not have all attributes (e.g. an entry without any "input", just with "instruction" and "output"), you'd have to instead set the attribute in the JSON entry to an empty string:

```json
[
{
	"instruction": "Do stuff",
	"input": "Additional Input",
	"output": "Answer stuff"
},
{
	"instruction": "Do stuff2",
        "input": "",
	"output": "Answer stuff2"
}
]
```

The fix in my pull request just checks if the attribute of an entry in the dataset is None before checking if it's an empty string, thus also making it possible to just leave out attributes in the JSON if they're empty anyways.

However, I have meanwhile found out that there's another issue my current fix doesn't fix, it is cause by how the huggingface datasets package works. In its default configuration if you load a JSON file, it only uses the first entry in the file to determine which columns exist in the dataset. This means that if you have a file like this:


```json
[
{
	"instruction": "Do stuff",
	"output": "Answer stuff"
},
{
	"instruction": "Do stuff2",
        "input": "Additional Input2",
	"output": "Answer stuff2"
}
]
```

The loaded dataset will look like this:


| **Instruction** | **Output**      |
|-----------------|-----------------|
| "Do stuff"      | "Answer Stuff"  |
| "Do stuff2"    | "Answer Stuff2" |

This won't cause the program to crash, instead it will just train with the data missing (because it assumes there is no "input" data in the first place) and the model you get out of training will then be very bad obviously.

This seems to be intended behavior of the datasets library to improve performance, because otherwise it would have to check for each entry in the JSON file if there are more attributes.

However, there's an easy fix we can apply. It is possible to pass a list of column names to load_dataset(), so it knows which columns exist that it has to load. We already have the column names specified in the formatting file, so we can just get the column names from the formatting file and pass them to load_dataset() to make sure it loads all the columns we need.

I'll update this merge request soon with the fix for this issue.

mcmonkey4eva:
Oh, oops, that looks like one bug accidentally hid the other. That'd be my bad.

Regarding the last point (datasets being weird) tbh the best solution is probably just to ignore `datasets` for loading and building the dataset, and then shove the result into `datasets` after (similar to how the raw text input does it). ie just bypass the oddities of HF datasets code entirely.

myluki2000:
> However, there's an easy fix we can apply. It is possible to pass a list of column names to load_dataset(), so it knows which columns exist that it has to load. We already have the column names specified in the formatting file, so we can just get the column names from the formatting file and pass them to load_dataset() to make sure it loads all the columns we need.

Upon further inspection of the hf datasets code, this won't work, because the list of columns for some reason isn't passed further by the datasets package's internal methods (at least for JSON files).

As it stands now, this pull request could be merged as-is, in my opinion. We can resolve the remaining issue caused by the datasets package seperately.

> Regarding the last point (datasets being weird) tbh the best solution is probably just to ignore `datasets` for loading and building the dataset, and then shove the result into `datasets` after (similar to how the raw text input does it). ie just bypass the oddities of HF datasets code entirely.

I've identified the line of code in the datasets package now. Fixing it probably shouldn't be too complicated. They just I could open an issue in their repo and hope they fix it sooner or later. It seems to me like they just forgot to pass on the columns to load to pyarrow's load method.

mcmonkey4eva (approved):
Looks good to me.

oobabooga:
Thank you @myluki2000 for the PR and thanks @mcmonkey4eva for the review :)

Issue #542, opened 2023-03-24
Add command line flags for loading character, chat history, etc


tensiondriven:
**Description**

As a user of Text Generation UI, I want to be able to start and stop the app and have the conversation pick up where it left off.  Right now, every time I start, I have a small script that loads a specific model and some settings.  However, I then have to go into the webUI and manually import a recent chat.json, and also load a character manually.

The enhancement is to add command-line flags for loading a chat.json and character from the command line, such that when i open the webUI,  the existing context is there and I can pick up with my session.

**Additional Context**

Here's the current command I use to start the server.  If I want TTS on/off, I'll comment the 

```
python server.py \
	--cai-chat \
	--gptq-bits 4 \
	--model llama-13b \
	--listen --listen-port 8000 \
	--auto-launch \
	--verbose \
| tee -a ./output.log
```

I'd like additional options, something like:

```
--load-character <filename> - Loads a file named filename from the `characters` directory
--load-chat <filename> - Loads a chat.json file from the `logs` directory
```

Additionally, I'd like to be able to turn parts of the WebUI on/off so that I can have a cleaner experience.  This could be done by a config file, or perhaps some builtin gradio settings.  If gradio supports themes, adding support for that would be a big improvement.  I'd still want to be able to rename strings (similar to localization, but custom, for example, using emojis for tab names so they display nicely on mobile).  I'd also like to be able to completely disable or hide sections of the UI.  That way, I could run a command from the command line, and have a specific web address and port that would be specific to a certain character or purpose.

I'm going to log some additional feature requests here, which should probably be split out into separate issues, but for now:

General Usability:

- The 'listen' flag binds to 0.0.0.0, which is great, except that the URL isn't clickable when this is run as a VM, which is a common use case.  As a user I'd like to be able to have the link displayed for the eth0, so that I can easily open up the UI after starting the app.
- Related, when running in a vm, the `--auto-launch` flag doesn't work.  I don't know if it's possible for python to pass an instruction all the way through ssh to the top level terminal, but if so, that would allow a user who is running via VM to have the auto launch feature work.  It would also be nice if auto-launch could detect that it was running headless or something, so you'd get a warning
- Support for theming, and all these additional options, could get tedious to use as command line flags.  To solve this, support a `config` file, such that i could run `python server.py --config my_config.json` - This could load all my existing settings, and easily support new settings down the road, as well.

Chat Persistence:
[...]

PR #976, opened 2023-04-09
Add ability to load a character via args

bmoconno:
* Add new `--character` argument to `server.py` to load a selected default character
* Also make character choice persistent, so on refresh the last chosen character will still be selected.

This PR is fulfilling the requested enhancement in #542, and adds the ability to load a character by including the filename of a character in the `characters` directory (with or without extension).

Example: `python server.py --chat --character test`

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -414,7 +414,7 @@ def load_character(character, name1, name2, mode):
 
 
 def load_default_history(name1, name2):
-    load_character("None", name1, name2, "chat")
+    return load_character(shared.character or "None", name1, name2, "chat")
 
 
 def upload_character(json_file, img, tavern=False):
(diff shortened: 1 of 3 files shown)


Malrama:
May I suggest having this as a setting in settigs.json instead of a starting argument?
What way we could even use 	use something like this, as we do with Presets:
```
//file settings.json
[...]
"characters": {
		"default": "Default"
                ".*alpaca": "Alpaca"
                ".*vicuna": "Vicuna"
}
[...]
```
That way we won't bring in unnecessary stating arguments and cleanly separate model specific running parameters (e.g. preset, character, parameters) and model loading starting arguments, which later are essential for loading a model.

I hope my suggestion makes sense.

Thank you

bmoconno:
@Malrama 

I think maybe having both would be the way to go? That way it can support either method and people can choose whichever they prefer, or even use both (giving priority to the `--character` argument if it is used)?

I'll see if I can get this suggestion implemented today or tomorrow.

Malrama:
> @Malrama
> 
> I think maybe having both would be the way to go? That way it can support either method and people can choose whichever they prefer, or even use both (giving priority to the `--character` argument if it is used)?
> 
> I'll see if I can get this suggestion implemented today or tomorrow.

I am really not a fan of having model running parameters in the starting arguments. I do think that the settings.json file is the perfect place for that and it will only make starting arguments unnecessary bloated.

Ofc adding both would be an option. But I suggest, in my humble opinion, to just add the .json config support. I don't see the benefit in adding it as a starting argument. It would also introduce running parameters into the starting arguments, which are currently separated nicely. (User-friendliness)

Greeting

bmoconno:
Closing this PR to break into smaller parts that might get merged.

PR #977, opened 2023-04-09
Update character log logic

bmoconno:
Currently, if you press the clear chat history button then refresh the page, the old logs will still be there. This PR solves that bug by saving the cleared log after the confirm button is pressed.

Additionally, this saves a character's initial log on first character load, instead of waiting for chat to happen.

* When logs are cleared, save the cleared log over the old log files
* Generate a log file when a character is loaded the first time


oobabooga:
Seems reasonable

PR #986, opened 2023-04-10
Added dropdown to character bias.

BlueprintCoding:
Implemented a new feature to the script that incorporates a checkbox and a dropdown to enable users to switch between predefined biases in the dropdown or input custom bias via a textbox. Upon execution, the code generates a new "bias_options.txt" file in the extensions directory, which stores the option values for the dropdown. This enables users to modify the file and add their frequently used biases for quick access. Please note that restarting the webUI is necessary for changes made to the txt file to reflect in the dropdown.

### DIFF
--- a/extensions/character_bias/script.py
+++ b/extensions/character_bias/script.py
@@ -20,6 +20,7 @@ params = {
     "use custom string": False,
 }
 
+
 def input_modifier(string):
     """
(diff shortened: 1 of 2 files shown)


oobabooga:
Thank you! That's a great addition to the extension.

PR #997, opened 2023-04-10
UI improvements

xanthousm:
I got tired of scrolling so much, so I made some changes that I might as well share:
- Page tabs now remain at the top of the page while scrolling (small main.css edit)
- Text generation panel now moves to the side on wide screens, scaling to a single row like before when on a narrow screen (added rows, columns to main tab, css edits to remove side padding)
- Skinny scroll bars

I could not figure out how to move the extensions block to a new tab, or move them into the part of the text generation panel (since extension gradio elements need to be defined after all other gradio elements) without making major changes, so I left them alone to remain at the bottom of the text generation page and will move the tabs when scrolled to. 

## Text generation page previews:
Small Narrow Screen:

Small Wide Screen:

Half PC Screen:

Full PC Screen:

### DIFF
--- a/extensions/api/script.py
+++ b/extensions/api/script.py
@@ -57,6 +57,7 @@ class Handler(BaseHTTPRequestHandler):
                 'length_penalty': float(body.get('length_penalty', 1)),
                 'early_stopping': bool(body.get('early_stopping', False)),
                 'seed': int(body.get('seed', -1)),
+                'add_bos_token': int(body.get('add_bos_token', True)),
             }
 
             generator = generate_reply(


Ph0rk0z:
Yea.. mixed feelings about that (esp moving the input box). Also with no more generation params preset on chat screen.

MaiJames:
> 

I have to agree on that. In terms of UX, I think that most people is used to have the input box underneath the conversation when chatting, as seen on most chat apps. The other settings, I'm not sure where would be better to have them.
What I would like to see, on the other hand, is to have some options on the side to tweak while testing, like changing models, characters, temp, and such.
Or maybe also presets of those settings, to make it more creative or more precise.
Just my opinion.

Ph0rk0z:
Yea.. there is space right next to ui mode that is empty.

xanthousm:
Thanks for the feedback guys. I guess I've been using auto1111 too much and got used to the input box at the top. 
I'm still figuring out how to organise things with Gradio (its missing some documentation) without making big changes, so I think I'll remove the text generation tab changes and just leave the "tabs remaining on page" part.

PR #1004, opened 2023-04-10
fix random seeds to actually randomize

mcmonkey4eva:
Without this fix, manual seeds get locked in.

Background: a user on Discord commented that `-1` doesn't seem to randomize, but typing different manual seeds did, I looked into it and - yeah if `seed` is `-1` it just does nothing, meaning any seed ever set in theory just gets locked in place after the first time one is set.

This code change adapts so that when `seed` is `-1`, a random seed is generated within the valid range, and manually fed in - thus every call to generate gets a seed freshly applied.

In testing, this works exactly as intended, results with seed `-1` are randomized properly now when regenerating.

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -1,3 +1,4 @@
+import random
 import re
 import time
 import traceback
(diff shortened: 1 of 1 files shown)


oobabooga:
We can use this opportunity to print the seed that was used like this:

> Output generated in 2.12 seconds (93.67 tokens/s, 199 tokens, context 16, seed 1808394723)

This was requested by someone at some point. It's a way to increase reproducibility.

Issue #1001, opened 2023-04-10
Ignoring all previous conversation when setting max_new_tokens to 2000


andresdelcampo:
### Describe the bug

I observed that after setting max_new_tokens to 2000 the model will not keep into account any of the previous questions / answers -even if they were very short! 
See repro steps for sample short conversation. 

**Benefits of fixing**: The default max_new_tokens is about 5 lines of reply, so it should be possible to increase it... -models allow a much higher token number in context.

Related but not identical: 
https://github.com/oobabooga/text-generation-webui/issues/897
https://github.com/oobabooga/text-generation-webui/issues/576

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Got latest version as of April 10 (commit 1911504f8205bfae5203185d89d7860a0f930974).
Open a Vicuna 4-bit model (llama.cpp or not, Chat or Instruct model, with or without instruction prompt) 
I usually change Generation parameter preset to Default, and temperature to 0,34, but that probably does not matter too much.

Try a conversation like:
> What is the name of the Dragonlance continent?
...Ansalon... (2 lines of response)
> What is the name of the world?
Krynn

If you then change max_new_tokens to 2000 and start over that conversation, the 

> What is the name of the Dragonlance continent?
...Ansalon... (2 lines of response)
> What is the name of the world?
The world is called Earth. (!!!)

Vicuna has a context size of 2048 and even so, the context used by that conversation was roughly 100 tokens. Why would it not keep the previous question and answer anymore? 

### Logs

```shell
No errors in logs
```

### System Info

```shell
Local, RTX 3080 -I doubt this has to do with the card?
```

andresdelcampo:
The issue seems related to [this line](https://github.com/oobabooga/text-generation-webui/blob/1dc464dcb0bad8a7221bd18198c8ff50bcb00b41/modules/chat.py#:~:text=max_length%20%3D%20min(get_max_prompt_length(max_new_tokens)%2C%20chat_prompt_size))
`   max_length = min(get_max_prompt_length(max_new_tokens), chat_prompt_size)`

[...]

PR #1005, opened 2023-04-10
Issue 1001: max_new_tokens maxed ignores previous conversation

andresdelcampo:
Fix Issue 1001 https://github.com/oobabooga/text-generation-webui/issues/1001

The issue seems related to [this line](https://github.com/oobabooga/text-generation-webui/blob/1dc464dcb0bad8a7221bd18198c8ff50bcb00b41/modules/chat.py#:~:text=max_length%20%3D%20min(get_max_prompt_length(max_new_tokens)%2C%20chat_prompt_size))
   max_length = min(get_max_prompt_length(max_new_tokens), chat_prompt_size)

which figures out the max_length of the context to send. Calling [get_max_prompt_length](https://github.com/oobabooga/text-generation-webui/blob/1dc464dcb0bad8a7221bd18198c8ff50bcb00b41/modules/text_generation.py#L18:~:text=def-,get_max_prompt_length(,-tokens)%3A) with max_new_tokens seems wrong, as it will effectively cut down the length to almost 0 when max_new_tokens is 2000 (2048-2000). That is the opposite it was supposed to do. The method is reused, so the issue is calculating it that way in this specific call -not in the method called.

Future improvements should remove the 2048 hardcoded number inside get_max_prompt_length and put the max prompt instead. But that is outside the scope of this issue and fix.

Tested verifying that the max_new_tokens is still respected, and it also keeps context when maxed.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -29,7 +29,7 @@ def generate_chat_prompt(user_input, max_new_tokens, name1, name2, context, chat
     # Finding the maximum prompt size
     if shared.soft_prompt:
         chat_prompt_size -= shared.soft_prompt_tensor.shape[1]
-    max_length = min(get_max_prompt_length(max_new_tokens), chat_prompt_size)
+    max_length = min(get_max_prompt_length(0), chat_prompt_size)
 
     if is_instruct:
         prefix1 = f"{name1}\n"


andresdelcampo:
Sure, that may justify that we keep 2048 hardcoded... but the fix is still required? 
Does it make sense that you get no past context when you ask for long answers? -and when those answers were actually short? 

Reduce memory would be accomplished reducing the max_new_tokens and "Maximum prompt size in tokens", right?

It sounds counterintuitive that you ask for more and you get less instead... (more answer length, less context length??)

Ph0rk0z:
I tried to fix this before to use the file from settings and it just never updated. I think that's why it was hardcoded. For RWKV you can have beyond 2048 context.. and really this doesn't help anything at all because if you load a character with 1800 context you will still get OOM even when you reduce the "max" . All you can do is increase the # of generated tokens to try to truncate things.

andresdelcampo:
On the hardcoded value, sure, I am fine with that. Thanks for the explanation. 

On the main issue:
I understand that long answers + long contexts will potentially cause out of memory. 
I also understand that by reducing the context when asking for longer answers we may keep memory lower...

But, why should asking for long answers mean that you get a reduced context? -less than the 2048 that the model allows?
It should be possible to ask for long answers AND long context -it is what the model was designed for, right? 

Why is max_new_tokens is even affecting the context size? The context should be 2048 in Vicuna no matter what -unless explicitly reduced by the user in the max prompt size in tokens?

Ok, I think I get it... because the prompt size could not be applied for some reason, the workaround seems to have been to use the max tokens to determine the max prompt size -and in inverse amount (weeeeeird...). Doesn't seem ideal.

I do not have a personal attachment to this -I mean, I would use my branch :D. But it seems extremely counterintuitive how it is currently implemented. As a user I would not expect to lose all the conversation -even short ones- because I asked for longer answers...

oobabooga:
The prompt size includes the generated tokens. Try this (any model/tokenizer will work as most models today use 2048 tokens of context)

```python
>>> from transformers import AutoTokenizer, AutoModelForCausalLM

tokenizer = AutoTokenizer.from_pretrained('models/facebook_galactica-125m')
model = AutoModelForCausalLM.from_pretrained('models/facebook_galactica-125m')
input_ids = tokenizer.encode("lorem ipsum"*512, return_tensors='pt')
print(input_ids.shape)
# prints torch.Size([1, 2048])

model.generate(input_ids, max_new_tokens=10)
```

```
File ~/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/functional.py:2210, in embedding(input, weight, padding_idx, max_norm, norm_type, scale_grad_by_freq, sparse)
   2204     # Note [embedding_renorm set_grad_enabled]
   2205     # XXX: equivalent to
   2206     # with torch.no_grad():
   2207     #   torch.embedding_renorm_
   2208     # remove once script supports set_grad_enabled
   2209     _no_grad_embedding_renorm_(weight, input, max_norm, norm_type)
-> 2210 return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)

IndexError: index out of range in self

```

andresdelcampo:
I suspect we have a misunderstanding. 

You do realize that calling [get_max_prompt_length](https://github.com/oobabooga/text-generation-webui/blob/1dc464dcb0bad8a7221bd18198c8ff50bcb00b41/modules/text_generation.py#L18:~:text=def-,get_max_prompt_length(,-tokens)%3A) with 0 will return 2048, right? That was my intention at least. I wanted long answers while keeping 2048 context, which seems impossible in the current implementation. 

Thanks, I won't insist further.

oobabooga:
What I have just shown you is that if you try to generate 200 tokens with 2048 tokens of context, that generates an error.

andresdelcampo:
Oops, I see now. Thanks!

PR #1006, opened 2023-04-10
SD_API UI additions for SD workflow  

if-ai:
Updated models, samplers, upscaler hi res fix, denoise strength, minimal changes to the code will add a way to select VAEs and loRas if I find where to get them from the SD_api #batch size is working but I commented out because is duplicating the images, maybe some bug on the seed parameters. https://youtu.be/kAUQ04iNfjI

### DIFF
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -1,43 +1,53 @@
 import base64
 import io
 import re
-from pathlib import Path
-from datetime import date
 import time
-
+from datetime import date
+from pathlib import Path
+from PIL import Image
+import torch
 import gradio as gr
 import modules.shared as shared
 import modules.chat as chat
 import requests
-import torch
 from modules.models import reload_model, unload_model
-from PIL import Image
+
+
 
 torch._C._jit_set_profiling_mode(False)
 
 # parameters which can be customized in settings.json of webui
 params = {
-    "address": "http://127.0.0.1:7860",
-    "mode": 0,  # modes of operation: 0 (Manual only), 1 (Immersive/Interactive - looks for words to trigger), 2 (Picturebook Adventure - Always on)
-    "manage_VRAM": False,
-    "save_img": False,
-    "sd_model_chekpoint": "dreamlikeDiffusion10_10.ckpt [0aecbcfa2c]",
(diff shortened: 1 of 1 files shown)


if-ai:
I changed the name so is easier to see

extensions/sd_api_pictures/script.py
```diff
@@ -111,31 +129,57 @@ def input_modifier(string):
...
+
+
     payload = {
+
         "prompt": params['prompt_prefix'] + description,
+        "negative_prompt": params['negative_prompt'],
         "seed": params['seed'],
-        "sampler_name": params['sampler_name'],
-        "steps": params['steps'],
+        "sampler_index": params['sampler_index'],
```
mcmonkey4eva, line 161: This change is incorrect, `index` is a very old legacy key-name. `sampler_name` was right.

extensions/sd_api_pictures/script.py
```diff
@@ -111,31 +129,57 @@ def input_modifier(string):
...
+        "hr_second_pass_steps": params['hr_second_pass_steps']
+                 
     }
 
     print(f'Prompting the image generator via the API on {params["address"]}...')
     response = requests.post(url=f'{params["address"]}/sdapi/v1/txt2img', json=payload)
     response.raise_for_status()
     r = response.json()
 
+    # This is broken for me please check where I messed up. I am sorry.
```
mcmonkey4eva, line 182: wh... what?

extensions/sd_api_pictures/script.py
```diff
@@ -16,23 +17,28 @@
...
-    'prompt_prefix': '(Masterpiece:1.1), detailed, intricate, colorful',
-    'negative_prompt': '(worst quality, low quality:1.3)',
-    'width': 512,
-    'height': 512,
-    'restore_faces': False,
-    'seed': -1,
-    'sampler_name': 'DDIM',
-    'steps': 32,
-    'cfg_scale': 7
+    "address": "http://127.0.0.1:7860",
```
mcmonkey4eva, line 20: This set of changes is... odd? Why did you reformat the quotes here?

extensions/sd_api_pictures/script.py
if-ai: I was trying to get the samplers to change with name but somehow, they only picked up with index but if name works leave name.

PR #1016, opened 2023-04-10
Update readme.md to reflect llama-cpp-python requirements

m40l:
The new CLI installation process fails if you don't follow this step, so it should be added.


oobabooga:
Duplicate of https://github.com/oobabooga/text-generation-webui/pull/1013

PR #1030, opened 2023-04-11
fix add_bos_token err

disarmyouwitha:
I was getting this error since new changes to `text_generation.py`, not sure how you wanted to handle it, but here was my patch!

```
File "/home/nap/Documents/llm-api/modules/text_generation.py", line 167, in generate_reply
    input_ids = encode(question, generate_state['max_new_tokens'], add_bos_token=generate_state['add_bos_token'])
KeyError: 'add_bos_token'
```


oobabooga:
How did this error happen? add_bos_token is a parameter of the encode function.

disarmyouwitha:
Upon further inspection this must be caused from the way I am calling it, running it from server.py is fine.

Sorry!

PR #1034, opened 2023-04-11
SD_API_Pictures: Character json tags processing, suffix processing and translations triggers

GuizzyQC:
@Brawlence Moved the character json and NSFW tag processing to a separate suffix processing function. Created a function that, upon matching specific words in text model prompt or text model response will add specific tags to the SD prompt. With a populated translations.json file, this can make for a much more seamless experience than adding tags manually to the prefix when something more complex and precise is requested from SD.

The new character description request to the text model I have in there is giving good results with Alpaca in getting the character to first state what it's wearing, then describe its environment and finally itself doing an action.

### DIFF
--- a/characters/Example.yaml
+++ b/characters/Example.yaml
@@ -14,5 +14,5 @@ example_dialogue: |-
   {{char}}: Motherboards, they're like puzzles and the backbone of any system.
   {{user}}: That sounds great!
   {{char}}: Yeah, it's really fun. I'm lucky to be able to do this as a job.
-positive_sd: "manga-style, 20 year old woman, anime, red square glasses"
-negative_sd: "angry, old, elderly, child, deformed, cross-eyed"
+sd_tags_positive: "manga-style, japanese, 20 year old woman, anime, red square glasses"
+sd_tags_negative: "angry, old, elderly, child, deformed, cross-eyed"
--- a/characters/Example.yaml
+++ b/characters/Example.yaml
@@ -14,3 +14,5 @@ example_dialogue: |-
   {{char}}: Motherboards, they're like puzzles and the backbone of any system.
   {{user}}: That sounds great!
   {{char}}: Yeah, it's really fun. I'm lucky to be able to do this as a job.
+positive_sd: "manga-style, 20 year old woman, anime, red square glasses"
+negative_sd: "angry, old, elderly, child, deformed, cross-eyed"
--- a/extensions/sd_api_pictures/README.MD
+++ b/extensions/sd_api_pictures/README.MD
@@ -54,11 +54,11 @@ Whenever the Activate SD translations box is checked, the extension will load th
 {
 	"pairs": [{
 		"descriptive_word": ["tennis"],
-		"SD_positive_translation": "1ball, 2rackets, (net)",
+		"SD_positive_translation": "tennis ball, rackets, (net), <lora:povTennisPlaying_lora:0.5>",
 		"SD_negative_translation": ""
 	        },
 		{"descriptive_word": ["soccer","football"],
-		"SD_positive_translation": "((soccer)), 2nets",
+		"SD_positive_translation": "((soccer)), nets",
 		"SD_negative_translation": ""
 	    }]
 }
--- a/extensions/sd_api_pictures/README.MD
+++ b/extensions/sd_api_pictures/README.MD
@@ -76,8 +76,8 @@ present in script.py, ex:
     "sd_api_pictures-manage_VRAM": 1,
     "sd_api_pictures-save_img": 1,
     "sd_api_pictures-prompt_prefix": "(Masterpiece:1.1), detailed, intricate, colorful, (solo:1.1)",
-    "sd_api_pictures-nsfw_prompts": "nsfw, naked",
-    "sd_api_pictures-anti_nsfw_prompts": "young, old",
+    "sd_api_pictures-secondary_positive_prompt": "nsfw, naked",
+    "sd_api_pictures-secondary_negative_prompt": "young, old",
     "sd_api_pictures-sampler_name": "DPM++ 2M Karras"
 }
 ```
--- a/extensions/sd_api_pictures/README.MD
+++ b/extensions/sd_api_pictures/README.MD
@@ -40,8 +40,8 @@ To help your character maintain a better fixed image, add positive_sd and negati
 
 ```json
 {
-	"positive_sd": "24 year old, asian, long blond hair, ((twintail)), blue eyes, soft skin, height 5'8, woman, <lora:shojovibe_v11:0.1>",
-	"negative_sd": "old, elderly, child, deformed, cross-eyed"
+	"sd_tags_positive": "24 year old, asian, long blond hair, ((twintail)), blue eyes, soft skin, height 5'8, woman, <lora:shojovibe_v11:0.1>",
+	"sd_tags_negative": "old, elderly, child, deformed, cross-eyed"
 }
 ```
 If clothing and accessories are permanently affixed and will never be changed in any picture you will request of that character, feel free to add them to this tag too. The extension prompts the character to describe what it is wearing whenever a picture of itself is requested as to keep that aspect dynamic, so adding it in the character json makes it more static.
--- a/extensions/sd_api_pictures/README.MD
+++ b/extensions/sd_api_pictures/README.MD
@@ -34,6 +34,38 @@ The model input is modified only in the interactive mode; other two are unaffect
 Connection check (insert the Auto1111's address and press Enter):  
 ![API-check](https://raw.githubusercontent.com/Brawlence/texgen-webui-SD_api_pics/main/illust/API-check.gif) 
 
+### Character sheet Stable Diffusion tags
+
+To help your character maintain a better fixed image, add positive_sd and negative_sd to your character's json file to have Stable Diffusion tags that define their appearance automatically added to Stable Diffusion prompts whenever the extension detects the character is referring to itself, ex:
+
+```json
+{
+	"positive_sd": "24 year old, asian, long blond hair, ((twintail)), blue eyes, soft skin, height 5'8, woman, <lora:shojovibe_v11:0.1>",
+	"negative_sd": "old, elderly, child, deformed, cross-eyed"
+}
+```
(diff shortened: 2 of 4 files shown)


GuizzyQC:
Another point to note: I've been testing the extension and tuning my prompts with Alpaca-7b. I would be surprised if the character self description prompt would work significantly worse on a better model, but I haven't yet managed to get anything better to run on my paltry RTX3070 yet, so it would be great if people could test it on more recent models like Vicuna and GPT 4 x Alpaca, and with more parameters, to check that it behaves satisfactorily with them too.

ClayShoaf:
I'm testing it a little more and I'm getting some little errors. I'll send a PR to your repo soon.

GuizzyQC:
@ClayShoaf Excellent, I had some weird results testing that I couldn't identify and couldn't rule out as just being a quirky model or messy context/history, and I wasn't set up to inspect the final prompt sent to the language model. I made a slight edit to the non-nsfw yes-characterfocus negative suffix to remove the nsfw tags; I had edited that out a bit earlier (maybe before you started to work on your changes) as 1) it's not in-line with the non-nsfw non-characterfocus result and 2) it can give counterintuitive results if a user puts a LORA in the nsfw_prompt (according to the Auto1111 documentation: "Lora cannot be added to the negative prompt.", so we should probably make sure we avoid users inadvertently doing it).

GuizzyQC:
Moved string evaluation outside of input_modifier, as it was interfering with picturebook / adventure mode. For now I'm leaving characterfocus outside of picturebook / adventure mode, so the entire weight of describing itself for that mode is on the language model. Maybe in the future a toggle for that mode or string evaluation for that mode could allow it to trigger characterfocus.

ClayShoaf:
> I'm mostly copying examples from better prompters than me for the sample translations, I'm not sure what are the best examples

A little unrelated, but I've probably done over 100K SD generations at this point. From what I've seen, a lot of the "better prompters" are very good at making one specific thing with one specific model and setup. A lot of the stuff that is included in generations is superfluous. I've done a lot of XYZ grids testing out the different popular tags, and while they have some effect it's not necessarily "better" in most cases, it works more like an extra bit of random entropy.

I am, by no means, the authoritative voice on the matter, but I have a much better understanding of how SD generation parameters/prompts work than I do, for example, the mechanics of git, haha.

I look forward to having even an XY grid for oobabooga. I would try to write it myself, but I'm worried that by the time I have something presentable, someone else will have made one that is better and all my time will have been wasted.

> Moved string evaluation outside of input_modifier, as it was interfering with picturebook / adventure mode. For now I'm leaving characterfocus outside of picturebook / adventure mode, so the entire weight of describing itself for that mode is on the language model. Maybe in the future a toggle for that mode or string evaluation for that mode could allow it to trigger characterfocus.

I'm ashamed to say, I hadn't even tested picturebook mode. I don't have the code up right now, but it seems like something that could be handled with a `bool` and and `if` statement, maybe?

ClayShoaf:
I won't be working on anything for the next few hours at least, so you don't have to worry about any random PRs coming through.

characters/Example.yaml
```diff
@@ -14,3 +14,5 @@ example_dialogue: |-
   {{char}}: Motherboards, they're like puzzles and the backbone of any system.
   {{user}}: That sounds great!
   {{char}}: Yeah, it's really fun. I'm lucky to be able to do this as a job.
+positive_sd: "manga-style, 20 year old woman, anime, red square glasses"
```
Brawlence, line 17: IMO should be `sd_tags_positive:` & `sd_tags_negative` as we're referring to a subset of properties applied to SD only. Also this lets them stay close if one to sort the properties alphabetically

extensions/sd_api_pictures/script.py
```diff
@@ -97,19 +152,51 @@ def input_modifier(string):
...
+                break
+        file_contents = open(filepath, 'r', encoding='utf-8').read()
+        data = json.loads(file_contents) if extension == "json" else yaml.safe_load(file_contents)
+
+    if params['nsfw']:
+        params['positive_suffix'] = params['nsfw_prompts'] + ", " + params['prompt_translation_positive']
+        params['negative_suffix'] = params['anti_nsfw_prompts'] + ", " + params['prompt_translation_negative']
+    if params['characterfocus']:
+        params['positive_suffix'] = data['positive_sd'] + ", " + params['prompt_translation_positive'] if 'positive_sd' in data else params['prompt_translation_positive']
+        params['negative_suffix'] = data['negative_sd'] + ", " + params['prompt_translation_negative'] if 'positive_sd' in data else params['prompt_translation_negative']
```
Brawlence, line 195: Again, IMO '`sd_tags_positive`' & '`sd_tags_negative`' are better

extensions/sd_api_pictures/script.py
```diff
@@ -97,19 +152,51 @@ def input_modifier(string):
 
     global params
 
-    if not params['mode'] == 1:  # if not in immersive/interactive mode, do nothing
-        return string
-
-    if triggers_are_in(string):  # if we're in it, check for trigger words
+    if params['mode'] == 1:  # if not in immersive/interactive mode, do nothing
```
Brawlence, line 155: Right now it should be `#For immersive/interactive mode, use the separate logic`

extensions/sd_api_pictures/script.py
Brawlence, line 195: Also, we should probably check for `negative` (not `positive`) in `data` when dealing with negative suffixes

extensions/sd_api_pictures/README.MD
```diff
@@ -34,6 +34,38 @@ The model input is modified only in the interactive mode; other two are unaffect
 Connection check (insert the Auto1111's address and press Enter):  
 ![API-check](https://raw.githubusercontent.com/Brawlence/texgen-webui-SD_api_pics/main/illust/API-check.gif) 
 
+### Character sheet Stable Diffusion tags
```
Brawlence, line 37: I especially like how you documented the changes in README.MD, nice touch!

extensions/sd_api_pictures/script.py
```diff
@@ -97,19 +152,51 @@ def input_modifier(string):
...
         toggle_generation(True)
         string = string.lower()
-        if "of" in string:
-            subject = string.split('of', 1)[1]  # subdivide the string once by the first 'of' instance and get what's coming after it
-            string = "Please provide a detailed and vivid description of " + subject
-        else:
-            string = "Please provide a detailed description of your appearance, your surroundings and what you are doing right now"
+        return string
+    if params['mode'] == 0:
+        toggle_generation(False)
```
Brawlence, line 163: You shouldn't toggle the generation here at all because the current state could be FORCED (user has pressed the 'Force pic' button while in Manual mode before submitting the text)

extensions/sd_api_pictures/script.py
```diff
@@ -97,19 +152,51 @@ def input_modifier(string):
...
 
-    if not params['mode'] == 1:  # if not in immersive/interactive mode, do nothing
-        return string
-
-    if triggers_are_in(string):  # if we're in it, check for trigger words
+    if params['mode'] == 1:  # if not in immersive/interactive mode, do nothing
+            return if_of_is_in(string)
+    if params['mode'] == 2:
+        params['characterfocus'] = False
         toggle_generation(True)
```
Brawlence, line 159: You shouldn't toggle the generation here at all because the current state could be SUPPRESSED (user has pressed the 'Supress pic' button while in Picturebook/Adventure mode)

extensions/sd_api_pictures/script.py
```diff
@@ -29,7 +31,18 @@
...
     'sampler_name': 'DDIM',
     'steps': 32,
-    'cfg_scale': 7
+    'cfg_scale': 7,
+    'characterfocus': False,
+    'nsfw': False,
+    'translations': False,
+    'character': "",
+    'anti_nsfw_prompts': '',
+    'nsfw_prompts': '(nsfw, nude, naked:1.2)',
```
Brawlence, line 40: IMO better to leave this as 'nsfw' by default.

My reasoning:
- NSFW isn't strictly nude\naked, could be other things beside those two;
- attention of 1.2 is too strong when tags are scarce;
- users would be able to override it via settings.json anyway.

Brawlence (commented):
One small bug introduced with `toggle_generation()` (which renders force_pic and suppress_pic buttons non-operational), other comments are my personal preferences in naming and defaults.

Other than that, LGTM, great work!

Anyone else is willing to look at those changes?

Brawlence:
Also cross-linking: https://github.com/oobabooga/text-generation-webui/pull/1038#issuecomment-1504776203

GuizzyQC:
> Also cross-linking: [#1038 (comment)](https://github.com/oobabooga/text-generation-webui/pull/1038#issuecomment-1504776203) .
> 
> In part I asked to factor out the logic for `if_of_is_in` to incorporate those future changes easier. Any thoughts on em?
> 
> Also, I really need to plan the separation of script.py into submodules as it's already too big to grasp on a glace.

@Brawlence I like the idea of those changes, it would help make the experience more immersive than a character being completely submissive to every request for pictures. I'm not sure of the flow of the inputs from the main text-generation-webui so I'd be out of my depth writing that for now. Of course there's also many improvements we could write for a more immersive experience around input evaluation. For now I just separated out the generation of the request to textgen from the evaluation of the input, that should make that new change you were talking around easier as you can add a case that won't toggle generation but will toggle a trigger for special inspection on the next message for terms of agreement or disagreement.

I've renamed "if_of_is_in" to a more helpful "string_evaluation". With regards to separation in submodules, it's getting to that point yes. I think input evaluation, UI, payload generation could all be separate.

GuizzyQC:
I think I'm done with improvements on this PR for now, unless it's bug fixes. If it's merged the community will probably come up with other ideas for improvements. For the couple of weeks I've been playing with it, this extension has been such a game changer for local models; hopefully these improvements will take it a bit further

GuizzyQC:
Fixes #1316 and obsoletes #1358

Fixed many bugs found by @altoiddealer and has latest improvement recommended by @Brawlence

oobabooga:
I think that this PR has valid changes that can improve the immersion of the SD extension, but it

1. Adds a lot of complexity to the extension that will make it harder for me to maintain it in the future.
2. Departs from the core functionality of the extension, which is to allow the LLM to generate stable diffusion prompts on its own.
3. Depends on modifications to the character yamls, when stable diffusion integration is not a core functionality of the web UI.

I encourage you to create your own customized fork of the extension and submit it here for others to download: https://github.com/oobabooga/text-generation-webui-extensions

GuizzyQC:
Makes a lot of sense, I've moved it to its own repo and am submitting it

PR #1038, opened 2023-04-11
update to make sd_api_pictures more dynamic

ClayShoaf:
I rewrote `input_modifier()` to take advantage of the LLM in formatting a new image prompt. This should allow you to ask your bot for images of whatever you want.

Also stopped `fix_newlines()` from operating on `user_input` as far as I can tell, nothing breaks from doing this, and it makes the final `prompt` more cohesive. Rather than
```
You: text

Assisstant: text

You: text
Assisstant: 
```
The prompt that will be submitted will look like
```
You: text
Assisstant: text
You: text
Assisstant: 
```

### DIFF
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -35,9 +35,6 @@ params = {
 }
 
 
-
-
-
 def give_VRAM_priority(actor):
     global shared, params
 
@@ -130,7 +127,7 @@ def input_modifier(string):
 
     if triggers_are_in(string):  # if we're in it, check for trigger words
         toggle_generation(True)
-        string = f"\n{name2}: *Translates 3 INPUTs to image prompt format. The prompts are descriptive and ALWAYS include a detailed description the subject of the image in the prompt. Any references in the second person like \"you\", \"your\", or \"yourself\" are about {name2} and include detailed information about {name2}'s hair, body type, clothing, surroundings, etc.*\nINPUT: send me a picture of a cool looking crystal\n{name2}: (crystal), diamond cut, reflective rock, transparent gemstone, beautiful sparkling gem, ultra detailed, volumetric lighting, professional photography\nINPUT: show me an image of Brad Pitt\n{name2}: portrait photo of (Brad Pitt), man, brown hair, short hair, athletic build, handsome, fashionable clothes, photo realistic, detailed skin, masterpiece, cinematic lighting\nINPUT: " + string
+        string = f"\n{name2}: *Translates 3 INPUTs to image prompt format. The prompts are descriptive and ALWAYS include a detailed description of the subject of the image in the prompt. Any references in the second person like \"you\", \"your\", or \"yourself\" are about {name2} and include detailed information about {name2}'s hair, body type, clothing, surroundings, etc.*\nINPUT: send me a picture of a cool looking crystal\n{name2}: (crystal), diamond cut, reflective rock, transparent gemstone, beautiful sparkling gem, ultra detailed, volumetric lighting, professional photography\nINPUT: show me an image of Brad Pitt\n{name2}: portrait photo of (Brad Pitt), man, brown hair, short hair, athletic build, handsome, fashionable clothes, photo realistic, detailed skin, masterpiece, cinematic lighting\nINPUT: " + string
 
     return string
 
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -2,6 +2,8 @@ import base64
 import io
 import re
 import time
+import json
+import yaml
 from datetime import date
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


ClayShoaf:
An example of outputs 
image

ClayShoaf:
I checked out https://github.com/oobabooga/text-generation-webui/pull/1034 and it seems to be working well. If that one gets merged I will rework this code, as I think the reformatted prompt will give better results. I will make sure I test it on models from 7B to 30B

extensions/sd_api_pictures/script.py
```diff
@@ -86,14 +88,37 @@ def triggers_are_in(string):
...
+                break
+        file_contents = open(filepath, 'r', encoding='utf-8').read()
+        data = json.loads(file_contents) if extension == "json" else yaml.safe_load(file_contents)
+    
+        if 'your_name' in data and data['your_name'] != '':
+            name1 = data['your_name']
+        name2 = data['name'] if 'name' in data else (data['char_name'] if 'char_name' in data else "Assistant")
+    else:
+        name2 = "Assistant"
+
```
Brawlence, line 121: I like the workaround to get name1 and name2 @lines99-121 but I feel like it should be factored out into separate function.

After all, this does mostly unrelated things to input_modifier and pulls info from the filesystem, solving a problem which ideally shouldn't even exist

extensions/sd_api_pictures/script.py
```diff
@@ -102,12 +127,7 @@ def input_modifier(string):
...
-            subject = string.split('of', 1)[1]  # subdivide the string once by the first 'of' instance and get what's coming after it
-            string = "Please provide a detailed and vivid description of " + subject
-        else:
-            string = "Please provide a detailed description of your appearance, your surroundings and what you are doing right now"
+        string = f"\n{name2}: *Translates 3 INPUTs to image prompt format. The prompts are descriptive and ALWAYS include a detailed description of the subject of the image in the prompt. Any references in the second person like \"you\", \"your\", or \"yourself\" are about {name2} and include detailed information about {name2}'s hair, body type, clothing, surroundings, etc.*\nINPUT: send me a picture of a cool looking crystal\n{name2}: (crystal), diamond cut, reflective rock, transparent gemstone, beautiful sparkling gem, ultra detailed, volumetric lighting, professional photography\nINPUT: show me an image of Brad Pitt\n{name2}: portrait photo of (Brad Pitt), man, brown hair, short hair, athletic build, handsome, fashionable clothes, photo realistic, detailed skin, masterpiece, cinematic lighting\nINPUT: " + string
```
Brawlence, line 130: This looks like a great way to get a more comprehensive pictures out of the model.

There's a downside, though. The dialogue flow is broken, it's like an instruct-injection right in the middle of the conversation. Wouldn't this contaminate future contexts for the model?

@oobabooga Is there a way to pass this to LLM hidden from the history; or can we somehow remove it once the response is made?

Brawlence (changes requested):
The main block I see is clean-up of that `INPUT:` from the chat history.

You can convince me that it's a non-issue.
Maybe one can link the source code proving that the model won't actually include those in its future contexts?

Brawlence:
Additional thoughs:

> This should allow you to ask your bot for images of whatever you want.

I know many would probably disagree with me on this one, but I think that one shouldn't always get what one wants. The next idea on my list for this extension is to implement a more natural two-step request system, ex.
```
- Would you please send me a picture?
- Sure!
<automatic hidden prompt for a pic here>
*sends a picture* <hidden code\link to image here>
```
or
```
- Would you please send me a picture?
- No chance in hell!
<picture response is set to false>
<nothing is sent, flow returns to dialogue>
```
Essentially, this allows the bot to imitate _**refusal**_ to provide user with a pic (_the option to force the gen anyway will be present_).

Maybe I should split `Immersive / Interactive` mode into just `Interactive` (with current behaviour) and `Immersive` (with this 2-step flow for future waifu imitations)

extensions/sd_api_pictures/script.py
ClayShoaf: Also, how do you get these code snippets inline with the rest of the comments?

ClayShoaf:
> The next idea on my list for this extension (for Immersive mode) is to implement a more natural two-step request system

Yea, I kind of like that idea. You might want to use some sort of sentiment analysis, rather than trying to hand code every possible output.

I'm going to close the PR for now and wait to see what happens with the other one.

AngelTs:
> I rewrote `input_modifier()` to take advantage of the LLM in formatting a new image prompt. This should allow you to ask your bot for images of whatever you want.
> 
> Also stopped `fix_newlines()` from operating on `user_input` as far as I can tell, nothing breaks from doing this, and it makes the final `prompt` more cohesive. Rather than
> 
> ```
> You: text
> 
> Assisstant: text
> 
> You: text
> Assisstant: 
> ```
> 
> The prompt that will be submitted will look like
> 
> ```
> You: text
> Assisstant: text
> You: text
> Assisstant: 
> ```

Not work. Th egenerated picture not shown in the chat?

PR #1075, opened 2023-04-12
Added an audible notification after text generation.

HappyWorldGames:
If there is a file "notification.mp3" in the root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ venv/
 *.log
 
 settings.json
+notification.mp3
 img_bot*
 img_me*
 prompts/[0-9]*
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
Not sure how often this would matter in practice, but: This is coded to play a sound where the python program, not where the webpage is open.
In other words, if you have the program running on an external server (say it's on your computer but you're accessing from a phone, or it's on google colab, or...) it won't function.

Implementing properly via webpage would likely be a bit more complicated though.

oobabooga:
As @mcmonkey4eva, it would be best to play the sound in the web UI side and not in the server side. There should be a gradio way of doing it.

HappyWorldGames:
Oh, really. I'm using the program on the same computer, so I didn't think of that... I think I will try to work on a web solution.

Issue #853, opened 2023-04-07
Multi-adapter PEFT loading


lxe:
**Description**

https://github.com/huggingface/peft/pull/263 just got merged into PEFT. I think we can take a stab ad multi-lora loading. I'm using it here: https://github.com/lxe/simple-llm-finetuner/blob/master/trainer.py#L59

**Additional Context**

oobabooga:
Very nice! Thanks for the heads up. Cc @mcmonkey4eva

mcmonkey4eva:
Woo!

Main question in my head is how to build the interface for multi-lora-loading?

Few options:

1: Literally a text line where you type model names, `Loras: lora_a, my_lora_b, etc`. Maybe some javascript hack to validate it as you type.

2: The same current dropdown, but when you select a new LoRA, it just applies over the top of the last one, and only resets if you hit `None`

3: Just list every LoRA as checkboxes. This is the worst option as it'd be hell to make gradio work with it and also it prevents ordering.

4: An interface like what Auto WebUI uses for _styles_:
![image](https://user-images.githubusercontent.com/4000772/230691953-5107668e-2650-487d-80e8-ecad4be59399.png)

Where it's a dropdown that you can select things in, and they get populated into a list of lil bubbles, and you can hit `X` to remove ones.

All of the above options other than `2` would require an `Apply` button or similar activation such that LoRAs only load and apply once the user is ready for it.
(Unless we can find a way to accelerate rapid load/unload of LoRAs? ie process them during inference rather than applying over the model? Might be dirty. Would cost VRAM and speed too, so not great.)

oobabooga:
I think that we could have a `gr.CheckboxGroup` where the user can select/deselect the desired LoRAs.

https://gradio.app/docs/#checkboxgroup

![cg](https://user-images.githubusercontent.com/112222186/230694834-1223a1e9-8cc4-4e0a-b94a-7db85f2d0985.png)

In principle, we could parse the `adapter_config.json` for all available LoRAs and only show those that match the currently selected model, since there is a `base_model_name_or_path` field. The problem is that the model names are all over the place. Here are two examples:

```
"base_model_name_or_path": "zpn/llama-7b",
"base_model_name_or_path": "decapoda-research/llama-7b-hf",
```

mcmonkey4eva:
Yknow, I was thinking LoRA order mattered and therefore options like that were insufficient, but in hindsight I'm not sure that order actually does matter.
[...]

PR #1098, opened 2023-04-12
Lora trainer improvements part 3

mcmonkey4eva:
- Adds support for other model types.
    - Technically the existing code worked for GPT-J and OPT, other than having the "untested" warning.
    - I trained GPT-J without issue.
    - I trained OPT without issue but it borked on actually *using* the LoRA.
    - I wrote the code in a generic way such that future model types can be easily supported via peft. Peft has a mapping of model types to required data, we just have to feed the correct model type - easiest way to do that rn is a classname mapping. The peft mapping is very new, and not actually in the normal release yet. We might want to update to a git link just to be able to add #853 .
        - See https://github.com/huggingface/peft/blob/cc82b674b5db38b9a393463d38afe66e8f48ac1c/src/peft/utils/other.py#L202-L222 for module mapping in peft.
        - This code is technically redundant right now as all 3 supported model types have the same standard modules, but I wanted the code to be actively ready for other types, at least what peft supports, which as you can see from that link, is a lot!
        - As was mentioned in passing on the multi-adapter issue, we could really use a clean standard way to identify model types accurately.
- Uses `encoding="utf8"` in the other file read.
- Added more description to `dropout`
- Made shuffling optional
- added `Evaluate every n steps` when using proper-dataset input
- added `Save every n steps` to control how often backups are saved
- cleaned code to not jank globals so much
- tried to add the ability to load and resume existing models, but something's going wrong and I'm not sure what.
    - Saving works, loading works, but then when _that_ model is saved and then loaded, that next loading is what errors out. It seems to think the model has become empty? But it's not. I don't know why.
- Added saving/loading parameters for easy reuse via simple json files.
    - `get_available_loras` is copied from `server.py`, can't be imported directly as that is a circular import. Moving it to `LoRA.py` or something would probably be cleaner. Or a generic file-utils file maybe.
- Added a checkbox to raise the Rank limit to `2048` instead of `1024`. It's not default that high because the slider becomes difficult to work with for normal ranges if you keep going higher. Anything above `1024` is effectively unreachable on consumer GPUs due to VRAM limits, but I was able to get 1500 with batchsize=1, a tiny cutoff limit, on a 7B model.
    - Note that the checkbox still gets input to the `do_train` function so it can be saved in the parameter list for save/load. Note that runtime alterations of the maximum value are not normally intended by gradio but you can still do it anyway via the `{"__type__": "update"}` thingy.


USBhost:
So I have been testing this pull and I am running into some issues.
```
export CUDA_VISIBLE_DEVICES=0; python server.py --listen --model llama-13b --load-in-8bit
bin /UI/text-generation-webui/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so
Loading llama-13b...
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████| 3/3 [00:31<00:00, 10.49s/it]Loaded the model in 34.00 seconds.

Loading raw text file dataset...
Getting model ready...
Prepping for training...
Creating new LoRA model...
Starting training...
/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/optimization.py:391: FutureWarning: This implementation of AdamW is deprecated and will be removed in a future version. Use the PyTorch implementation torch.optim.AdamW instead, or set `no_deprecation_warning=True` to disable this warning
  warnings.warn(
Exception in thread Thread-4 (threaded_run):
  File "/usr/local/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.10/threading.py", line 953, in run
  File "/UI/text-generation-webui/modules/training.py", line 335, in threaded_run
    trainer.train()
  File "/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/trainer.py", line 1662, in train
    return inner_training_loop(
  File "/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/trainer.py", line 2053, in _inner_training_loop
    train_loss = self._total_loss_scalar / self.state.global_step
ZeroDivisionError: float division by zero
Training complete, saving...
Training interrupted.
```
So I have a A6000 and other than this error is it normal when it's trying to save the lora it uses 2x more vram than what it takes to train?

mcmonkey4eva:
That error is a code error within the underlying peft/Transformers trainer itself that happens if you interrupt after it starts training but before it completes a single step. I tried to mitigate this by having an interrupt check immediately before starting, but it's still possible to hit if you time the button press just wrong enough. It's literally just the internal code breaking if you tell it to stop with 0 steps completed.

The VRAM jump is not something I've observed in my testing at all. I've done a few tests that were *right on the brink* of running out of VRAM so a genuine VRAM jump when saving would have been an instant crash.
I also, uh, can't think of any logical reason that would happen? Other than maybe some wild bug in the depths of peft where it copies the entire model... twice? (for a jump from 15 to 40) in VRAM before saving it. That doesn't make sense though.

USBhost:
image
image
When It does not error out do to float division by zero error, it saves? but for a 32 rank is 4k adapter_model.bin size a little small?

USBhost:
This is what I'm talking about, something's not right. @mcmonkey4eva 
image

```
export CUDA_VISIBLE_DEVICES=0; python server.py --listen --model llama-13b --load-in-8bit
bin /UI/text-generation-webui/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so
Loading llama-13b...
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:12<00:00,  4.05s/it]
Loaded the model in 14.78 seconds.

Loading raw text file dataset...
Getting model ready...
Prepping for training...
Creating new LoRA model...
Starting training...
/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/optimization.py:391: FutureWarning: This implementation of AdamW is deprecated and will be removed in a future version. Use the PyTorch implementation torch.optim.AdamW instead, or set `no_deprecation_warning=True` to disable this warning
  warnings.warn(
{'loss': 2.3382, 'learning_rate': 1.4999999999999999e-05, 'epoch': 0.11}
{'loss': 2.3343, 'learning_rate': 2.9999999999999997e-05, 'epoch': 0.21}
{'loss': 2.3235, 'learning_rate': 4.4999999999999996e-05, 'epoch': 0.32}
{'loss': 2.2757, 'learning_rate': 5.9999999999999995e-05, 'epoch': 0.43}
{'loss': 2.2326, 'learning_rate': 7.5e-05, 'epoch': 0.53}
{'loss': 2.1719, 'learning_rate': 8.999999999999999e-05, 'epoch': 0.64}
{'loss': 2.1609, 'learning_rate': 0.00010499999999999999, 'epoch': 0.75}
{'loss': 2.1264, 'learning_rate': 0.00011999999999999999, 'epoch': 0.85}
{'loss': 2.1106, 'learning_rate': 0.000135, 'epoch': 0.96}
{'train_runtime': 3049.5937, 'train_samples_per_second': 1.966, 'train_steps_per_second': 0.015, 'train_loss': 2.226774759914564, 'epoch': 0.98}
Training complete, saving...
Training complete!
```
image

USBhost:
image
I oom training 30b because the vram jumps 2x from what it takes to train.

mcmonkey4eva:
... I'm at a loss actually, something is indeed wrong with saving. `adapter_model.bin` is outputting a 443 byte file - an empty pickle with no data. Wtf happened? I don't think anything in this PR or other recent code changes could have / should have caused that? It might be an issue with the peft update?

USBhost:
Did you see any VRAM issues? Could peft have caused it as well?

oobabooga:
If the issue of `adapter_model.bin` being incorrectly saved is due to the PEFT library, it should be fine to merge this PR. Eventually it will be fixed on its own.

mcmonkey4eva:
No I didn't replicate any VRAM issues while testing.

Yes I'm not aware of any issue blocking this PR from merging, should be good to go.

All current possible issues are underlying peft things (if the VRAM issue @USBhost is having is confirmed, that is very much also a peft issue).

oobabooga:
Thank you for this PR @mcmonkey4eva, the new features here are outstanding. It's also very nice that you submitted a ticket to peft for the issue that we cannot handle here.

GamerUntouch:
> So I have been testing this pull and I am running into some issues.
> 
> ```
> export CUDA_VISIBLE_DEVICES=0; python server.py --listen --model llama-13b --load-in-8bit
> bin /UI/text-generation-webui/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so
> Loading llama-13b...
> Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████| 3/3 [00:31<00:00, 10.49s/it]Loaded the model in 34.00 seconds.
> Running on local URL:  http://0.0.0.0:7860
> 
> To create a public link, set `share=True` in `launch()`.
> Loading raw text file dataset...
> Getting model ready...
> Prepping for training...
> Creating new LoRA model...
> Starting training...
> /UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/optimization.py:391: FutureWarning: This implementation of AdamW is deprecated and will be removed in a future version. Use the PyTorch implementation torch.optim.AdamW instead, or set `no_deprecation_warning=True` to disable this warning
>   warnings.warn(
> Exception in thread Thread-4 (threaded_run):
>   File "/usr/local/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
>     self.run()
>   File "/usr/local/lib/python3.10/threading.py", line 953, in run
>     self._target(*self._args, **self._kwargs)
>   File "/UI/text-generation-webui/modules/training.py", line 335, in threaded_run
>     trainer.train()
>   File "/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/trainer.py", line 1662, in train
>     return inner_training_loop(
>   File "/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/trainer.py", line 2053, in _inner_training_loop
>     train_loss = self._total_loss_scalar / self.state.global_step
> ZeroDivisionError: float division by zero
> Training complete, saving...
> Training interrupted.
> ```
> 
> So I have a A6000 and other than this error is it normal when it's trying to save the lora it uses 2x more vram than what it takes to train?
> 
> 13B 8bit lora training just using ooba's defualts I changed nothing. 15648MiB / 49140MiB, The moment It tries to save that 15GiB jumps to 40GiB.

Did some testing, this appears to be a problem with bitsandbytes, the recent push (0.38.0 five days ago) seems to have messed something up with LoRA training. 0.37.2 is working for me.

mcmonkey4eva:
Oo, that explains why I never saw it! I haven't tried updating bitsandbytes for a while. That's good info, thank you, we can notify the other users getting VRAM spikes to downdate it.

USBhost:
I can confirm going back to 0.37.2 fixed my vram jump.

USBhost:
https://github.com/TimDettmers/bitsandbytes/issues/324
Finally got reported to bitsandbytes

Issue #1091, opened 2023-04-12
"--listen" not working (?!)


Malrama:
### Describe the bug

So, I am not 100% sure what "--listen" is supposed to do, but I assume it makes my client visible for other browsers in my network (or even on the internet with port forwarding)?

So I used "--listen" and everything is setup 0.0.0.0:7860. So dar so good.
Localhost:7860 is working perfectly.

However, I can not access Text-generation-webui either within my homenetwork or outside my homenetwork with my IP-Adresse.

Neither 192.XXX[...]:7860 (Homenetwork) nor 77:XXX[...]:7860 (Internet) is working when I try to connect from my phone.
All ports are free and setup.
Am I missing something? Maybe I don't quite understand what "--listen" does but I hope and assume that it let's me run this software from different devices/networks? Like "--web" in InvokeAI?

I am not a big fan of "--share" because that involves a (not mine) webserver inbetween.
Thank you and greetings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run with "--listen" and then try to connect from a different device to your Network-Adress or Internet-IP adress.

### Screenshot

Run with "--listen" and then try to connect from a different device to your Homenetwork-Adressor Internet-IP adress.

### Logs

```shell
Nothing bad
```

### System Info

```shell
All up-to-date
12GB Nvidia RTX 3060 OC
Ryzen 7 3700X
Everything working good!
```

dblacknc:
It has worked well for me all along, including today after a git pull.  OS is Ubuntu 22.04.

A firewall could block access from the LAN, while allowing localhost to work.

Malrama:
> It has worked well for me all along, including today after a git pull. OS is Ubuntu 22.04.
> 
> A firewall could block access from the LAN, while allowing localhost to work.

Weird. I have port 7860 open in my firewall and my Router is forwading the port to my PC. I have the same setup as in InvokeAI and it works there but not for this software. Not locally, not on the internet. Am I missing something at this point? Am I crazy?

dblacknc:
On either Windows or Linux, you could try "netstat -a -n" to verify what address a socket is bound to.  I'm using --listen --listen-port 80.  For your netstat -a -an output, you'd look for :7860.  On Linux "netstat -an | fgrep :80" returns for the first line:

tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN     

[...]

PR #1122, opened 2023-04-13
Add --listen-host launch option

v0xie:
For #1091 

This enables specifying the host name of your listen server with `--listen-host`.

WSL2 has weird networking which causes the default '0.0.0.0' option to fail in my case. Changing '0.0.0.0' to 'localhost' in server.py fixes the issue for me.

This is just a convenience option for those who need it.

### DIFF
--- a/server.py
+++ b/server.py
@@ -809,7 +809,7 @@ def create_interface():
     # Launch the interface
     shared.gradio['interface'].queue()
     if shared.args.listen:
-        shared.gradio['interface'].launch(prevent_thread_lock=True, share=shared.args.share, server_name=shared.args.listen_host if shared.args.listen_host else '0.0.0.0', server_port=shared.args.listen_port, inbrowser=shared.args.auto_launch, auth=auth)
+        shared.gradio['interface'].launch(prevent_thread_lock=True, share=shared.args.share, server_name=shared.args.listen_host or '0.0.0.0', server_port=shared.args.listen_port, inbrowser=shared.args.auto_launch, auth=auth)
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
If `localhost` works, just drop `--listen`. The default interface is localhost, `--listen` changes it to `0.0.0.0`

That said, being able to configure a host is still potentially useful in some obscure cases so 👍

server.py
```diff
@@ -809,7 +809,7 @@ def create_interface():
     # Launch the interface
     shared.gradio['interface'].queue()
     if shared.args.listen:
-        shared.gradio['interface'].launch(prevent_thread_lock=True, share=shared.args.share, server_name='0.0.0.0', server_port=shared.args.listen_port, inbrowser=shared.args.auto_launch, auth=auth)
+        shared.gradio['interface'].launch(prevent_thread_lock=True, share=shared.args.share, server_name=shared.args.listen_host if shared.args.listen_host else '0.0.0.0', server_port=shared.args.listen_port, inbrowser=shared.args.auto_launch, auth=auth)
```
mcmonkey4eva, line 812: Minor Python syntax thing: should be able to just do `shared.args.listen_host or '0.0.0.0'` and it should work the same as this

v0xie: I've updated the PR with the new syntax.

oobabooga:
Looks good. This is not useful to me, but it might be useful for others. 

Thank you @v0xie and thanks @mcmonkey4eva for the review :D

PR #1147, opened 2023-04-13
Use current input for Impersonate (continue impersonate feature)

kamyker:
Makes Impersonate use current input instead of replacing it.

Also changed the space to be inserted after generation attempts. Not sure what was the purpose in doing it before them.


kamyker:
merged with recent changes

mcmonkey4eva (approved):
This will need to be rebased/merged onto main before it can be pulled -- but I tested and it's working exactly as intended, and is definitely a nice feature to have.

kamyker:
> This will need to be rebased/merged onto main before it can be pulled -- but I tested and it's working exactly as intended, and is definitely a nice feature to have.

I prob resolved it using github gui

kamyker:
well, not anymore

PR #1166, opened 2023-04-14
changing the params max_new_tokens/repetition_penalty to match the api documentation

Regrant-Dev:
Noticed the max_new_tokens/repetition_penalty params were not matching the documentation. Changed the values to match documentation parameter names.

max_length to max_new_tokens
rep_pen to repetition_penalty

### DIFF
--- a/extensions/api/script.py
+++ b/extensions/api/script.py
@@ -47,7 +47,7 @@ class Handler(BaseHTTPRequestHandler):
                 'temperature': float(body.get('temperature', 0.5)),
                 'top_p': float(body.get('top_p', 1)),
                 'typical_p': float(body.get('typical', 1)),
-                'repetition_penalty': float(body.get('rep_pen', 1.1)),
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
You're mixing things up here- the `api` extension is for a Kobold-like API, as used by programs such as TavernAI. The names you're editing correspond to the names those programs expect, as seen eg @ https://github.com/TavernAI/TavernAI/blob/6e6d84f020ac963f452dc4b97679a3bfe7848d68/server.js#L261

Regrant-Dev:
> You're mixing things up here- the `api` extension is text-gen-webui's own API, it's an alternative extension for a Kobold-like API, as used by programs such as TavernAI. The names you're editing correspond to the names those programs expect, as seen eg @ https://github.com/TavernAI/TavernAI/blob/6e6d84f020ac963f452dc4b97679a3bfe7848d68/server.js#L261

Ah gotcha! My bad! Ill go ahead and close this. Thanks!

PR #1167, opened 2023-04-14
Add model-specific parameters

oobabooga:
These are defined through regex patterns in `models/config.yaml`.

interface parameters like `wbits`, `groupsize`, `Mode` (instruct/cai-chat) are automatically determined and set based on the model name.

Currently only works for models selected from the dropdown menu in the UI.

- [x] Make this also work for models selected through `--model` or `--model-menu`
- [ ] ~~Add a button called "Save current settings" to the Models tab that saves user settings to `model/config-user.yaml`, for user-specific stuff like always loading some model in 8-bit mode~~ I'll do this later

### DIFF
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -8,10 +8,10 @@ llama-[0-9]*b-4bit$:
   wbits: 4
   groupsize: -1
   model_type: 'llama'
-.*-4bit-(gr128|128g):
+.*-(4bit|int4)-(gr128|128g):
   wbits: 4
   groupsize: 128
-.*-(gr128|128g)-4bit:
+.*-(gr128|128g)-(4bit|int4):
   wbits: 4
   groupsize: 128
 .*-3bit-(gr128|128g):
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -4,16 +4,22 @@
   groupsize: 'None'
   pre_layer: 0
   mode: 'cai-chat'
-.*-4bit-128g:
-  wbits: 4
-  groupsize: 128
-.*-128g-4bit:
-  wbits: 4
-  groupsize: 128
 llama-[0-9]*b-4bit$:
   wbits: 4
   groupsize: -1
   model_type: 'llama'
+.*-4bit-(gr128|128g):
+  wbits: 4
+  groupsize: 128
+.*-(gr128|128g)-4bit:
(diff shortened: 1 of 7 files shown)


server.py
```diff
@@ -801,13 +676,133 @@ def create_interface():
...
+                ui.gather_interface_values, [shared.gradio[k] for k in shared.input_elements], shared.gradio['interface_state']).then(
+                generate_reply, shared.input_params, output_params, show_progress=shared.args.no_stream)  # .then(
+                # None, None, None, _js="() => {element = document.getElementsByTagName('textarea')[0]; element.scrollTop = element.scrollHeight}")
+            )
+
+            shared.gradio['Stop'].click(stop_everything_event, None, None, queue=False, cancels=gen_events if shared.args.no_stream else None)
+            shared.gradio['prompt_menu'].change(load_prompt, [shared.gradio['prompt_menu']], [shared.gradio['textbox']], show_progress=False)
+            shared.gradio['save_prompt'].click(save_prompt, [shared.gradio['textbox']], [shared.gradio['status']], show_progress=False)
+            shared.gradio['interface'].load(None, None, None, _js=f"() => {{{ui.main_js}}}")
+
```
oobabooga, line 804: This looks like a lot of changes, but all I did was move the event handlers below all the element definitions, and group the elements that notebook and default modes had in common.

models/config.yaml
```diff
@@ -0,0 +1,25 @@
+.*:
+  wbits: 'None'
+  model_type: 'None'
+  groupsize: 'None'
+  pre_layer: 0
+  mode: 'cai-chat'
+.*-4bit-128g:
+  wbits: 4
+  groupsize: 128
+.*-128g-4bit:
```
mcmonkey4eva, line 10: I'd swap to `(gr128|128g)` to account for both ways it's commonly written.

Also, `3bit` gr128 models also exist and are picking up popularity (Notably, llama-30b-3bit-gr128 can fit max context length on a 3090/4090, whereas 4bit hits VRAM limit at just over 1000 token context, so 3bit is preferable for that)

oobabooga: Done here https://github.com/oobabooga/text-generation-webui/pull/1167/commits/2f27df702a1c765586b298422db8a0118fafa643

PR #1204, opened 2023-04-14
XY Grid

ClayShoaf:
I finally have something functioning. It's not pretty and it still has some vestigial stuff in it from when I was testing and figuring things out. It's not ready to be merged in yet, but I wanted to put something out there just so people could see that it's being worked on and to get any feedback.

So far, the only two axes you can use are paremeter presets and comma separated prompts. If you want to use commas in a prompt, then put your prompt in quotes like this:
```
this is a prompt, "this is a prompt, but with quotes", this is a third prompt
```

There's a lot of stuff that I am itching to add to it, but I have to get out the door right now and I'm already late. If you have any questions or comments, please let me know.

I'll put more explanation soon. Probably tomorrow.

Here's an example output:
image
Gradio doesn't seem to like HTML tables. I'll work on that tomorrow.


ClayShoaf:
oh god, all my commits... I think maybe I need to do `rebase`, but I'm not exactly sure how that works.

ClayShoaf:
I think I have to make a new branch to fix it. Closing this

PR #1210, opened 2023-04-15
XY Grid

ClayShoaf:
**EDIT:** It should be ready to go. I would like for at least one other person to test it before I take this out of draft again. I still kind of want to add LoRA to the axis options, but I'm hesistant to add it until the whole monkeypatch thing is fully worked out.

For prompts, they are comma separated, but if you want to include commas in a prompt, you can enclose it in quotes, like so:
```
this is a prompt, "this is a prompt, but with a comma", this is a third prompt
```
There is still some vestigial stuff in here from trying to figure out how to get the information correctly. It will be cleaned up and improved in the coming days.

Here is an example output:
image

(if you want to track my progress, I'm working on [this branch](https://github.com/ClayShoaf/text-generation-webui/commits/testing_xy) before merging "major" commits.)

Feedback is welcome!


Loufe:
You may consider adding the "draft" label if you don't mean for this to be merged right away.

ClayShoaf:
@Loufe I wasn't aware of that. Thank you!

ClayShoaf:
Still working hard on this, but I don't use the computer on Sundays, so my next commit to this branch will probably be on monday. If you are watching this thread, so far I have prompts, presets, and characters for options (in an, admittedly, dirty implementation for parts of it). You can see how far along I am here: https://github.com/ClayShoaf/text-generation-webui/commits/testing_xy

If there are any settings that you'd like available for axis, please let me know.

ClayShoaf:
Made a major improvement to the UI. It's coming along nicely, but it works a lot better now. Hopefully someone can test it out and tell me what they think so far.

ClayShoaf:
Ok, I think I'm finally ready for review. I'll probably add more axis options tomorrow. Specifically, I want to add options for individual parameter presets (shouldn't be hard at all). I'd also like to add LoRAs and models to the mix, but I think I'll wait until they are a little easier to use and the standards are a little more nailed down.

I think it looks pretty good in the web interface, and I did a fair bit of bug testing, but there's nothing like sending something out to the public to really figure out where the problems are. Would appreciate any help with tightening  up my code, as I'm still a novice. It looked a lot jankier before than it does now, though.

kaiokendev:
@ClayShoaf 
I think you need to clear the history buffer before each run, as it is contaminating the prompt for subsequent generations. A user noted the results of my LoRA from the grid implied the LoRA was overfitted, but from my debugging it is populating the context with the first response generated, which inherently fits future generations to produce a similar response

grid bug fixed

ClayShoaf:
@kaiokendev Good catch! That fix will work for this particular use case, but it's a little more complicated, because we don't necessarily want  to remove all of the history in all cases. This would make the "use chat history" button non-functional, for when people want to include the whole history to influence a prompt. I'll figure out where and what exactly needs to be changed on Monday. I just kind of threw instruct mode in there as an afterthought, because I had a feeling people might want to use that mode.

I'll go ahead and throw this back into draft mode until it is working properly. I also need to fix the way that "-1" is handled for the seed options. I didn't think about this until today, but if you put in a seed of "-1" on one of the axes, it's going to generate a new seed for each cell, rather than the whole column/row.

I also want to change how the table is output to make the columns uniform in size.

ClayShoaf:
@kaiokendev I am not able to test it right now, but I took a quick look, and I suspect, in line 347 (as well as it's respective duplicates in other parts of the code), if `>` is changed to `>=`, that may fix it. I know I didn't include that for no reason, though. So, it might cause other errors. I remember I was getting errors while working on it, but I can't remember the exact variables. Fixing it may end up being a little more complicated. I might break out the logic into a couple of different files, for instruct vs. chat. I also just want to clean it up a bit as well. The mass of blocks is becoming a little unwieldy. Maybe that's just the nature of the beast. This is the most complicated thing I've worked on (not a programmer by trade).

extensions/xy_grid/script.py
```diff
@@ -0,0 +1,551 @@
...
+                    bot_output = convert_to_markdown(gen_output[-1][1])
+
+                    if custom_state['mode'] == 'instruct':
+                        output = output + f"<td><b>{custom_state['name1']}</b> {user_output}<b>{custom_state['name2']}</b> {bot_output}</td>"
+                    else:
+                        output = output + f"<td><h3><b>{custom_state['name1']}:</b></h3> {user_output}<h3><b>{custom_state['name2']}:</b></h3> {bot_output}</td>"
+
+                    # Remove the last outputs, so they don't influence future generations
+                    gen_output.pop()
+                    if len(shared.history['internal']) > 1:
```
ClayShoaf, line 347: This was the fix for the context poisoning that was happening with chat mode

extensions/xy_grid/script.py
```diff
@@ -146,274 +148,131 @@ def parse_axis(axis, value):
...
-            return "<h1><span style=\"color: red;\">ERROR: both fields are empty</span>"
-
+                elif len(shared.history['internal']) == 1:
+                    if shared.history['internal'] == gen_output:
+                        shared.history['internal'].clear()
+                if len(shared.history['visible']) > 1:
+                    shared.history['visible'].pop()
+                elif len(shared.history['visible']) == 1:
+                    if shared.history['visible'] == gen_output:
+                        shared.history['visible'].clear()
```
ClayShoaf, line 273: @kaiokendev I fixed the context poisoning with this block here. Can you test the new script and see if you find any bugs?

ClayShoaf: I'm not sure how or if I can show more lines of code with my comment, but the instruct portion is at the top of this block, a few lines up.

kaiokendev: Works correctly! one thing I forgot to mention before -- in the output file saving you should change from
>mkdir

to 
>os.makedirs(output_folder)

So it can make the intermediate folders if they don't already exists
Thanks for fixing this

ClayShoaf: Alright, I'll do that. I was considering putting this in an `outputs/xy_grid` folder in the root directory, but I don't know how @oobabooga would feel about that. Personally, I hate when extensions output things to their own folder in SD, but it felt a little presumptuous.

I also would like to include models as an axis option, but it's gonna take a little work to figure out exactly how I want to go about getting the correct instruct characters lined up with the correct models. If you were to load three different models with 3 different instruct characters, it would be pretty much useless for 2/3 of the generations.

oobabooga:
Some comments:

1. This needs a refactor. In the web UI, the parameters are not updated in the backend for every change. Instead, they are all collected at once by the `ui.gather_interface_values()` function. In `server.py`, several examples of use of this function can be found.
2. I agree that having the option of using different models as one of the axis would be very useful. It's not obvious to me how to do that. The closest thing so far is the perplexity evaluation tab, where you can select multiple models from the dropdown and then click on "Evaluate" to evaluate all of them, one at a time.
3. Saving outputs to the root folder: all extensions so far use the subfolder for saved files (silero comes to mind). It's not too important to worry about that as it can be changed later

ClayShoaf:
> 1. This needs a refactor. In the web UI, the parameters are not updated in the backend for every change. Instead, they are all collected at once by the `ui.gather_interface_values()` function. In `server.py`, several examples of use of this function can be found.

I'll play around with that. The method I'm using now came from a comment in a discussion thread, because I couldn't figure out how to access those values. I know the way I'm doing it now is pretty hacky.

> 2. I agree that having the option of using different models as one of the axis would be very useful. It's not obvious to me how to do that. The closest thing so far is the perplexity evaluation tab, where you can select multiple models from the dropdown and then click on "Evaluate" to evaluate all of them, one at a time.

My initial thought was to either try to parse the model names and auto select the instruct character and include it in the axis textbox with some sort of delimiter, but I think that would look really messy. I also thought about having another dropdown appear for models that would have respective characters for models, but that might be too confusing for an end user, not to mention easy to mess up (although adding pause/stop buttons and live loading the grid should help with that). I'll take a look at the perplexity evaluation and see if I can draw any inspiration (or, you know, just take)

> 3. Saving outputs to the root folder: all extensions so far use the subfolder for saved files (silero comes to mind). It's not too important to worry about that as it can be changed later

Yea, that's pretty much been my method so far. Just try to format things the way everyone else is doing it. Personally I'm a fan of camel case, but when in Rome...

I appreciate the reply.

oobabooga:
> My initial thought was to either try to parse the model names and auto select the instruct character 

This is what the web UI does when you select a model from the dropdown. For instance, if you select a model with "alpaca" in its name, `instruct` mode with the `Alpaca` template will be automatically set. It does that based on the information in [models/config.yaml](https://github.com/oobabooga/text-generation-webui/blob/main/models/config.yaml)

The solution is probably to reuse the `get_model_specific_settings` function to find the appropriate templates and then manually load them.

https://github.com/oobabooga/text-generation-webui/blob/main/server.py#L254

ClayShoaf:
I'm waiting for https://github.com/huggingface/peft/issues/430 to be resolved before I continue working on this.

oobabooga:
Closing this for inactivity.

PR #1215, opened 2023-04-15
Suppress "WARNING: Did not find..." for Whisper_STT  pip install

Loufe:
image

This annoying warning is caused when referencing the "short" form of a git commit. [Using the full, 40 character, version suppresses it. ](https://github.com/pypa/pip/pull/4674)


oobabooga:
Thanks!

Issue #1161, opened 2023-04-14
Improve Databricks Dolly support (special tokens?)


mcmonkey4eva:
Dolly: https://www.databricks.com/blog/2023/04/12/dolly-first-open-commercially-viable-instruction-tuned-llm
Can grab 12B variant from https://huggingface.co/databricks/dolly-v2-12b

-----

I think Dolly has its format inputs encoded as special tokens, or something? `### Instruction:`, `### Response:`,` ### End` are its 3 magic labels.

I noticed while running that they seem to "disappear" from the `--verbose` log. My initial guess is, what happens is they get encoded to a special token, and the decode stage doesn't restore them properly.

This theory is backed up, by, yknow, `special_tokens_map.json` having them marked as special tokens. (I notice it also has bos/pad/unk set as `<|endoftext|>` which seems questionable but idk).

The tokenizer is a `json` file, so I can actually see that these are tokens 50277, 50278, 50279, all considered `"special": true`.
I imagine the `special` status affects either the encode or decode stage in some way.

----

More importantly than `verbose` having them missing, I've noticed that after it's done generating, it will output additional lines starting with `ing` or similar. My theory is that it outputs `### End` and that token isn't recognized as the end piece, and so it tries to continue from there as `Ending (... more sentence here)`.
I tried setting stop phrase to `### End` but that doesn't appear to work as the stop.

Based on that, my secondary guess is maybe `encode` entirely fails to process the special tokens, and they're being omitted entirely?

(Which would explain why `### End` as stop input would be emitted by the LLM but not found by the stop token search)

----

I'm submitting this as a feature request as (A) Dolly is a new model and so _adding support_ is a new feature, and (B) the bug-report issue template isn't particularly conducive to this topic.

oobabooga:
This is also an issue with galactica and open assistant, which depend on special tokens to perform certain tasks. I have hard coded a regex for the model names in the decode function as a workaround:

https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L59

which is not ideal.

mcmonkey4eva:
Adding `dolly` to that matcher indeed fixes the `--verbose` output to contain the text properly. However while testing it generated:
```
### End!

More unwanted text here
```

Despite having
![image](https://user-images.githubusercontent.com/4000772/231927461-567bdc1b-979b-486e-866e-f8702cd01e74.png)

[...]

PR #1218, opened 2023-04-15
Add skip_special_tokens checkbox for Dolly model

oobabooga:
It is disabled by default for Dolly, GALACTICA, and Open Assistant.

For #1161

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -56,14 +56,13 @@ def encode(prompt, add_special_tokens=True, add_bos_token=True, truncation_lengt
         return input_ids.cuda()
 
 
-def decode(output_ids):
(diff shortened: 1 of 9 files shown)


modules/text_generation.py
```diff
@@ -183,7 +182,7 @@ def generate_reply(question, state, eos_token=None, stopping_strings=[]):
     output = input_ids[0]
 
     if shared.args.verbose:
-        print(f'\n\n{decode(input_ids[0])}\n--------------------\n')
+        print('\n\n{decode(input_ids[0], state["skip_special_tokens"])}\n--------------------\n')
```
mcmonkey4eva, line 185: The removal of `f` here breaks verbose output

oobabooga: Oops, fixed it, thank you.

PR #1240, opened 2023-04-15
Add French TTS

johngtrs:
I tried to update TTS voice dropdown with another "TTS Language" dropdown but I don't know how to update the choices list Dropdown in Gradio.

So, I'm sorry for this bad code :/

### DIFF
--- a/extensions/silero_tts/script.py
+++ b/extensions/silero_tts/script.py
@@ -31,6 +31,13 @@ voice_pitches = ['x-low', 'low', 'medium', 'high', 'x-high']
 voice_speeds = ['x-slow', 'slow', 'medium', 'fast', 'x-fast']
 streaming_state = shared.args.no_stream  # remember if chat streaming was enabled
 
+# Add french TTS
+if 'tts_french' in shared.args.extensions:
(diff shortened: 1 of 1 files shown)


da3dsoul:
Personally, I would build a map of models, languages, and speakers, so that we don't have args cluttered with irrelevant options. Silero is also not the only TTS system, so making args that are only relevant to it is not good.

Gradio dropdowns are pretty easy (note: this is pseudocode and isn't done):
```py
model_dropdown = gr.Dropdown(value=params['model_id'], choices=models, label='Model')
speaker_dropdown = gr.Dropdown(value=params['speaker'], choices=speakers['model_id'], label='Speaker')

model_dropdown.change(lambda x: update_model(x, speaker_dropdown), model_dropdown, None)
speaker_dropdown.change(lambda x: params.update({"speaker": x}), speaker_dropdown, None)

def update_model(x, speaker_dropdown):
    global params, language, speaker
    params.update({"model_id": x})
    # update language
    language = pull it from a map or something
    # update speakers
    speaker = find a default speaker for the model and set it
    speaker_dropdown.update(value=speaker, choices=speakers['model_id'])
```

PR #1242, opened 2023-04-15
Allow extensions to have custom display_name

ye7iaserag:
Mainly title, extensions are displayed in gradio by folder name which is sometimes ugly
This PR introduces a new field called `display_name` that is read from params dict when the extension is loaded


Brawlence (approved):
Small but convenient change, LGTM

PR #1254, opened 2023-04-16
Searx Integration Extension

GuizzyQC:
Created a script that allows users who have access to a Searx instance's API to forward search requests, that will feed the prompt to the AI with new information about a topic.


GuizzyQC:
Improved on in #1427

Closing

PR #1256, opened 2023-04-16
Update models.py to support chatGLM

Ph0rk0z:
It generates for me in cai-chat. I hope I didn't make any typos. Shouldn't affect other models and make them do insecure things.


oobabooga:
Does this model require the prompt to be formatted in a special way like Alpaca?

oobabooga:
Here is the usage

```
python download-model.py THUDM/chatglm-6b
python server.py --model THUDM_chatglm-6b --chat --trust-remote-code
```

PR #1277, opened 2023-04-16
Added an audible notification after text generation in web.

HappyWorldGames:
If there is a file "notification.mp3" in the root.


mcmonkey4eva:
Could you rebase/merge against the new main? There's been changes to the same sections of code which prevent it from merging.

Also, perhaps add a comment in the docs somewhere about this?

When I implemented the same feature for a different project I tossed in a link to https://bigsoundbank.com/search?q=notification as an example of where to get files for it.

mcmonkey4eva (approved):
Aside from the above-noted merge issue and suggested docs, I tested this PR and it works as intended. Notification sound plays in-web-browser when generating is done.

PR #1290, opened 2023-04-17
Fix TypeError in browser console

DavG25:
This should fix `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')` appearing in the browser console when the interface loads

When JS is loaded using `Blocks`, it seems that Gradio expects an output, so in this case we create a dummy and hidden label as the output


DavG25:
Closing this PR as the issue has been fixed in Gradio 3.28

PR #1321, opened 2023-04-17
Update google_translate extension (save settings, LibreTranslate support, LocalAlpaca support)

janvarev:
- now it saves language settings in settings.json file (no needs to edit Language setting every start)
- choose between GoogleTranslator and LibreTranslator
- option to set a custom url for LibreTranslator server (for example, for local offline LibreTranslator installation)

LibreTranslate is Free and Open Source Machine Translation API, entirely self-hosted. Unlike other APIs, it doesn't rely on proprietary providers such as Google or Azure to perform translations. (installing as simple as pip install libretranslate )
How to setup your own LibreTranslate translation server: https://github.com/LibreTranslate/LibreTranslate

image


janvarev:
UPD: I've added new option - LocalAlpaca translator.
If you have Alpaca model loaded in memory and choose this option, engine will try to translate your query using loaded model. Also you can check results in console.

So, the workflow is:
- Run textgen with "Translate phrase from Lang to English"
- Run prompt
- Run textgen with "Translate phrase from English to Lang" on result

It sucessfully works, at least in some cases.

Also you can add Vicuna tpl etc., I just didn't have a time to do it.

janvarev:
UPD: added checkbox for enable/disable user/system translation, and also to add English original response to output
image

janvarev:
@oobabooga This PR **copy old default google_translate behaviour** (only change is that default language is English (no translation), instead of Japanese), so I hope it will be merged in close future.

visteras:
I am also looking forward to these changes. The code looks normal to me, is there any reason for the delay?

janvarev:
Moved to separate extension https://github.com/janvarev/multi_translate

PR #1322, opened 2023-04-17
Add an "Evaluate" tab to calculate the perplexities of models

oobabooga:
A large number of models and LoRAs are being released every week, and it has become confusing to tell which one performs best.

The purpose of this PR is to introduce an "Evaluate" sub-tab inside the "Training" tab where the user can calculate the perplexity of the currently loaded model when applied to some some arbitrary text file. This is inspired by the tables in https://github.com/qwopqwop200/GPTQ-for-LLaMa#result.

My implementation is probably very dumb, and is sent here with the goal of annoying someone who actually knows what they are doing so that they will tell me how to do this the right way.

This is what the interface currently looks like:

print

Some ways of improving this could be:

- [ ] Somehow define multiple model/LoRA combinations at once and print a table as the output. Right now, you have manually select each model and copy and paste the perplexities that are printed.
- [ ] Make it also work with structured datasets like one used for Alpaca.

Cc @mcmonkey4eva


oobabooga:
https://github.com/oobabooga/GPTQ-for-LLaMa/blob/cuda/llama.py#L301

modules/evaluate.py
```diff
@@ -0,0 +1,38 @@
...
+        text = f.read()
+
+    # Tokenize the text
+    tokens = encode(text, add_special_tokens=False)
+
+    # Loop through the tokens and generate text token-by-token
+    yield "Evaluating..."
+    perplexity_sum = 0.0
+    print(context_size, tokens.shape[1], window)
+    for i in tqdm(range(context_size, tokens.shape[1], window)):
```
mcmonkey4eva, line 24: I'm confused about the `window` value here - at first from the var name I thought it referred to the size of the `context window`, but, `context_size` is that... instead `window` appears to be implemented as the `step` input to `range(...)`, which if I'm not mistaken means that if it's set to any value higher than `1`, tokens will simply be skipped? Eg if `window` == `2`, then only every other token will be tested (and the other skipped).

mcmonkey4eva:
At a glance, the perplexity code looks to be about right in terms of logic (haven't verified the small details).

One thing you might add is to use the streaming generator to improve performance - so eg for context size `256`, input `200` tokens, then stream for the remaining `56` (noting to of course force it to the ground-truth token each time as it goes). This would be a more valid usage of that `window` input. (Looks like maybe that's what you intended to implement there, but just didn't finish?)
(This is labeled as a "strided sliding window" by HuggingFace <https://huggingface.co/docs/transformers/perplexity#calculating-ppl-with-fixedlength-models>)

---

Building a table out of it shouldn't be difficult, mostly a matter of wrangling the code to plug a bunch of difference things together and formatting the input well. If it's just models, or just loras, can easily just do a multiselect Dropdown and be done with it. Being able to select pairings (model X + lora Y) is harder.
Past that it's just, (pseudocode)
```py
table = "Model | Score\n--- | ---\n"
for model_variant in model_inputs:
  load(model_variant)
  table += f"model_variant | {evaluate(context)}\n"
print(table)
```

---

Structured datasets... I mean, option A, just reuse the existing dataset code and loop over it. Or B, reimplement the dataset code since its pretty simple to do (just a JSON list-of-dicts, and loop over the list) - only reason I didn't for the training code is it wants a `Dataset` object as the input anyway so might as well let the code do what it already knows how to do, since it needs to wind up as a Dataset anyway. That doesn't apply to this eval code, though, so you can just load it directly as json.

Ph0rk0z:
Very cool. I have 1.2TB of models now and I'd like to get rid of ones that aren't really worth it. Sometimes after quantizing the evaluations for one reason or another, depending on who wrote the GPTQ file that converts them.

modules/training.py
```diff
@@ -102,28 +103,48 @@ def create_train_interface():
 
         output = gr.Markdown(value="Ready")
 
-        def do_copy_params(lora_name: str):
-            with open(f"{shared.args.lora_dir}/{clean_path(None, lora_name)}/training_parameters.json", 'r', encoding='utf-8') as formatFile:
-                params: dict[str, str] = json.load(formatFile)
+    with gr.Tab('Evaluate the current model', elem_id='lora-train-tab'):
```
mcmonkey4eva, line 106: Bad copypaste on this `elem_id`

oobabooga:
I adapted the transformers code (thanks for the link @mcmonkey4eva), and now the output numbers seem to make more sense. I should probably add some default options like wikitext to the dropdown.

adapt

oobabooga:
It seems to be working well now. The past evaluations are automatically saved/loaded so that you don't have to run the same test more than once.

I tested some 20-30 models on a small dataset that I created, and the general conclusion is that llama-30b 4bit is better than everything else. Even fine tuned models lose to llama-30b at areas that llama-30b was not fine tuned to do.

modules/evaluate.py
oobabooga: You are right, it did not make sense to name that parameter `window`. The proper term is `stride`.

oobabooga:
Some comments about previous open problems:

* LoRAs: I didn't handle them directly. Instead, I created an `Evaluate current model` button that will use the current model as it is for the test. If LoRAs are applied to the model, their names will be added to the pandas dataframe along with the model name.
* This is also the way to handle the problem of testing a model in 8-bit precision vs 16-bit precision. You reload the model manually, and force a new test with `Evaluate current model`. Two identical entries will be added to the dataframe in this case. To allow the user to distinguish between the two, I added a "comment" column that can be edited manually. So I simply write "8-bit" in this column when necessary.
* Structured datasets: it seems to me that the prompt tokens should be ignored in this case, and the evaluation should only happen on the "reply" tokens. I don't know how to implement that, so I left it as an open problem.
* Streaming tokens: same as above, I don't understand what I am doing enough to depart too much from the Hugging Face example. It outputs perplexities that make perfect sense, so I feel safer to just stick to its code.
* Also worth noting: the models are loaded with the default settings, exactly the same as when you load them from the Models tab. So wbits, groupsize, etc, are automatically set.

Ph0rk0z:
This has been a godsend and will show quite a few things.

Worth it to add how long the evaluation took and maybe how many it/s?

PR #1324, opened 2023-04-17
Added custom_stopping_strings option for preset files

Fire-Input:
Added support for `custom_stopping_strings` into preset files.
Make sure to wrap the value in **single quotes**.

Example preset
```
do_sample=True
top_p=0.5
top_k=40
temperature=0.7
repetition_penalty=1.2
typical_p=1.0
custom_stopping_strings='"\nYou:", "\nChatBot:"'
```


oobabooga:
I don't think they belong there. Maybe they would make sense in character yamls, but presets are for generation parameters that affect how new tokens are sampled

PR #1333, opened 2023-04-18
add support for pip installable alpaca_lora_4bit

johnsmith0031:
1. add support for pip installable alpaca_lora_4bit
2. make the monkey patch usable for other 4bit model
3. make find_quantized_model_file independent

Temporarily I will leave:
pip install git+https://github.com/johnsmith0031/alpaca_lora_4bit.git@winglian-setup_pip
in the information, If the branch is merged into main, I think I'll make another PR to adjust it.


Ph0rk0z:
I made it work with dispatching too: https://github.com/Ph0rk0z/text-generation-webui-testing/commit/a2c9bb0e1cbc668402740e82320a23d6c72b1f1d

oobabooga:
@johnsmith0031 what is the benefit of using the pip package vs cloning the repository as is done now?

johnsmith0031:
Just because of using a independent cuda kernel inside it and can be separated from original GPTQ. But currently I see that many code here are already based on current alpaca_lora_4bit which would be broken If I make it pip installable. I think the best choice for me is to keep alpaca_lora_4bit@winglian-setup_pip as another branch. Feel free to use the code here for increasing compatibility with other 4bit models in monkey patch.

```
def load_model_4bit_low_ram(config_path, model_path, groupsize=-1, device_map="auto", is_v1_model=False):

    print(Style.BRIGHT + Fore.CYAN + "Loading Model ...")
    t0 = time.time()

    with accelerate.init_empty_weights():
        config = AutoConfig.from_pretrained(config_path)
        model = AutoModelForCausalLM.from_config(config)
        model = model.eval()
        layers = find_layers(model)
        for name in ['lm_head']: # can add some other modules like 'embed_out' for compatibility for other models
            if name in layers:
                del layers[name]
        make_quant_for_4bit_autograd(model, layers, groupsize=groupsize, is_v1_model=is_v1_model)
    model = accelerate.load_checkpoint_and_dispatch(
        model=model,
        checkpoint=model_path,
        device_map=device_map
    )

    model.half()
    for n, m in model.named_modules():
        if isinstance(m, Autograd4bitQuantLinear):
            if m.is_v1_model:
                m.zeros = m.zeros.half()
            m.scales = m.scales.half()
            m.bias = m.bias.half()
    
    # reduce context VRAM usage by half
    wrapper = AMPWrapper(model)
    wrapper.apply_generate()
    
    try:
        tokenizer = AutoTokenizer.from_pretrained(config_path)
    except HFValidationError as e:
        tokenizer = AutoTokenizer.from_pretrained(model)
    tokenizer.truncation_side = 'left'
    print(Style.BRIGHT + Fore.GREEN + f"Loaded the model in {(time.time()-t0):.2f} seconds.")

    return model, tokenizer
```

oobabooga:
@johnsmith0031 It seems like the `load_model_4bit_low_ram` function is already used in `modules/monkey_patch_gptq_lora.py`, since it gets directly imported there.

Also, what is a v1 model? All my models seem to be v2.

Ph0rk0z:
V1 model is the old GPTQ that didn't support group size or act-order, etc.

johnsmith0031:
> @johnsmith0031 It seems like the `load_model_4bit_low_ram` function is already used in `modules/monkey_patch_gptq_lora.py`, since it gets directly imported there.
> 
> Also, what is a v1 model? All my models seem to be v2.

V1 model is old model format that does not use multiple groups across zeros and scales, and consume less memory than v2 model which use them. The reason to support v1 model is just because lower memory consumption of the model itself can support longer context input on consumer GPUs with limited VRAM.

PR #1334, opened 2023-04-18
Lora Trainer Improvements Part 4: 4-Bit Support and more!

mcmonkey4eva:
Main update: 4-Bit LoRA training support!
tldr: apply the monkeypatch when training, copy a few references from the alpaca-lora-4bit repo, and, boom, working 4bit training!

Notes:

- ~~Monkeypatches have to apply next to the imports, so, uh, imports have been moved inside the train function. This is not a nice or pretty way of doing it and I'd love a better way.~~ no longer applicable.
- ~~The alpaca-lora-4bit repo came with options for flash attention and triton support that seem useful potentially useful but I'm not sure how to translate to the text-gen-webui repo properly, so I've left them commented out but in-place for the moment.~~ Discarded, not relevant to training. If applicable they're better placed in the model loading code anyway.

----

Also in this PR:
- Because 4bit monkeypatch requires up-to-date peft, I couldn't just backdate peft to check, and had to find a workaround to make saving work - and I did and it works!
~~So, in this PR's second commit, I applied the workaround to saving trained LoRA (and reported the workaround I found on <https://github.com/huggingface/peft/issues/317> to help peft figure out the root cause better hopefully).
tldr of that: literally just manually call `torch.save` on the raw `state_dict`, since that's all the file needs to be. This means in practice right now an empty `adapter_model.bin` is saved immediately followed by being overwritten with a valid one.~~
**EDIT:** Got a reply on the issue, was told the actual solution is to remove some outdated state_dict wrangling that was needed before but is now invalid. Tested and working.

- `Save every n steps` now actually saves in the expected format, by way of bypassing HF Trainer save steps and instead using peft to save.

- I got reloading prior saves working reliably now! I've swapped the default of "Override" back to false as I originally intended with it now that it's not broken. You can reliably just reuse an old training run with new or different data whenever you want (as long as your Rank matches, will be a problem if you try to change it). Works by ignoring peft's load method and just using the normal create-new method then shoving the state-dict from file into it. In my testing I see the loss value is fairly consistent when resuming so I do believe it's working fully as intended.

- ~~I reduced warmup steps to be less annoying for short runs. The high value was passable for the original alpaca_lora's intention of a 5 hour megarun of a huge dataset, but doesn't make much sense for anything smaller. Might be better to set it to zero even? Or make configurable. Put that and dropout under an "Advanced" accordion.~~ Added Warmup Steps as a setting in UI.

- Added an `Advanced` accordion now that there are enough applicable settings. Contains Shuffle, Warmup Steps, Dropout Rate, Enable Higher Ranks.

- Added some code to protect against `Copy parameters from` erroring out from eg having new keys that aren't in old files, or clicking a LoRA name that doesn't exist (eg `None`, or one recently deleted). Gradio doesn't provide `**kwargs`, so I had to use `*args` (positional index) which required the code to be a little bit weirder than I'd like, but it works as intended.

- Added a safety check against empty name input.

- Added LR scheduler input. Don't have an easy way to shove a link onto the `info=`. It'd be nice to have a way to contribute to the wiki - a wiki page going into more detail about training would be very valuable. (For the LR Scheduler info, a [link to HF docs](https://moon-ci-docs.huggingface.co/docs/transformers/pr_1/en/main_classes/optimizer_schedules#transformers.SchedulerType) or a copy/paste of the charts from there would be handy to provide a more visual reference of what each option does).

- Now saves in the training thread, to ensure the result is saved even if you close the browser, since gradio decides to kill threads when you do that.

- Added advanced option to select which optimizer to use. Not sure how useful it is in practice but worth having the option to experiment with.

----

I've done only basic tests at the moment - loading up a 4bit LLaMA-13B with `--monkeypatch` and then training a small text file dataset, then loading the LoRA and trying it. The result was an immediately success, unquestionably learned real information straight away, worked flawlessly straight out of the box! (I tested with a copy of some documentation from a project that normal LLaMA doesn't know about at all, gave it no character file and asked a question, and it immediately started dumping exact quotes from the docs I fed it. It learned *so well, so fast* that I'm wondering if maybe learning rate acts differently with the 4bit patch or something?)

I also tested 8bit to make sure it didn't get broken by the changes. Seems to work as intended, including saving now working as normal.

In my initial tests, I got CUDA OOMs and had to turn batch size down. It seems like peft is possibly creating float32s for training? But with batch size low there's no trouble, VRAM usage isn't particularly higher than it is for inference, only few gigs added, which is about the same as having any input context tokens anyway.
We might lower the default batch size on the assumption that the average user probably will go for 4bit going forward?


mcmonkey4eva:
Actual numbers since I know some people will want them.

Tested on an RTX 3090, on Linux, using ~1.3 GiB while nothing is loaded.
VRAM (Inference): 800 tokens input, generating 200 more, no LoRAs loaded.
VRAM (Training): MicroBatch = 1, Batch = 4, Cutoff = 256

Model | VRAM (Idle) | VRAM (Inference) | VRAM (Training)
--- | --- | --- | ---
LLaMA-7B-4Bit-128g | 7.2 GiB | 9.1 to 12.2 GiB | 9.74 GiB
LLaMA-13B-4Bit-128g | 10.6 GiB | 13.24 to 18 GiB | 14.48 GiB

Conclusion: if you can run the model in inference with loaded context, you can train it too (at least with small batch size and cutoff).

janvarev:
Thanks for your work!!
Question (if you have a time to test, of course) : does your solution work with --auto-devices? or --cpu?

mcmonkey4eva:
@janvarev `--auto-devices` / `--gpu-memory`, you have to use `--pre_layer` ... and seemingly neither works _at all_ with `--monkey-patch`. Like it doesn't even process the pre_layer at all it just tries to load the full raw model into VRAM and overloads. So, for now, no, that won't work :(
Same story with `--cpu` it just ignores that and loads into VRAM anyway.
Those problems are external to this PR and are something to do with `monkey-patch` (which, uh, despite the naming coincidence is not directly related to me lol)

janvarev:
@mcmonkey4eva Thanks for answering!

Ph0rk0z:
I wonder if it's possible to use the monkeypatch out of the lora repo and combine it with the new PEFT to stack loras.

If we used the dispatching function out of the repo we can definitely offload.

USBhost:
> Or make configurable. Put that and dropout under an "Advanced" accordion.
@mcmonkey4eva 
Yes please. Also just now reading that and looks at my almost done 14h 30b training... oops

USBhost:
`export CUDA_VISIBLE_DEVICES=0; python server.py --listen --model llama-30b --load-in-8bit
Gradio HTTP request redirected to localhost :)
Loading llama-30b...
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████| 7/7 [01:03<00:00,  9.09s/it]
Loaded the model in 67.77 seconds.

Loading raw text file dataset...
Getting model ready...
Prepping for training...
Creating LoRA model...
Starting training...
/UI/text-generation-webui/venv/lib/python3.10/site-packages/transformers/optimization.py:391: FutureWarning: This implementation of AdamW is deprecated and will be removed in a future version. Use the PyTorch implementation torch.optim.AdamW instead, or set `no_deprecation_warning=True` to disable this warning
  warnings.warn(
{'loss': 2.3527, 'learning_rate': 0.0002294117647058823, 'epoch': 0.8}
{'loss': 2.2529, 'learning_rate': 0.00014117647058823528, 'epoch': 1.6}
{'loss': 2.2158, 'learning_rate': 5.294117647058824e-05, 'epoch': 2.4}
{'train_runtime': 45148.7136, 'train_samples_per_second': 0.376, 'train_steps_per_second': 0.0, 'train_loss': 2.2620022031995983, 'epoch': 2.88}
`

I would like to report a bug. I think when you close the browser it will not save finished lora. checkpoints save as expected but not the finished one. See the above output it does not save the finished LoRA.

mcmonkey4eva:
`Tried to load multiple loras and won't. Even on latest hijacked PEFT.` I mean the main thing on that is just getting the patched hacks pushed upstream so that we're not dealing with dirty hacks that break things. But, until that happens, there might be a way to get them working smarter aka breaking fewer things.

Loading of LoRAs in general is a bit shaky with `monkeypatch` enabled it seems. I've had some weird experiences while playing with it, including my vicuna-4bit model deciding it never wants to run again and act like its full of NaNs now (not sure how/why, I definitely didn't modify the model itself - or at least my OS assures me the file hasn't been modified).

-----

`Yes please.` 👍 

`Since im new to LLM training is that option needed on large datasets?` tbh it's pretty rare you'd want to turn it off, that should probably be put under "Advanced" as well. "Needed" is a hard question, but, I don't think it should ever hurt? Alpaca's trainer just always shuffles by default without an option for it.

`I think when you close the browser it will not save the finished lora. checkpoints save as expected but not the finished one.` yknow I was aware of that but didn't really look into it. I guess it's pretty important. Likely just a matter of making sure the save code happens in the trainer thread rather than the main one. Bit annoying that gradio just silently drops the main thread.

`Also can we chose the type of LR we use?` that's a good question, it clearly has a scheduler going, but not one I've selected. I'll look into finding a way to control it.

`.. then jumps super high.` that's not jumping high, that's going very very low. `5.2e-5` is `0.00005`, just the `e` notation suddenly appearing because of how low it got.
It's probably a linear or cosine scheduler, such that it starts high and slowly drops down over time (this is a standard technique when training AI models, it greatly improves the quality of results, at the cost of taking a bit longer to run). Having an option to control that would be good for sure.

USBhost:
> just the `e` notation suddenly appearing because of how low it got.

Oh I did not catch that.

> Alpaca's trainer just always shuffles by default without an option for it.

Got it.

> yknow I was aware of that but didn't really look into it

Looks at my 14 hour train not finishing. :(

Also It would be nice to control target_modules. I would imagen using all 4 is better than just the default two.

USBhost:
https://moon-ci-docs.huggingface.co/docs/transformers/pr_1/en/main_classes/trainer#transformers.TrainingArguments

Interesting https://moon-ci-docs.huggingface.co/docs/transformers/pr_1/en/main_classes/trainer#transformers.TrainingArguments.lr_scheduler_type

mcmonkey4eva:
`target_modules` is mostly just identifying where the LoRA connects to because it's unique based on model type. I don't _think_ there's any reason to muck with that (unless there's room to do what LyCORIS did and target normally-ignored parts of the model to improve results?)

Ph0rk0z:
> I mean the main thing on that is just getting the patched hacks pushed upstream so that we're not dealing with dirty hacks that break things

I agree. But mine have been pretty much flawless.. just want the stacking to work. Maybe I more deeply integrated the patch so that's the difference. I also reload the model after removing the lora.

mcmonkey4eva:
Full model reload wasn't needed for 8-bit, but might be for 4-bit, or at least with the patches? If your version works better, should PR improvements.

USBhost:
For warm up steps that should only apply to constant with warm up scheduler. Iirc.

mcmonkey4eva:
@USBhost  `For warm up steps that should only apply to constant with warm up scheduler. Iirc.`
image

All schedulers other than `constant` support a warmup. (inverse_sqrt does too even though it's not labeled there)

USBhost:
> @USBhost `For warm up steps that should only apply to constant with warm up scheduler. Iirc.` image
> 
> All schedulers other than `constant` support a warmup. (inverse_sqrt does too even though it's not labeled there)

Oh, that's good to know. Thanks for the correction.

mcmonkey4eva:
All above requests (unless I missed one) are now done, and updated in the OP. Open questions I had previously are addressed and handled now.

Saving will happen in the training thread, with a backup to the browser's gradio thread if it crashes. (If it crashes while your browser is closed, er, ... better have checkpoint saving on I guess)

-----

@oobabooga this PR is pretty packed with improvements, not to mention the base commit adding support for 4-bit training that a few users on Discord and elsewhere are already frothing about.
To my knowledge, there are no standing issues preventing a pull, so ready for your review whenever you're able.

Only open question is whether maybe default batch size should reduced, or if there's another way to reduce VRAM usage while training? (4bit's requirements are currently higher than 8bit, relatively speaking, in terms of how much extra VRAM is added when you bump batch size up)

USBhost:
https://github.com/kohya-ss/sd-scripts/blob/main/library/train_util.py#L2451 example?
AdamW8bit would be nice to have. It will save some vram but idk how much.

mcmonkey4eva:
Added an option to select the optimizer, `adamw_bnb_8bit` doesn't seem to actually reduce VRAM over the default `adamw_hf`.
Might be valuable to do some research/testing to see if maybe one of the options offers any noticeable improvements.

oobabooga:
I have read everything and made some simple tests. Seemingly the changes make sense and nothing exploded.

Thanks @mcmonkey4eva for the God's work that you are doing with the Training tab, and thanks @Ph0rk0z and @USBhost for the reviews.

oobabooga:
@mcmonkey4eva about a wiki entry, one workaround could be to create a README.md inside the `training` folder and then link it in the main README, as was done for docker here:

https://github.com/oobabooga/text-generation-webui/tree/main/docker

mcmonkey4eva:
Oh, that's a great idea!

tensiondriven:
This is fan freaking tasktic!

An additional feature to consider:  If we had the ability to do training runs via command line or via a simple python api, it would enable a whole other level of usage for text-generation-webui.  Currently I am doing this with a hacked up version of the alpaca_lora_4bit repo, which is working, but it would be great if I could rely on text-generation-webui for this.

One of the last pieces of automation I need after that is being able to load multiple loras from the command line.  (I see that we have `--lora`, which may accept multiple parameters, but I've only tried it with one so far.)

Ph0rk0z:
I don't think you can stack 4bit loras.

EmerJK:
I'd love to be able to load multiple lora from the command line as well.

PR #1348, opened 2023-04-18
Fix stopping strings in api

karlovm:
By default, stopping strings is ignored when used in default API. That's simple fix for that.

### DIFF
--- a/modules/api.py
+++ b/modules/api.py
@@ -36,7 +36,7 @@ def generate_reply_wrapper(string):
     }
     params = json.loads(string)
     generate_params.update(params[1])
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Is this why silly tavern keeps generating?

CyberTimon:
It still doesn't work

Ph0rk0z:
It's working for me without this PR on sillytavern. Just update the gradio ID to 43.

oobabooga:
I think that this is already working after https://github.com/oobabooga/text-generation-webui/pull/990

PR #1365, opened 2023-04-19
Adjust to "instruct" mode

umLu:
Modification in the behavior of the API extension to work as in the web interface in "instruct" mode

- Fixes #1137

### DIFF
--- a/extensions/api/script.py
+++ b/extensions/api/script.py
@@ -4,6 +4,7 @@ from threading import Thread
 
 from modules import shared
 from modules.text_generation import encode, generate_reply
+from modules.chat import extract_message_from_reply
 
 params = {
     'port': 5000,
@@ -58,10 +59,14 @@ class Handler(BaseHTTPRequestHandler):
                 'early_stopping': bool(body.get('early_stopping', False)),
                 'seed': int(body.get('seed', -1)),
                 'add_bos_token': int(body.get('add_bos_token', True)),
-                'custom_stopping_strings': body.get('custom_stopping_strings', []),
+                'custom_stopping_strings': body.get('custom_stopping_strings', ''),
                 'truncation_length': int(body.get('truncation_length', 2048)),
                 'ban_eos_token': bool(body.get('ban_eos_token', False)),
                 'skip_special_tokens': bool(body.get('skip_special_tokens', True)),
+                'mode': 'instruct',
+                'name1': '### Human:',
+                'name2': '### Assistant:',
+                'stop_at_newline': False,
             }
 
             generator = generate_reply(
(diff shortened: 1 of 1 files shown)


umLu:
Request example:
```python
import requests
import json

api_url = "http://127.0.0.1:5000/api/v1/generate"
prompt_text = "Say hello!"

complete_prompt = f'Below is an instruction that describes a task. Write a response that appropriately completes the request.\n### Human:\n{prompt_text}\n### Assistant:\n'

payload = {
    "prompt": complete_prompt
}

headers = {
    "Content-Type": "application/json",
}

response = requests.post(api_url, data=json.dumps(payload), headers=headers)

if response.status_code == 200:
    result = json.loads(response.text)
    generated_text = result["results"][0]["text"]
    print("Generated text:", generated_text)
else:
    print("Error. Status code:", response.status_code)
```

oobabooga:
The arguments that you added are not used by the generate_reply function

umLu:
The elements are not used by the generate_reply function, but rather by the extension itself. They are carried by the generate_params variable:
from modules.chat import extract_message_from_reply
...
reply, _ = extract_message_from_reply(answer, generate_params).

I suggest testing both the original version and the proposal to compare the differences in the output.

sumeetv:
I've been running into this issue over the past couple days and have to agree with umLu's experience here. I think there are a couple issues:

1. The API extension has mismatched params with the server itself (e.g. `max_length` in the API gets converted to `max_new_tokens`)
2. The caller cannot set the mode to instruct from chat, which results in all responses as truncated conversations that keep going

I would recommend instead adding additional parameters to params and trying to get them from the request or defaulting back to what's in the `shared` module

extensions/api/script.py
```diff
@@ -58,10 +59,14 @@ def do_POST(self):
...
                 'seed': int(body.get('seed', -1)),
                 'add_bos_token': int(body.get('add_bos_token', True)),
-                'custom_stopping_strings': body.get('custom_stopping_strings', []),
+                'custom_stopping_strings': body.get('custom_stopping_strings', ''),
                 'truncation_length': int(body.get('truncation_length', 2048)),
                 'ban_eos_token': bool(body.get('ban_eos_token', False)),
                 'skip_special_tokens': bool(body.get('skip_special_tokens', True)),
+                'mode': 'instruct',
+                'name1': '### Human:',
+                'name2': '### Assistant:',
```
sumeetv, line 68: Could we instead do something like:
`'mode': body.get('mode', shared.mode`

oobabooga:
The current api state is kind of a mess. What I want is to rename the current "api" extension to "kobold-api", then create a new api extension that has two endpoints, one for streaming and another for not streaming. After that, this extension should be expanded to handle the chat case, a subcategory of which is the instruction-following case that this PR tried to handle.

Issue #1371, opened 2023-04-19
KeyError: 0


Rudd-O:
### Describe the bug

```
[root@roxanne oobabooga_linux]# ./start_linux.sh 
Select the model that you want to download:

A) OPT 6.7B
B) OPT 2.7B
C) OPT 1.3B
D) OPT 350M
E) GALACTICA 6.7B
F) GALACTICA 1.3B
G) GALACTICA 125M
H) Pythia-6.9B-deduped
I) Pythia-2.8B-deduped
J) Pythia-1.4B-deduped
K) Pythia-410M-deduped
L) None of the above

Input> h
Traceback (most recent call last):
  File "/opt/oobabooga_linux/text-generation-webui/download-model.py", line 260, in <module>
    links, sha256, is_lora = get_download_links_from_huggingface(model, branch, text_only=args.text_only)
  File "/opt/oobabooga_linux/text-generation-webui/download-model.py", line 103, in get_download_links_from_huggingface
    fname = dict[i]['path']
KeyError: 0
Gradio HTTP request redirected to localhost :)
bin /opt/oobabooga_linux/installer_files/env/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117.so
No models are available! Please download at least one.

Done!
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Just entered a model choice and that happened.

### Screenshot

The above is a complete shot.

### Logs

```shell
The stacktrace is above.
```

### System Info

```shell
Fedora 37 x86_64.
```

Rudd-O:
Yep, adding an assert shows that the data downloaded from `https://huggingface.co/api/models/EleutherAI/pythia-6.9b-deduped/tree/main?cursor=` is not valid API data.

```
Traceback (most recent call last):
  File "/opt/oobabooga_linux/text-generation-webui/download-model.py", line 261, in <module>
    links, sha256, is_lora = get_download_links_from_huggingface(model, branch, text_only=args.text_only)
  File "/opt/oobabooga_linux/text-generation-webui/download-model.py", line 103, in get_download_links_from_huggingface
    assert 0, dict
AssertionError: {'error': '"cursor" is not allowed to be empty'}
```

h3ndrik:
thanks. fixes this for me.

Rudd-O:
Please proceed in merging.

BeaverInGreenland:
Thanks! It fixes the problem for me aswell!

[...]

PR #1373, opened 2023-04-19
Fix misuse of empty cursor in downloader

Rudd-O:
Fixes https://github.com/oobabooga/text-generation-webui/issues/1371

Maybe they changed their API.  This, however, works.


VanIseghemThomas:
Can confirm this fix worksi

Issue #1383, opened 2023-04-19
⭐ Support StableLM From StabilityAI


MarkSchmidty:
[Blog Post Announcement](https://stability.ai/blog/stability-ai-launches-the-first-of-its-stablelm-suite-of-language-models)
(It may be using the same architecture as GPT-NeoX)

[![Launch In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/Stability-AI/StableLM/blob/main/notebooks/stablelm-alpha.ipynb) 

[GitHub Repo](https://github.com/Stability-AI/StableLM)

In case these links 404 due to being posted early by accident:
https://archive.is/ZQszO
https://archive.ph/U0Pr8

(Checkpoint links are Hugging Face repos with model weights)
| Size | StableLM-Base-Alpha                        | StableLM-Tuned-Alpha                      | Training Tokens | Context Window | Web Demo |
|----------------|-------------------------------|-----------------------------|----------|----------|----------|
|3B| [checkpoint](https://huggingface.co/stabilityai/stablelm-base-alpha-3b/) |[checkpoint](https://huggingface.co/stabilityai/stablelm-tuned-alpha-3b/)            | 800B | 4096 | |
|7B| [checkpoint](https://huggingface.co/stabilityai/stablelm-base-alpha-7b) |[checkpoint](https://huggingface.co/stabilityai/stablelm-tuned-alpha-7b)            | 800B | 4096 | [HuggingFace](https://huggingface.co/spaces/stabilityai/stablelm-tuned-alpha-chat) |
|15B| (in progress) | (pending) | 1.5T | | |
|30B| (in progress) | (pending) | 1.5T | | |
|65B| (in progress) | (pending) | 1.5T | | |
|175B| (planned) | | | | |

underwaterbepis:
Seems to work fine for me as is

Ph0rk0z:
Just have to set the longer contexts.. these are GPT-J or NeoX as far as I know.

eiery:
The 3B base model works out of the box for me on oobabooga but the quality of the generated text is visibly worse than the 3B Pythia and Pygmalion. It's not really coherent and likes to go off-topic.

CarlKenner:
> Seems to work fine for me as is

Then please provide the steps you used to make it work.

MarkSchmidty:
Put it in a folder next to the rest of your models and then select it from the dropdown. 
[...]

PR #1395, opened 2023-04-20
Support StableLM

CarlKenner:
This addresses Issue #1383
Adds StableLM Instruction Following support in Chat mode and Notebook mode for Tuned models.
Adds support for 4096 tokens with StableLM models.

I tested the 3 GB version on my 4GB Nvidia GTX 970, and it works fine and follows instructions at a slow but usable speed on the default settings.


models/config.yaml
```diff
@@ -48,3 +48,6 @@ llama-[0-9]*b-4bit$:
 .*chatglm:
   mode: 'instruct'
   instruction_template: 'ChatGLM'
+.*stablelm-tuned:
+  mode: 'instruct'
+  instruction_template: 'StableLM'
```
mcmonkey4eva, line 53: This should be updated to:
- recognize ALL the standard filenames of StableLM. eg `stablelm-base-alpha-7b` is one such standard filename.
- configure base as normal and the Instruct model, separately
- configure them all to 4096 token max inputs (that needs a code update I think?)

characters/instruction-following/StableLM.yaml
```diff
@@ -0,0 +1,3 @@
+name: "<|ASSISTANT|>"
+your_name: "<|USER|>"
+context: "<|SYSTEM|># StableLM Tuned (Alpha version)\n- StableLM is a helpful and harmless open-source AI language model developed by StabilityAI.\n- StableLM is excited to be able to help the user, but will refuse to do anything that could be considered harmful to the user.\n- StableLM is more than just an information source, StableLM is also able to write poetry, short stories, and make jokes.\n- StableLM will refuse to participate in anything that could harm a human.\n"
```
mcmonkey4eva, line 3: Please look at how `Example.yaml` deals with multi-line strings, and use the same format here, it's much cleaner. (tldr, `|-` and then spaced out lines, no need for quotes even when doing that).

models/config.yaml
CarlKenner: `stablelm-base-alpha-7b` is not instruction tuned and does not support that instruction_template or instruct mode, so there's nothing needed here for it.
Are the token max inputs specified in this file? I thought they were specified somewhere else, like in presets.

CarlKenner: done

characters/instruction-following/StableLM.yaml
CarlKenner: done

models/config.yaml
mcmonkey4eva: need `.*` on the end to match the standard folder names

CarlKenner: No, I really don't think it needs that.

mcmonkey4eva: The upload from Stability is <https://huggingface.co/stabilityai/stablelm-base-alpha-7b>
The folder is named `stablelm-base-alpha-7b`
If you use the UI download, git clone, or any similar method, that is the name that will be used.
The regexes here must match these standard names. Your current regex will not, as `-alpha-7b` is not accounted for.

CarlKenner: Incorrect. My current regex works fine on names like stablelm-base-alpha-7b. Notice how none of the other regex (eg. oasst) end with .* even though their model names all have something after the oasst. I've tested it (on tuned rather than based, but it's the same issue) and it works fine how it is. It will still be recognised even if the filename contains text after the regular expression.

mcmonkey4eva (approved):
Gave it a try and it works perfectly.
(It's rather odd that the regex matchers in the model config appear to apparently be a startsWith? ie the end doesn't need to match?) but it works.

PR #1403, opened 2023-04-20
Support the GPTCache to reduce the model Response Time

SimFG:
Added a cache function to LLM, which can reduce request response time if the cache is hit. This feature is disabled by default and can be enabled through the `--enable-init-gptcache` configuration. Hope this feature will improve the current project

By default, the ui will not be changed, and there will be two checkboxes after opening.
image

testing
WZOAHYqc7q

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ img_bot*
 img_me*
 prompts/[0-9]*
 models/config-user.yaml
+cache_data/*
+.idea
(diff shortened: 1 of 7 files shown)


SimFG:
@oobabooga please help me check it if you think it's suitable this project

Ph0rk0z:
So this caches replies and returns the same answer if you ask the same question? So only useful for a multi-user environment?

SimFG:
The questions don’t need to be exactly the same, just similar. If you are not satisfied with the cached answer, you can skip the cache so that you can get the answer again

Ph0rk0z:
Definitely looks helpful for a multi-user service and then not letting them skip the cache.

SimFG:
@Ph0rk0z have something updated? :)

Ph0rk0z:
Lol, not yet.

oobabooga:
This is an interesting idea, but I think that it should be an extension. It should be possible to create a custom_generate_reply that returns the cached reply if "question" is in the database, and otherwise calls the regular generate_reply 

https://github.com/oobabooga/text-generation-webui/blob/main/docs/Extensions.md#custom_generate_reply-example

PR #1408, opened 2023-04-20
Compatable with newer gptq 

catalpaaa:
yea they are changing up shits again

### DIFF
--- a/modules/GPTQ_loader.py
+++ b/modules/GPTQ_loader.py
@@ -12,13 +12,14 @@ import modules.shared as shared
 
 sys.path.insert(0, str(Path("repositories/GPTQ-for-LLaMa")))
 import llama_inference_offload
-from utils.modelutils import find_layers
 
 try:
     from quant import make_quant
+    from modelutils import find_layers
     is_triton = False
 except ImportError:
     import quant
+    from utils.modelutils import find_layers
     is_triton = True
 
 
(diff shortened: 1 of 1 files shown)


catalpaaa:
> Thanks. I don't have a triton installation to test, so I'll tag @sgsdxzy and @EyeDeck for review.
sure, they just moved modelutils into utils/
i just made it so it looks for that too

modules/GPTQ_loader.py
```diff
@@ -12,13 +12,14 @@
...
 import llama_inference_offload
-from modelutils import find_layers
 
 try:
     from quant import make_quant
+    from modelutils import find_layers
     is_triton = False
 except ImportError:
     import quant
+    from utils.modelutils import find_layers
```
sgsdxzy, line 22: I think it is fine to just `from utils import find_layers`

EyeDeck:
Works for me.

Without this patch, the latest GPTQ-for-LLaMA crashes with:
```
  File "/home/anon/text-generation-webui/server.py", line 912, in <module>
  File "/home/anon/text-generation-webui/modules/models.py", line 125, in load_model
    from modules.GPTQ_loader import load_quantized
  File "/home/anon/text-generation-webui/modules/GPTQ_loader.py", line 15, in <module>
    from modelutils import find_layers
ModuleNotFoundError: No module named 'modelutils'
```
Maybe some day we'll stop needing to fight with random breaking changes every three days :')

oobabooga:
I merged #1451 which had the simpler import syntax

USBhost:
I did not see this pull mybad.

Issue #1014, opened 2023-04-10
New UI Layout sketch


KirillRepinArt:
**OpenAI like WEB UI**

I'm not sure if this would be of interest to anybody else but me, but I've made some sketches of a new UI design. As a template I've used Chat GPT, but adapted their layout for this project and slightly changed the color scheme not to look exactly the same.

By studying OpenAI's design it was obvious that they've used Golden Ratio proportions, so I did just that.

I've attached a couple of examples how it could look.

**Additional Context**
Just to clarify, these are just sketches, but I think it's possible to transfer all of the current features into this format.

![TextGen_UI_Sketch_01](https://user-images.githubusercontent.com/118350327/230966321-4d460a9b-5f81-419e-952e-6957e2bc0c56.jpg)
![TextGen_UI_Sketch_02](https://user-images.githubusercontent.com/118350327/230966325-b6597bfb-a888-401d-8156-5765ee030f9f.jpg)
![TextGen_UI_Sketch_03](https://user-images.githubusercontent.com/118350327/230966327-23e824d1-28e1-408f-8cfb-295d28ecd2fc.jpg)

bmoconno:
> Just to clarify, these are just sketches, but I think it's possible to transfer all of the current features into this format.

I'm still new to all of this python/Gradio stuff, but I'm not entirely sure that is true. I know some css can be applied to Gradio, but I don't know if what has been outlined here is possible.

That being said, I do like the look of this. I think you'd maybe need to add an accordion or something to bring back the rest of the buttons (and extensions) that seem to have been removed from this layout to make things look cleaner.

KirillRepinArt:
> > Just to clarify, these are just sketches, but I think it's possible to transfer all of the current features into this format.
> 
> I'm still new to all of this python/Gradio stuff, but I'm not entirely sure that is true. I know some css can be applied to Gradio, but I don't know if what has been outlined here is possible.
> 
> That being said, I do like the look of this. I think you'd maybe need to add an accordion or something to bring back the rest of the buttons (and extensions) that seem to have been removed from this layout to make things look cleaner.

Unfortunately my experience with coding is very limited, but I have a degree in graphic design, so I thought I could be helpful in that way.
These are just relatively quick sketches of how it could look like. Of course if this design was to be implemented, it should be significantly expanded. 
And since I love using this text gen, I selfishly want it to be as comfortable and enjoyable as possible.

j4ded:
[...]

PR #1414, opened 2023-04-20
Added Extensions Tab, Manifest and Bootstrap4 CSS

ChobPT:
Added an extensions tab meant to look like an app gallery with basic info about each plugin, the code is prepared to show images, more links, features etc but depends on the community's appetite for this. This would be the first step to future proof-ing and make the expansion easier (maybe who know add remote extensions or download form other extension repos?)

Added a manifest.json  that uses the same format for an OpenAI ChatGPT plugin. Devs can change theirs and provide more info. 
Edit: Upon reflecting, this could be used to turn on and off extensions (duh) and show which options each extension needs enabled + auto enable them. Could even help guide through possible conflicts where 2 extensions need 2 different values for the same option

Added bootstrap 4 min, which currently is getting heavily overwritten by GRadio but in the future would allow for stuff like new themes or interfaces like https://github.com/oobabooga/text-generation-webui/issues/1014

Pretty new to community contributions, 🐻 with me


ChobPT:
Been working with the API to make some experiments while still using the interface with the same model loaded. 

Because the API had some defaults that in the interface were easily tweaked by using the Preset (Kobold, LLaMa Precise, etc), I've added a preset parameter where you can choose the same as the dropdown. 

```json
{	
	"prompt":"Tell me a joke",
	"preset":"LLaMA-Precise",
	"top_k":66,
	"stopping_strings":"### Human:"
}
```

In this example, we'd be loading the LLaMa-Precise model, and then overriding the preset's top_k to a custom_value. If neither the preset or the top_k were set on the request, the default for the extension will be loaded

oobabooga:
I think that a better way to do this would be to mimic the automatic1111 extensions tab in a pure gradio way without any bootstrap code and without the manifest files. In his case, the extensions are pulled from a json located here:

https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/blob/master/index.json

And this interface is generated based on it:

print

Something similar could be hosted here:

https://github.com/oobabooga/text-generation-webui-extensions

ChobPT:
Bootstrap is literally there only for future proofing. The code uses gr.Row and gr.Columb for the grid with the onyl thing being used from bootstrap being the button floating right. If Id remove it, would it be a worthwhile submission?

PR #1439, opened 2023-04-21
Clarified "download from hugging face" and added "do not download" to download-model.py

Loufe:
The new one-click installers are fantastic. Small nit: as the model-download.py script is now run automatically in the new install chain, there's a forced error unless a model is specified.

image

Many LLaMa/other users will just want to install a model manually. The script currently doesn't allow no choice of download, instead assuming the user wants to specify a Hugging Face model. This PR clears it up by renaming that and adding a "_Do not download_" option (direct exit).

image

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -50,8 +50,7 @@ def select_model_from_default_options():
     print("Input> ", end='')
     choice = input()[0].strip().upper()
     if choice == char_exit:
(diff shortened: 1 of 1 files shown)


oobabooga:
That's a good idea. It was bothering me to force users to download a model during the installation process.

PR #1464, opened 2023-04-21
add optional text between generation attempts to promote better responses

matthewmeyer:
Frequently when using multiple generation attempts the second generation would not generate an additional usable tokens in chat. This will first append a string to the end of the previous generation to help the model with longer completions. Some ideas for strings to use:

- ' And then...'
- ' After that'
- '...'

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -158,6 +158,7 @@ def chatbot_wrapper(text, state, regenerate=False, _continue=False):
     # Generate
     for i in range(state['chat_generation_attempts']):
         reply = None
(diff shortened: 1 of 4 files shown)


modules/chat.py
```diff
@@ -158,6 +158,7 @@ def chatbot_wrapper(text, state, regenerate=False, _continue=False):
     # Generate
     for i in range(state['chat_generation_attempts']):
         reply = None
+        if i > 0 : cumulative_reply += state['chat_generation_improv_phrase']
```
mcmonkey4eva, line 161: this syntax is a lil weird.

oobabooga:
I find that this adds complexity for little return.

PR #1476, opened 2023-04-22
ci(github actions): add github action to build and push docker image.

LonghronShen:
Add github action to build and push docker image to the GHCR.

* [Multi-platform build] Build for **arm64v8**, tested with failure, commented out for future.
* [CUDA capability] By setting **GPTQ_FOR_LLAMA_VER** repository envrionment to **841feedde876785bc8022ca48fd9c3ff626587e2**, and **TORCH_CUDA_ARCH_LIST** to **3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX**, the older gpu could work.


oobabooga:
I'm not a docker user and wouldn't be able to maintain this

LonghronShen:
Oh, thank you for your reply. 
I think my change has been relatively simple and will not need much change in the future. I would like to help the maintance for these Docker staff. I think for the normal users, it is better to have a good docker based solution.

PR #1484, opened 2023-04-22
Apply the settings for extensions after import, but before setup()

da3dsoul:
I had an issue where there wasn't an easy way to init an extension based on settings. Before, you could theoretically do it by performing that logic in ui(), but setup() should have the settings loaded, imo


oobabooga:
That definitely makes sense. I have merged the main branch which had some changes to the extensions framework since your PR was submitted. Everything seems to be working, but please let me know if you encounter any issues.

LaaZa:
I think this broke loading settings for LLaVA. Do I have to change the settings format somehow?

oobabooga:
No, the format should be the same

LaaZa:
Stopped loading the settings though.

oobabooga:
Can you check if this fixed it? https://github.com/oobabooga/text-generation-webui/commit/da812600f49309cd7090464d7ffcfa1246d3a309

LaaZa:
Yes, fixed.

PR #1485, opened 2023-04-23
Better GPT-4chan defaults

Brawlence:
+ability to move gpt-j-6B tokenizer files (as a whole directory) into GPT-4chan's model directory.

The intent is to avoid cluttering the model selection menu if one does not use `gpt-j-6B`. I.e., I keep the tokenizer files for `gpt-j-6B` only to use them with `GPT-4chan` and there is no checkpoint present in `gpt-j-6B` folder.

### DIFF
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -9,6 +9,9 @@
 llama-[0-9]*b-4bit$:
   wbits: 4
   model_type: 'llama'
+gpt[-_]?4chan.*:
+  model_type: 'gptj'
+  mode: 'default'
 .*-(4bit|int4)-(gr128|128g):
   wbits: 4
   groupsize: 128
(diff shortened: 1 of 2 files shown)


Ph0rk0z:
I just copied all the GPT-J stuff into GPT4chan.. I'm not sure why it even pulls from another folder rather than having users complete the model.

gitihobo:
Where is the gpt4chan model available at? The original huggingface seems to be disabled.

Brawlence:
@gitihobo https://github.com/oobabooga/text-generation-webui#gpt-4chan

oobabooga:
I think that this adds complexity to models.py for little return. You can move gpt-j tokenizer files directly into the gpt4chan folder as @Ph0rk0z pointed out

PR #1493, opened 2023-04-23
Lora trainer docs

mcmonkey4eva:
- Added a full guide to LoRA training into the docs.
    - I moved it to its own file, and kept the old content, but shoved it to the bottom.
    - View rendered markdown preview @ https://github.com/mcmonkey4eva/text-generation-webui/blob/lora-trainer-docs/docs/Training-LoRAs.md
    - I may have gone a little overboard. I went with the idea of making it idiot-proof. Essentially there are 3 categories of user I can see
        - 1: "I can figure it out with some technical docs" users - they'll just open the training tab and use the on-page docs to figure it all out
        - 2: "I almost got it but i need some hints" users - they'll skim the docs page to eg the parameters section to get the hints they need.
        - 3: "I'm new here please hold my hand" users - they can read the whole page and follow it step by step.
        - All 3 types of user should end up having no trouble at all making their own LoRAs now!
    - I am open to feedback on content and/or format. Right now it's a bit of a wall-of-text. I could break it up with some images or similar, but I'm not sure how appropriate that'd be within this documentation base.
- Also, I added a link to the new docs (or where they'll be assuming the PR is accepted) to the top of the Training tab (it still weirds me out that `a` links render like plaintext, makes it non-obvious when they're clickable)
- Also, I slightly improved error messages to be clearer when training over top of Peft model objects as there was a user confused by it.
- Alsoalso, I tossed in recognition for GPT NeoX as users are asking about training StableLM.

### DIFF
--- a/docs/Training-LoRAs.md
+++ b/docs/Training-LoRAs.md
@@ -2,39 +2,43 @@
 
 The WebUI seeks to make training your own LoRAs as easy as possible. It comes down to just a few simple steps:
 
-- **Step 1**: Make a plan.
-    - What base model do you want to use? The LoRA you make has to be matched up to a single architecture (eg LLaMA-13B) and cannot be transferred to others (eg LLaMA-7B, StableLM, etc. would all be different). Derivatives of the same model (eg Alpaca finetune of LLaMA-13B) might be transferrable, but even then it's best to train exactly on what you plan to use.
-    - What model format do you want? At time of writing, 8-bit models are most stable, and 4-bit are supported but experimental. In the near future it is likely that 4-bit will be the best option for most users.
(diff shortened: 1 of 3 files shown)


oobabooga:
Thank you for this! The format is fine as it is and I like that you went into lots of detail. I believe that most users are the 3: type. Anecdote: the web UI had a `--gpu-memory` parameter since the beginning, but barely anyone used it because there wasn't a slider in the interface initially.

The weird `<a>` style that you noticed is an instance of a bigger problem, which is that `gr.Markdown()` styles are weird overall. I want to modify the CSS to make it match the style in the main notebook mode textbox at some point.

PR #1496, opened 2023-04-23
Feature Request: Response streaming with HTTP API

im-not-tom:
This PR extends API extension with `/api/v1/generate_stream` endpoint. This endpoint takes same input as `/api/v1/generate`, but responds with plain text while outputting tokens as soon as they are ready.

This would allow 3rd party scripts and applications to stream tokens wihtout having to use websocket and guessing gradio function IDs.

It would also allow for simple way to use api from curl and shellscripts:

https://asciinema.org/a/LXzNgmMbdLEv2mg7aCbLJkbMf

### DIFF
--- a/extensions/api/script.py
+++ b/extensions/api/script.py
@@ -23,6 +23,40 @@ class Handler(BaseHTTPRequestHandler):
         else:
             self.send_error(404)
 
+    def parse_params(self, body):
+        prompt = body['prompt']
+        prompt_lines = [k.strip() for k in prompt.split('\n')]
+        max_context = body.get('max_context_length', 2048)
+        while len(prompt_lines) >= 0 and len(encode('\n'.join(prompt_lines))) > max_context:
+            prompt_lines.pop(0)
+
+        prompt = '\n'.join(prompt_lines)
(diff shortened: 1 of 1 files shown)


oobabooga:
Closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/990 (it was an older PR that I was in the process of merging). If you can find improvements to this new API implementation, please feel free to submit them in a new PR.

im-not-tom:
That's bit unfortunate for me, as that one still uses websocket. Is that intended direction or just something that no one else considered as a issue?

oobabooga:
Why is websocket an issue?

im-not-tom:
Well, they are certainly more complicated than just reading plain text from the HTTP response.

But what I should have been asking is whether you would be open for PR which adds non-websocket version of `api/v1/generate`, just like this one but over current version of API. I've also noticed that there's no api for "stop" function.

oobabooga:
I feel like the best solution would be for you to host your alternative api in its own repository and submit it to https://github.com/oobabooga/text-generation-webui-extensions. I would be interested in testing it.

PR #1502, opened 2023-04-23
Preserve selected character on refresh/reload

bmoconno:
This will cause the selected character to remind selected if the UI is refreshed or closed/re-opened without shutting down the server.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -484,7 +484,7 @@ def load_character(character, name1, name2, mode):
 
 
 def load_default_history(name1, name2):
-    load_character("None", name1, name2, "chat")
+    load_character(shared.character or "None", name1, name2, "chat")
 
 
 def upload_character(json_file, img, tavern=False):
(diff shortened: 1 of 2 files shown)


bmoconno:
Found a bug.

bmoconno:
Fixed.

oobabooga:
The problem is that `shared.settings['instruction_template']` can be dynamically set while switching models, while the event handler definition is static

oobabooga:
Not a good solution. If I load the interface in chat mode, select instruct mode, and refresh, it returns to chat mode

If/else statements don't belong to event handler definitions

bmoconno:
Thanks for the input. I'll think about a different solution, unless you think it's not worth pursuing.

oobabooga:
After this change the character already persists, as the "character_menu" dropdown is updated with the current character every time the page reloads https://github.com/oobabooga/text-generation-webui/commit/b1ee674d75fca92c638e967118addb796a34a9b0

bmoconno:
Ahh nice, didn't see that yet. Thanks!

PR #1507, opened 2023-04-24
LLaVA: fix Continue button, unify internal and visible representation of images 

Wojtab:
Previously internal representation was `<image:{base64string}>`, while visible was ``. For some reason the continue button replaces visible with internal for the message from user, so the pictures were disappearing. This unifies both of the representations, and in turn fixes it.
Also - by unifying them `fix_picture_after_remove_last` function is no longer needed, as user will be able to re-send the image in the correct representation


oobabooga:
It seems to work as expected.

> For some reason the continue button replaces visible with internal for the message from user

This is a bug, I'm looking into it but since it's outside the scope of this PR, I'll merge and fix it afterwards

Issue #1357, opened 2023-04-19
No Google T5 based models (including FLAN-T5) work on oobabooga


raziel2001au:
### Describe the bug

Not sure what the issue is, but no models based on Google's T5 architecture work on oobabooga, which is extremely unfortunate since the FLAN-T5 models (the versions officially fine-tuned by Google) are extremely good. It's basically Google's LLaMA except it's officially released by Google under an extremely permissive Apache 2.0 license.

When loading a T5 based model you get the following error:
```
Traceback (most recent call last):
  File "D:\Development\oobabooga\text-generation-webui\server.py", line 913, in <module>
    shared.model, shared.tokenizer = load_model(shared.model_name)
  File "D:\Development\oobabooga\text-generation-webui\modules\models.py", line 186, in load_model
    model = LoaderClass.from_pretrained(checkpoint, **params)
  File "D:\Development\oobabooga\installer_files\env\lib\site-packages\transformers\models\auto\auto_factory.py", line 474, in from_pretrained
    raise ValueError(
ValueError: Unrecognized configuration class <class 'transformers.models.t5.configuration_t5.T5Config'> for this kind of AutoModel: AutoModelForCausalLM.
Model type should be one of BartConfig, BertConfig, BertGenerationConfig, BigBirdConfig, BigBirdPegasusConfig, BioGptConfig, BlenderbotConfig, BlenderbotSmallConfig, BloomConfig, CamembertConfig, CodeGenConfig, CpmAntConfig, CTRLConfig, Data2VecTextConfig, ElectraConfig, ErnieConfig, GitConfig, GPT2Config, GPT2Config, GPTBigCodeConfig, GPTNeoConfig, GPTNeoXConfig, GPTNeoXJapaneseConfig, GPTJConfig, LlamaConfig, MarianConfig, MBartConfig, MegaConfig, MegatronBertConfig, MvpConfig, OpenAIGPTConfig, OPTConfig, PegasusConfig, PLBartConfig, ProphetNetConfig, QDQBertConfig, ReformerConfig, RemBertConfig, RobertaConfig, RobertaPreLayerNormConfig, RoCBertConfig, RoFormerConfig, Speech2Text2Config, TransfoXLConfig, TrOCRConfig, XGLMConfig, XLMConfig, XLMProphetNetConfig, XLMRobertaConfig, XLMRobertaXLConfig, XLNetConfig, XmodConfig.
```

[...]

PR #1535, opened 2023-04-25
Add support for Seq2Seq LMs

Vinno97:
Hi,

This PR adds support for seq2seq (encoder-decoder) models like T5, UL2 and their FLAN variants.
The FLAN models are pretty good few/zero shot learners and found it was a shame we couldn't use them here.

About the 'Not a "catch all"', I don't know for certain that all encoder-decoder models contain this key in their config. The ones I checked do. However, most decoder models also contain a "is-decoder" key, except GPT-2.

I kin of just merged the support in there, but don't think I could do much better without rewriting stuff.

Let me know what you guys think.

### DIFF
--- a/modules/models.py
+++ b/modules/models.py
@@ -10,7 +10,7 @@ import numpy as np
 import torch
 import transformers
 from accelerate import infer_auto_device_map, init_empty_weights
-from transformers import (AutoConfig, AutoModel, AutoModelForCausalLM,
+from transformers import (AutoConfig, AutoModel, AutoModelForCausalLM, AutoModelForSeq2SeqLM,
                           AutoTokenizer, BitsAndBytesConfig, LlamaTokenizer)
 
 import modules.shared as shared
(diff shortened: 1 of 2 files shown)


Ph0rk0z:
Maybe I can finally use that reddit model that has been taking space on my HDD.

oobabooga:
I have tested it with `google/flan-t5-xl` and it worked. It worked even in chat mode. Thank you!

modules/models.py
```diff
@@ -55,7 +56,12 @@ def find_model_type(model_name):
     elif any((k in model_name for k in ['gpt4chan', 'gpt-4chan'])):
         return 'gpt4chan'
     else:
-        return 'HF_generic'
+        config = AutoConfig.from_pretrained(f"{shared.args.model_dir}/{model_name}")
```
aaroncoffey, line 59: It appears that model_dir or model_name already has a `/`. This is breaking the install for me. Consider os.path.join for this type of concatenation.

oobabooga: I missed that, should be fixed now https://github.com/oobabooga/text-generation-webui/commit/9c2e7c0fab385c8208a17a19feaa43cacf44bcf1

Ph0rk0z:
Finally https://huggingface.co/hyunwoongko/reddit-9B works.

evangineer:
Fixes #1357

PR #1546, opened 2023-04-25
LoRA trainer improvements part 5

mcmonkey4eva:
- If you have latest PEFT, it now automagically supports any model type, even ones never explicitly tested (eg gpt2, etc). This was half-implemented previously but I apparently forgot half of it.
- Removed the shuffle option. HF trainer actually inherently shuffles as it goes and recommends against trying to bypass that <https://stackoverflow.com/questions/74415426/how-to-stop-data-shuffling-while-training-the-huggingface-bert-model> / <https://github.com/huggingface/transformers/issues/11693>

For once I don't have any big list of things, just a couple small improvements. Seems like it just works and solves most use-cases now. Or at least I haven't seen any comments about issues / wanted features in a while (other than a user asking about GPT-2 support, thus this PR).

I'm also happy to report that I've been hearing an increasing number of success stories from users using the WebUI LoRA Trainer tab to train their own unique data into models and getting working results!

### DIFF
--- a/modules/training.py
+++ b/modules/training.py
@@ -37,7 +37,7 @@ except:
 
 WANT_INTERRUPT = False
 
-PARAMETERS = ["lora_name", "always_override", "save_steps", "micro_batch_size", "batch_size", "epochs", "learning_rate", "lr_scheduler_type", "lora_rank", "lora_alpha", "lora_dropout", "cutoff_len", "dataset", "eval_dataset", "format", "eval_steps", "raw_text_file", "overlap_len", "newline_favor_len", "do_shuffle", "higher_rank_limit", "warmup_steps", "optimizer"]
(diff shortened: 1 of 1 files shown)


USBhost:
Could we save the tokenized dataset so if we wanted to change a setting we wouldn't have to reconvert it. Because I had the pleasure working with a 850MB raw text file and processing that took forever. So if we could cache it I could change the dim size and if the cutoff doesn't effect the tokens we could play with that to fine the max before oom and save countless hours.

oobabooga:
Thanks for the update!

mcmonkey4eva:
@USBhost caching would be possible in theory, but a bit complicated.
For your issue you could also just, test with a very small dataset to find the settings you need, _then_ load the full dataset for a real run after you have the right settings.

PR #1579, opened 2023-04-26
Fixing Vicuna text generation

TiagoGF:
Vicuna based models stopped working for me after latest updates. I was able to get them working again with this small change.

### DIFF
--- a/characters/instruction-following/Vicuna-v0.yaml
+++ b/characters/instruction-following/Vicuna-v0.yaml
@@ -1,4 +1,4 @@
 name: "### Assistant:"
 your_name: "### Human:"
 context: "A chat between a human and an assistant.\n\n"
-turn_template: "<|user|> <|user-message|>\n<|bot|> <|bot-message|>\n"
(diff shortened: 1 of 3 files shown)


oobabooga:
Their example prompts have spaces in them: https://github.com/lm-sys/FastChat/blob/main/docs/vicuna_weights_version.md#example-prompt-weight-v11

But I confirm that this fix seems necessary, although it doesn't make sense.

mcmonkey4eva:
There's been several comments recently relating to spaces in various sources (eg one user discovered API calls were doubling space inputs after the API update). There might have been an unintended change somewhere in message handling leaving an extra space somewhere it shouldn't be?

oobabooga:
I think that's exactly what happened @mcmonkey4eva. Apparently models don't like prompts with trailing spaces for whatever reason, even though the spaces are supposed to be there. I think that https://github.com/oobabooga/text-generation-webui/commit/1d8b8222e95e7669dd431a60866ebdc786ea76dd and https://github.com/oobabooga/text-generation-webui/commit/c83210c46022818db6999fb162ec1350705baab0 should have fixed the issue.

PR #1612, opened 2023-04-28
Add ssl cert support

garrettsutula:
Adds CLI args to enable use of TLS when running the web UI. Inspired by Automatic1111 extension [SDWUI Auto TLS-HTTPS Extension](https://github.com/papuSpartan/stable-diffusion-webui-auto-tls-https).

Tested by running the same changes locally w/ self-signed certificates.

### DIFF
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -43,23 +43,23 @@ def give_VRAM_priority(actor):
     if actor == 'SD':
         unload_model()
         print("Requesting Auto1111 to re-load last checkpoint used...")
-        response = requests.post(url=f'{params["address"]}/sdapi/v1/reload-checkpoint', json='', verify=shared.args.ssl_verify)
+        response = requests.post(url=f'{params["address"]}/sdapi/v1/reload-checkpoint', json='', verify=shared.args.disable_ssl_verify)
         response.raise_for_status()
 
     elif actor == 'LLM':
         print("Requesting Auto1111 to vacate VRAM...")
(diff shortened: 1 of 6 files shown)


garrettsutula:
@oobabooga sorry guess I shoulda tagged you in some way - afaik this is still good to go & would add proper tls support for self-signed certs.

requirements.txt
```diff
@@ -2,7 +2,7 @@ accelerate==0.18.0
 colorama
 datasets
 flexgen==0.1.7
-gradio==3.25.0
+gradio==3.28.0
```
oobabooga, line 5: I need 3.25.0, 3.28.0 is bugged

extensions/sd_api_pictures/script.py
```diff
@@ -43,23 +43,23 @@ def give_VRAM_priority(actor):
     if actor == 'SD':
         unload_model()
         print("Requesting Auto1111 to re-load last checkpoint used...")
-        response = requests.post(url=f'{params["address"]}/sdapi/v1/reload-checkpoint', json='')
+        response = requests.post(url=f'{params["address"]}/sdapi/v1/reload-checkpoint', json='', verify=shared.args.disable_ssl_verify)
```
oobabooga, line 46: I don't like adding global flags that only apply to the sd extension. Those should be removed and only the gradio flags should be kept. 

The sd flags could instead be turned into booleans into the params dict for the extension, making them customizable through settings.json

kierenalt255:
We need https to use the microphone in Chrome, with the `listen` parameter. i.e. UI reachable from your local network. The microphone is not permitted in Chrome for a local site without https. More [here](https://stackoverflow.com/questions/67663961/cannot-read-property-getusermedia-of-undefined-over-https/67664025#67664025)

PR #1641, opened 2023-04-28
Update README.md

bgrmwbys:


briancw:
"docker compose up" is actually correct on newer versions of docker. Compose is just part of docker now.

bgrmwbys:
Then this is wrong https://github.com/oobabooga/text-generation-webui/blob/main/docs/Docker.md#8-startup-docker-container

briancw:
Not really wrong. It just depends on if you have "docker-compose" installed on your system, or a newer version of Docker. Many people who have been using docker for a while probably have both even.
Screenshot_20230501_113828

If you wanted to normalize it across the project, I'd suggest changing it to "docker compose" rather than "docker-compose" since its the newer one.

oobabooga:
Done https://github.com/oobabooga/text-generation-webui/commit/e6a78c00f2ecfeb040f4a45b7760274f5cf7a4f6

Issue #1072, opened 2023-04-12
Support for --no-mmap for llamacpp


fgdfgfthgr-fox:
https://github.com/ggerganov/llama.cpp/issues/864

With --no-mmap, there could be a potential performance gain for a system with large enough RAM to fit the entire model. (although it's arguable at the moment)

fgdfgfthgr-fox:
It's done.

PR #1649, opened 2023-04-29
added no_mmap & mlock parameters to llama.cpp and removed llamacpp_model_alternative

axolotlite:
llamacpp users can now specify if they wish to use mmap or mlock
I've removed the llamacpp_model module and renamed llamacpp_module_alternative into it, since the "alternative" module is the default.
I've also changed the default patch size into 512 tokens, which is the current default used by llama-cpp-python library. 
There was an issue opened for --no-mmap support #1072


Priestru:
Thank you! Just what we need

PR #1660, opened 2023-04-29
GPTQ: Allow groupsize 1024 in the UI, helpful for larger models eg 30B to lower VRAM usage

TheBloke:
A trivial fix to allow groupsize=1024 in the UI for GPTQ models.

1024g is already supported when one uses command line arguments, but can't be set in the UI.

I only realised this after I pushed a new models to HF that use 1024g (https://huggingface.co/TheBloke/OpenAssistant-SFT-7-Llama-30B-GPTQ) so I'd really appreciate it if this could be merged :)

I normally use 128g for my GPTQs, but at 30B it makes a useful difference in VRAM which allows the user to get longer responses before OOMing on a 24GB GPU.


badr-jpg:
Very helpful addition.

CarlKenner (approved):
Looks good to me.

PR #1662, opened 2023-04-29
Allow downloading model from HF branch via UI

TheBloke:
A simple fix that allows the user to download models from HF branches, using `repo/model:branch` notation.

Eg a user could enter `TheBloke/stable-vicuna-13B-GPTQ:latest` to download from https://huggingface.co/TheBloke/stable-vicuna-13B-GPTQ/tree/latest

This is useful for me as I've started using HF branches to create separate versions of my GPTQ models.

eg in stable-vicuna-13B-GPT:
* main branch = GPTQ 4bit 128g without act-order
* 'latest' branch = GPTQ 4bit 128g with --act-order, requiring recent GPTQ-for-LLaMa code

It'd be great if I could tell users in my README that they can use the text-generation-webui download feature for whatever branch they want.

### DIFF
--- a/server.py
+++ b/server.py
@@ -183,8 +183,11 @@ def download_model_wrapper(repo_id):
     try:
         downloader = importlib.import_module("download-model")
 
-        model = repo_id
-        branch = "main"
+        repo_id_parts = repo_id.split(":")
+
+        model = repo_id_parts[0] if len(repo_id_parts) > 0 else repo_id
+        branch = repo_id_parts[1] if len(repo_id_parts) > 1 else "main"
+
         check = False
 
         yield ("Cleaning up the model/branch names")


CarlKenner:
You also need to implement this in `download-model.py` for downloading from the command line otherwise it will have inconsistent behaviour.

> This is useful for me as I've started using HF branches to create separate versions of my GPTQ models.

I think you're supposed to create separate models for separate model versions. So you created the problem yourself.
But since there's already code for dealing with branches, we might as well allow them to specify the branch.

However, there's still the more common problem of people putting multiple model files inside one huggingface model. There should also be some way to specify the model file within the folder, and having it download that model file along with the configuration files, but not the other model files. I'm thinking maybe we should use another slash if they want to specify the path within the model folder for the actual model file. We'd need some way to handle multiple shards of the same model.

oobabooga:
Good idea, thank you @TheBloke.

TheBloke:
Thanks ooba! And, sorry Carl, I only just saw the feedback. I must have missed the notification.  

I'm happy to do those suggested changes as well. It's a good idea to allow the same format to work on CLI also. I'll do another PR for that as this is merged now.

PR #1663, opened 2023-04-29
Remove mutable defaults from function signatures.

IJumpAround:
Mutable defaults can cause sneaky bugs, so instead assign them in the function body using truthy short circuiting.
https://docs.python-guide.org/writing/gotchas/


oobabooga:
Good idea, thank you

Issue #1533, opened 2023-04-25
Error using LLaVA on Mac M1, MPS not supported?


edwios:
### Describe the bug

An error "No module named 'llama_inference_offload' occurred when starting server.py.

If there doesn't exist a non-CUDA model, then this feature should be documented as not supported on Mac M1/M2.
Otherwise, documentation is needed for Mac installations.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
python download-model.py wojtab/llava-13b-v0-4bit-128g
python3 server.py --model wojtab_llava-13b-v0-4bit-128g --chat  --extensions llava
```

### Logs

```shell
[soro:/Users … ment/ml/pytorch/text-generation-webui] [pytorch-m1] main(+0/-1)+* 16d17h44m9s ± python download-model.py wojtab/llava-13b-v0-4bit-128g
Downloading the model to models/wojtab_llava-13b-v0-4bit-128g
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████| 631   /631    105kiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 88.0  /88.0   13.6kiB/s
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████| 754   /754    162kiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 137   /137    24.5kiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 7.45G /7.45G  59.8MiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 33.6k /33.6k  4.14MiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 97.0  /97.0   56.7kiB/s
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 500k  /500k   11.8MiB/s
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████| 244   /244    118kiB/s
[soro:/Users … ment/ml/pytorch/text-generation-webui] [pytorch-m1] main(+0/-1)+* 16d17h47m33s ± python3 server.py --model wojtab_llava-13b-v0-4bit-128g --chat  --extensions llava
Gradio HTTP request redirected to localhost :)
bin /Users/soro/miniforge3/envs/pytorch-m1/lib/python3.9/site-packages/bitsandbytes/libbitsandbytes_cpu.so
[...]

PR #1664, opened 2023-04-30
LLaVA: small fixes

Wojtab:
1. Haotian Liu - the author of LLaVA noticed that the projector weights I used are from before the finetuning stage (see: https://github.com/haotian-liu/LLaVA/discussions/42#discussioncomment-5725359), whereas the model weights are after finetuning. This PR fixes it, and also makes them configurable by the user. (The projector is a separate, ~10MB model, working as "glue" between vision model and LLM)
After running both versions on the same prompts, the difference isn't huge, but the model seems less likely to hallucinate objects which aren't in the images.
2. I removed the step with adding `"\n###"` to custom stopping strings from readme, as https://github.com/oobabooga/text-generation-webui/commit/d87ca8f2af2458e8b57b1ec9915c72a4ca5ca19f added it to `models/config.yaml`, and added a note that the default model needs GPTQ, per this suggestion: https://github.com/oobabooga/text-generation-webui/issues/1533#issuecomment-1522806803
3. I fixed stopping_strings in no-stream mode, `starting_idx=len(input_ids[0])` broke if we modified `input_ids` in tokenizer extensions. No idea why it works without --no-stream
4. I added example usage through API in docs
5. LLaVA 7B just got released, it requires changing a few consts, and that's it, but I'd like to make multimodality support more generic first - see: https://github.com/oobabooga/text-generation-webui/discussions/1687. For now added to readme that it isn't supported


Wojtab:
@Ph0rk0z no, the projector is a separate 1-layer model, 10MB in total, I added it to description

Ph0rk0z:
Awesome!

PR #1263, opened 2023-04-16
Add support for GPTQ-triton using the --gptq-triton flag.

fpgaminer:
GPTQ-triton is my WIP implementation of the GPTQ kernels in Triton, which improve inference speed on GPTQ quantized models.  On short prompts it provides on average a 10% boost in performance relative to the CUDA kernels.  On large prompts it is over 10x faster.  It's the source of the Triton branch in GPTQ-for-LLaMa.

While GPTQ-triton is still a work in progress, I did integrate support for it in text-generation-webui and thought it might be useful to share the code, since it's a relatively simple addition.  So far I've tested this integration against the latest transformers, llama 7B at 4-bit quantization and groupsize -1.  Support for multi-gpu, other wbit settings, and Flash attention are not yet implemented.

If you don't feel comfortable integrating support yet, that's totally fine, just let me know what features are blockers.

Thank you.

PR #1668, opened 2023-04-30
Implement support for AutoGPTQ for loading GPTQ quantized models.

LaaZa:
This is a quick implementation of PanQiWei /[AutoGPTQ](https://github.com/PanQiWei/AutoGPTQ) for inference.

This is an alternative to the current GPTQ-for-LLaMA hopefully offering a more universally supported option which is not limited to one platform like linux.

AutoGPTQ supports cuda, triton(on linux) and ~cpu~. Splitting using pre_layer is not supported.

#1263 implements another alternative for GPTQ, but due to relying on triton it is not universal. This pr should be compatible with it though as an option.

Right now requires newer than 0.0.5 version of AutoGPTQ in pypi, so build from source is required at the time of writing.

From my testing appears to be slightly slower than GPTQ-for-LLaMA triton branch and slower still with cuda. I have not compared against cuda versions of GPTQ-for-LLaMA. Probably slower than #1263

But AutoGPTQ is seeing rapid development and likely will have better performance while maintaining compatibility and I think this is the main benefit of this implementation.

Please give feedback and testing is appreciated.


TheBloke:
Awesome! I strongly believe that AutoGPTQ is the way forward for GPTQ and yeah it's been seeing rapid progress recently.

I release a lot of GPTQs on HF and am really hoping that in the future it will be easier for users to use them. And I think AutoGPTQ is the right repo for the community to gather around for future development.

I will try your PR shortly with my GPTQs

Ph0rk0z:
Great it's finally working. I did the same thing 3 days ago and would get errors loading the state_dict.

LaaZa:
Added support for offloading and multiple devices.
Uses --gpu-memory and --cpu-memory and added --autogptq-device-map  to use these features [Accelerate: device_map](https://huggingface.co/docs/accelerate/main/en/usage_guides/big_modeling#designing-a-device-map) setting memory will use 'auto' unless something else is specified.

However this requires more testing and I don't think it was working properly for me.
My testing showed that vram was filled with the whole model regardless what memory was specified and ram was fluctuating during loading while vram was already filled(not full but normal memory usage). And loading was **very** slow.

To use these features, disable triton (--autogptq-triton flag).

Especially multiple gpus would be useful to test, for that use the --gpu-memory flag since the ui has only one device.
--auto-devices will use ooba's automatic memory setting like for normal non-gptq model loading.

You will likely need `protobuf==3.20`

TheBloke:
OK I've done some testing.  Firstly, thanks so much for getting this PR'd - it's awesome to see my models loading with AutoGPTQ!

My findings so far. I'll do some more testing tomorrow.

## All testing with:
* https://huggingface.co/TheBloke/wizardLM-7B-GPTQ
  * Bits = 4. Groupsize = 128. One model with desc_act, one without. 
* Ubuntu 20.04
* NVidia 4090 24GB
* CUDA 11.6
* AutoGPTQ installed with `pip install .` as of `e2c7cd4fb3765538569f903ca7f81563fce70c6e`
* LaaZa:AutoGPTQ as of `0fd4857646c56deb7b7a2b7df561e35e0a172d0b`

## AutoGPTQ testing

### Triton
`--model XYZ --wbits 4 --groupsize 128 --model_type llama --autogptq --autogptq-triton`

* `act-order (desc_act = True)` : WORKS. 11-12 token/s
* `no-act-order (desc_act = False)` : WORKS. 11-12 token/s

For comparison, GPTQ-for-LLaMa Triton records 15-16 token/s on this same GPU + model.

### CUDA
`--model XYZ --wbits 4 --groupsize 128 --model_type llama --autogptq`

* `act-order (desc_act = True)` :  **GIBBERISH** - [AutoGPTQ issue](https://github.com/PanQiWei/AutoGPTQ/issues/22#issuecomment-1529167155)
* `no-act-order (desc_act = False)` : **GIBBERISH** - [AutoGPTQ issue](https://github.com/PanQiWei/AutoGPTQ/issues/22#issuecomment-1529167155)

## text-generation-webui specific

### When running server.py with `--autogptq --autogptq-triton` and no models yet downloaded

* Download a model that has `quantize_config.json` defined then Refresh Model and choose model
  * UI throws error "no file named pytorch_model.bin" - it's trying to load it as a non-GPTQ model because no GPTQ params are defined. But there is a `quantize_config.json` so it could read this to know it's a GPTQ.
  * UI GPTQ params also should update to reflect the contents of `quantize_config.json`

### When running server.py with `--autogptq --autogptq-triton` and one or more GPTQ model downloaded which has `quantize_config.json`

* UI fails to load because it doesn't recognise the model as a GPTQ model:
```
root@cee7d387f502:~/textgen-auto# python server.py  --autogptq --autogptq-triton  --listen
Gradio HTTP request redirected to localhost :)
bin /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cuda116.so
Loading act...
  File "/root/textgen-auto/server.py", line 914, in <module>
  File "/root/textgen-auto/modules/models.py", line 84, in load_model
    model = LoaderClass.from_pretrained(Path(f"{shared.args.model_dir}/{model_name}"), low_cpu_mem_usage=True, torch_dtype=torch.bfloat16 if shared.args.bf16 else torch.float16, trust_remote_code=trust_remote_code)
  File "/usr/local/lib/python3.10/dist-packages/transformers/models/auto/auto_factory.py", line 471, in from_pretrained
    return model_class.from_pretrained(
  File "/usr/local/lib/python3.10/dist-packages/transformers/modeling_utils.py", line 2405, in from_pretrained
    raise EnvironmentError(
OSError: Error no file named pytorch_model.bin, tf_model.h5, model.ckpt.index or flax_model.msgpack found in directory models/act.
``` 

It should recognise the `quantize_config.json` and detect it as a GPTQ.

Otherwise the user must always specify `--wbits --groupsize --model_type` params in order to launch the server, even when `quantize_config.json` exists.


## Next steps

I'll do some more testing tomorrow, eg with GPU splitting, offloading, etc.

qwopqwop200:
https://github.com/qwopqwop200/AutoGPTQ-no-act-order
Here's a fork of auto-gptq that takes an existing old cuda and provides faster speeds.
I'm sure you'll probably get a faster speed.
However, this version does not support using act-order and groupsize at the same time.

TheBloke:
Just tried loading one of my models which doesn't use `safetensor`: https://huggingface.co/TheBloke/vicuna-13B-1.1-GPTQ-4bit-128g/tree/main

And I get: `FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g`

Folder structure:
```
root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g# ll
total 3809388
drwxrwxrwx  2 root root    3000362 Apr 30 23:45 ./
drwxrwxrwx 34 root root    3066244 Apr 30 23:28 ../
-rw-rw-rw-  1 root root        581 Apr 28 08:00 config.json
-rw-rw-rw-  1 root root        136 Apr 26 16:21 generation_config.json
-rw-rw-rw-  1 root root        411 Apr 26 16:21 special_tokens_map.json
-rw-rw-rw-  1 root root     499723 Apr 26 16:21 tokenizer.model
-rw-rw-rw-  1 root root        700 Apr 28 10:16 tokenizer_config.json
-rw-rw-rw-  1 root root 3894242469 Apr 12 23:08 vicuna-13B-1.1-GPTQ-4bit-128g.compat.no-act-order.pt
root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g#
```

Error:
```
root@cee7d387f502:~/textgen-auto# python server.py --model TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g  --autogptq --autogptq-triton  --listen --wbits 4 --groupsize 128 --model_type llama
Gradio HTTP request redirected to localhost :)
bin /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cuda116.so
Loading TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g...
  File "/root/textgen-auto/server.py", line 914, in <module>
  File "/root/textgen-auto/modules/models.py", line 158, in load_model
    model = load_quantized(model_name)
  File "/root/textgen-auto/modules/AutoGPTQ_loader.py", line 46, in load_quantized
    raise FileNotFoundError(f'No quantized model found for {model_name}')
FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g
root@cee7d387f502:~/textgen-auto#
```

I tried creating a `quantize_config.json` as well but that didn't help.

qwopqwop200:
Additionally, the speed difference between AutoGPTQ and GPTQ-for-LLaMA is due to fused-attn in cuda. Also, in the case of triton, it occurs because of fused-attn,fused-mlp.

LaaZa:
Thank you for your reports.

Only `--wbits` should matter whether it knows to load a GPTQ model.
`--model_type` has no effect on AutoGPTQ, it is automatically detected and completely ignored by this.
 `quantize_config.json`  will override any other settings. I'll look into updating the UI.

I didn't want to change the detection, just put this as an alternative when it tries to load GfL. The check for `quantize_config.json` would have to happen in a different place. I'll look into it if I can change the loading criteria without changing the code too much.


@qwopqwop200 I think a separate fork would go against the idea of implementing AutoGPTQ as a universal solution. Would it be possible to implement those optimisations for the AutoGPTQ main?

qwopqwop200:
I think it will be possible. I'll try later.

LaaZa:
> Just tried loading one of my models which doesn't use `safetensor`: https://huggingface.co/TheBloke/vicuna-13B-1.1-GPTQ-4bit-128g/tree/main
> 
> And I get: `FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g`
> 
> Folder structure:
> 
> ```
> root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g# ll
> total 3809388
> drwxrwxrwx  2 root root    3000362 Apr 30 23:45 ./
> drwxrwxrwx 34 root root    3066244 Apr 30 23:28 ../
> -rw-rw-rw-  1 root root        581 Apr 28 08:00 config.json
> -rw-rw-rw-  1 root root        136 Apr 26 16:21 generation_config.json
> -rw-rw-rw-  1 root root        411 Apr 26 16:21 special_tokens_map.json
> -rw-rw-rw-  1 root root     499723 Apr 26 16:21 tokenizer.model
> -rw-rw-rw-  1 root root        700 Apr 28 10:16 tokenizer_config.json
> -rw-rw-rw-  1 root root 3894242469 Apr 12 23:08 vicuna-13B-1.1-GPTQ-4bit-128g.compat.no-act-order.pt
> root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g#
> ```
> 
> Command:
> 
> ```
> python server.py --model TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g  --autogptq --autogptq-triton  --listen --wbits 4 --groupsize 128 --model_type llama
> ```
> 
> Error:
> 
> ```
> Gradio HTTP request redirected to localhost :)
> bin /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cuda116.so
> Loading TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g...
>   File "/root/textgen-auto/server.py", line 914, in <module>
>     shared.model, shared.tokenizer = load_model(shared.model_name)
>   File "/root/textgen-auto/modules/models.py", line 158, in load_model
>     model = load_quantized(model_name)
>   File "/root/textgen-auto/modules/AutoGPTQ_loader.py", line 46, in load_quantized
>     raise FileNotFoundError(f'No quantized model found for {model_name}')
> FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g
> root@cee7d387f502:~/textgen-auto#
> ```
> 
> I tried creating a `quantize_config.json` as well but that didn't help.

AutoGPTQ only supports `.safetensors` and `.bin` I simply cannot pass the other extensions since it only takes the name without the extension and adds either .bin or .safetensors to it based on the use_safetensors.

LaaZa:
Now checking for quantize_config.json and if it exists wbits does not need to be manually set.

I don't think it is related to anything in this commit but certain models fail to load.
For whatever reason it happens with pygmalion models, including the LLaMA based pygmalion-7B, so it isn't about gptj.
```
ValueError: QuantLinear() does not have a parameter or a buffer named bias.
coming from auto_gptq/modeling/_base.py at line 540: model = accelerate.load_checkpoint_and_dispatch(...)
```
Example model [`gozfarb/pygmalion-7b-4bit-128g-cuda`](https://huggingface.co/gozfarb/pygmalion-7b-4bit-128g-cuda)

qwopqwop200:
Created a PR to update AutoGPTQ to provide optimizations.
This is enabled automatically if act-order and groupsize are not used at the same time.
https://github.com/PanQiWei/AutoGPTQ/tree/faster-cuda-no-actorder

LaaZa:
> Created a PR to update AutoGPTQ to provide optimizations.
> This is enabled automatically if act-order and groupsize are not used at the same time.
> https://github.com/PanQiWei/AutoGPTQ/tree/faster-cuda-no-actorder

Nice. But does that mean that act-order needs to be passed? It can't be automatically checked for?

qwopqwop200:
An automatic check seems very hard to implement.

LaaZa:
> An automatic check seems very hard to implement.

Probably just have to do best effort trying to check the filename then. Do you have any ideas about the issue with pygmalion models?

qwopqwop200:
1.Given the way AutoGPTQ currently loads models, it's not a good idea to check whether or not by the name of a file.
2.It seems that the model was made with the old version of GPTQ. You will need to do a fresh conversion of your model to the latest GPTQ.

TheBloke:
We can specify `desc_act = true` in `quantize_config.json` and @LaaZa your code should check for this.

If there is going to be a difference in inference with/without desc_act, I think there probably needs to be a new command line parameter and UI GPTQ parameter in ooba to specify if the model is desc_act or not, eg `--desc_act` to specify model is desc_act.  So that older models can be loaded that don't have `quantize_config.json`

I will soon go back through all my GPTQ models on HF and add a quantize_config.json to every one.  One issue is that in my older repos I have two model files per repo - one with desc_act (act-order), and one without.  So I will have to move to using separate branches, one for the desc_act/act-order model, one for the no-desc_act/no-act-order model.

@qwopqwop200 how big of an improvement is using desc_act?  Maybe I should not even bother making desc_act models any more if they are always going to have problems for users on CUDA?

TheBloke:
> > Just tried loading one of my models which doesn't use `safetensor`: https://huggingface.co/TheBloke/vicuna-13B-1.1-GPTQ-4bit-128g/tree/main
> > And I get: `FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g`
> > Folder structure:
> > ```
> > root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g# ll
> > total 3809388
> > drwxrwxrwx  2 root root    3000362 Apr 30 23:45 ./
> > drwxrwxrwx 34 root root    3066244 Apr 30 23:28 ../
> > -rw-rw-rw-  1 root root        581 Apr 28 08:00 config.json
> > -rw-rw-rw-  1 root root        136 Apr 26 16:21 generation_config.json
> > -rw-rw-rw-  1 root root        411 Apr 26 16:21 special_tokens_map.json
> > -rw-rw-rw-  1 root root     499723 Apr 26 16:21 tokenizer.model
> > -rw-rw-rw-  1 root root        700 Apr 28 10:16 tokenizer_config.json
> > -rw-rw-rw-  1 root root 3894242469 Apr 12 23:08 vicuna-13B-1.1-GPTQ-4bit-128g.compat.no-act-order.pt
> > root@cee7d387f502:/workspace/TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g#
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Command:
> > ```
> > python server.py --model TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g  --autogptq --autogptq-triton  --listen --wbits 4 --groupsize 128 --model_type llama
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Error:
> > ```
> > Gradio HTTP request redirected to localhost :)
> > bin /usr/local/lib/python3.10/dist-packages/bitsandbytes/libbitsandbytes_cuda116.so
> > Loading TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g...
> > Traceback (most recent call last):
> >   File "/root/textgen-auto/server.py", line 914, in <module>
> >     shared.model, shared.tokenizer = load_model(shared.model_name)
> >   File "/root/textgen-auto/modules/models.py", line 158, in load_model
> >     model = load_quantized(model_name)
> >   File "/root/textgen-auto/modules/AutoGPTQ_loader.py", line 46, in load_quantized
> >     raise FileNotFoundError(f'No quantized model found for {model_name}')
> > FileNotFoundError: No quantized model found for TheBloke_vicuna-7B-1.1-GPTQ-4bit-128g
> > root@cee7d387f502:~/textgen-auto#
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I tried creating a `quantize_config.json` as well but that didn't help.
> 
> AutoGPTQ only supports `.safetensors` and `.bin` I simply cannot pass the other extensions since it only takes the name without the extension and adds either .bin or .safetensors to it based on the use_safetensors.

Oh yeah of course! OK. I guess I can go back through my older models and rename `.pt` files to `.bin`.  In my later models I only use `safetensors` anyway.

Or maybe I could PR a change to AutoGPTQ to also check for .pt files.. I will ask PanQiWei what they think.

TheBloke:
> @qwopqwop200 I think a separate fork would go against the idea of implementing AutoGPTQ as an universal solution. Would it be possible to implement those optimisations for the AutoGPTQ main?

I agree completely. I think AutoGPTQ should support everything. Otherwise it is really confusing for users.

> Only `--wbits` should matter whether it knows to load a GPTQ model. `--model_type` has no effect on AutoGPTQ, it is automatically detected and completely ignored by this. `quantize_config.json` will override any other settings. I'll look into updating the UI.
> 
> I didn't want to change the detection, just put this as an alternative when it tries to load GfL. The check for `quantize_config.json` would have to happen in a different place. I'll look into it if I can change the loading criteria without changing the code too much.

Yeah I understand. Maybe it is beyond the scope of this PR.  But I do think the UI should support the features I mention. With GPTQ-for-Llama, a text-gen-ui user had to either specify `--wbits --groupsize --model_type` command line params, or else fill in the "GPTQ Params" in the UI and then use "Save settings for this model" to save config to YAML.  If they didn't do that, the GPTQ model would fail to load because the UI would think it was an HF format model.

But now with AutoGPTQ, we can avoid that extra work for users because we have `quantize_config.json`.  So I think it will be very helpful to users if text-gen-ui can make use of this file to automatically load the right settings for GPTQ models in all scenarios.

But that could be done in a separate PR, after yours is merged. Or maybe ooba will do it himself.

qwopqwop200:
1.It's already implemented that way.
2.The current old cuda kernel is written assuming that desc_act is not enabled, so inference won't work.
3.There is a very small additional improvement of 0.01 to 0.03 when using act-order. However, for a specific model (ex:opt-66b) we experienced a significant improvement of around 0.2.

TheBloke:
> 1.It's already implemented that way. 
Thank you!

> 2.The current old cuda kernel is written assuming that desc_act is not enabled, so inference won't work.

But desc_act does seem to work with the current AutoGPTQ CUDA code? Using the latest AutoGPTQ in CUDA mode I can run inference on models I created with --act-order in GPTQ-for-Llama, and it does work.

So is it only old versions of GPTQ-for-LLaMa CUDA that can't use desc_act/--act-order?

>  3.There is a very small additional improvement of 0.01 to 0.03 when using act-order. However, for a specific model (ex:opt-66b) we experienced a significant improvement of around 0.2.

OK thank you. So for Llama, maybe it would be easier not to keep making desc_act/--act-order models, if it is going to cause problems for some users.

qwopqwop200:
> But desc_act does seem to work with the current AutoGPTQ CUDA code? Using the latest AutoGPTQ in CUDA mode I can run inference on models I created with --act-order in GPTQ-for-Llama, and it does work.
> 
> So is it only old versions of GPTQ-for-LLaMa CUDA that can't use desc_act/--act-order?

yes

Ph0rk0z:
Group size breaks 30b models for 24gb vram. So I liked act order to smarten them up slightly.

LaaZa:
> 1.Given the way AutoGPTQ currently loads models, it's not a good idea to check whether or not by the name of a file.
> 2.It seems that the model was made with the old version of GPTQ. You will need to do a fresh conversion of your model to the latest GPTQ.

1. What I meant was to try and check for no act order in the name if it is not specified elsewhere (setting or conf). Does desc_act=False trigger the faster cuda?

2. Weird if they are all using old versions. Is there any way to load them anyway? GPTQ for LLaMA does.

TheBloke:
Hi @LaaZa 

I was just reviewing the quantize_config.json code

In these lines: https://github.com/LaaZa/text-generation-webui/blob/b173274c63c7c133e6a71986051017e5cc9ff918/modules/AutoGPTQ_loader.py#L75-L80

If the user already has `quantize_config.json`, this code will not use it.

You only run `quantize_config=BaseQuantizeConfig()` if the user doesn't have a `quantize_config.json`. If they do have one, this is not called and you will then pass `quantize_config=None` to `AutoGPTQForCausalLM.from_quantized()`

Given you are already reading quantize_config.json in `set_quantize_config` and saving its contents to `shared.args`, I think you can remove the `if` from line 77 and always set quantize_config like so:
```
    quantize_config = BaseQuantizeConfig(
        bits=shared.args.wbits,
        group_size=shared.args.groupsize,
        desc_act=shared.args.desc_act
    )
```

(I assumed here that desc_act would be added as well)

TheBloke:
> Does desc_act=False trigger the faster cuda?

As of qwopqwop's latest PR (https://github.com/PanQiWei/AutoGPTQ/pull/38) desc_act=False models will load and run fine and fast in CUDA, but desc_act=True models will return gibberish.

So right now, CUDA doesn't support desc_act models at all I think. I've raised a question on that PR as to whether it would be possible to have CUDA support groupsize + desc_act: https://github.com/PanQiWei/AutoGPTQ/pull/38#issuecomment-1529704057

LaaZa:
> > @qwopqwop200 I think a separate fork would go against the idea of implementing AutoGPTQ as an universal solution. Would it be possible to implement those optimisations for the AutoGPTQ main?
> 
> I agree completely. I think AutoGPTQ should support everything. Otherwise it is really confusing for users.
> 
> > Only `--wbits` should matter whether it knows to load a GPTQ model. `--model_type` has no effect on AutoGPTQ, it is automatically detected and completely ignored by this. `quantize_config.json` will override any other settings. I'll look into updating the UI.
> > 
> > I didn't want to change the detection, just put this as an alternative when it tries to load GfL. The check for `quantize_config.json` would have to happen in a different place. I'll look into it if I can change the loading criteria without changing the code too much.
> 
> Yeah I understand. Maybe it is beyond the scope of this PR.  But I do think the UI should support the features I mention. With GPTQ-for-Llama, a text-gen-ui user had to either specify `--wbits --groupsize --model_type` command line params, or else fill in the "GPTQ Params" in the UI and then use "Save settings for this model" to save config to YAML.  If they didn't do that, the GPTQ model would fail to load because the UI would think it was an HF format model.
> 
> But now with AutoGPTQ, we can avoid that extra work for users because we have `quantize_config.json`.  So I think it will be very helpful to users if text-gen-ui can make use of this file to automatically load the right settings for GPTQ models in all scenarios.
> 
> But that could be done in a separate PR, after yours is merged. Or maybe ooba will do it himself.

I already used the quantize_config.json but I also added a function to update wbits and groupsize during model load so GPTQ is triggered even without manually set options if the file is present. Setting wbits without a file will still trigger GPTQ like before. AutoGPTQ will use the file if it exists regardless of settings. Currently the UI doesn't get updated because of the order of things.

LaaZa:
> Hi @LaaZa 
> 
> I was just reviewing the quantize_config.json code
> 
> In these lines: https://github.com/LaaZa/text-generation-webui/blob/b173274c63c7c133e6a71986051017e5cc9ff918/modules/AutoGPTQ_loader.py#L75-L80
> 
> If the user already has `quantize_config.json`, this code will not use it.
> 
> You only run `quantize_config=BaseQuantizeConfig()` if the user doesn't have a `quantize_config.json`. If they do have one, this is not called and you will then pass `quantize_config=None` to `AutoGPTQForCausalLM.from_quantized()`
> 
> Given you are already reading quantize_config.json in `set_quantize_config` and saving its contents to `shared.args`, I think you can remove the `if` from line 77 and always set quantize_config like so:
> ```
>     quantize_config = BaseQuantizeConfig(
>         bits=shared.args.wbits,
>         group_size=shared.args.groupsize,
>         desc_act=shared.args.desc_act
>     )
> ```
> 
> (I assumed here that desc_act would be added as well)

It should use the file automatically if quantize_config is None since that is default. I'm not manually passing the file in any situation.

TheBloke:
Right, sorry, yeah just saw your other comment.

So AutoGPTQ handles quantize_config.json if it's there, and we need to pass quantize_config to it only if we want to override the config file or if the config file is not there.

OK great, makes sense then.

qwopqwop200:
> > 1.Given the way AutoGPTQ currently loads models, it's not a good idea to check whether or not by the name of a file.
> > 2.It seems that the model was made with the old version of GPTQ. You will need to do a fresh conversion of your model to the latest GPTQ.
> 
> 1. What I meant was to try and check for no act order in the name if it is not specified elsewhere (setting or conf). Does desc_act=False trigger the faster cuda?
> 2. Weird if they are all using old versions. Is there any way to load them anyway? GPTQ for LLaMA does.

1. The trigger is `not(desc_act) or groupsize == -1`
2. First of all, in the case of AutoGPTQ, it seems difficult to load. In the case of GPTQ-for-LLaMA, it can be loaded with
`model.load_state_dict(torch.load(path),strict=False)`

PanQiWei:
> > AutoGPTQ only supports `.safetensors` and `.bin` I simply cannot pass the other extensions since it only takes the name without the extension and adds either .bin or .safetensors to it based on the use_safetensors.
> 
> Oh yeah of course! OK. I guess I can go back through my older models and rename `.pt` files to `.bin`. In my later models I only use `safetensors` anyway.
> 
> Or maybe I could PR a change to AutoGPTQ to also check for .pt files.. I will ask PanQiWei what they think.

Hi, I think using only .bin suffix for normal pytorch model and preserve .pt and .pth for the long-term future extension if we are going to support torchscript. I'm open to add an optional argument that allows users to decide whether to replace .bin to .pt internally when loading model at that time, a pr is welcomed.

TheBloke:
> Hi, I think using only .bin suffix for normal pytorch model and preserve .pt and .pth for the long-term future extension if we are going to support torchscript. I'm open to add an optional argument that allows users to decide whether to replace .bin to .pt internally when loading model at that time, a pr is welcomed.

OK thanks. No need for anything complex. I am happy to go through my HF model repos and rename .pt to .bin. Or I may re-make my early GPTQs anyway, to ensure they are compatible with the latest inference code.

Newer models I have only been using .safetensors. It's only the early ones I did where I used .pt for some of them.

Ph0rk0z:
How about support for V1 models? The ones that did not support GS.. like decapoda. Then you truly support it all.

LaaZa:
> How about support for V1 models? The ones that did not support GS.. like decapoda. Then you truly support it all.

Models without groupsize should be supported, but I'm not sure if the old models are.

Ph0rk0z:
I assumed they are not. Only have been working in https://github.com/johnsmith0031/alpaca_lora_4bit or much older GPTQ for me. They are a bit faster and use less memory.. of course their perplexity is lower too.

Wojtab:
quick question: if I quantized a model on old cuda, with `--wbits 4 --true-sequential --groupsize 128`, is this `quantize_config.json` correct?
```
{
  "bits": 4,
  "desc_act": false,
  "group_size": 128
}
```
if so I'll add it to my llava quant(s)

TheBloke:
Yes that's right

LaaZa:
I don't think the model loads though, because it is quantized with old GPTQ.

Wojtab:
> I don't think the model loads though, because it is quantized with old GPTQ.
> 
> The support in GPTQ-for-LLaMA might be a double edged sword, you can load old GPTQ models but on the other hand people may keep using the old GPTQ quantization methods because they seem to work.
> 
> I think the best might be to be able to load the old models but show a big warning. But since it might be difficult to implement for AutoGPTQ, this warning should be shown for GPTQ-for-LLaMA maybe.
> 
> Also any input on whether it should be `--autogpq-act-order` or `--autogptq-desc-act`

I just tested it(this PR + AutoGPTQ from `faster-cuda-no-actorder`), you're right, I get `ValueError: QuantLinear() does not have a parameter or a buffer named bias.`, so old format doesn't work

LaaZa:
It is not just the old cuda, it happens with every combination. Triton, cuda with act-order and without(with faster-cuda-no-actorder).

Ph0rk0z:
I just tested the "old_cuda" branch. It is actually 2x as slow as the ooba commit and what I have in merged GPTQ. 

I used the same 30b model and now, not only do I go OOM, but I get under 1it/s for small contexts. 

Maybe it is faster on newer arch than pascal or something. Response time went from 10s to 23s @430 cont. 20s used to be the response time for ~1800 cont. I was all excited too :(

>--autogpq-act-order

I like act order for consistency.

LaaZa:
No group size is supported.
Old cuda refers to the older cuda kernel that is faster than current but does not support act-order together with group size, that is currently the recommended GPTQ-for-LLaMA from ooba. Another is newer cuda branch in GfL which is slower but does support act-order, it isn't being updated though. The updated main branch is Triton and is faster than cuda and supports everything(except 3bit) but Triton itself is not supported on Windows and that is the main issue for textgen.

AutoGPTQ supports all of these in the main branch. But it does not support models made with some particular older version of GPTQ but I don't know what this specific difference is.

Ph0rk0z:
Act order by itself or with true sequential is supported by the ooba version. v1 isn't some particular version, but the original GPTQ when llama first came out, eg models from https://huggingface.co/decapoda-research. That one isn't so much important anymore.

originally when I tried to load models made with the ooba/v2 version using autogptq I had this error

```
  File "/home/mint/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/modules/module.py", line 2041, in load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for LlamaForCausalLM:
	Missing key(s) in state_dict: "model.layers.0.self_attn.k_proj.g_idx", "model.layers.0.self_attn.o_proj.g_idx",
```

It's the same error I got when trying to load a model using ooba that was made with the current cuda branch. The format is changed.

Triton is also unsupported by pascal cards.. which are the most accessible 24gb GPU at the moment, and can perform acceptable 3-6 it/s on a quantized 30b model. In essence triton cuts out anyone who didn't pay up for a 3090+, besides windows.

Wojtab:
> why are they still using that?

Simple answer: the recommended gptq in this repo (oobas cuda) quantizes them in this format. If someone(for example me with llava) wants to make a quant to be used with text-generation-ui, and they don't want to do a deep dive into gptq, they will select the default gptq from here.
If autogptq becomes the default, most people will probably quantize with autogptq

TheBloke:
I think this is definitely something that should be resolved if it can be.  One of the most popular models on HF, anon8231's Vicuna 13B GPTQ - which has 32,227 downloads - is also affected.

I raised it on the AutoGPTQ issues as referenced above.

I note that the latest GPTQ-for-LLaMa code can load the model, in both Triton and CUDA branches. So hopefully there's some compat fix that can be made to AutoGPTQ without a huge problem.

LaaZa:
Added support for [#43](https://github.com/PanQiWei/AutoGPTQ/pull/43) but also requires it, because I expect it to be merged by the time this PR is ready. 

Also enabled cpu support with the normal cpu flag, **but it shouldn't work**. Enabled it because it's optional and people can then test it so we can maybe track it down.

abdsrd:
Hello just wanted to follow up on @TheBloke I was able to load the model on the ooba's webui using an oder version (a few weeks back), after the update and refresh of the env, I am not able to load the GPTQ model in the webui even when setting group size to 128 wbit to 4 which generally solved this issue before.

What can I do to fix it please?

TheBloke:
> Hello just wanted to follow up on @TheBloke I was able to load the model on the ooba's webui using an oder version (a few weeks back), after the update and refresh of the env, I am not able to load the GPTQ model in the webui even when setting group size to 128 wbit to 4 which generally solved this issue before.
> 
> What can I do to fix it please?

What model are you talking about?  If it's one of my ones on HF and you need support, it's better to post on its discussion page and I'll help you there.

This thread is for discussing AutoGPTQ, a new GPTQ system which is still in development. It's unlikely to be the best solution for your issues right now.

Ph0rk0z:
@LaaZa @Wojtab 

It's not just that it's default. It's also the new code performance is worse and didn't get better. Other projects using GPTQ also stayed with the old version that you are complaining about.

TheBloke:
I'm working on a comprehensive series of benchmarks. Be ready in 30-60 mins,

TheBloke:
Right, benchmarks are done.

The full information is available in an AutoGPTQ Issue, which will be where I put any further updates and discussion: [Comprehensive benchmarking of AutoGPTQ; Triton vs CUDA; vs old 'ooba' GfL](https://github.com/PanQiWei/AutoGPTQ/issues/49)

I'll put the summary information here as well:

# Benchmarks of: ooba CUDA; AutoGPTQ CUDA; AutoGPTQ Triton

## Implementations tested
* AutoGPTQ as of https://github.com/PanQiWei/AutoGPTQ/pull/43, using text-gen-ui from [LaaZa's AutoGPTQ PR]( https://github.com/LaaZa/text-generation-webui/tree/AutoGPTQ)
* GPTQ-for-LLaMa using [ooba's GfL fork (oobabooga/GPTQ-for-LLaMa)](https://github.com/oobabooga/GPTQ-for-LLaMa), using text-gen-ui from `main`, commit: `875da16b7bc2a676d1a9d389bf22ee4579722073`.

## Test system
* Ubuntu 20.04, with Docker https://hub.docker.com/repository/docker/thebloke/runpod-pytorch-new/general
* CUDA toolkit 11.6
* NVidia 4090 24GB
* [WizardLM 7B 128g](https://huggingface.co/TheBloke/wizardLM-7B-GPTQ)

## Test method
* All benchmarking done in text-gen-ui, using the output time and token/s reported by it.
* text-gen-ui restarted between each test.
* Output limit = 512 tokens.
* 'Default' paramater set used + 'ban eos token' set so as to always get 512 tokens returned
* Run one test and discard results as warm up, then record 4 results.

## Results spreadsheet and overview charts

### Spreadsheet
[Google sheets with results and charts](https://docs.google.com/spreadsheets/d/1_VWHD-5jgVq5_nlP_0KJeKLsWDkZrZNBFdGR7YRMuME/edit?usp=sharing)

### Charts

benchmark no-stream, model: no-act-order

benchmark no-stream, model: act-order

A chart showing streaming performance is in the spreadsheet.

## Description of results

### AutoGPTQ vs 'ooba' CUDA
* AutoGPTQ CUDA outperforms GfL 'ooba' CUDA by 15% on a no-act-order model
* AutoGPTQ CUDA outperforms GfL 'ooba' CUDA by 10% on an act-order model (comparing AutoGPTQ on act-order model to GfL on no-act-order model)
* AutoGPTQ Triton is 5% slower than GfL 'ooba' CUDA
* AutoGPTQ Triton is 20% slower than AutoGPTQ CUDA

### desc_act models vs non-desc_act models
* AutoGPTQ CUDA is 4.5% slower on a desc_act model vs not
* AutoGPTQ Triton has no performance difference between desc_act model vs not
* AutoGPTQ CUDA records significantly higher GPU usage % on desc_act models
  * 80% usage with desc_act + fused_attn vs 30% with no-desc_act model
  * This might be a problem on weaker cards? That needs testing.
* AutoGPTQ Triton has only a few % extra GPU usage with desc_act models.

Ph0rk0z:
Pretty cool. You didn't have models with act order but no group size to test on ooba? And I assume this is benchmarks for ampere. I will have to try on earlier archs and see if the story is different.

Another thing I like to do is to do one at full context and one at low context like you did. The story can change. So you get "fast" performance at first and then by the time you are many messages in it becomes a crawl.

LaaZa:
I think the performance concerns are mostly fixed. 
Remaining issues are:
1. CPU infrerence does not work?
2. Offloading and splitting between devices acts weirdly
3. ooba GPTQ-for-LLaMA quantized models do not load.

Minor issue with this PR: quantize_config.json values are not reflected in the UI.

Ph0rk0z:
Fixed because it outperforms on a 4090? I hardly think that's fair.

Ph0rk0z:
I plan to. I just have to find/convert a model to be compatible with AutoGPTQ. The impact was always the lowest on the top of the line cards.

>I doubt it would be worse than GPTQ-for-LLaMA.

qwopqwop200's 20% "faster" old_cuda was 50% slower on my P40. Unless it's all works for me, forget everyone else.

LaaZa:
> I plan to. I just have to find/convert a model to be compatible with AutoGPTQ. The impact was always the lowest on the top of the line cards.
> 
> > I doubt it would be worse than GPTQ-for-LLaMA.
> 
> qwopqwop200's 20% "faster" old_cuda was 50% slower on my P40. Unless it's all works for me, forget everyone else.

AutoGPTQ uses the "old cuda" if you use cuda and load a model with no act-order or groupsize. Could you test how act-order + groupsize and no act-order + groupsize perform on your hardware? If it performs worse without act-order then we might look into some options how we select it. The purpose is absolutely not to exclude any hardware as long as it is reasonable. The aim is to offer best possible performance in every situation. If an optimisation comes at the sacrifice of something else, then it must be optional. And best would be if this can be done automatically.

Ph0rk0z:
Ok.. I am cooking my hat to eat.. the performance seems to be identical on the P6000 system. It has a slightly different env than the P40 server so I will test there too.  

a/b comparison  https://pastebin.com/8UkC7nJ2

Will have to look for a group size + act order model somewhere as well. Out of the 45 in the models folder on this machine, there is not a single one. Not a single model has failed to load yet either so kudos to that.

Ph0rk0z:
Now that I'm done being brain damaged. I merged all the PRs and made a proper autogptq.
Quantized a new model too so that I have *something* native.

Does autogptq not cache models? They take the full time to load every time.

```
AutoGPTQ (llama-7b-g128-newly quantized)
Output generated in 4.34 seconds (3.92 tokens/s, 17 tokens, context 36, seed 12026778)
Output generated in 4.33 seconds (3.93 tokens/s, 17 tokens, context 36, seed 2113006546)
Output generated in 4.34 seconds (3.92 tokens/s, 17 tokens, context 36, seed 235251226)

Output generated in 8.32 seconds (2.16 tokens/s, 18 tokens, context 1848, seed 753132182)
Output generated in 8.29 seconds (2.17 tokens/s, 18 tokens, context 1848, seed 2060631158)
Output generated in 8.29 seconds (2.17 tokens/s, 18 tokens, context 1848, seed 170653282)

AutoGPTQ (llama-7b-g128-newly quantized) fused_mlp
Output generated in 4.59 seconds (3.70 tokens/s, 17 tokens, context 36, seed 1372040907)
Output generated in 4.30 seconds (3.96 tokens/s, 17 tokens, context 36, seed 1913005001)
Output generated in 4.31 seconds (3.94 tokens/s, 17 tokens, context 36, seed 1390257500)

Output generated in 8.28 seconds (2.17 tokens/s, 18 tokens, context 1848, seed 433353161)
Output generated in 8.28 seconds (2.17 tokens/s, 18 tokens, context 1848, seed 485984634)
Output generated in 8.26 seconds (2.18 tokens/s, 18 tokens, context 1848, seed 1910686177)

OobaGPTQ (llama-7b-g128-newly quantized) Loads but generates nonsense
Output generated in 12.83 seconds (15.51 tokens/s, 199 tokens, context 36, seed 322898765)

OobaGPTQ (llama-7b-4bit) [no anything] 
Output generated in 2.44 seconds (9.84 tokens/s, 24 tokens, context 36, seed 162104450)
Output generated in 2.14 seconds (11.20 tokens/s, 24 tokens, context 36, seed 2028303563)
Output generated in 2.19 seconds (10.97 tokens/s, 24 tokens, context 36, seed 454894630)

Output generated in 5.57 seconds (3.95 tokens/s, 22 tokens, context 1848, seed 2048174070)
Output generated in 5.63 seconds (3.91 tokens/s, 22 tokens, context 1848, seed 1237116859)
Output generated in 5.62 seconds (3.91 tokens/s, 22 tokens, context 1848, seed 1358685030)
```

Everything else doesn't want to load and errors at: 
`ValueError: QuantLinear() does not have a parameter or a buffer named bias.`

Is there some trick to load old cuda models?

But as you see I go from literally 10it/s down to 4it/s on a 7b with no context. On a 30b this becomes unusable on my card. The fused mlp does nothing. 

Need to see if I can convert a 13b with no group size and see what happens there.

LaaZa:
`fused_mlp` is only for Triton.

Ph0rk0z:
according to https://github.com/PanQiWei/AutoGPTQ/pull/43#issuecomment-1531451088 it was supposed to work.

TheBloke:
fused_attn works also on CUDA (and provides a significant perf boost, as my benchmarks showed)

fused_mlp is Triton only.

Ph0rk0z:
I will run it again when this 13b finishes. Make sure I didn't have brain worms again.

LaaZa:
Try to quantize it without act-order or groupsize so it can run on the old cuda.

Ph0rk0z:
Yea.. I deleted groupsize so I have 1:1 comparison. It finished and then error-ed when packing. I'm just using basic_usage.py and changing the paths. It's definitely much faster at quantizing.. if it saves this time.

TheBloke:
You can use groupsize on the old ooba CUDA. Just not act-order.

LaaZa:
> You can use groupsize on the old ooba CUDA. Just not act-order.

Yeah but isn't it that you can't use them together the only limitation?

Ph0rk0z:
Yea.. but I have nothing else with group size that I have FP16 for.  I want 2 models without any group size.

Ph0rk0z:
With the bug fixed in _utils.py I made a model with no groups or AO

`91:        if not(desc_act) or group_size == -1:`


Normal tests with the UI in --chat.
```
>AutoGPTQ (llama-13b-new-4bit 7.0gb) Loaded the model in 175.54 seconds
Output generated in 15.68 seconds (1.02 tokens/s, 16 tokens, context 36, seed 1286732055)
Output generated in 15.30 seconds (1.05 tokens/s, 16 tokens, context 36, seed 1550791109)
Output generated in 15.30 seconds (1.05 tokens/s, 16 tokens, context 36, seed 1175855933)

Output generated in 13.93 seconds (1.29 tokens/s, 18 tokens, context 1848, seed 2129637862)
Output generated in 13.86 seconds (1.30 tokens/s, 18 tokens, context 1848, seed 758650593)
Output generated in 13.91 seconds (1.29 tokens/s, 18 tokens, context 1848, seed 1861719483)


>AutoGPTQ (llama-13b-new-4bit 7.0gb) --quant_attn Loaded the model in 176.98 seconds
Output generated in 15.49 seconds (1.03 tokens/s, 16 tokens, context 36, seed 1035701343)
Output generated in 15.22 seconds (1.05 tokens/s, 16 tokens, context 36, seed 1536583082)
Output generated in 15.24 seconds (1.05 tokens/s, 16 tokens, context 36, seed 1145894960)

Output generated in 13.37 seconds (1.35 tokens/s, 18 tokens, context 1848, seed 1162984787)
Output generated in 13.46 seconds (1.34 tokens/s, 18 tokens, context 1848, seed 929173489)
Output generated in 13.49 seconds (1.33 tokens/s, 18 tokens, context 1848, seed 257706299)


>OobaGPTQ (llama-13b-new-4bit 7.0gb) Loaded the model in 3.83 seconds. And it works this time.
Output generated in 2.67 seconds (5.61 tokens/s, 15 tokens, context 36, seed 2048670938)
Output generated in 2.32 seconds (6.47 tokens/s, 15 tokens, context 36, seed 918606258)
Output generated in 2.30 seconds (6.51 tokens/s, 15 tokens, context 36, seed 1597699954)

Output generated in 9.33 seconds (1.82 tokens/s, 17 tokens, context 1848, seed 708157762)
Output generated in 9.29 seconds (1.83 tokens/s, 17 tokens, context 1848, seed 220563899)
Output generated in 9.31 seconds (1.83 tokens/s, 17 tokens, context 1848, seed 2008428659)

>OobaGPTQ (llama-13b-4bit) old,no groups, doesn't work in AutoGPTQ, Loaded the model in 107.71 seconds.
Output generated in 2.61 seconds (5.74 tokens/s, 15 tokens, context 36, seed 921361998)
Output generated in 2.29 seconds (6.56 tokens/s, 15 tokens, context 36, seed 184740008)
Output generated in 2.31 seconds (6.49 tokens/s, 15 tokens, context 36, seed 1407407442)

Output generated in 9.00 seconds (1.44 tokens/s, 13 tokens, context 1848, seed 340666632)
Output generated in 8.98 seconds (1.45 tokens/s, 13 tokens, context 1848, seed 44663723)
Output generated in 9.06 seconds (1.43 tokens/s, 13 tokens, context 1848, seed 1723085104)

```


And now we test with silly tavern + proxy for sustained speeds.

```
>AutoGPTQ (llama-13b-new-4bit 7.0gb) -Outputs "instruction" for some reason into chat

Output generated in 93.79 seconds (2.65 tokens/s, 249 tokens, context 1139, seed 19927686)
Output generated in 93.84 seconds (2.65 tokens/s, 249 tokens, context 1139, seed 1022083023)
Output generated in 93.06 seconds (2.68 tokens/s, 249 tokens, context 1139, seed 1436956536)

>OobaGPTQ (llama-13b-new-4bit 7.0gb), had to switch chars because aqua didn't talk enough

Output generated in 15.21 seconds (9.21 tokens/s, 140 tokens, context 1139, seed 1223496528)
Output generated in 7.53 seconds (4.65 tokens/s, 35 tokens, context 1139, seed 549852981)
Output generated in 10.87 seconds (7.45 tokens/s, 81 tokens, context 1139, seed 451611516)
Output generated in 25.61 seconds (9.72 tokens/s, 249 tokens, context 1362, seed 2010231807)
Output generated in 24.66 seconds (9.53 tokens/s, 235 tokens, context 1362, seed 92789824)
```


So between perf loss and having to re-encode all of the models. AutoGPTQ and the newer version of GPTQ cuda has not been appealing for people to release models or to use. Hence almost all of my models I download make `ValueError: QuantLinear() does not have a parameter or a buffer named bias.`

Now with triton, I think more people made triton models. And the good news is that without group size regular GPTQ can load the autoGPTQ model.


For shits and giggles I tested autograd from the 4bit lora repo for inference.. before it was made incompatible.
The newly converted model made:

`RuntimeError: probability tensor contains either `inf`, `nan` or element < 0`

The original llama-13b-4bit works but his kernel is also slower compared to the previous version.

```
Output generated in 42.76 seconds (2.50 tokens/s, 107 tokens, context 1362, seed 2009122068)
Output generated in 92.89 seconds (2.68 tokens/s, 249 tokens, context 1362, seed 1303447712)
Output generated in 92.08 seconds (2.70 tokens/s, 249 tokens, context 1362, seed 587758066)
```

So something introduced in newer versions of the cuda kernel (s) have caused a performance issue.

LaaZa:
I think we have to summon @qwopqwop200 from his lair of terrible commit messages to give some input on this.

Ph0rk0z:
It seems like it happened around the end of march. I have looked through the kernels too to see what my brain could come up with.. have come up empty.

I can literally *HEAR* the difference as there is no coil whine when using  a "slow" kernel.

qwopqwop200:
Please let us know what bugs and improvements you have.

LaaZa:
> Please let i know what bugs and improvements you have.

We need to find a way to load old models. I don't think there is a way around it, there are too many.

What is the main reason the loading behaviour in GPTQ-for-LLaMA cannot be replicated in AutoGPTQ?
And why is @Ph0rk0z having bad performance compared to ooba's fork?

TheBloke:
> Please let i know what bugs and improvements you have.

Please see this bug raised in the AutoGPTQ repo yesterday regarding issues loading older GPTQ models in AutoGPTQ:  https://github.com/PanQiWei/AutoGPTQ/issues/45

This seems to be the main compatibility issue being experienced.

qwopqwop200:
> We need to find a way to load old models. I don't think there is a way around it, there are too many.

This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.

> What is the main reason the loading behaviour in GPTQ-for-LLaMA cannot be replicated in AutoGPTQ? And why is @Ph0rk0z having bad performance compared to ooba's fork?

I guess this is because the parameter called faster is enabled by default. I'm going to make this selectable as a setting.
https://github.com/PanQiWei/AutoGPTQ/blob/faster-cuda-no-actorder/auto_gptq/nn_modules/qlinear_old.py#L21

TheBloke:
> > We need to find a way to load old models. I don't think there is a way around it, there are too many.
> 
> This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.
>
Can we do this conditionally only for the models that need it? Meaning, can we detect which models will throw `ValueError: QuantLinear() does not have a parameter or a buffer named bias`?

Worst case if it could be a setting and the user can be told to enable the setting for old models that need it.  But if it could be detected automatically, that would be ideal.

LaaZa:
> > We need to find a way to load old models. I don't think there is a way around it, there are too many.
> 
> This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.
> 
> > What is the main reason the loading behaviour in GPTQ-for-LLaMA cannot be replicated in AutoGPTQ? And why is @Ph0rk0z having bad performance compared to ooba's fork?
> 
> I guess this is because the parameter called faster is enabled by default. I'm going to make this selectable as a setting. https://github.com/PanQiWei/AutoGPTQ/blob/faster-cuda-no-actorder/auto_gptq/nn_modules/qlinear_old.py#L21

1. I think it's okay, if it means the model can be loaded in at all. But obviously it should be an option or preferrably automatic.
2. Is it possible the "faster" is slower on all pascal cards?

LaaZa:
> > > We need to find a way to load old models. I don't think there is a way around it, there are too many.
> > 
> > 
> > This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.
> 
> Can we do this conditionally only for the models that need it? Meaning, can we detect which models will throw `ValueError: QuantLinear() does not have a parameter or a buffer named bias`?
> 
> Worst case if it could be a setting and the user can be told to enable the setting for old models that need it. But if it could be detected automatically, that would be ideal.

That would be a bit slow to do every time, since it needs to do all the other things first before that error is raised. Some other way to test would be ideal.

qwopqwop200:
> > > We need to find a way to load old models. I don't think there is a way around it, there are too many.
> > 
> > 
> > This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.
> 
> Can we do this conditionally only for the models that need it? Meaning, can we detect which models will throw `ValueError: QuantLinear() does not have a parameter or a buffer named bias`?
> 
> Worst case if it could be a setting and the user can be told to enable the setting for old models that need it. But if it could be detected automatically, that would be ideal.

You can make this check automatically. But since you have to load the model to check it. Not a very good idea.

Ph0rk0z:
Well if it's the fault of faster I can try to disable it on both autograd and in autogptq and see if speed returns to ooba levels.

So does this faster being disabled hurt newer cards? I have the situation of running 65b between P40 and 3090.

qwopqwop200:
> > > We need to find a way to load old models. I don't think there is a way around it, there are too many.
> > 
> > 
> > This can be fixed with torch.load_state_dict(,strict=False) but it seems that it will make the model load slower.
> > > What is the main reason the loading behaviour in GPTQ-for-LLaMA cannot be replicated in AutoGPTQ? And why is @Ph0rk0z having bad performance compared to ooba's fork?
> > 
> > 
> > I guess this is because the parameter called faster is enabled by default. I'm going to make this selectable as a setting. https://github.com/PanQiWei/AutoGPTQ/blob/faster-cuda-no-actorder/auto_gptq/nn_modules/qlinear_old.py#L21
> 
> 1. I think it's okay, if it means the model can be loaded in at all. But obviously it should be an option or preferrably automatic.
> 2. Is it possible the "faster" is slower on all pascal cards?

1. I think auto will make it slower, so I'm adding it as an option.
2. By default "faster" accelerates the speed using fp16. However, I'm not sure if this slows down the GPU on the Pascal architecture.

Ph0rk0z:
I will change it manually and give it a shot.. in the other repo it looks like faster are separate functions.. But I will benchmark again after that.

LaaZa:
Maybe I'll make in textgen a check for this error and then try reloading with the option enabled. That way, we can infrom the user to set the option to load the model faster in the future. But there could also be an opportunity to set that setting for the user when this happens for the first time for a model. Maybe you can raise a more clear error saying that it is an old model and what to do to load it.

Ph0rk0z:
Took off "faster". Speed is improved. Model loading was slow before and it's slow now. 
Reload is always just as long. I'm starting to feel it :)

```
>llama-13b-new-4bit - Loaded the model in 217.14 seconds.

Output generated in 7.68 seconds (2.08 tokens/s, 16 tokens, context 36, seed 1738797549)
Output generated in 2.18 seconds (7.33 tokens/s, 16 tokens, context 36, seed 889332095)
Output generated in 2.25 seconds (7.12 tokens/s, 16 tokens, context 36, seed 998761703)

Output generated in 9.10 seconds (1.98 tokens/s, 18 tokens, context 1848, seed 719422260)
Output generated in 9.09 seconds (1.98 tokens/s, 18 tokens, context 1848, seed 2056004316)
Output generated in 9.55 seconds (1.89 tokens/s, 18 tokens, context 1848, seed 717474184)

```
For some reason the first generation was slow.

And then again in tavern/proxy.

```
Output generated in 27.94 seconds (8.91 tokens/s, 249 tokens, context 1362, seed 837243053)
Output generated in 27.62 seconds (9.02 tokens/s, 249 tokens, context 1362, seed 1533629634)
Output generated in 26.73 seconds (9.32 tokens/s, 249 tokens, context 1362, seed 1670684038)
```

quant_attn is slower?

```
Output generated in 27.99 seconds (8.89 tokens/s, 249 tokens, context 1362, seed 1176969996)
Output generated in 28.13 seconds (8.85 tokens/s, 249 tokens, context 1362, seed 534975496)
Output generated in 28.04 seconds (8.88 tokens/s, 249 tokens, context 1362, seed 600618292)
```

qwopqwop200:
Added an option called strict for loading outdated models.
Added use_cuda_fp16 (faster). Enabled by default.

LaaZa:
Added those things with `--autoqptq-compat` and `--autogptq-cuda-tweak`

But I could not get model unloading to work automatically(you need to use the unload or reload button) so there is no automatic compatibility mode fallback for now. If anyone has ideas how to unlod the model in the exept, then maybe I can add it.

LaaZa:
So with very quick and lazy testing. Compatibility mode(strict=False) seems to be much faster to load, especially on cuda.
So, is there are reason not to always use it? Or is something weird on my end?

Ph0rk0z:
Well I'm going to try it today on the dual card server.. I hope that it loads models like the old GPTQ did without strict dict. A reload of even a 30b was 20s and now it's 2-3 minutes.

LaaZa:
> Well I'm going to try it today on the dual card server.. I hope that it loads models like the old GPTQ did without strict dict. A reload of even a 30b was 20s and now it's 2-3 minutes. 

I don't understand what that reload thing is about. The whole model should be removed from memory. Because the reload unloads the model first. For me reload isn't any faster than loading with GfL.

Ph0rk0z:
I get module` 'quant_cuda' has no attribute 'vecquant4matmul_old'` on the other system...

The way this installs with the repo is not compatible with ooba GPTQ.. but it does work with the sterlind and my fork being installed at the same time. It will also cause issues if you want to use kobold 4bit in the same environment. Maybe it needs to be autogptq_cuda or something to make less headache.

```
Autogptq Loading vicuna-13b-free..
Model loaded in 243 seconds.

Quit and load in old GPTQ
Model loaded in 5.3s

Quit and load in Autogptq
Model loaded in 242.3s
```

TheBloke:
Yes I have been planning to raise an issue on AutoGPTQ suggesting the CUDA part is renamed to autogptq_cuda or something, to avoid conflicts with GfL.

I'll do it in a minute

oobabooga:
Question 1: does AutoGPTQ support the same CUDA behavior as this branch? https://github.com/oobabooga/gptq-for-llama

There are 3 branches that I have tested: my old fork, the upstream cuda branch, and the upstream old-cuda branch that @qwopqwop200 recently created (thank you for that!). I have compared the performance of my fork and `old-cuda`, and they are apparently the same in tokens/s.

-----

Question 2: does this work on Windows? Are there precompiled binaries, or does it require `nvcc` to be available?

-----

About this PR, I couldn't test it so far because I get this error

```
    model = AutoGPTQForCausalLM.from_quantized(path_to_model,
TypeError: AutoGPTQForCausalLM.from_quantized() got an unexpected keyword argument 'fused_attn'
```

I'm using the latest `auto-gptq`.

Ph0rk0z:
You need to merge all the PRs into autogptq.

oobabooga:
What do you mean?

Ph0rk0z:
Merge all the PRs in the AutoGPTQ repo for it to work.

TheBloke:
1. Yes. There is still some testing to be done, but generally now all models that load with ooba CUDA load with AutoGPTQ

As Phorkoz is alluding to, this PR currently relies on this PR: https://github.com/PanQiWei/AutoGPTQ/pull/43

This PR shouldn't be merged until that AutoGPTQ PR is merged, which should be in the next few days

I'm planning to do a range of further model tests in the next few days and will post results here, to ensure all bases are covered wrt compatibility.

2. Not sure but I think this has been tested yet.

Ph0rk0z:
Its still got a ways to go.. model loading takes forever and a day. On the 65b it fills the GPUs and sits there for additional time.


Also.. the memory splitting is weird. On normal GPTQ, I load 17/23 so that it fills the first GPU to 17G and then fills the 2nd. It would then come back and fill in the rest as needed for a nice 19/19 split. Here it doesn't. I get 17/17 and lopsided inference which causes OOM at 500 context. Using normal GPTQ I can do around 12-1500 context using the same split. To it's credit it is up to 1t/s now.

LaaZa:
@oobabooga I would like input on adding a few model specific options to the model page. I have not yet touched the UI. Settings like act-order and compatibility would be useful on the model page and saved for each model.

TheBloke:
> Yes, CPU offloading / GPU splitting is still not fully in AutoGPTQ I believe. So if this PR is merged before that's ready in AutoGPTQ it would probably need to be made clear it doesn't work yet in AutoGPTQ.

Just checked @PanQiWei 's Project list (https://github.com/users/PanQiWei/projects/1) and he lists CPU offload and multi-device inference as on the to-do for the next version, v0.2.0.  So this should mean he plans to start looking at it soon.

v0.2.0 is also the version he plans to merge the faster-llama PR, which this PR is already dependent on.  

So once AutoGPTQ is at 0.2.0 hopefully it will provide all the features needed for text-gen-UI, subject to testing.

oobabooga:
With the PR code at https://github.com/PanQiWei/AutoGPTQ/tree/faster-llama I managed to load a model with

```
python server.py --model llama-30b-4bit-128g --listen --autogptq --autogptq-compat
```

The performance in tokens/s seems to be the same, but it's a bummer to have to use an additional `autogptq-compat` flag.

> IMHO it's probably wise to wait at least a little while longer before merging this PR, for further testing and fine tuning to be done in AutoGPTQ. It wouldn't be good to create a bad first impression in users.

That's what I am inclined towards. Ideally, this would act as a replacement to the current GPTQ loaders rather than something to use in parallel.

@LaaZa about adding more settings to the interface: couldn't parameters like autogptq-compat and act-order be included in `quantize_config.json`? AutoGPTQ removes the need for `--wbits` and `--groupsize` but introduces new parameters that have to be guessed for a model to be loaded.

oobabooga:
As for `--pre_layer`, +1 that this is essential. With `--pre_layer 40` I can run `llama-65b` with 1 token every 8 seconds on my RTX 3090, and with `--pre_layer 20` I can run llama-7b with 4GB VRAM at about 1 token/second.

TheBloke:
> The performance in tokens/s seems to be the same, but it's a bummer to have to use an additional `autogptq-compat` flag.

This may not be needed. We're talking about that over in AutoGPTQ.  it may be possible to set the default to what autogptq-compat currently does.
 
> That's what I am inclined towards. Ideally, this would act as a replacement to the current GPTQ loaders rather than something to use in parallel.

I think that would be awesome.

> @LaaZa about adding more settings to the interface: couldn't parameters like autogptq-compat and act-order be included in `quantize_config.json`? AutoGPTQ removes the need for `--wbits` and `--groupsize` but introduces new parameters that have to be guessed for a model to be loaded.

* desc_act/act-order is already in quantize_config.json
*compat may not be needed as discussed above, hopefully we'll have more clarity on that soon
* For all new models, quantize_config should replace all need for manual params.  The only issue is existing models that don't have those files, and for which the uploader doesn't know to or is not interested in providing them.

Personally I will provide quantize_config.json for all my models and will encourage others to do the same, and PR them to HF for other people's models if they don't provide them themselves. 

But there's bound to be a number of popular legacy models where the uploader is no longer interested in maintaining it, so for at least a while the manual params will still be needed

oobabooga:
For models that don't have quantize_config.json, we can keep inferring the parameters from the model name through https://github.com/oobabooga/text-generation-webui/blob/main/models/config.yaml.

LaaZa:
@oobabooga quantize_config.json is the file that AutoGPTQ creates when it is used for quantization, you can of course manually make one. If that file is present, it'll override any relevant settings in textgen. However, since the compatibility mode is needed for models made with the old recommended GPTQ-for-LLaMA this option is not native for quantize_config.json. It is not impossible to add it but it is a bit weird place to have it in, since those old models wouldn't have that file in the first place and would have to be manually added. act-order is a setting that similarly to wbits and groupsize would only matter in the UI when the quantize_config.json does not exist.

Ph0rk0z:
I just turned off the compat flag and the 16bit in the loader. I see no difference on the 3090 or the P40.

So does act order by itself not work? Because most people didn't use act order with group size until triton and all those models are labeled as such. I should try some models that I did true sequential + act order on and see if anything changes. Get the feeling it will work like it does in autograd or oobaGPTQ.

oobabooga:
I see @LaaZa. In any case, the end goal is to be able to load the llama models with `--model` and no additional flag. 

The current recommended llama models are the ones in these torrents:

https://github.com/oobabooga/text-generation-webui/pull/530#issuecomment-1484235789 

which require an additional `--autogptq-compat`. I will wait for the updates mentioned by @TheBloke.

This is going in the right direction, it's refreshing to see a standardization effort for GPTQ.

LaaZa:
If we determine that the compatiblity mode is better to have on always, then I can flip the flag to disable it instead. Then you can load all models without any additional options(assuming wbit, groupsize and act-order are defined somewhere).

Ph0rk0z:
You could also save a config for models that don't have one if they loaded successfully, right?

LaaZa:
> You could also save a config for models that don't have one if they loaded successfully, right?

That is possible. I could do that if you think it would be beneficial over webui configs that you could save anyway. But I think that technically would prevent you form loading any other model from the same directory because you cannot set wbits to none.(which currently determines if it is supposed to load gptq model)

Ph0rk0z:
Yea.. I suppose it will make no difference since usually the parameters will get specified anyway.

LaaZa:
@oobabooga ??

MillionthOdin16:
Yeah this is one of the most important PRs I've been following...

abdsrd:
I dont understand will this be merged and added to base code? Why was it closed ?

LaaZa:
@oobabooga now seriously, please explain why you closed this out of nowhere?

Malrama:
@oobabooga ? 😢

LaaZa:
I will keep an updated fork [here](https://github.com/LaaZa/text-generation-webui/tree/AutoGPTQ) so if you want to use AutoGPTQ with textgen, you can use that fork.

TheBloke:
Thanks LaaZa.

My guess is that ooba closed it by mistake. Clicked the wrong button or something. There's not been any purge of other open PRs. And he said he was definitely interested.

If we've not heard soon, maybe just open a new PR.

LaaZa:
> Thanks LaaZa.
> 
> My guess is that ooba closed it by mistake. Clicked the wrong button or something. There's not been any purge of other open PRs. And he said he was definitely interested.
> 
> If we've not heard soon, maybe just open a new PR.

[.](its-not-a-mistake)

TheBloke:
Based on..?

LaaZa:
> Based on..?

Message me on discord

MillionthOdin16:
He just merged his own version lol. It's pretty bad that he didn't even comment on this at all...

LaaZa:
> He just merged his own version lol. It's pretty bad that he didn't even comment on this at all...

Because the truth doesn't look too good for him...

MillionthOdin16:
He just instantly closed an issue pointing it out lol. This dude... #2142

MillionthOdin16:
I wonder if you did the same thing with super booga? Because the PR for that was originally called SuperBig, then he did his own implementation and called it superbooga. He seems to push away people who contribute cool things and try to implement them himself...

LaaZa:
If you want the truth, message me on discord LaaZa#9463

oobabooga:
I closed this because the author decided it would be funny to mock and insult me in another PR. This kind of behavior is not accepted in this repository. Cut the drama and try to be constructive instead.

oobabooga:
I needed code created here to load models without a `quantize_config.json`. To give @LaaZa due credit, I am merging the PR.

PR #1675, opened 2023-04-30
Fix Training Pad Token

practicaldreamer:
Currently padding with `0` the character vs `0` the token id (`<unk>` in the case of llama)


practicaldreamer:
I'll reopen this PR under another branch of my fork

PR #1678, opened 2023-04-30
Fix Training Pad Token

practicaldreamer:
Currently modules/training.py is padding with `0` the **token** vs `0` the **token id** 
token_id 0 = "`<unk>`" (for llama at least)
token 0 = "`0`"

### DIFF
--- a/modules/training.py
+++ b/modules/training.py
@@ -243,7 +243,7 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
         return
 
     gradient_accumulation_steps = batch_size // micro_batch_size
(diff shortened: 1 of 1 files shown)


oobabooga:
I don't know what difference this makes, but it seems to make sense

practicaldreamer:
I’m not an expert at this but I believe the difference is padding with special tokens that tell the model to ignore vs padding with legitimate tokens (zeros) 

It’s kind of like the difference between null and 0. They’re similar but while null is the absence of a value 0 is a legitimate value.

The way it was it was prepending “0000000000000000000” etc… to the start of every example fed to the model

PR #1679, opened 2023-04-30
Added instruct-following template for Metharme

minipasila:
Adds support for a new instruct model called Metharme from PygmalionAI. https://huggingface.co/PygmalionAI/metharme-7b


Wojtab:
I think the name should be `<|model|>` not system. Also the turn template looks different than the example prompt on hugging face (additional `\n`-s, no endoftext). 
You can also add the model config here: https://github.com/oobabooga/text-generation-webui/blob/main/models/config.yaml to automatically go to instruct mode and select this template

minipasila:
does it look right now?

Wojtab:
I think so

CarlKenner:
> does it look right now?

No. It's wrong. There's not supposed to be `</s>` in the turn_template.
Here's an example from the documentation:
```
<|system|>This is a text adventure game. Describe the scenario to the user and give him three options to pick from on each turn.<|user|>Start!<|model|>You are standing in front of an old, abandoned house. The windows are boarded up, and there's no sign of life around it. As you approach, you notice a strange feeling emanating from within. Suddenly, you hear a voice calling out to you... 'Come inside!'

- Go inside the house.
- Ignore the call and move away.
- Run as fast as you can.<|user|>go inside<|model|>
```
Notice how it has no `</s>` after the bot's response (and before the next turn) when passing the prompt back to the model?
Yes, the bot emits `</s>` when it has finished its response, but it doesn't want that `</s>` passed back to it in the next query.

CarlKenner:
BTW, your changes add support for `chat` mode, but oobabooga also has a `notebook` mode.
Notebook mode uses `.txt` files in the `prompts` folder, and you must specify the default prompt (the name of the `.txt` file but without the `.txt` ending) for each model in two places: `settings-template.json` and `modules/shared.py`. Note that settings-template is a json file so it uses double quotes and a comma after each line except the last one, while shared.py is a python file so it uses either double or single quotes and you can have a comma after the last line. But other than that they should contain the same information.
If it was me, I'd use the text adventure example above (from their documentation) as the `prompts/Metharme.txt` file, since we don't really have any other examples.

The `README.md` file also has a list of supported instruction modes, which you can add to:
`* Instruct mode compatible with Alpaca, Vicuna, Open Assistant, `...

And when TheBloke (and others) inevitably releases his GPTQ version of Metharme, oobabooga's GPTQ loader needs to know that Metharme is a `llama` model (rather than OPT or GPT-J). So you can modify the `load_quantized` function in `modules/GPTQ_loader.py` to also recognise metharme as `llama` model type.

minipasila:
Is it good now?

CarlKenner (approved):

Dampfinchen:
Metharme looks to be very promising. I hope this will be implemented soon. The formatting is indeed correct.

Issue #1648, opened 2023-04-29
Issue with Elevenlabs TTS on one-click install


Siigari:
### Describe the bug

Hello, I am getting this error:

Loading the extension "gallery"... Ok.
Loading the extension "send_pictures"... Ok.
Loading the extension "sd_api_pictures"... Ok.
Open error: is JACK server running?
Loading the extension "elevenlabs_tts"... Fail.
Traceback (most recent call last):
  File "F:\oobabooga_windows\text-generation-webui\modules\extensions.py", line 33, in load_extensions
    exec(f"import extensions.{name}.script")
  File "<string>", line 1, in <module>
  File "F:\oobabooga_windows\text-generation-webui\extensions\elevenlabs_tts\script.py", line 6, in <module>
    from elevenlabslib.helpers import save_bytes_to_path
ImportError: cannot import name 'save_bytes_to_path' from 'elevenlabslib.helpers' (F:\oobabooga_windows\installer_files\env\lib\site-packages\elevenlabslib\helpers.py)

I do not know what to do about this. I use an ASIO driver for audio, and it is running (so the Jack server *is* installed and online.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Start web-ui.

### Logs

```shell
Loading the extension "gallery"... Ok.
Loading the extension "send_pictures"... Ok.
Loading the extension "sd_api_pictures"... Ok.
Open error: is JACK server running?
Loading the extension "elevenlabs_tts"... Fail.
Traceback (most recent call last):
  File "F:\oobabooga_windows\text-generation-webui\modules\extensions.py", line 33, in load_extensions
    exec(f"import extensions.{name}.script")
  File "<string>", line 1, in <module>
  File "F:\oobabooga_windows\text-generation-webui\extensions\elevenlabs_tts\script.py", line 6, in <module>
    from elevenlabslib.helpers import save_bytes_to_path
ImportError: cannot import name 'save_bytes_to_path' from 'elevenlabslib.helpers' (F:\oobabooga_windows\installer_files\env\lib\site-packages\elevenlabslib\helpers.py)
```

### System Info

```shell
13900KS
2x 4090
64GB DDR5 6000
```

gopfranco:
[...]

PR #1685, opened 2023-04-30
Proposal to fix loading errors with elevenlabslib extension

ezra-ch:
Several users have reported issues with loading the elevenlabslib extension while using the one-click installation process with the current settings. as seen from https://github.com/oobabooga/text-generation-webui/issues/1648

I propose a simple fix by changing the import statement to 'from elevenlabslib.helpers import *' to improve stability.


ezra-ch:
actually this just fixed the loading issue, there's still something wrong with the save_bytes_to_path function

PR #1716, opened 2023-05-01
fixed variable name "context" to "prompt"

Dhaladom:


CarlKenner:
I think the downvote might have misunderstood what you were doing.

You fixed a bug in the example code, which would have given an error before. Your fix looks good to me.

CarlKenner (approved):
Looks good to me.

Issue #1715, opened 2023-05-01
Unable to open any language models. 


DuckersMcQuack:
### Describe the bug

1. Aitrepreneur_wizardLM-7B-GPTQ-4bit-128g
2. facebook_opt-350m
3. TheBloke_stable-vicuna-13B-GPTQ

Which one do you want to load? 1-3

3

Loading TheBloke_stable-vicuna-13B-GPTQ...
Loading checkpoint shards:   0%|                                                                 | 0/3 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\transformers\modeling_utils.py", line 442, in load_state_dict
    return torch.load(checkpoint_file, map_location="cpu")
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\torch\serialization.py", line 791, in load    with _open_file_like(f, 'rb') as opened_file:
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\torch\serialization.py", line 271, in _open_file_like
    return _open_file(name_or_buffer, mode)
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\torch\serialization.py", line 252, in __init__
    super().__init__(open(name, mode))
FileNotFoundError: [Errno 2] No such file or directory: 'models\\TheBloke_stable-vicuna-13B-GPTQ\\pytorch_model-00001-of-00003.bin'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "G:\Stablechat\oobabooga_windows\text-generation-webui\server.py", line 914, in <module>
    shared.model, shared.tokenizer = load_model(shared.model_name)
  File "G:\Stablechat\oobabooga_windows\text-generation-webui\modules\models.py", line 217, in load_model
    model = LoaderClass.from_pretrained(checkpoint, **params)
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\transformers\models\auto\auto_factory.py", line 471, in from_pretrained
    return model_class.from_pretrained(
  File "G:\Stablechat\oobabooga_windows\installer_files\env\lib\site-packages\transformers\modeling_utils.py", line 2795, in from_pretrained
    ) = cls._load_pretrained_model(
[...]

PR #1737, opened 2023-05-02
Automatically detect wbits and groupsize

CarlKenner:
Addresses issue #1715 (in theory)


oobabooga:
I think that going case by case is better, being too general can cause unintended consequences in the future

CarlKenner:
> I think that going case by case is better, being too general can cause unintended consequences in the future

You're wrong. Going case by case will cause unintended consequences in the future. Consistently following the standard naming schemes that people use will cause it to work as expected in the future.

PR #1744, opened 2023-05-02
Extension elevenlabs: fix save_audio after library update (#1648)

subversion23:
fix for: #1648  
elevenlabslib.helpers library changed save_bytes_to_path to  save_audio_bytes.   
Also needs now Path as str and a fileformat.

Tested under linux, python 3.10.

### DIFF
--- a/extensions/elevenlabs_tts/script.py
+++ b/extensions/elevenlabs_tts/script.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 import gradio as gr
 from elevenlabslib import ElevenLabsUser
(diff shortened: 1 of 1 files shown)


ezra-ch:
nice, this fixed the problem for me.

Joemaku:
got it loading without errors on windows 11, API also loading but voice dropdown list shows None and i can't enter any messages. do i have to use the api extension in oobabooga or install the requirements for elevenlabs_tts even though it's built in? i wonder what im missing.. maybe use public share for it to work? thanks for everything.

PR #1753, opened 2023-05-03
Fix openai extension script.py - TypeError: '_Environ' object is not …

Thireus:
…callable

Fixes the following issue:

  File "text-generation-webui/modules/extensions.py", line 33, in load_extensions
    exec(f"import extensions.{name}.script")
  File "text-generation-webui/extensions/openai/script.py", line 11, in <module>
    'port': int(os.environ('OPENEDAI_PORT')) if 'OPENEDAI_PORT' in os.environ else 5001,
TypeError: '_Environ' object is not callable

### DIFF
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -8,7 +8,7 @@ from modules import shared
 from modules.text_generation import encode, generate_reply
 
 params = {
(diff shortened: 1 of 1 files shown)


matatonic:
Thanks, missed that.

matatonic (approved):

PR #1754, opened 2023-05-03
[openai] add edits & image endpoints & fix prompt return in non --chat modes

matatonic:
This was pretty simple and works well for translations, spelling, "professional emails", code reformatting, etc.

PR #1757, opened 2023-05-03
Openai Chat Completion support improvement

Thireus:
Fixes:
- Uses Alpaca prompt template
- Standard stopping strings reduced to Alpaca format
- Completion no longer returns the prompt but the answer only
- Role name with first letter uppercase

Compatible and tested with Vicuna v1.1

Tested using the following python script:
```

# Makes sure the requests don't go to your local proxy
import os
os.environ['NO_PROXY'] = '127.0.0.1'

import openai

# Port can be changed with OPENEDAI_PORT env variable
openai.api_base = "http://127.0.0.1:5001/v1"

# Enter any non-empty API key to pass the client library's check.
openai.api_key = "xxx"

# Enter any non-empty model name to pass the client library's check.
completion = openai.ChatCompletion.create(
    model="Thireus_Vicuna13B-v1.1-8bit-64g", messages = [{"role": "system", "content": "Below is an instruction that describes a task. As a helpful assistant, write a response that appropriately completes the request."},{"role": "Human", "content": "What is 1+1?"},{"role": "Assistant", "content": "The sum of 1 and 1 is 2."},{"role": "Human", "content": "Why is it 2?"}], max_tokens=256, echo=False
)

# Print the completion.
print(completion.choices[0].message)
```

Will return:
```
{
  "content": "Because when you add two numbers together, the result is called their \"sum.\" The sum of 1 and 1 is therefore 2.",
  "role": "Assistant"
}
```

### DIFF
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -205,10 +205,10 @@ class Handler(BaseHTTPRequestHandler):
                 chat_msgs = []
 
                 for m in messages:
-                    role = m['role']
+                    role = m['role'].capitalize()
                     content = m['content']
                     # name = m.get('name', 'user')
-                    if role == 'system':
+                    if role == 'System':
                         system_msg += content.strip()+'\n'
                     else:
                         chat_msgs.extend([f"### {role}:\n{content.strip()}\n"])
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -304,9 +304,15 @@ class Handler(BaseHTTPRequestHandler):
 
             for a in generator:
                 if isinstance(a, str):
-                    answer = a[len(prompt):]
+                    if shared.args.chat:
+                        answer = a
+                    else:
+                        answer = a[len(prompt):]
                 else:
-                    answer = a[0][len(prompt):]
+                    if shared.args.chat:
+                        answer = a[0]
+                    else:
+                        answer = a[0][len(prompt):]
 
                 stop_string_found = False
                 len_seen = len(seen_content)
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -8,7 +8,7 @@ from modules import shared
 from modules.text_generation import encode, generate_reply
 
 params = {
-    'port': int(os.environ('OPENEDAI_PORT')) if 'OPENEDAI_PORT' in os.environ else 5001,
+    'port': int(os.environ.get('OPENEDAI_PORT')) if 'OPENEDAI_PORT' in os.environ else 5001,
 }
 
 debug = True if 'OPENEDAI_DEBUG' in os.environ else False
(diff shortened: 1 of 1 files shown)


matatonic:
I'm a bit confused by this change, and perhaps it's a problem specific to a model. When I run your example without your code change, I get essentially the same output (using Alpaca-30B-4bit).

>>> completion = openai.ChatCompletion.create(
...     model="Alpaca-30B-4bit", messages = [{"role": "system", "content": "Below is an instruction that describes a task. As a helpful assistant, write a response that appropriately completes the request."},{"role": "Human", "content": "What is 1+1?"},{"role": "Assistant", "content": "The sum of 1 and 1 is 2."},{"role": "Human", "content": "Why is it 2?"}], max_tokens=256, echo=True
... )
>>> print(completion.choices[0].message)
{
  "content": "1 + 1 = 2 because when you add two numbers together, their total will always be equal to the sum of both individual numbers added together.",
  "role": "assistant"
}

How can I reproduce the problem you're having? Just use a vicuna model?

Also, from the openai API documentation, the "assistant" should be lowercase, see here: https://platform.openai.com/docs/api-reference/chat - why change that?

If it's just a model difference, then I guess it may require some shuffling of prompting formats based on the models to get good support across many models. If that's the road we go down, I think it's worth some discussion about how to support multiple models flexibly. I use a yaml config with various per-model settings in another project I'm working on, maybe something like that.

Thireus:
@matatonic, I was launching text-generation-webui without "--chat". When chat mode is not enabled, the answer was returning prompt+answer. I made the relevant changes in 9d8d32b to take that into account, now it works with and without chat mode launch.

OpenAI's indeed uses lowercase roles, but Alpaca was trained with "### Role:" format with the first letter being uppercase. I have made the relevant change in e4d7030 to capitalize the first letter so that you can now use openai's lowercase roles in your request, the script will transform it to uppercase to match the Alpaca format "### Role:".

The versatility of the role that is injected as such in the prompt is great, I like the fact the role can be changed to Human, Instruction, Input, etc. This is a great hack that improves Alpaca's format compatibility. More on the matter here: https://github.com/oobabooga/text-generation-webui/pull/1757/commits/e4d7030108ec6d520c348def5729ebda01e297a5

Thireus:
@matatonic, fantastic work btw! :)

matatonic:
Aha! --chat yes, I can see the problem now, I also found another bug I've been looking for - a mysterious extra space at the beginning of generated sequences. I think however I want to fix this differently and use lower level model functions to generate the text. I'm looking into this today and will update asap. Thanks!

matatonic:
@Thireus I've updated my edits pull request to include a fix for --chat mode only, can you try it out? It's at https://github.com/oobabooga/text-generation-webui/pull/1754,  It should work in any mode now.

As for the chat endpoint and role names, especially with other models, I think that requires a bit more work.

oobabooga:
Closing this for the same reason as https://github.com/oobabooga/text-generation-webui/pull/1754#issuecomment-1539299329

Could any of you create a PR solely with the bug fixes and without the endpoint that hard codes alpaca?

matatonic:
totally agree, I literally just finished this: https://github.com/oobabooga/text-generation-webui/pull/1935
@oobabooga @Thireus

PR #1772, opened 2023-05-03
Make --trust-remote-code actually use trust_remote_code=True

gitmylo:
## Summary
I was having issues with some models not loading because `trust_remote_code` was `False`. Specifically with [bigcode/santacoder](https://huggingface.co/bigcode/santacoder). Which will get an error saying that `trust_remote_code` is false. Even when using the `--trust-remote-code` command-line parameter.

## In this PR
`modules/models.py` has been modified to always use the `--trust-remote-code` command-line parameter's existence as the value for the `trust_remote_code` argument for `from_pretrained()` in the AutoConfig and the LoaderClass.

## What does this fix?
This fixes the loading of models that didn't load even with `--trust-remote-code` set.

### DIFF
--- a/modules/models.py
+++ b/modules/models.py
@@ -56,7 +56,7 @@ def find_model_type(model_name):
     elif any((k in model_name_lower for k in ['gpt4chan', 'gpt-4chan'])):
         return 'gpt4chan'
     else:
-        config = AutoConfig.from_pretrained(Path(f'{shared.args.model_dir}/{model_name}'))
+        config = AutoConfig.from_pretrained(Path(f'{shared.args.model_dir}/{model_name}'), trust_remote_code=shared.args.trust_remote_code)
         # Not a "catch all", but fairly accurate
         if config.to_dict().get("is_encoder_decoder", False):
             return 'HF_seq2seq'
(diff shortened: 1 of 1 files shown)


modules/models.py
```diff
@@ -56,7 +56,7 @@ def find_model_type(model_name):
     elif any((k in model_name_lower for k in ['gpt4chan', 'gpt-4chan'])):
         return 'gpt4chan'
     else:
-        config = AutoConfig.from_pretrained(Path(f'{shared.args.model_dir}/{model_name}'))
+        config = AutoConfig.from_pretrained(Path(f'{shared.args.model_dir}/{model_name}'), trust_remote_code=shared.args.trust_remote_code)
```
oobabooga, line 59: Can you test if this addition is necessary? All this line does is read config.json to determine if the model is a seq2seq or not.

kenfink:
Chiming in here to be able to load https://huggingface.co/replit/replit-code-v1-3b as well.  The line 80 deletion should hopefully remedy this?  Or does it need stay in but check args and set to true?

modules/models.py
Maykeye: Can confirm it  is necessary for  `mosaicml/mpt-1b-redpajama-200b`

oobabooga: Thanks, fair enough.

PR #1817, opened 2023-05-05
Refactor text_generation.py, add support for custom generation functions

oobabooga:
This PR implements two new extension functions:

* `custom_generate_reply`: executed in place of the main generation functions. You can use it to connect the web UI to an external API, or to load a custom model that is not supported yet.
* `state_modifier`: modifies the input parameters before they are used by the text generation functions. For instance, it can be used to disable streaming in STT extensions, or to randomize parameters like temperature.

Additionally, it:

* Turns `shared.args.no_stream` into a checkbox in the Parameters tab
* Refactors `modules/text-generation.py` to make it more modular

See the updated documentation for details and examples: https://github.com/oobabooga/text-generation-webui/blob/refactor-text-generation/docs/Extensions.md#how-to-write-an-extension


Wojtab:
please ping me when you merge it, so I can fix conflicts in #1741
(I added `custom_tokenized_length` to extensions API, and `text_generation.get_encoded_length(prompt)` to replace `len(encode(prompt)[0])`, so it takes into account the length after tokenizer extensions modify input ids)

oobabooga:
All done @Wojtab. This PR was possible thanks for the framework that you came up with :)

Issue #1825, opened 2023-05-06
Server.py Won't Run KeyError: 'serialized_input'


chigkim:
### Describe the bug

Update: For me, upgrading gradio with `pip install -U gradio` fixed the problem.

I'm running on Colab, and when I run server.py, it throws an error.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run server.py on Colab.

### Logs

```shell
INFO:Loaded the model in 130.66 seconds.
INFO:Loading the extension "silero_tts"...

Using Silero TTS cached checkpoint found at /root/.cache/torch/hub
INFO:Loading the extension "gallery"...
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /content/GPTQ-for-LLaMa/server.py:930 in <module>                            │
│                                                                              │
│   927 │   │   })                                                             │
│   928 │                                                                      │
│   929 │   # Launch the web UI                                                │
│ ❱ 930 │   create_interface()                                                 │
│   931 │   while True:                                                        │
│   932 │   │   time.sleep(0.5)                                                │
│   933 │   │   if shared.need_restart:                                        │
│                                                                              │
│ /content/GPTQ-for-LLaMa/server.py:517 in create_interface                    │
│                                                                              │
│   514 │   if shared.args.extensions is not None and len(shared.args.extensio │
│   515 │   │   extensions_module.load_extensions()                            │
│   516 │                                                                      │
│ ❱ 517 │   with gr.Blocks(css=ui.css if not shared.is_chat() else ui.css + ui │
│   518 │   │                                                                  │
│   519 │   │   # Create chat mode interface                                   │
│   520 │   │   if shared.is_chat():                                           │
│                                                                              │
│ /usr/local/lib/python3.10/dist-packages/gradio/blocks.py:1285 in __exit__    │
│                                                                              │
│   1282 │   │   │   Context.root_block = None                                 │
│   1283 │   │   else:                                                         │
│   1284 │   │   │   self.parent.children.extend(self.children)                │
│ ❱ 1285 │   │   self.config = self.get_config_file()                          │
│   1286 │   │   self.app = routes.App.create_app(self)                        │
│   1287 │   │   self.progress_tracking = any(block_fn.tracks_progress for blo │
│   1288 │   │   self.exited = True                                            │
│                                                                              │
│ /usr/local/lib/python3.10/dist-packages/gradio/blocks.py:1261 in             │
│ get_config_file                                                              │
[...]

PR #1827, opened 2023-05-06
Update gradio dependency, to fix KeyError issue

mateusz:
Fixes https://github.com/oobabooga/text-generation-webui/issues/1825 for me. I'm on Ubuntu 20, and installing onto a clean micromamba environment - reinstalled from scratch using this, and it seems to work.

More testing probably needed :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ accelerate==0.18.0
 colorama
 datasets
 flexgen==0.1.7
-gradio==3.25.0
+gradio==3.28.3
 markdown
 numpy
 pandas


mateusz:
Might be a bigger effort, Silero TTS and Whisper STT extensions don't work (although I'm not sure if gradio upgrade is really the culprit here).

XenioxYT:
Works for me, after bumping gradio version. Agreed more testing is needed, but seems to boot and text generation works.

oobabooga:
Last time I checked the character gallery was broken in 3.28

XenioxYT:
> Last time I checked the character gallery was broken in 3.28 (clicking on a character wouldn't make it load)

Can confirm gallery doesn't work. Have to manually load in the character tab

oobabooga:
Try this instead https://github.com/oobabooga/text-generation-webui/commit/81be7c2dd49708cf1c225ac01ea64f4e46e972ee

```
pip install gradio_client==0.1.4
```

I can't upgrade gradio because the character gallery is broken in 3.28

PR #1787, opened 2023-05-04
Elevenlabs Extension Improvement and migration to official API

TechnoTaff:
I have refactored the ElevenLabs API to use the official API and implemented the following benefits:

* No API key needed for limited testing
* Borrowed a lot of functionality from the excellent silero_tts extension
* Added in-line playback with flags to control automatic playback
* cope with ElevenLabs API exceeding quota
* Works out of the box now, as soon as you enable the extension

PR #1829, opened 2023-05-06
Elevenlabs Extension Improvement and migration to official API

TechnoTaff:
I have refactored the ElevenLabs API to use the official API and implemented the following benefits:

No API key needed for limited testing
Borrowed a lot of functionality from the excellent silero_tts extension
Added in-line playback with flags to control automatic playback
cope with ElevenLabs API exceeding quota
Works out of the box now, as soon as you enable the extension

Replay of PR[ #1787](https://github.com/oobabooga/text-generation-webui/pull/1787)


TechnoTaff:
image

Solves https://github.com/oobabooga/text-generation-webui/issues/1648

extensions/elevenlabs_tts/script.py
```diff
@@ -55,17 +60,20 @@ def remove_surrounded_chars(string):
     return re.sub('\*[^\*]*?(\*|$)', '', string)
 
 
-def state_modifier(state):
```
oobabooga, line 58: This is now necessary, and the lines that change shared.args.no_stream have to be removed

PR #1830, opened 2023-05-06
Elevenlabs Extension Improvement and migration to official API

TechnoTaff:
I have refactored the ElevenLabs API to use the official API and implemented the following benefits:

No API key needed for limited testing
Borrowed a lot of functionality from the excellent silero_tts extension
Added in-line playback with flags to control automatic playback
cope with ElevenLabs API exceeding quota
Works out of the box now, as soon as you enable the extension


oobabooga:
It works nicely, thank you!

bubbabug:
After updating to this version of ElevenLabs I  am receiving this error:

"AttributeError: module 'modules.shared' has no attribute 'reload_inputs'"

PR #1846, opened 2023-05-06
Fix character "�" being sent with the streaming API.

### LINKED ISSUE #1603: �/U+FFFD character sent when using the streaming api.
### Describe the bug

The streaming api sends "�" when the output contains some non-English characters. It can be seen too when using the Gradio UI but it quickly fixes itself when more characters come, but that doesn't happen when using the api.

### Reproduction

The character I use is named c.ⱯI but appears as c�� in the output.

### Screenshot

screenshot1

I use this as a workaround:
```diff
diff --git a/extensions/api/streaming_api.py b/extensions/api/streaming_api.py
index c06facc..10e6e6e 100644
--- a/extensions/api/streaming_api.py
+++ b/extensions/api/streaming_api.py
@@ -38,6 +38,9 @@ async def _handle_connection(websocket, path):
             else:
                 to_send = a[0][skip_index:]

+            if to_send.endswith("�"):
+                continue
+
             await websocket.send(json.dumps({
                 'event': 'text_stream',
                 'message_num': message_num,
```

screenshot2

### Logs

```shell
None.
```

### System Info

```shell
Linux.
```

ghost:
Closes #1603.
Not sure if this is the correct way to fix it since I'm not a Python programmer but it works for me.

### DIFF
--- a/extensions/api/streaming_api.py
+++ b/extensions/api/streaming_api.py
@@ -39,6 +39,9 @@ async def _handle_connection(websocket, path):
             else:
                 to_send = a[0][skip_index:]
 
+            if to_send.endswith("�"):
+                continue
+
             await websocket.send(json.dumps({
                 'event': 'text_stream',
                 'message_num': message_num,


LaaZa:
I don't think this is the best way to fix this if it is an issue. We should fix the source of this instead.

LaaZa:
That's what I thought it was. Maybe there could be a setting to combine tokens with the streaming. But I also think this could be fixed on the client side.

PR #1847, opened 2023-05-06
Use a random seed, instead of the fixed seed 0

tr7zw:
I recently started using this repo and downloaded https://huggingface.co/MetaIX/Alpaca-30B-Int4 as my model. Quickly noticed that the model always gives the same output for a given input. Turns out, llama models are always run with 0 as seed, instead of -1 for random.

### DIFF
--- a/modules/llamacpp_model.py
+++ b/modules/llamacpp_model.py
@@ -23,7 +23,7 @@ class LlamaCppModel:
         params = {
             'model_path': str(path),
             'n_ctx': 2048,
-            'seed': 0,
+            'seed': -1,
             'n_threads': shared.args.threads or None,
             'n_batch': shared.args.n_batch,
             'use_mmap': not shared.args.no_mmap,


LaaZa:
In llama-cpp-python [0 is random](https://github.com/abetlen/llama-cpp-python/blob/2f2ea00a3db0b1dd45af56bfa0bf9464b819982b/llama_cpp/llama.py#LL93C13-L93C45)

tr7zw:
Changed it for a draft for now, will check more later. Random seeds are certainly not working correctly(at least with this model). I thought it worked correctly yesterday after I changed it to -1 instead of 0, but right now when asking it "Write a funny joke." it always prints the exact same answer, no matter how often I prompt it/change the parameters.

LaaZa:
Might be your other params. Increase the temperature or something.

tr7zw:
Hm after reverting to the 0 and cranking up the temperature a bit, now it works. Guess at 0.7 the answers were just so repetitively the same, that I assumed it was broken. Guess don't code while fully exhausted.

Issue #1820, opened 2023-05-05
LoRAs are not loading correctly after the first load (demo LoRAs included)


ClayShoaf:
### Describe the bug

* When a single LoRA is loaded, it works as expected.
* If multiple LoRAs are loaded, it appears that only one is applied. If, after this, both LoRAs are removed from the dropdown and "Apply the selected LoRAs" is clicked, the LoRAs will be removed from `shared.lora_names`, but the first LoRA will still be applied.
* If a LoRA is loaded, then removed and another LoRA is loaded, it is not actually applied and the base model will generate as if no LoRA is applied. `shared.lora_names` will still show the LoRA that is supposed to be applied.

This appears to be the case with both 8-bit and 4-bit with `--monkey-patch`

I tried to figure this out on my own, but I'm not familiar enough with how peft loads these and I don't have enough time to learn it right now. If this is still open on Monday, I'll try to step through the program and see what I can figure out.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

I created some LoRAs that you can easily test this with to see what I'm talking about. You can find them here: https://huggingface.co/clayshoaf/AB-Lora-Test

Just try loading and unloading them and you'll see what I'm talking about.

### Logs

There are no error messages in the logs. Everything appears as if the LoRAs are loading correctly.

```shell



INFO:Gradio HTTP request redirected to localhost :)
bin /home/clay/anaconda3/envs/textgen/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117.so
INFO:Loading settings from settings.json...
INFO:Loading the extension "xy_grid"...
INFO:Loading the extension "gallery"...
Running on local URL:  http://0.0.0.0:7860
To create a public link, set `share=True` in `launch()`.
INFO:Loading LLaMA-7B-4bit-128g...
WARNING:Applying the monkey patch for using LoRAs in 4-bit mode. It may cause undefined behavior outside its intended scope.
Loading Model ...
WARNING:The safetensors archive passed at models/LLaMA-7B-4bit-128g/llama-7b-4bit-128g.safetensors does not contain metadata. Make sure to save your model with the `save_pretrained` method. Defaulting to 'pt' metadata.
Loaded the model in 4.03 seconds.
INFO:Loaded the model in 4.16 seconds.
INFO:Applying the following LoRAs to LLaMA-7B-4bit-128g: AAAAAAA
[...]

PR #1865, opened 2023-05-07
fixed LoRA loading issue

ClayShoaf:
This resolves https://github.com/oobabooga/text-generation-webui/issues/1820

### DIFF
--- a/modules/LoRA.py
+++ b/modules/LoRA.py
@@ -28,6 +28,7 @@ def add_lora_to_model(lora_names):
     # If any LoRA needs to be removed, start over
     if len(removed_set) > 0:
         shared.model.disable_adapter()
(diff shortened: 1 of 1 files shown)


ClayShoaf:
Nice now I can work on putting LoRAs in the xy-grid

mcmonkey4eva:
Good fix! That had been bothering me too, but i was getting inconsistent results between 4bit/4bitmonkeypatch/8bit and was too confused to try to resolve it. I was wondering if the solution might be to give up and just trigger a full model reload, but your solution here of digging the original model back out is very clever!

ClayShoaf:
Thanks. It was a pain in butt trying to figure out what was going on. Thankfully, pycharm is very helpful.

PR #1868, opened 2023-05-07
Fix character loading

matthew-mcallister:
Pretty self-explanatory diff


matthew-mcallister:
Why is this closed? Could you at least leave a comment? Is this already fixed in another branch?

matthew-mcallister:
Ah, I see the issue. While this code does look a little buggy, it is not. The actual bug I encountered was that loading a character when the chat mode is 'instruct' causes character loading to fail.

PR #1871, opened 2023-05-07
Add error message when GPTQ-for-LLaMa import fails

matthew-mcallister:
I'm of the opinion that users should *not* have to Google random error messages for 30 minutes when a program crashes, so I decided to do something constructive and add helpful instructions for when loading a GPTQ model fails.

EDIT: FWIW, a super helpful improvement would be to check in GPTQ-for-LLaMa as a git submodule; that way `git submodule update --init` would automatically download the external dependency without having to follow an extra documentation page in the first place.


oobabooga:
> without having to follow an extra documentation page in the first place.

The user would still have to compile the cuda kernel, which requires a c++ compiler and nvcc to be installed. gptq is not trivial to distribute when an official pypi package and official wheels are not available

PR #1873, opened 2023-05-07
Fix character menu when default chat mode is 'instruct'

matthew-mcallister:
When changing the chat mode to 'instruct' via the UI, this menu is properly marked non-interactive. However, if the settings file or default model cause the mode to start as 'instruct', the menu was not initially marked non-interactive, leading to a confusing error.


Malrama:
Great fix! I got a lot of confusion because of it. Thank you :D

PR #1876, opened 2023-05-07
[extensions/openai] Support undocumented base64 'encoding_format' param for compatibility with official OpenAI client

jeffreylin:
The official Python OpenAI API client uses an undocumented `encoding_format` parameter on the `/v1/embeddings` endpoint by default.

https://github.com/openai/openai-python/blob/d6fa3bfaae69d639b0dd2e9251b375d7070bbef1/openai/api_resources/embedding.py#L25-L29

This PR adds support for the `encoding_format` parameter to maintain support as a drop-in replacement for the OpenAI API. Without this parameter, [llama_index](https://github.com/jerryjliu/llama_index), and possibly other langchain apps, may fail to work.

### DIFF
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -1,4 +1,6 @@
+import base64
 import json
+import numpy as np
 import os
 import time
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
@@ -44,6 +46,20 @@ def default(dic, key, default):
 def clamp(value, minvalue, maxvalue):
     return max(minvalue, min(value, maxvalue))
 
+
+def float_list_to_base64(float_list):
+    # Convert the list to a float32 array that the OpenAPI client expects
+    float_array = np.array(float_list, dtype="float32")
+
+    # Get raw bytes
+    bytes_array = float_array.tobytes()
+
+    # Encode bytes into base64
+    encoded_bytes = base64.b64encode(bytes_array)
+
+    # Turn raw base64 encoded bytes into ASCII
+    ascii_string = encoded_bytes.decode('ascii')
+    return ascii_string
 
 class Handler(BaseHTTPRequestHandler):
     def do_GET(self):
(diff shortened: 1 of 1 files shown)


matatonic (approved):
Cool, nice find, thanks!

matatonic:
@jeffreylin 
Because the parameter is undocumented and python-openai isn't the only client, can you update the fetching of the parameter from body['encoding_format'] to body.get('encoding_format', '')  (or some other actual default). The node-js client may be different and also people using curl or swagger or other api clients should be able to use the API as documented.

jeffreylin:
@matatonic - Oh nice catch, thanks!! JS doesn't throw there, so I totally forgot and I'm glad you caught it.

Added a fix to address your comment, and verified new code using `curl` to make sure we get a string in the base64 case, and a list of floats in the default case:

{"encoding_format": "base64"}
```
$ curl -H "Authorization: Bearer foobar" -H 'Content-Type: application/json' -X POST http://localhost:5001/v1/embeddings -d '{"input": "Your text goes here", "model": "text-embedding-ada-002", "encoding_format": "base64"}'

{"object": "list", "data": [{"object": "embedding", "embedding": "BNthuwUy ... QoXb2YvWO8", "index": 0}], "model": "all-mpnet-base-v2", "usage": {"prompt_tokens": 0, "total_tokens": 0}}
```

(no encoding_format param)
```
$ curl -H "Authorization: Bearer foobar" -H 'Content-Type: application/json' -X POST http://localhost:5001/v1/embeddings -d '{"input": "Your text goes here", "model": "text-embedding-ada-002"}'

{"object": "list", "data": [{"object": "embedding", "embedding": [-0.003446281887590885, -0.058153171092271805, 0.002769032958894968, ... 0.04360939562320709, -0.053994014859199524, -0.013900183141231537], "index": 0}], "model": "all-mpnet-base-v2", "usage": {"prompt_tokens": 0, "total_tokens": 0}}
```

jeffreylin:
Shipping this now~

jeffreylin:
Ooops, wrong button - anyways, this should be ready for merge now - LMK if there's anything else!

matatonic (approved):

oobabooga:
Thank you!

PR #1887, opened 2023-05-07
Add config for pygmalion-7b and metharme-7b

matthew-mcallister:


CarlKenner:
Metharme is already handled (mostly?) in Pull Request #1679. I'm not sure where/if model_type is used outside the load_quantized function. Unfortunately, oobabooga takes his sweet time pulling pull requests, so we end up with multiple pull requests for the same thing, and pull requests that are out of date (like that one now is), and code that needs to be based on code in other pull requests.

oobabooga:
@CarlKenner don't be rude. You are not entitled to anything.

PR #1899, opened 2023-05-08
More trust_remote_code

EgrorBs:
I'm not entirely sure if trust_remote_code should be used here, but i got error here when loading MPT, and it can't break anything anyway. :)


oobabooga:
Thanks! I missed some 8 of those.

PR #1910, opened 2023-05-08
Allow extensions to wrap `generate_reply`

Vinno97:
**What**

I added a `wrap_method` decorator to `modules/extensions.py`, which can be used to easily mark methods to be "wrappable" by extensions.

For now, I only marked `generate_reply`, which allows extensions to modify either the input or the full output of this method. However, in our own code, we also wrap `extensions.api.util.build_parameters` to add some extra data to requests.

**Why**

I'm using this app to host models within my team and we currently monkey-patch utilities into it, which obviously breaks with each update. The current extension methods don't fulfill our needs and I I also can't really upstream our changes, because they don't really belong in the core of this server.

If this is merged, I could also possibly share the extension we use to log all prompts+results into MLFlow

---

I'm not sure this PR fits the "philosopy" of how this app currently works, but it works quite well for us, to create simple extensions.

### DIFF
--- a/modules/extensions.py
+++ b/modules/extensions.py
@@ -1,6 +1,7 @@
 import logging
+import inspect
 import traceback
-from functools import partial
+from functools import partial, wraps
 
 import gradio as gr
 
@@ -93,7 +94,7 @@ def _apply_state_modifier_extensions(state):
             state = getattr(extension, "state_modifier")(state)
 
     return state
- 
+
 
 # Extension functions that override the default tokenizer output
 def _apply_tokenizer_extensions(function_name, state, prompt, input_ids, input_embeds):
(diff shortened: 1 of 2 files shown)


Vinno97:
Well firstly I must admit that I wrote this before `custom_generate_reply` existed, but I don't think they serve the same purpose. 
This implementation does not replace, but decorate the original function. This allows it to enhance the original behaviour, or trigger some other function like logging. 

It also is more general. For example, I added a `@wrap_method` to  `extensions.api.utils.build_parameters`, which now allows me to ensure some sort of authentication (used to log the user into MLFlow). The general "wrappability" makes it possible to easily patch into various parts of the server, without having to rewrite parts of it.

However, I do now think that this might not be the best implementation for that purpose, since currently every "wrappable" function needs to be decorated first. This means that you still have to modify the source file of the function that you want to wrap (unless it already has the decorator applied). 

If you see this PR have value, I think I'd want to rewrite it to work something like this:

```py
@wrap_method("modules.text_generation.generate_reply") # or "extend_method?"
def foo(func, args, kwargs):
    print("Starting generation!")
    return func(*args, **kwargs)
```

oobabooga:
I do see a huge value in this PR. I want every function in the web UI to be modifiable by an extension, and this seems to create a framework for that that is more elegant than defining several `custom_...` in `modules/extensions.py`.

For instance, it would be nice to be able to hijack `modules.html_generator.chat_html_wrapper` to render entirely different chat interfaces, or `modules.models.load_model` to add support for new models before they are officially supported.

I look forward to merging your implementation once you think it's ready.

Vinno97:
Perfect! Swamped at work right now, but will get back to it during the weekend!

oobabooga:
Closing due to inactivity. If you ever get back to this please feel free to @ me

PR #1914, opened 2023-05-08
Fix - .DS_Store files should be ignored in the models and loras lists

Manoz:
For some reason on macOS, `.DS_Store` files are shown in the models and loras lists.  
The reason is the wildcard in the `.glob('*')` method.

ds-models ds-loras

This PR add conditions to ignore them. There's probably a better way to to that globally?
Code tested on my Mac and it works.

### DIFF
--- a/modules/utils.py
+++ b/modules/utils.py
@@ -15,9 +15,9 @@ def natural_keys(text):
 
 def get_available_models():
     if shared.args.flexgen:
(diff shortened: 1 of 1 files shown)


oobabooga:
A solution through a function that could be applies to all functions in `modules/utils.py` would be better

Manoz:
Yeah that's what I thought. We can even probably handle that using the `Path().glob()`. I'll try to submit a better PR to do that :)

PR #1935, opened 2023-05-09
[extension/openai] add edits & image endpoints & fix prompt return in non --chat modes 

matatonic:
This includes correct instruction templates for each model (not just Alpaca). For now they're in the code, but I will change them to use the on disk format when they're fully compatible. 

I found some small differences with different instruction formats, but I don't want to include those changes here. I will update those in another PR.


matatonic:
Hold off merging this for now, I have an improved instruction prompt selection coming today based on the instruction-following character templates.

matatonic:
@oobabooga 
Ok, this works more flexibly (not just alpaca) and will work as new instruction following templates are added for new models. 
I also added wizard vicuna to models/config.yaml, it's a Vicuna v1.1 without the 1.1 in the name.
While looking into prompt formats I also found that ChatGLM uses a non-ascii ":" character (I found that from the training code), so I update the instruction template for that also.

oobabooga:
Thanks for the updates, @matatonic. It's impressive that you noticed that ChatGLM detail.

PR #1936, opened 2023-05-09
Add in-memory cache support for llama.cpp

abetlen:
This change adds the new in-memory cache for llama-cpp-python.

To make use of the cache I had to change the LlamaCppModel to use `create_completion` API which takes advantage of the cache (generate can only rewind the current text generation session). They should behave identically here as `create_completion` already checks for the end of sequence token and terminates when `max_token` tokens are gneerated.

One open question I have is regarding the cache, it currently has a `cache_capacity` option which is the in-memory cache size in bytes, the defaultt is 2Gb, this is probably something users will want to tweak though, would it make sense to add this to your shared args file?


oobabooga:
Thanks a lot for the PR!

>  would it make sense to add this to your shared args file?

Yes it would. You could make it similar to how `--gpu-memory` is used. The accepted formats are "2GiB" or "2140MiB", and the string is parsed like this:
 
```python
            if 'mib' in i.lower():
                default_gpu_mem.append(int(re.sub('[a-zA-Z ]', '', i)))
            else:
                default_gpu_mem.append(int(re.sub('[a-zA-Z ]', '', i)) * 1000)
```

By the way, if there are other performance-enhancing parameters that you think are worth adding, or defaults that you think are bad, please let me know.

abetlen:
@oobabooga thanks, I'll add the option like that.

The biggest recent performance improvement has been the OpenBLAS / cuBLAS / clBlast support added to llama.cpp, those can be enabled by setting the environment variables when installing from source ([see here](https://github.com/abetlen/llama-cpp-python/tree/main#installation-with-openblas--cublas--clblast))

oobabooga:
@abetlen I have added a `--cache-capacity` argument to this PR that can be used like this:

```
--cache-capacity 2142MiB
--cache-capacity 2GiB
--cache-capacity 2  # Assumes GiB by default
```

I couldn't find any info on how to use it in the documentation (https://abetlen.github.io/llama-cpp-python/). Is it a new feature?

abetlen:
@oobabooga merged in the changes for the gpu offloading and tested this out, works well for me. Made a small change to the cache capacity parsing to default to bytes but feel free to change that.

oobabooga:
Looks good to me as it is now. Thanks for the updates to the `generate` function, by the way!

PR #1964, opened 2023-05-10
Add custom prompt format for SD API pictures

Minecrafter20:
Added additional textbox to modify the prefix fed into the text generator when generating image prompts.

This way it is possible to customize how the model makes the prompt, e.g. some models respond best to certain key words, this allows those to be added.

The current format is to write the prefix however you want, placing [subject] wherever you want the subject to appear in the prompt. e.g. if "a bird" is detected as the subject, all instances of [subject] will be replaced with "a bird" when fed into the prompt generator. This allows a much greater degree of freedom in regard to image generation.


Brawlence:
Probably okay-ish but the changes are messy. Gonna review later.

Minecrafter20:
Do you mind letting me know which parts should be changed when you have time?

Brawlence:
@Minecrafter20 check out the 'files changed' tab, should be visible there

PR #1965, opened 2023-05-10
Fix `custom_generate_chat_prompt`

dibrale:
Fix to allow `custom_generate_chat_prompt` to be loaded from extensions again rather than throwing a TypeError.


GiusTex:
Which extension are you using?

TruthSearchers:
I am talking about extensions.py in module file
Just changed to previous code instead of latest one
Here this one 
9754d6a8119a4c48226d40f69fe83d9f03079016

GiusTex:
Thanks 👍🏻

oobabooga:
Thanks

PR #1985, opened 2023-05-10
Retry model download if it fails

Vuizur:
It might that I am the only person with pretty bad internet, but the downloads from huggingface sometimes keep hanging forever for me. This pull request adds a timeout limit of 20 seconds and then retries the download if the download times out, continuing from the last point the download left off. 
grafik

(I am on Windows, so I didn't test it on other systems.)

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -164,7 +164,7 @@ def get_output_folder(model, branch, is_lora, base_folder=None):
     return output_folder
 
 
-def get_single_file(url, output_folder, start_from_scratch=False):
+def get_single_file(url, output_folder, start_from_scratch=False, attempt = 0):
     filename = Path(url.rsplit('/', 1)[1])
     output_path = output_folder / filename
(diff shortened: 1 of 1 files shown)


oobabooga:
You can just re-run the download script again. It will restart from where it stopped, including when it's executed from within the ui

Vuizur:
When I tested it with the batch file for Windows (without having downloaded any model), I got the behaviour that it prompts you to download a model, you give it the letter to select the model you want + you then select it and it starts to download. But then it hangs forever and you need to close the window. If you then open the batch file again it tries to load the unfinished model in the webui and crashes. So the only way to get it going again is deleting the entire model folder. This was what initially 
motivated me to open the pull request.

I also would recommend to add the timeout parameter regardless of this PR, because crashing is better than hanging indefinitely and [the requests docs recommend it.](https://docs.python-requests.org/en/latest/user/quickstart/#timeouts)

PR #1988, opened 2023-05-10
 Fixes the gradio bug when "--share" is used

gavin660:
The change checks for presence of  the --share flag. If --share IS present it will import Gradio using the normal requests.get. If --share is NOT present  it will import with my_get like before.


oobabooga:
--share works fine for me

PR #1990, opened 2023-05-11
Add api port options

atriantafy:
I found it useful to be able to set a custom port for the API (instead of the default 5000 and 5005 for the blocking and streaming APIs respectively).
I simply added two argument parser options (one for blocking API port and one for streaming API port) and used this argument in the right place (script.py)


oobabooga:
Good idea

PR #1993, opened 2023-05-11
Add api endpoint to initiate stop

briancw:
Simple change to make it possible to generation via an API route.
I made it a get method rather than a post so that you don't have send a payload or change the existing parsing on post requests.


oobabooga:
i am closing this PR in favor of https://github.com/oobabooga/text-generation-webui/pull/2392

PR #1997, opened 2023-05-11
Add ability to load all text files from a subdirectory for training

kizinfo:
minor tweak to allow a subdirectory to be selected in the UI with multiple files to be loaded for raw txt training


USBhost:
This sounds useful. So I could mass convert epubs to text then just dump all the txt of them in a folder and just train off that.

kizinfo:
yes, but it seems like the training code needs to be modified to handle larger datasets more easily, so it's still limited.

oobabooga:
I think that it would be better to do this using pathlib

PR #2003, opened 2023-05-11
Refactor chat functions

oobabooga:


matatonic:
@oobabooga I've really got to figure out how to generate tokens without relying on chat mode... but I think it's a trivial fix, testing now.

matatonic:
Fixed in https://github.com/oobabooga/text-generation-webui/pull/2008

PR #2008, opened 2023-05-11
[extensions/openai] is_chat changes fix

matatonic:
@oobabooga fix for is_chat changes here.


matatonic:
Re: https://github.com/oobabooga/text-generation-webui/pull/2003

oobabooga:
Thanks! A caveat is that `is_chat=True` causes `output_modifier` extensions to not be applied, but that can be ignored for now.

PR #2009, opened 2023-05-11
revert inserting extra space in get_reply_from_output_ids(). Fixes #1791

### LINKED ISSUE #1791: Extra space at the beginning of generation
### Describe the bug

An extra space is generated at the beginning of text,

It is done here: https://github.com/oobabooga/text-generation-webui/commit/15940e762e9f9a257fb8ce4f711b5e1ca7740616

This breaks a lot of use cases, like python code formatting and sequence completions.

Why is this done?

### Reproduction

In default mode, with max_tokens set to 2 try the following input (the next number should be 13):

0,1,1,2,3,5,8,

Instead it produces

0,1,1,2,3,5,8, 1

It should be:

0,1,1,2,3,5,8,13

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
Ubuntu
```

matatonic:
I don't really understand why this is here, but it breaks a variety of formatted text outputs like python code, continuations, etc. So remove it.

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -109,11 +109,6 @@ def get_reply_from_output_ids(output_ids, input_ids, original_question, state, i
     else:
         new_tokens = len(output_ids) - len(input_ids[0])
         reply = decode(output_ids[-new_tokens:], state['skip_special_tokens'])
-
-        if type(shared.tokenizer) is transformers.LlamaTokenizer:
-            if len(original_question) > 0 and original_question[-1] not in [' ', '\n']:
-                reply = ' ' + reply
-
         if not is_chat:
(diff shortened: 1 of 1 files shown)


oobabooga:
Without this, `My name is` results in `My name isJill and I am a 30-something year old woman.` (no space added) in regular/notebook modes. It's an inadequacy in the llama tokenizer. There could be a better solution than the current one, but removing the code causes another bigger problem.

matatonic:
hrm, I see... so llama can't start with a space? tricky. broke either way... ok.

oobabooga:
Yep. I agree that this is ugly and am trying to figure out an alternative solution.

PR #2012, opened 2023-05-12
Vicuna instruction templates space removal, huge quality improvement

matatonic:
@oobabooga 
I'm seeing massive quality improvements after removing the space after the <|bot|> entry in the 3 vicuna instruction templates - it's now Koala like in format and performance. The bot always leads with the space itself.

Additionally, in the openai/edits I remove the extra leading space when using templates with no leading space.

With this change I'm getting 'perfect' output now from Vicuna's using the openai.Edit() interface. As well as correct, no extra space, outputs from the normal completion interfaces - since your changes.

Working great!

For reference, compare the Koala and Vicuna templates here:
https://github.com/lm-sys/FastChat/blob/main/fastchat/conversation.py

I tested against:
TheBloke_wizard-vicuna-13B-GPTQ
TheBloke_stable-vicuna-13B-GPTQ
TheBloke_vicuna-13B-1.1-GPTQ-4bit-128g

### DIFF
--- a/characters/instruction-following/StableVicuna.yaml
+++ b/characters/instruction-following/StableVicuna.yaml
@@ -1,4 +1,4 @@
 user: "### Human:"
 bot: "### Assistant:"
-turn_template: "<|user|> <|user-message|>\n<|bot|> <|bot-message|>\n\n"
+turn_template: "<|user|> <|user-message|>\n<|bot|><|bot-message|>\n\n"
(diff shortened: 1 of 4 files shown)


oobabooga:
This breaks the formatting of old messages in the prompt, since they are stored in stripped form in shared.history

The proper way to do it is to `.rstrip(' ')` the prompt before it is sent to the model, as is done in `modules.chat.generate_chat_prompt`

PR #2034, opened 2023-05-13
Fix the API example for missing stream key error

mzbac:
Missing stream value in request parameters will cause an exception in text_generation.py.

```bash
text-generation-webui/modules/text_generation.py", line 245, in generate_reply_HF
    if not state['stream']:
KeyError: 'stream'
```


oobabooga:
This is already set both in blocking_api.py and streaming_api.py

PR #2042, opened 2023-05-13
[extensions/openai] clip extra leading space

matatonic:
with rstrip in the template expansion now, we can safely clip an extra space if it's returned for some models.


13631368:
Thanks for your help and support ❤️

PR #2051, opened 2023-05-14
Add superbooga chunk separator option

toast22a:
This PR adds a custom chunk separator option for manually specifying how data should be split (instead of chunk length windows). If a separator is specified, the chunk length will be ignored. For example, with the following parameters:

Text input:
```
You are wearing a red shirt.
You are wearing blue pants.
You are wearing white shoes.
```

Chunk separator: `\n`

The data chunks would look like this:
`["You are wearing a red shirt.", "You are wearing blue pants.", "You are wearing white shoes."]`

### DIFF
--- a/extensions/superbooga/script.py
+++ b/extensions/superbooga/script.py
@@ -13,6 +13,7 @@ from .download_urls import download_urls
 params = {
     'chunk_count': 5,
     'chunk_length': 700,
+    'chunk_separator': '',
     'strong_cleanup': False,
     'threads': 4,
 }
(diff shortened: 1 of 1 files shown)


oobabooga:
I think that this should be used in conjunction with the chunk length. If a given paragraph is too long (say, 3000 tokens long), that will lead to an OOM error if it gets matched and injected into a prompt

toast22a:
Got it -- I pushed commit c631d9fe7e0e812b3ddb5bcc4ee965b26ab1dfe2 with new behavior. If a separator-split chunk is longer than chunk length, it is further split into sub-chunks of length <= chunk length, and then the whole list is flattened. For example, with the following parameters:

Text input:
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

Chunk length: 100
Chunk separator: `\n`

The data chunks would look like this:
|length|text|
|-|-|
|100|"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore "|
|100|"et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut a"|
|31|"liquip ex ea commodo consequat."|
|100|"Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatu"|
|100|"r. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id"|
|13|" est laborum."|

In this case, the user can probably just specify an arbitrarily long chunk length if they don't care about chunk length.

oobabooga:
That's perfect now, thank you. This is very useful for formatted web pages.

Issue #2046, opened 2023-05-14
is this really a GGML file? llama_init_from_file: failed to load model


2tYfX8sgnrx:
Error

KingofMidnight:
i got the same error on my windows vm

llama.cpp: loading model from models\TheBloke_Wizard-Vicuna-13B-Uncensored-GGML\Wizard-Vicuna-13B-Uncensored.ggml.q4_0.bin
error loading model: unknown (magic, version) combination: 67676a74, 00000002; is this really a GGML file?
llama_init_from_file: failed to load model
Traceback (most recent call last):
  File "D:\oobabooga_windows\text-generation-webui\server.py", line 948, in <module>
    shared.model, shared.tokenizer = load_model(shared.model_name)
  File "D:\oobabooga_windows\text-generation-webui\modules\models.py", line 142, in load_model
    model, tokenizer = LlamaCppModel.from_pretrained(model_file)
  File "D:\oobabooga_windows\text-generation-webui\modules\llamacpp_model.py", line 32, in from_pretrained
    self.model = Llama(**params)
  File "D:\oobabooga_windows\installer_files\env\lib\site-packages\llama_cpp\llama.py", line 159, in __init__
    assert self.ctx is not None
AssertionError

m-spangenberg:
Not a bug. Your model isn't compatible with the version of llama.cpp being used by the web-ui.

See: https://github.com/oobabooga/text-generation-webui/issues/2020#issuecomment-1546656696

asif-ansari:
You can try updating llama.cpp
1. goto conda environment by running the `cmd_windows.bat` (assuming you're on windows because, the path in the error looks like windows one)
2. uninstall existing llama.cpp `pip uninstall -y llama-cpp-python`
3. re-install the latest llama.cpp `pip install llama-cpp-python --no-cache-dir`

You can optionally enable llama.cpp cmake flags like LLAMA_NATIVE, LLAMA_LTO or thirdparty libraries like CUBLAS

`CMAKE_ARGS="-DLLAMA_CUBLAS=ON -DCUDAToolkit_ROOT=/opt/cuda -DCMAKE_CUDA_ARCHITECTURES=86 -DCMAKE_CUDA_COMPILER=/opt/cuda/bin/nvcc -DLLAMA_NATIVE=ON -DLLAMA_LTO=ON" FORCE_CMAKE=1 pip install llama-cpp-python --no-cache-dir`

amaze28:
[...]

Issue #2020, opened 2023-05-12
llama.cpp: loading model from models\Chinese-LLaMA-7B\ggml-model-q4_0.bin error loading model: unknown (magic, version) combination: 67676a74, 00000002; is this really a GGML file? llama_init_from_file: failed to load model


wzgrx:
### Describe the bug

llama.cpp: loading model from models\Chinese-LLaMA-7B\ggml-model-q4_0.bin
error loading model: unknown (magic, version) combination: 67676a74, 00000002; is this really a GGML file?
llama_init_from_file: failed to load model

Traceback (most recent call last):
File “G:\Soft\text-generation-webui\[server.py](http://server.py/)”, line 67, in load_model_wrapper
shared.model, shared.tokenizer = load_model(shared.model_name)
File “G:\Soft\text-generation-webui\modules\[models.py](http://models.py/)”, line 142, in load_model
model, tokenizer = LlamaCppModel.from_pretrained(model_file)
File “G:\Soft\text-generation-webui\modules\llamacpp_model.py”, line 32, in from_pretrained
self.model = Llama(**params)
File “G:\Soft\text-generation-webui\python310\lib\site-packages\llama_cpp\[llama.py](http://llama.py/)”, line 148, in init
assert self.ctx is not None
AssertionError

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Cannot be used after updating the version

### Logs

```shell
llama.cpp: loading model from models\Chinese-LLaMA-7B\ggml-model-q4_0.bin
error loading model: unknown (magic, version) combination: 67676a74, 00000002; is this really a GGML file?
llama_init_from_file: failed to load model

Traceback (most recent call last):
File “G:\Soft\text-generation-webui\server.py”, line 67, in load_model_wrapper
shared.model, shared.tokenizer = load_model(shared.model_name)
File “G:\Soft\text-generation-webui\modules\models.py”, line 142, in load_model
model, tokenizer = LlamaCppModel.from_pretrained(model_file)
File “G:\Soft\text-generation-webui\modules\llamacpp_model.py”, line 32, in from_pretrained
self.model = Llama(**params)
File “G:\Soft\text-generation-webui\python310\lib\site-packages\llama_cpp\llama.py”, line 148, in init
assert self.ctx is not None
AssertionError
```

### System Info

```shell
win11
```

NightFuryPrime:
i got the same error today, i installed on fresh computer and got the same error message

[...]

Issue #2000, opened 2023-05-11
Says, "Is Typing..." But doesn't and resets


MikhaelLoo:
### Describe the bug

It's trying to chat with me, but can't get out a single word and then clears the screen and starts over. The first time it tries sometimes takes a while, but subsequent attempts are really fast.  (as if generating input is actually the Clear History button)

My machine is old, but I was hoping I could get by with slow performance.  I'm not sure what the reset behavior means.

I've tried 3 or 4 models with various settings but this behavior is consistent.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Install.  Set Flags: 
`run_cmd("python server.py --chat --auto-devices --gpu-memory 3300MiB", environment=True)  # put your flags here!`

Run with various Models.

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/18599445/ee5484b6-2f4a-49ea-bd94-eb3abec082e7)

### Logs

```shell
INFO:Gradio HTTP request redirected to localhost :)
bin D:\AI2\oobabooga_windows\installer_files\env\lib\site-packages\bitsandbytes\libbitsandbytes_cuda117_nocublaslt.dll
INFO:Loading the extension "gallery"...
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
ERROR:No model is loaded! Select one in the Model tab.
ERROR:No model is loaded! Select one in the Model tab.
INFO:Loading 4bit_WizardLM-7B-uncensored-GPTQ...
INFO:Found the following quantized model: models\4bit_WizardLM-7B-uncensored-GPTQ\WizardLM-7B-uncensored-GPTQ-4bit-128g.compat.no-act-order.safetensors
INFO:Using the following device map for the quantized model:
INFO:Loaded the model in 86.58 seconds.

Traceback (most recent call last):
  File "D:\AI2\oobabooga_windows\text-generation-webui\modules\callbacks.py", line 73, in gentask
    ret = self.mfunc(callback=_callback, **self.kwargs)
  File "D:\AI2\oobabooga_windows\text-generation-webui\modules\text_generation.py", line 259, in generate_with_callback
    shared.model.generate(**kwargs)
  File "D:\AI2\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
[...]

PR #2058, opened 2023-05-14
Update llama-cpp-python from 0.1.45 to 0.1.50

feeelX:
The ggml format was updated again so newly created models might require a newer llama.cpp version.

This should fix the following open issues:

- https://github.com/oobabooga/text-generation-webui/issues/2020
- https://github.com/oobabooga/text-generation-webui/issues/2046
- https://github.com/oobabooga/text-generation-webui/issues/2000

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,5 +17,5 @@ tqdm
 git+https://github.com/huggingface/peft
 transformers==4.28.1
 bitsandbytes==0.38.1; platform_system != "Windows"
-llama-cpp-python==0.1.45; platform_system != "Windows"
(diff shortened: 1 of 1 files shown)


eiery (changes requested):
You have a double h in "hhttps://github..."

feeelX:
> You have a double h in "hhttps://github..." (line 21), please fix.
> 
> But honestly though we should get this fix in sooner rather than later as llama.cpp upstream as well as many of the GGML HF repos have thoroughly embraced this change. I've been using the UI with v0.1.50 and it's working fine for me.

My bad. Fixed it now.

oobabooga:
I had to reconvert my llama 7b weights to be able to use this new version.

PR #2080, opened 2023-05-15
Add superbooga time weighted history retrieval

toast22a:
# Problem
During a long chat session, information about the characters and their environment may be described or change more than once. For example:
```
Assistant: I am wearing a red shirt.
...a few messages later...
Assistant: I am wearing a blue shirt.
```
If the user were to later query `What color is your shirt?`, the first message `...red shirt`, which is now outdated, may be retrieved as context instead of the second `...blue shirt`.

# Solution
This PR adds an option to apply time weighting to distance values. This might help break ties between similar history input/reply pairs by prioritizing more recent ones.

The time weight is linear within [`min_time_weight`, 1] where more recent history pairs are given smaller time weight values (to decrease the distance). Time weighting is disabled when `min_time_weight` = 1.

# Procedure
1. Retrieve top `n_initial` (where `n_initial` >= `n_results`) most similar documents from vector store.
2. Apply time weights to these documents' distances.
3. Sort the documents in ascending order by time-weighted distance.
4. Truncate the list of documents. Keep only the first `n_results` documents.
5. Sort the list again in ascending order by id.
6. Return the list of ids.

# Limitations
Suggestions for these, if any, are appreciated.
1. There might be better weighting schemes than this that I'm not aware of.

2. Ideally, we would do the time weighting on **all** distances, not just the subset retrieved by hnswlib/chromadb. Referring back to the example above, if the character changes clothes 500 times throughout your chat, but your `n_initial` = 50, that subset of 50 could still be anything from that 500 regardless of recency. I can only think of two ways to deal with this:
* Retrieve everything from the vector store (equivalent to `n_initial` = vector store size) and do all the calculations in Python. This won't scale well, but probably fine for small amounts of data.
* Fork hnswlib and insert the time weighting there. More complicated because hnswlib's custom distance functions accept the vectors only, and not ids, which we would need to calculate the time weights.

  As far as existing approaches, [Langchain's time weighted retriever](https://github.com/hwchase17/langchain/blob/master/langchain/retrievers/time_weighted_retriever.py#L23) just uses an initial k = 100 by default and does weighting on that subset.

### DIFF
--- a/extensions/superbooga/chromadb.py
+++ b/extensions/superbooga/chromadb.py
@@ -76,7 +76,7 @@ class ChromaCollector(Collecter):
         return [x for _, x in sorted(zip(ids, documents))]
 
     # Multiply distance by factor within [min_time_weight, 1] where more recent is lower
-    def apply_time_weight_to_distances(self, ids: list[int], distances: list[float], min_time_weight: float = 1.0):
+    def apply_time_weight_to_distances(self, ids: list[int], distances: list[float], min_time_weight: float = 1.0) -> list[float]:
         if len(self.ids) <= 1: return distances.copy()
         return [distance * (_id / (len(self.ids)-1) * min_time_weight - _id / (len(self.ids)-1) + 1) for _id, distance in zip(ids, distances)]
 
--- a/extensions/superbooga/chromadb.py
+++ b/extensions/superbooga/chromadb.py
@@ -45,16 +45,9 @@ class ChromaCollector(Collecter):
         self.ids = [f"id{i}" for i in range(len(texts))]
         self.collection.add(documents=texts, ids=self.ids)
 
-    def get_documents_and_ids(self, search_strings: list[str], n_results: int):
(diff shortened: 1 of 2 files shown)


oobabooga:
> Retrieve everything from the vector store (equivalent to n_initial = vector store size) and do all the calculations in Python. This won't scale well, but probably fine for small amounts of data.

Maybe we could make it such that `n_initial=0`, all ids are retrieved?

toast22a:
Just fixed two minor things:
* Fixed `get_documents_ids_distances()` to get the distances from the db
* If n_initial = -1, all ids are retrieved
  * chose -1 instead of 0 for this to stay consistent with "default value" convention elsewhere e.g. the random seed setter; feel free to change this if you want it as 0 anyway
  * n_initial = 0 now just raises `ValueError: n_initial 0 should be >= n_results <n_results>`

Everything else still works as intended, just tested.

oobabooga:
Thank you!

PR #2087, opened 2023-05-15
Add settings UI for llama.cpp and fixed reloading of llama.cpp models

strnad:
I added settings for llama.cpp models to the UI tab.
I also fixed unloading and reloading of llama.cpp models, previously it didnt free the memory. Unfortunately it releases only RAM but not VRAM (when using offloading to GPU) because [llama.cpp library doesn't release the memory in llama_free method](https://github.com/ggerganov/llama.cpp/issues/1456).

### DIFF
--- a/modules/llamacpp_model.py
+++ b/modules/llamacpp_model.py
@@ -16,6 +16,9 @@ class LlamaCppModel:
     def __init__(self):
         self.initialized = False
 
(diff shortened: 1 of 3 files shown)


oobabooga:
Looks good, thank you. I hadn't noticed that llama.cpp was not releasing the allocated RAM.

PR #2095, opened 2023-05-16
Add superbooga option to set embedder model

toast22a:
Current behavior: The embedder in superbooga defaults to using a SentenceTransformer with a fixed model string `sentence-transformers/all-mpnet-base-v2`.

This PR adds an option to set the embedder model. Two options are exposed to settings.json:
* `superbooga-embedder_model_type`: `sentence_transformer` or `instructor`
* `superbooga-embedder_model_name_or_path`: HF-style model string passed to constructor

Support for [InstructorEmbedding](https://github.com/HKUNLP/instructor-embedding) models is also added.

Additional changes:
* Embedder and collector initialization in `script.py` are moved into function `setup()` so that the updated params (using values from settings.json) are used.
* The default global embedder in script.py is renamed `embedder_default`, and is initially `None` to save memory. It is only initialized when a default `Embedder` is requested through `get_default_embedder()`.

### DIFF
--- a/extensions/superbooga/chromadb.py
+++ b/extensions/superbooga/chromadb.py
@@ -27,10 +27,20 @@ class Collecter():
 
 
 class Embedder():
-    def __init__(self):
-        pass
+    DEFAULT_DOCUMENT_TEMPLATE = '<|text|>'
+    DEFAULT_QUERY_TEMPLATE = '<|text|>'
+
+    def __init__(self, document_template: Optional[str] = None, query_template: Optional[str] = None):
+        self.document_template = document_template or self.DEFAULT_DOCUMENT_TEMPLATE
+        self.query_template = query_template or self.DEFAULT_QUERY_TEMPLATE
 
     def embed(self, text: str) -> list[torch.Tensor]:
+        pass
+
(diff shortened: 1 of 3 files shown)


toast22a:
Added document and query templating for zero-shot capable embedders that need prompting (e.g. Instructor). We separate these two because, depending on the task, the templates for documents and queries might have to be different. An example from the Instructor paper:

Task: Question answering
Document template: `Represent the evidence document for retrieval; Input: <|text|>`
Query template: `Represent the question for retrieving evidence documents; Input: <|text|>`

To support this behavior,
* I added methods `embed_document()` and `embed_query()` to the `Embedder` class.
* So that the raw document texts (not the templated ones) get added to/retrieved later from the DB, I also moved embedding from chromadb's `embedding_function` to the `ChromaCollector` class. This way, we can call `self.collection.add(documents=<raw texts>, embeddings=<embeddings of templated texts>, ...)` and `self.collection.query(query_embeddings=<embeddings of templated texts>, ...)`.
* Two new options are exposed to settings.json, but can be set in the UI as well:
  * `superbooga-document_template`
  * `superbooga-query_template`

Issue #2078, opened 2023-05-15
4-bit Lora training fail to start when Evaluation dataset is not None


janvarev:
### Describe the bug

4-bit Lora training fail to start when Evaluation dataset is not None.

IMPORTANT: I could start Train with Evaluation dataset by change line in training.py at 383
`save_strategy="steps"`,
but this is temporary solution.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Start with 4bit model
2. Try to train Lora WITH Evaluation dataset (all works ok WITHOUT it)

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/18393788/e65afde1-fb14-43d4-8891-aa917aae17c2)

### Logs

```shell
INFO:Prepping for training...
INFO:Creating LoRA model...
Traceback (most recent call last):
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\gradio\routes.py", line 395, in run_predict
    output = await app.get_blocks().process_api(
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\gradio\blocks.py", line 1193, in process_api
    result = await self.call_function(
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\gradio\blocks.py", line 930, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\anyio\to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\anyio\_backends\_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\anyio\_backends\_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\gradio\utils.py", line 491, in async_iteration
    return next(iterator)
  File "C:\Software\text-gen-webui\text-generation-webui\modules\training.py", line 372, in do_train
    args=transformers.TrainingArguments(
  File "<string>", line 110, in __init__
  File "C:\Software\text-gen-webui\installer_files\env\lib\site-packages\transformers\training_args.py", line 1180, in __post_init__
    raise ValueError(
[...]

PR #2099, opened 2023-05-16
Fix Training fails when evaluation dataset is selected

Ph0rk0z:
Fixes https://github.com/oobabooga/text-generation-webui/issues/2078 from Googulator


oobabooga:
Thank you

PR #2100, opened 2023-05-16
Experimental jank multiGPU inference that's 2x faster than native somehow

mcmonkey4eva:
Hello it's 4 am and I've committed attempted housefire

image

----

Anyway so this PR updates `pre_layer` to support a comma-separated list of integer layer IDs, such that if you use a `,` it enables new _multiGPU support_ in GPTQ. And by 'in GPTQ' I mean in the PR for GPTQ: https://github.com/oobabooga/GPTQ-for-LLaMa/pull/1

I've done my best to keep it all backwards-compatible, ie if you're using any fork of GPTQ that isn't my PR, `pre_layer` should still work as intended. Also if you use `pre_layer` without a `,` in it it will still use the old behavior.

It seems like due to <https://huggingface.co/docs/optimum/bettertransformer/overview> possibly? Native multi-GPU exists and half-works, but when loading LLaMA-30B-4Bit-gr128, native gets me 8 token/sec and this glorious 4 am creation gets me 19 token/sec. So, 2x perf on multi-GPU inference. I'm sure someone somewhere needs that!

It's not maxing out usage of the GPU at all, it splits load evenly. I'd bet a careful implementation of beam search could be made such that each GPU runs at full (ie GPU0 runs Beam 2 while GPU1 is finishing Beam 1). But that's future problems.

Here's how fast it runs for me using 1x 3090 and 1x 3080 TI with `pre_layer 50,60`

multigpulol

And here's what nvtop has to say about it:

Screenshot from 2023-05-16 03-29-40

using `30,60` splits much more evenly but runs slightly slower (17 token/sec).

---

Note: https://github.com/oobabooga/GPTQ-for-LLaMa/pull/1 must be pulled alongside this for it to work.

(ps I won't blame you if you call me insane and close these PRs, this is a mess)

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ Optionally, you can use the following command-line flags:
 | `--wbits WBITS`           | Load a pre-quantized model with specified precision in bits. 2, 3, 4 and 8 are supported. |
 | `--model_type MODEL_TYPE` | Model type of pre-quantized model. Currently LLaMA, OPT, and GPT-J are supported. |
 | `--groupsize GROUPSIZE`   | Group size. |
-| `--pre_layer PRE_LAYER`   | The number of layers to allocate to the GPU. Setting this parameter enables CPU offloading for 4-bit models. |
+| `--pre_layer PRE_LAYER`   | The number of layers to allocate to the GPU. Setting this parameter enables CPU offloading for 4-bit models. Use a comma-separated list for multi-GPU, eg `30,60`. |
 | `--checkpoint CHECKPOINT` | The path to the quantized checkpoint file. If not specified, it will be automatically detected. |
 | `--monkey-patch`          | Apply the monkey patch for using LoRAs with quantized models.
 | `--quant_attn`         | (triton) Enable quant attention. |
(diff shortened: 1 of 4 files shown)


oobabooga:
This is a nice addition. Tagging @qwopqwop200 and @PanQiWei for awareness.

oobabooga:
I have made a change to use `nargs="+"` instead of passing the arguments as a string. Can you check if it still works as expected @mcmonkey4eva? I hope your GPU survived, by the way.

Ph0rk0z:
Absolute kings! I get up to 2.6 it/s on the 65b split over P40 and 3090. It can do full context without OOM too. Even loaded up, it gets around 1.8 it/s. Finally made splitting usable as speed pretty much doubled.

Had to set some strict=False in the dict to load that particular model, but this is pretty big. No more playing around with guessing what # MiB will work.

In 0cc4m's GPTQ the offloading works for other models apparently so in theory this could be made for all and not just llama.

C000001:
Is Crossfire / SLI necessary?

mcmonkey4eva:
nope. As you can see in the picture in my original post, I don't have SLI... or anything sane, just a PCIe riser and a deathwish.

mcmonkey4eva:
Bonus update: thanks to @TheBloke for converting a LLaMA-65B-3Bit model for me, I now have 65B running, split across two GPUs, at 12 token/sec! Running with `--pre_layer 55 80` is redlining my available VRAM right now, and it starts visibly struggling when the context is long, but ... it works!

65b_story

oobabooga:
I'm jealous :)

Ph0rk0z:
That's the difference between 2 ampere GPU and 1 ampere + pascal, heh.

Don't even look at how much watts this uses, btw.

PR #2107, opened 2023-05-16
fix: elevenlabs cloned voices do not show up in webui after entering API key

PixelDJ:
The elevenlabs api key is not sent when refreshing voices. Therefore, only the default voices are shown. This fix adds a call to the elevenlabs api library to set the API key when changed, so that it is pushed to Elevenlabs when refreshing voices.

### DIFF
--- a/extensions/elevenlabs_tts/script.py
+++ b/extensions/elevenlabs_tts/script.py
@@ -17,6 +17,12 @@ voices = None
 wav_idx = 0
 
 
+def update_api_key(key):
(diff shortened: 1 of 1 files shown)


oobabooga:
I think that it would be better to create an `update_api_key(key)` function instead of having an unused output in the event handler.

PixelDJ:
Good point. I pushed a commit to fix that.

oobabooga:
Looks perfect now, thanks!

PR #2108, opened 2023-05-16
Lora Trainer improvements, part 6 - slightly better raw text inputs

mcmonkey4eva:
- Added an error check for if overlap/cutoff lengths don't make sense, that gives a clear error message (as a user on Discord was confused about it)
- Added `Hard Cut String`, a way to prevent unwanted overlaps.
This enables you to have text files like...
```
User: Question
Assistant: Answer


User: Question2
Assistant: Answer2
```
And it works just like a formatted json dataset wherein it properly cuts each chunk apart, but instead of messy json data you just split em with newlines. I made it generic enough that you can repurpose it to other split strings, other numbers of newlines, etc. easily within the UI.
Similarly useful if the raw text input is a series of different documents / quotes / whatever, separated by some consistent space block.
Prior to this change, models trained with this type of input would tend to learn that it's okay to suddenly and randomly change the subject (as, well, that's what the data did! And now, with this, your data doesn't, so theoretically your bot doesn't either).
I was originally going to use `ast` logic like `custom_stopping_strings` does but tbh I don't think it's worth the bother? You're unlikely to have multiple separator strings that need python-syntax input, so I just specialcased newlines and made it otherwise just a string.
- Also, added short bit of documentation about raw text file settings in the docs.
- ~~I was planning to test and implement the fix <https://github.com/oobabooga/text-generation-webui/discussions/1586> but I can't actually currently replicate the bug anymore ... and I do not not why. I also don't know why I was ever able to replicate the bug in the first place, as it's a strange bug that shouldn't happen. I might've updated packages at some point and that's why it's fine now? Not sure.~~ Confirmed as fixed by upstream
- I also wanted to look into multigpu training (per my jank housefire of a multigpu setup https://github.com/oobabooga/text-generation-webui/pull/2100 ) but, uh, frankly `torchrun` is an absolute nightmare, and pytorch internal splitting... exists, but is awful. And building my own multigpu logic as in that PR would be way more complicated, as it involves HF Transformers internals. So I'm not sure on how to do that, but it's a whole separate massive project.

### DIFF
--- a/docs/Training-LoRAs.md
+++ b/docs/Training-LoRAs.md
@@ -75,6 +75,13 @@ So for example if a dataset has `"instruction": "answer my question"`, then the 
 
 If you have different sets of key inputs, you can make your own format file to match it. This format-file is designed to be as simple as possible to enable easy editing to match your needs.
 
+## Raw Text File Settings
+
+When using raw text files as your dataset, the text is automatically split into chunks based on your `Cutoff Length` you get a few basic options to configure them.
+- `Overlap Length` is how much to overlap chunks by. Overlapping chunks helps prevent the model from learning strange mid-sentence cuts, and instead learn continual sentences that flow from earlier text.
+- `Prefer Newline Cut Length` sets a maximum distance in characters to shift the chunk cut towards newlines. Doing this helps prevent lines from starting or ending mid-sentence, preventing the model from learning to cut off sentences randomly.
+- `Hard Cut String` sets a string that indicates there must be a hard cut without overlap. This defaults to `\n\n\n`, meaning 3 newlines. No trained chunk will ever contain this string. This allows you to insert unrelated sections of text in the same text file, but still ensure the model won't be taught to randomly change the subject.
+
 ## Parameters
 
 The basic purpose and function of each parameter is documented on-page in the WebUI, so read through them in the UI to understand your options.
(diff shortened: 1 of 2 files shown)


Ph0rk0z:
FYI.. since you're working on this.. is it possible to fix this error for steps under 100?

```
Exception in thread Thread-11 (threaded_run):
  File "/home/mint/miniconda3/envs/textgen/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/mint/miniconda3/envs/textgen/lib/python3.10/threading.py", line 953, in run
  File "/home/mint/gptq2/text-generation-webui-testing/modules/training.py", line 416, in threaded_run
    trainer.train()
  File "/home/mint/miniconda3/envs/textgen/lib/python3.10/site-packages/transformers/trainer.py", line 1664, in train
    return inner_training_loop(
  File "/home/mint/miniconda3/envs/textgen/lib/python3.10/site-packages/transformers/trainer.py", line 2066, in _inner_training_loop
    train_loss = self._total_loss_scalar / self.state.global_step
ZeroDivisionError: float division by zero

```

mcmonkey4eva:
That'd be a reasonable idea. To my knowledge, there's nothing stopping that from being done within current code (assuming special tokens aren't stripped by default encode/decode calls)

oobabooga:
Fair enough, thank you.

Ph0rk0z:
Ok.. makes sense. Then it can't be fixed.

PR #2153, opened 2023-05-18
Remove irrelevant Metharme context about text adventure games

CarlKenner:


oobabooga:
Isn't this required (or at least desirable)?

https://huggingface.co/PygmalionAI/metharme-7b#prompting-example

CarlKenner:
Absolutely not. This isn't a text-adventure-game-where-you-have-three-options-each-turn model. It's a generic instruction-following model that can also be used for conversation, roleplaying and storywriting. That was just one example of millions of options the user could put in for a system prompt, if they wanted that specific thing. 99.99% of the time, the user won't want that. Much better to leave it blank and let the user put in their own choice if they want something specific, or use the default model behaviour if they don't want anything specific.

PR #2180, opened 2023-05-18
Update docker-compose.yml to reflect the move to the "docker" folder

ramblingcoder:
The docker related files appear to have started at the root of the repository and then were moved to the "docker" folder. However the docker compose file doesn't appear to have been updated to reflect this. 

I noticed the stablediffusion PR https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/7946 has a similar docker structure to this and applied similar changes. This ended up working on my local copy and I am able to run "docker compose build" to generate the "oobabooga/text-generation-webui" image for my use in Portainer.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,8 +1,10 @@
 version: "3.3"
 services:
   text-generation-webui:
+    image: oobabooga/text-generation-webui
     build:
-      context: .
+      context: ../
+      dockerfile: docker/Dockerfile
       args:
         # specify which cuda version your card supports: https://developer.nvidia.com/cuda-gpus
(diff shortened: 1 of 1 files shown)


oobabooga:
The readme instructs to create a symbolic link https://github.com/oobabooga/text-generation-webui/#alternative-docker

ramblingcoder:
Thank you, I see where I became confused then and believe there is a gap in the documentation. 

The documentation on the root readme.md mentions the symbolic link with the first command (ln -s etc.). 

Unfortunately that documentation wasn't the one I found first and was following the file at "text-generation-webui/docs
/Docker.md" which doesn't appear to mention symbolic links. 

This is probably due to me searching for "oobabooga docker" in google and that document was returned first.

Screenshot_20230518_193309

Thank you for the clarification going forward!

PR #2181, opened 2023-05-18
Update llama.cpp-models.md

Crimsonfart:
An additional tip for people using the windows one click installer.


oobabooga:
I think that the more instructions than that are needed, like how to install Visual C++. I don't know if some nvidia development library may also necessary. You are welcome to submit a new PR if you feel like writing a mode detailed guide

PR #2197, opened 2023-05-19
Use a libary for log coloring

gukoff:
Not only the current approach is outdated and hacky, but it fails to log the exceptions and breaks with the error:

```
  File "/Users/gukov/Downloads/oobabooga_macos/installer_files/env/lib/python3.10/logging/__init__.py", line 1624, in _log
    self.handle(record)
  File "/Users/gukov/Downloads/oobabooga_macos/installer_files/env/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/Users/gukov/Downloads/oobabooga_macos/installer_files/env/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/Users/gukov/Downloads/oobabooga_macos/installer_files/env/lib/python3.10/logging/__init__.py", line 968, in handle
    self.emit(record)
  File "/Users/gukov/Downloads/oobabooga_macos/text-generation-webui/modules/logging_colors.py", line 94, in new
    args[1].msg = color + args[1].msg + '\x1b[0m'  # normal
TypeError: can only concatenate str (not "OSError") to str
```

---

Note: the library `coloredlogs` has a different default log format.

Old format:
```
ERROR:No models are available! Please download at least one.
```

New format:
```
2023-05-19 12:07:50 gukoff root[97281] ERROR No models are available! Please download at least one.
```


oobabooga:
I find the current format more friendly. logging is only used for simple info/warning/error messages passed as strings in the web ui

gukoff:
> I find the current format more friendly. logging is only used for simple info/warning/error messages passed as strings in the web ui

The current implementation is literally broken?..

gukoff:
ping @oobabooga

oobabooga:
@gukoff I don't experience any errors with the current implementation. I have recently updated all occurrences of `logging.warning/info/error` in the code base to

```python
from modules.logging_colors import logger
logger.info(...)
```

So the only benefit would be able to print non-string objects through logger?

Issue #2160, opened 2023-05-18
Superbooga error in instruct mode


mykeehu:
### Describe the bug

If I am in instruct mode and superbooga is enabled, I get the error after sending the first message:
```
Traceback (most recent call last):
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\routes.py", line 414, in run_predict
    output = await app.get_blocks().process_api(
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\blocks.py", line 1323, in process_api
    result = await self.call_function(
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\blocks.py", line 1067, in call_function
    prediction = await utils.async_iteration(iterator)
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\utils.py", line 339, in async_iteration
    return await iterator.__anext__()
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\utils.py", line 332, in __anext__
    return await anyio.to_thread.run_sync(
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\anyio\to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\anyio\_backends\_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\anyio\_backends\_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "J:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\utils.py", line 315, in run_sync_iterator_async
    return next(iterator)
  File "J:\oobabooga_windows\text-generation-webui\modules\chat.py", line 307, in generate_chat_reply_wrapper
    for history in generate_chat_reply(text, state, regenerate, _continue):
  File "J:\oobabooga_windows\text-generation-webui\modules\chat.py", line 301, in generate_chat_reply
    for history in chatbot_wrapper(text, state, regenerate=regenerate, _continue=_continue):
  File "J:\oobabooga_windows\text-generation-webui\modules\chat.py", line 217, in chatbot_wrapper
    prompt = apply_extensions('custom_generate_chat_prompt', text, state, **kwargs)
[...]

PR #2212, opened 2023-05-20
Fix superbooga error when querying empty DB (Issue #2160)

toast22a:
# Current behavior
(Issue #2160) If the ChromaDB collection is queried (as in `ChromaCollector.get_documents_and_ids()`) before any documents have been added to it, it raises a `NoIndexException`.

# Fix
`ChromaCollector.get_documents_and_ids()` returns two empty lists without querying the DB if `n_results == 0`. Since `n_results` is set to 0 if `self.ids == 0`, this prevents the DB from being queried if it is empty.

# Additional changes
* `ChromaCollector.add()` returns immediately without adding to the DB if `len(texts) == 0`.
* `ChromaCollector.clear()` now clears its own `self.ids` attribute after clearing the DB itself, keeping consistency between `self.ids` and the actual contents of the DB.

### DIFF
--- a/extensions/superbooga/chromadb.py
+++ b/extensions/superbooga/chromadb.py
@@ -42,11 +42,13 @@ class ChromaCollector(Collecter):
         self.ids = []
 
     def add(self, texts: list[str]):
+        if len(texts) == 0: return
         self.ids = [f"id{i}" for i in range(len(texts))]
         self.collection.add(documents=texts, ids=self.ids)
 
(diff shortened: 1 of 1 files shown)


oobabooga:
That's perfect, thank you!

csprocket777:
What's the process for triggering `ChromaCollector.clear()`? Is there a UI for clearing the DB or a cli command?

PR #2213, opened 2023-05-20
Fix "ValueError: Queue needs to be enabled!" error on No_stream

ChobPT:
Added enable_queue to prevent no_stream errors


ChobPT:
Test was a false positive

Issue #2195, opened 2023-05-19
OpenAI Extension: The following AI output couldn't be converted to a JSON using Auto-GPT


Simplegram:
### Describe the bug

Auto-GPT can't convert any output from the API to JSON

Settings I've tried:
- python server.py --chat --listen --xformers --monkey-patch --bf16
- python server.py --chat --listen --extensions openai

Models I've tried:
- 4bit_gpt4-x-alpaca-13b-native-4bit-128g-cuda
- anon8231489123_vicuna-13b-GPTQ-4bit-128g
- TheBloke_wizard-mega-13B-GPTQ
- TheBloke_stable-vicuna-13B-GPTQ

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Name:  DataFinderGPT
Role:  An intelligent search engine AI designed specifically for finding relevant and accurate data related to any given topic quickly and efficiently.
Goals: []
API Budget: infinite

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/34500695/e2f8c9be-6205-4499-853f-4d8b35a71909)

### Logs

```shell
oobabooga
127.0.0.1 - - [19/May/2023 16:06:55] "POST /v1/chat/completions HTTP/1.1" 200 -
Output generated in 9.73 seconds (28.56 tokens/s, 278 tokens, context 1111, seed 1047326713)
127.0.0.1 - - [19/May/2023 16:07:05] "POST /v1/chat/completions HTTP/1.1" 200 -
Output generated in 0.57 seconds (7.06 tokens/s, 4 tokens, context 547, seed 1540217231)

auto-gpt
DataFinderGPT  has been created with the following details:
Name:  DataFinderGPT
Role:  An intelligent search engine AI designed specifically for finding relevant and accurate data related to any given
 topic quickly and efficiently.
Goals:
Using memory of type:  LocalCache
Using Browser:  chrome
Apparently json was fixed.
The JSON object is invalid.
Error:
  'str' object has no attribute 'get'
NEXT ACTION:  COMMAND = None ARGUMENTS = None
  Enter 'y' to authorise command, 'y -N' to run N continuous commands, 's' to run self-feedback commands, 'n' to exit program, or enter feedback for DataFinderGPT...
  Asking user via keyboard...
```

### System Info

```shell
Windows 11
CUDA branch
MSI RTX 4090 Suprim X
Local run
```

BarfingLemurs:
[...]

PR #2219, opened 2023-05-20
[extensions/openai]: assistant: space fix, system: prompt fix

matatonic:
Some more extra space related fixes, fix up system: prompt when multiline.

Related to https://github.com/oobabooga/text-generation-webui/issues/2195

### DIFF
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -243,9 +243,9 @@ class Handler(BaseHTTPRequestHandler):
 
                 messages = body['messages']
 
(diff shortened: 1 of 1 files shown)


extensions/openai/script.py
```diff
@@ -432,9 +442,15 @@ def do_POST(self):
                 self.wfile.write(response.encode('utf-8'))
                 # Finished if streaming.
                 if debug:
+                    if answer and answer[0] == ' ':
+                        answer = answer[1:]
```
oobabooga, line 446: I think that it's more elegant to use `.lstrip(' ')` in cases like this

matatonic: Hrm, in these cases I explicitly only want to remove a single space, nothing more. other spaces are valid and should remain.

oobabooga: fair enough

Issue #2228, opened 2023-05-20
Hugging Face peft and bitsandbytes mismatch


jasonmcaffee:
### Describe the bug

This just happened overnight.  Yesterday things were working great on google colab, but this morning I ran the same notebook and ran into an error:
```
AttributeError: module 'bitsandbytes.nn' has no attribute 'Linear4bit'
```

I did some digging, and found that:
HuggingFace Peft has just added 4-bit QLoRA via bitsandbytes 
https://github.com/huggingface/peft/pull/476

Since the requirements use latest git peft, but are locked to bitsandbytes version 0.38.1,  this error occurs due to bitsandbytes being outdated.

A temporary workaround is change requirements.txt to lock peft to a commit that was done a week ago, and does not include this change.

```
accelerate==0.19.0
colorama
datasets
flexgen==0.1.7
gradio_client==0.2.5
gradio==3.31.0
markdown
numpy
pandas
Pillow>=9.5.0
pyyaml
requests
rwkv==0.7.3
safetensors==0.3.1
sentencepiece
tqdm
git+https://github.com/huggingface/peft@27af2198225cbb9e049f548440f2bd0fba2204aa
transformers==4.29.1
bitsandbytes==0.38.1; platform_system != "Windows"
llama-cpp-python==0.1.51; platform_system != "Windows"
https://github.com/abetlen/llama-cpp-python/releases/download/v0.1.51/llama_cpp_python-0.1.51-cp310-cp310-win_amd64.whl; platform_system == "Windows"
```

You can simply add this echo command before installing requirements.txt:
```
!echo -e "accelerate==0.19.0\ncolorama\ndatasets\nflexgen==0.1.7\ngradio_client==0.2.5\ngradio==3.31.0\nmarkdown\nnumpy\npandas\nPillow>=9.5.0\npyyaml\nrequests\nrwkv==0.7.3\nsafetensors==0.3.1\nsentencepiece\ntqdm\ngit+https://github.com/huggingface/peft@27af2198225cbb9e049f548440f2bd0fba2204aa\ntransformers==4.29.1\nbitsandbytes==0.38.1; platform_system != \"Windows\"\nllama-cpp-python==0.1.51; platform_system != \"Windows\"\nhttps://github.com/abetlen/llama-cpp-python/releases/download/v0.1.51/llama_cpp_python-0.1.51-cp310-cp310-win_amd64.whl; platform_system == \"Windows\"" > requirements.txt
[...]

PR #2229, opened 2023-05-20
Fix broken version of peft

jllllll:
Peft updated to support an unreleased, closed beta version of bitsandbytes. Naturally, they also broke support for the current version of bitsandbytes.

See: #2228


oobabooga:
Thanks!

PR #2234, opened 2023-05-20
use pkg_resources to display the runtime version of llama-cpp-python

brandonmcclure:
small quality of life addition to the UI to display what version of llama-cpp-python is loaded at runtime. 

With the changes to quantization in llama.cpp I have some confusion regarding which models I need to run with which commit/release of text-generation-webui. Adding this gives me a quick reminder in the ui of what version I am on. 

I have only tested on linux. I would imagine it may fail on windows due to that package being installed directly from source instead of through pip.

### DIFF
--- a/server.py
+++ b/server.py
@@ -3,6 +3,7 @@ import os
 import requests
 import warnings
 import modules.logging_colors
+import pkg_resources
 
 os.environ['GRADIO_ANALYTICS_ENABLED'] = 'False'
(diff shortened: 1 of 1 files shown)


oobabooga:
I don't like the appearance. Also, even if you know the llama-cpp-python version, there is no easy way to tell what models will be compatible with it

print

I believe that soon llama.cpp will settle down and breaking changes will stop being an issue

PR #1521, opened 2023-04-24
Thread-safe API switch

Myp3a:
## Current behaviour
If an API request comes while processing another one, first request will be aborted.
<details>
<summary>Error log</summary>

```
100.75.108.28 - - [24/Apr/2023 23:26:48] "POST /api/v1/generate HTTP/1.1" 200 -
Invalidate trace cache @ step 11: expected module 15, but got module 0
Invalidate trace cache @ step 0: expected module 146, but got module 0
Traceback (most recent call last):
  File "/home/ivga/TasyaPugAI/text-generation-webui/modules/callbacks.py", line 66, in gentask
    ret = self.mfunc(callback=_callback, **self.kwargs)
  File "/home/ivga/TasyaPugAI/text-generation-webui/modules/text_generation.py", line 257, in generate_with_callback
    shared.model.generate(**kwargs)
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/transformers/generation/utils.py", line 1485, in generate
    return self.sample(
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/transformers/generation/utils.py", line 2524, in sample
    outputs = self(
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1538, in _call_impl
    result = forward_call(*args, **kwargs)
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/transformers/models/opt/modeling_opt.py", line 938, in forward
    outputs = self.model.decoder(
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1538, in _call_impl
    result = forward_call(*args, **kwargs)
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/transformers/models/opt/modeling_opt.py", line 704, in forward
    layer_outputs = decoder_layer(
  File "/home/ivga/TasyaPugAI/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1538, in _call_impl
    result = forward_call(*args, **kwargs)
[...]

PR #2262, opened 2023-05-21
Beginning of multi-user support

flurb18:
This PR makes it so the app doesn't crash on concurrent requests. It does so by forwarding generate_reply tasks to a queue with a service worker, which is initialized before creating the interface. The function generate_reply is rewritten to be a simple wrapper around the queue, ensuring backwards compatibility. The old blocking function is renamed as _generate_reply, which is called by the worker when a task is placed in the queue.

Behavior before PR:
Concurrent text generation requests via API, UI, extensions, etc. would crash the program entirely.

Behavior after PR:
Concurrent text generation requests are placed in a queue and executed in the order they came in.

Note that this doesn't separate session states or chat histories, so that's still an issue for multiple people trying to access through the native UI. 

One final important note is that I haven't _rigorously_ tested this. I have tested the functionality it promises - concurrent requests are in fact queued - but I'm not entirely sure if running the tasks in a single worker introduces a bottleneck on the CPU. Ideally we want multiple cores as far as I know (?), and I'm not sure if the threading library utilized here will automatically support that since _generate_reply is called in the worker. Something to look into. I get full speed, but I run GPTQ models on GPU so not sure if that's indicative, or if this is even a problem.

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -5,9 +5,7 @@ from pathlib import Path
 
 import yaml
 
-from modules.text_generation import GenerationQueue
-
-queue = GenerationQueue()
+queue = None
 model = None
 tokenizer = None
 model_name = "None"
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -5,6 +5,9 @@ from pathlib import Path
 
 import yaml
 
+from modules.text_generation import GenerationQueue
+
+queue = GenerationQueue()
 model = None
 tokenizer = None
 model_name = "None"
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -9,6 +9,34 @@ import numpy as np
 import torch
 import transformers
 
+import threading
+import queue
+
+class GenerationQueue:
+    def __init__(self):
+        self.task_queue = queue.Queue()
+        self.worker_thread = threading.Thread(target=self.process_generate_queue)
+        self.worker_thread.daemon = True
+        self.worker_thread.start()
+
+    def schedule_generation(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


flurb18:
Sorry about the extra commit, I accidentally introduced a circular import in the first one.

oobabooga:
Gradio already has a queue out of the box. The bottleneck in having multiuser support is not the queue system but not using global variables, which in turn would lead to other problems like the UI state not being persistent.

So the queue system would benefit only the API. https://github.com/oobabooga/text-generation-webui/pull/1521 alternatively proposes a lock approach. I don't know which one is preferable

flurb18:
Yeah I wrote this for simultaneous api calls and extension calls to generate_reply, which wasn't safe before. Queue and lock are essentially the same thing at a low level (queue will use a lock under the hood) as far as I understand; I chose the queue here because of the ease of implementation.

oobabooga:
Thanks for the clarification. What if we had something like this inside `generate_reply` that then calls `_generate_reply` once the lock is released?

https://stackoverflow.com/a/39147428

It seems simpler than having a queue that spawns new threads.

oobabooga:
I feel safer with the lock approach because even if just 1 new thread is created, past experience shows that it is easy to end up with obscure memory leaks in pytorch when threading is involved. Could you try it and update the PR if it works as expected?

flurb18:
Updated to lock based approach, works on my machine

oobabooga:
It works perfectly, thank you!

flurb18:
Cool! Thanks for merging.

PR #2265, opened 2023-05-22
Added API endpoints for superbooga

nabelekm:
This PR allows use of superbooga extension in the API extension while keeping current functionality intact.
PR implements these new API endpoints: 

**'/api/v1/superstream'** 
Adds "### Memory:" + chromadb chunks in front of the recieved user input before passing it to generator.

**'/api/v1/feed_urls'**
Accepts urls as string and passes them through current superbooga extension into the chromadb

Unfortunately, due to the current implementation of superbooga extension, collector variable had to be added to /superbooga/script.py functions, so it could be called externally from the API. 

Do you have some plans with the api extension? I'm using it on a regular basis and would be happy to help improving it further if possible.


oobabooga:
I am not a fan of mixing the api extension with superbooga. Superbooga should work automatically behind the scenes

In the chat API in 'chat' mode it already without any changes. In 'instruct' mode it would also work if there was something in the chromadb. So the question is how to add content to the chromadb without using the gradio UI.

nabelekm:
> So the question is how to add content to the chromadb without using the gradio UI.

That's whats the /feed_urls is doing. I was able to succesfully load some urls into chromadb through it.

I'm not proud of the extension mixing either, but I wanted to avoid duplicating the chromadb feeding code.

However, I would suggest making the database infusion optional, either through a flag in the api request, or even better by using some special string (which I remember seeing in some other discussion), that would be replaced with chunks and could be added anywhere in the prompt.

I will look into this further in the next couple of days.

nabelekm:
I have moved the data feeding functions from script.py into modules, so they can be accessed from api extension as well. Also, I've reworked the api to allow loading of data into chromadb, sharing the collection with gradio.

- [x] Changed api endpoint for feeding data into /feed_data. Now it accepts same inputs as superbooga in gradio.

```
/feed_data props

        data: string
        data_type: text, url or file
        chunk_size: 300,
        strong_cleanup: true,
        threads: 4,
```
- [x] Implemented file mode feeding into api

- [x]  Basic PDF extraction functionality through file mode which currently accepts file urls. Worked successfully with dropbox link to a ...file.pdf?dl=1

Todo:
- [ ]  Implement recieving number of chunks in the streaming request to be added to the the context.
- [ ]  Remove /superapi endpoint and include something like "infuse": True in the regular stream endpoint to enable disable database retrieval as necessary.
- [ ]  Update streaming_api.py further to include training endpoints and avoid duplicate code for every endpoint.

oobabooga:
The changes made don't address the point above

> I am not a fan of mixing the api extension with superbooga. Superbooga should work automatically behind the scenes
> 
> In the chat API in 'chat' mode it already works without any changes. In 'instruct' mode it would also work if there was something in the chromadb. So the question is how to add content to the chromadb without using the gradio UI.

I don't want to move the code in superbooga inside modules or mix it with the API. It should be left as a self-contained extension.

PR #2282, opened 2023-05-22
FIX `get_model_specific_settings` which currently is updated only on program start

RDeckard:
## Current behavior

The "Save settings for this model" save settings properly, but when we come back to the model, they are properly loaded **only after the app being restarted**.

## Fix

`get_model_specific_settings` currently relies on `shared.model_config`, which is computed just once on the program start.

The new behavior still loads `config.yaml` (which should be static indeed) on program start, but loads now dynamically the `config-user.yaml`.

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -214,18 +214,23 @@ def is_chat():
 # Loading model-specific settings
 with Path(f'{args.model_dir}/config.yaml') as p:
     if p.exists():
-        model_config = yaml.safe_load(open(p, 'r').read())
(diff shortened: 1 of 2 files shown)


oobabooga:
Why not update `shared.model_config` directly inside the `save_model_settings` function? It would be a simpler solution without additional disk IO.

RDeckard:
> Why not update shared.model_config directly inside the save_model_settings function?

@oobabooga damn you're right 👍, done! 💹

oobabooga:
Thank you!

PR #2283, opened 2023-05-22
Support of the `--gradio-auth` flag (like `--gradio-auth-path` but without the need of a file)

RDeckard:
## Context

To set up _gradio authentication_, we currently need to set a flag AND to create a file (`--gradio-auth-path GRADIO_AUTH_PATH`).

We could easily have a _flag-only_ solution like with [AUTOMATIC1111/stable-diffusion-webui](https://github.com/AUTOMATIC1111/stable-diffusion-webui).

## Proposed solution

Be able to set up _gradio authentication_ by simply passing a `--gradio-auth username:password` flag to the _server.py_ script (e.g. via the `CMD_FLAGS` constant).


oobabooga:
Thank you, this indeed makes things more convenient.

PR #2288, opened 2023-05-22
Add a `dark_theme` settings

RDeckard:
## Context

Currently, we can e.g. make the _autoload_ mode persistent through `settings.json` (which is great 🙏), but not the UI theme.
Bring this preference through a `__theme=dark` URL parameter is not convenient. The WebUI should be able to remember by itself our preference regarding the UI theme.

## Proposed solution

Here I propose to follow how the WebUI currently handles settings (such as how the _autoload_ mode for models are currently loaded).

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -33,6 +33,7 @@ reload_inputs = []  # Parameters for reloading the chat interface
 need_restart = False
 
 settings = {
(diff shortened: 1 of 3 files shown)


oobabooga:
This is a good idea, thank you

Issue #2118, opened 2023-05-17
LLaMA CPP - GPU Offloading - Windows, No Perf Difference regardless of --n-gpu-layers 


skatardude10:
### Describe the bug

I have noticed there is no difference in performance on the One-Click installer with the new GPU offloading w/ llama cpp. Updated to latest today to try again messing around, same with yesterday though see below:

I used VS2019 build tools Dev console, entered the micromamba-cmd.bat environment to confirm in venv and build tools/cmake are accessible, built llama cpp w/ set CMAKE_ARGS="-DLLAMA_CUBLAS=on" set FORCE_CMAKE=1 and all was successful. I also tried the instructions on the oobabooga llama cpp wiki (basically the same minus VS2019 dev console to install llama cpp w/ gpu offloading on Windows, see reproduction).   

--n-gpu-layers 0, 6, 16, 20, 22, 24, 26, 30, 36, etc... none result in any substantial difference in generation speed. Depending on context size and if its the first or subsequent message, it always ranges from 0.5 - 1.2 t/s 

The GPU layer offloading option does increase VRAM usage as I increase layers, and even at a certain point it OOMs, as you would expect, but generation speed is never affected.  

I find it strange that CUDA usage on my GPU is the same regardless of 0 layers offloaded or 20. 

When streaming generation, generation does not start until all CUDA processing has completed, even with 0 layers offloaded which I find strange as well. Is this normal?

I have tried changing the cache size around- and no cache size or having a cache regardless of size, mlock or not, enabling disabling hyperthreading, and never get any substantial difference in generation speed on llama cpp through oobabooga on Windows One-Click.

What kind of speed increases are you all getting, or should someone expect? I think something might need looking at with the new implementation and installing llama cpp for GPU offloading on the Windows One-Click installer.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

One Click Installer, then follow windows instructions at https://github.com/oobabooga/text-generation-webui/blob/main/docs/llama.cpp-models.md

### Screenshot

![Capture33](https://github.com/oobabooga/text-generation-webui/assets/12177832/9d1057ca-7a8f-4568-af69-08a991fad0b2)

### Logs

```shell
With no GPU offloading:  
Output generated in 46.76 seconds (1.22 tokens/s, 57 tokens, context 1830, seed 280607260)

With GPU offloading, 20 layers:
[...]

PR #2299, opened 2023-05-23
Make llama.cpp read prompt size and seed from settings

DGdev91:
Prompt settings and seed shouldn't be hardcoded, the user should be hardcoded, the user should be able to set them.
with this PR, the code uses the values from settings, and fall backs to the old hardcoded values if that setting value is not set.

Also, i was trying to make my AutoGPT fork and the WebUI work together using the "openai" plugin. AutoGPT is made with GPT3.5/4 in mind, and assumes a larger context size than llama, so the hard limit imposed by the webui prevents it from working (it's also written in the pugin's readme https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai)
But it should be actually possible to make it use a larger prompt context and (kinda) make it work, or at least make easier for people to try that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,10 +228,10 @@ Optionally, you can use the following command-line flags:
 |-------------|-------------|
 | `--threads` | Number of threads to use. |
 | `--n_batch` | Maximum number of prompt tokens to batch together when calling llama_eval. |
-| `--no_mmap` | Prevent mmap from being used. |
+| `--no-mmap` | Prevent mmap from being used. |
 | `--mlock`   | Force the system to keep the model in RAM. |
 | `--cache-capacity CACHE_CAPACITY`   | Maximum cache capacity. Examples: 2000MiB, 2GiB. When provided without units, bytes will be assumed. |
-| `--n_gpu_layers N_GPU_LAYERS` | Number of layers to offload to the GPU. Only works if llama-cpp-python was compiled with BLAS. Set this to 1000000000 to offload all layers to the GPU. |
+| `--n-gpu-layers N_GPU_LAYERS` | Number of layers to offload to the GPU. Only works if llama-cpp-python was compiled with BLAS. Set this to 1000000000 to offload all layers to the GPU. |
 | `--n_ctx N_CTX` | Size of the prompt context. |
 | `--llama_cpp_seed SEED` | Seed for llama-cpp models. Default 0 (random). |
 
--- a/README.md
+++ b/README.md
@@ -228,10 +228,12 @@ Optionally, you can use the following command-line flags:
 |-------------|-------------|
 | `--threads` | Number of threads to use. |
 | `--n_batch` | Maximum number of prompt tokens to batch together when calling llama_eval. |
-| `--no-mmap` | Prevent mmap from being used. |
+| `--no_mmap` | Prevent mmap from being used. |
 | `--mlock`   | Force the system to keep the model in RAM. |
 | `--cache-capacity CACHE_CAPACITY`   | Maximum cache capacity. Examples: 2000MiB, 2GiB. When provided without units, bytes will be assumed. |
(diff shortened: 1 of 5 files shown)


oobabooga:
settings.json is for interface defaults (slider values, checkboxes). this should be new command-line arguments, with corresponding checkboxes in the models tab

DGdev91:
> settings.json is for interface defaults (slider values, checkboxes). this should be new command-line arguments, with corresponding checkboxes in the models tab

Oh, right, sorry.
Anyway, these values can already be set from the interface. they are in the Parameters tab (not Models).
Should i use shared.gradio['chat_prompt_size'] and shared.gradio['seed'] or there's a more proper way to get those values?

DGdev91:
> > settings.json is for interface defaults (slider values, checkboxes). this should be new command-line arguments, with corresponding checkboxes in the models tab
> 
> Oh, right, sorry. Anyway, these values can already be set from the interface. they are in the Parameters tab (not Models). Should i use shared.gradio['chat_prompt_size'] and shared.gradio['seed'] or there's a more proper way to get those values?

I'm perfectly ok to change that code accordingly, there's no need to close the PR.

oobabooga:
The truncation_length parameter is only used in `generate_reply_HF`, not `generate_reply_custom` that uses `llamacpp_model.py`. This is a parameter that affects how the model is loaded, so for consistency with similar parameters it should be defined in modules/shared.py. See all occurences of `mlock` in the source files, adding those two new parameters is exactly the same

DGdev91:
> The truncation_length parameter is only used in `generate_reply_HF`, not `generate_reply_custom` that uses `llamacpp_model.py`. This is a parameter that affects how the model is loaded, so for consistency with similar parameters it should be defined in modules/shared.py. See all occurences of `mlock` in the source files, adding those two new parameters is exactly the same

Ok, i'll give it a look.

...But that's true for the n_ctx parameter, wouldn't make sense to just use the existing seed parameter for llama.cpp?

DGdev91:
@oobabooga So, i'm following your instructions... i found a possible bug.
in the cli there are no-mmap and n-gpu-layers parameters, while in the gradio config they are called no_mmap and n_gpu_layers. this means that changing these vaules don't really means anything in the software, and that can explain https://github.com/oobabooga/text-generation-webui/issues/2118.
(i also tried to set a different default value to n-gpu-layers and it's still at 0 in the UI)

i know that i opened this PR for a compleatly different matter, but since i'm touching the same code... or you prefer to have them as two separate PRs?

oobabooga:
--no-mmap --n-gpu-layers 20 works for me

params

If I change no-mmap in the interface and reload the model, it gets updated accordingly. --n-gpu-layers requires an additional special compilation step to work as described in the docs

oobabooga:
> > The truncation_length parameter is only used in `generate_reply_HF`, not `generate_reply_custom` that uses `llamacpp_model.py`. This is a parameter that affects how the model is loaded, so for consistency with similar parameters it should be defined in modules/shared.py. See all occurences of `mlock` in the source files, adding those two new parameters is exactly the same
> 
> Ok, i'll give it a look.
> 
> ...But that's true for the n_ctx parameter, wouldn't make sense to just use the existing seed parameter for llama.cpp?

It would, but seed is not a generation parameter in llamacpp (as far as I know). Instead, it is a parameter that is used at loading time

DGdev91:
> --no-mmap --n-gpu-layers 20 works for me
> 
> params
> 
> If I change no-mmap in the interface and reload the model, it gets updated accordingly. --n-gpu-layers requires an additional special compilation step to work as described in the docs

Oh, nevermind then. i'll just stick with those settings.

> It would, but seed is not a generation parameter in llamacpp (as far as I know). Instead, it is a parameter that is used at loading time

So you wanna keep it as a separate parameter? ok. i'll name it llamacpp_seed, to make it more obvious those are different things

DGdev91:
@oobabooga Is it fine now or there's something else to change?

oobabooga:
It looks good, thank you!

PR #2313, opened 2023-05-24
FIX `silero_tts` activation/deactivation

RDeckard:
## Current behavior

When we deactivate the `silero_tts` extension unchecking the "_Activate TTS_" checkbox, it still prevents the _stream_ and still modifies the input with "_*Is recording a voice message...*_".

## Proposed solution

Add guard clauses based on `params['activate']` in the beginning of both `state_modifier` and `input_modifier` function.


oobabooga:
We can also do the same for elevenlabs_tts

PR #2315, opened 2023-05-24
LoRA Trainer: `train_only_after` option to control which part of your input to train on

mcmonkey4eva:
tldr: adds new option `train_only_after` under Advanced that you can fill in a string and it will ignore text before that part and only count text after it for training.

Relevant backend `labels` with its magic value `-1` is documented @ https://huggingface.co/transformers/v3.1.0/custom_datasets.html

Labels seems to have other abilities that might be worth looking into? Not sure.

This feature was shown to be in use in the original Stanford Alpaca https://github.com/tatsu-lab/stanford_alpaca/blob/main/train.py#L123 but to my knowledge hasn't been ported to other trainers (eg the original alpaca-lora that was the starting point for this code, way back when, in the ancient times). Not sure if that means it's not that useful, or just nobody's really noticed it's missing before now.

So for training on an Alpaca formatted dataset, you can set this to `### Response:` 

Operates in string space because token space is cursed for some theoretically valid matchers (ie anything that lacks spaces).

The code is a bit longer than I'd like it to be, because operating in string space means replicating the padding code (as otherwise it would double-pad), but not too bad.

I tested and verified that training with/without the setting works as normal, and that the splitter does exactly what it's meant to, and that the labels are being received and processed by the HF Trainer.
I have not tested the effect on quality of training results with vs without.

---

Also, additional fixes:

- Fixed last tokens being stripped. Not sure if this was happening before or as a byproduct of the encoder call change, but either way, it definitely doesn't wrongly strip now!
- HF datasets are cursedbad. See prior issue <https://github.com/oobabooga/text-generation-webui/discussions/1586> but also new issue - they give errors when loading from cache when new labels data is present it seems. ... also after disabling that, it loads so much faster. Wow. Ow.

### DIFF
--- a/modules/training.py
+++ b/modules/training.py
@@ -257,9 +257,9 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
             labels = None
 
         else:
-            ind = prompt.index(train_only_after)
+            ind = prompt.index(train_only_after) + len(train_only_after)
             before_tokens = encode(prompt[:ind])[:-1]
-            after_tokens = encode(prompt[ind + len(train_only_after):])[1:]
+            after_tokens = encode(prompt[ind:])[1:]
 
             full_length = len(after_tokens) + len(before_tokens)
             if full_length > cutoff_len:
--- a/modules/training.py
+++ b/modules/training.py
@@ -247,12 +247,12 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
     shared.tokenizer.padding_side = "left"
 
     def encode(text):
-        return shared.tokenizer.encode(text, truncation=True, max_length=cutoff_len + 1)
+        return shared.tokenizer.encode(text, truncation=True, max_length=cutoff_len)
 
     def tokenize(prompt):
 
         if train_only_after == '' or train_only_after not in prompt:
-            input_ids = encode(prompt)[:-1]
+            input_ids = encode(prompt)
             input_ids = [shared.tokenizer.pad_token_id] * (cutoff_len - len(input_ids)) + input_ids
             labels = [1] * len(input_ids)
 
@@ -264,8 +264,8 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
             full_length = len(after_tokens) + len(before_tokens)
             if full_length > cutoff_len:
                 after_tokens = after_tokens[:cutoff_len - len(before_tokens)]
-                full_length = cutoff_len
-            before_tokens = [shared.tokenizer.pad_token_id] * (cutoff_len - full_length) + before_tokens
+            else:
+                before_tokens = [shared.tokenizer.pad_token_id] * (cutoff_len - full_length) + before_tokens
 
             input_ids = before_tokens + after_tokens
             labels = [-100] * len(before_tokens) + [1] * len(after_tokens)
--- a/modules/training.py
+++ b/modules/training.py
@@ -1,5 +1,6 @@
 import json
 import math
+import random
 import sys
 import threading
 import time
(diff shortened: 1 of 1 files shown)


modules/training.py
```diff
@@ -245,11 +247,35 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
...
+            labels = [1] * len(input_ids)
+
+        else:
+            ind = prompt.index(train_only_after) + len(train_only_after)
+            before_tokens = encode(prompt[:ind])[:-1]
+            after_tokens = encode(prompt[ind:])[1:]
+
+            full_length = len(after_tokens) + len(before_tokens)
+            if full_length > cutoff_len:
+                after_tokens = after_tokens[:cutoff_len - len(before_tokens)]
```
oobabooga, line 267: Could there be an edge case where `len(before_tokens) > cutoff_len`?

oobabooga (commented):
This looks excellent!

modules/training.py
```diff
@@ -245,11 +247,35 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
...
+
+        if train_only_after == '' or train_only_after not in prompt:
+            input_ids = encode(prompt)
+            input_ids = [shared.tokenizer.pad_token_id] * (cutoff_len - len(input_ids)) + input_ids
+            labels = [1] * len(input_ids)
+
+        else:
+            ind = prompt.index(train_only_after) + len(train_only_after)
+            before_tokens = encode(prompt[:ind])[:-1]
+            after_tokens = encode(prompt[ind:])[1:]
```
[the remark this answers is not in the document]
mcmonkey4eva: Oops - good catch on checking that, I had fixed it for the non-onlyafter case but not here. There is no eos token generated, it's a bos that needs to be stripped. I've only tested LLaMA - I *assume* every tokenizer emits one, I haven't verified that though.
Can toss `print(f"[Debug] before ({shared.tokenizer.decode(before_tokens)}) after ({shared.tokenizer.decode(after_tokens)})")` after the `after_tokens = ` line to see exactly what it's generating if in doubt.

modules/training.py
mcmonkey4eva: No, that's guaranteed against on line 251 `max_length=cutoff_len` in the tokenizer encode call.

modules/training.py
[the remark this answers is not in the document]
mcmonkey4eva: Formal guarantee doesn't hurt, added. Fits cleanly.

oobabooga:
Thank you for the fixes @mcmonkey4eva !

modules/training.py
```diff
@@ -245,11 +247,38 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
...
-        result = shared.tokenizer(prompt, truncation=True, max_length=cutoff_len + 1, padding="max_length")
+
+        if train_only_after == '' or train_only_after not in prompt:
+            input_ids = encode(prompt, True)
+            input_ids = [shared.tokenizer.pad_token_id] * (cutoff_len - len(input_ids)) + input_ids
+            labels = [1] * len(input_ids)
+
+        else:
+            ind = prompt.index(train_only_after) + len(train_only_after)
+            before_tokens = encode(prompt[:ind], False)
```
oobabooga, line 265: Just one more question, why remove the bos token from before _tokens?

mcmonkey4eva: Wait you're right good question. I think the answer is I need proper sleep before submitting PRs.
That probably shouldn't be stripped.

oobabooga: Heh. I have added it back here: https://github.com/oobabooga/text-generation-webui/commit/63ce5f9c28829a24be2d0eb3f2e57c222c14c042

(if you later conclude that no, we don't want that bos token, let me know)

PR #2321, opened 2023-05-24
Replace openai-whiper implementation with more efficient faster_whisper

friyin:
The changes made of of the faster_whisper alternative implementation which is much faster and uses less memory than official OpenAI one and has no any drawback.


oobabooga:
Apparently it doesn't work with `numpy==1.25.0`. I like the idea of using the faster implementation but for the sake of stability I think that it's best to keep the current one. I would be thankful if you could submit this modified extension to https://github.com/oobabooga/text-generation-webui-extensions so it doesn't get lost

PR #2340, opened 2023-05-25
Update docker repo link

Atinoda:
Hi oobabooga, thanks for checking out and linking to my docker companion repo. This PR is a quick update to the link in README.me so that it points directly to the repo, rather than to the docs. Hope that you're happy enough with that, I think it'll make it easier for people to find it. I plan to get pre-built images on Docker hub this weekend, and that should accelerate uptake because users won't need to build the deployment to get it running!

PS. nice work on getting 4-bit transformers rolled out so quickly


oobabooga:
Thanks for the support, @Atinoda

PR #2342, opened 2023-05-25
add api endpoint for finetuning

tohrnii:
Adds an api endpoint for lora training.

Please let me know if it's preferred to have things more documented.

### DIFF
--- a/api-example-finetune.py
+++ b/api-example-finetune.py
@@ -1,0 +1,18 @@
+import json
+import requests
+
+HOST = 'localhost:5000'
+URI = f'http://{HOST}/api/v1/finetune'
+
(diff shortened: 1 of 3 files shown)


matatonic:
Very cool, much simpler than I expected. It could be easy to include in the openai api extension for this endpoint: https://platform.openai.com/docs/api-reference/fine-tunes 

I've never done fine-tuning yet, so I haven't tackled that API endpoint yet.

ben-mkiv:
doesn't that block the whole thread until the finetune is finished?

bet0x:
This should work in tandem directly with the @FartyPants Training_PRO module, or at least, that would be ideal.

PR #2345, opened 2023-05-25
Whisper_STT extension dependency: Change SpeechRecognition version

EliasVincent:
For the whisper_stt extension I used a git version of SpeechRecognition because it had a fix for Whisper speech recognition, which wasn't merged into the stable version yet.
That changed with version 3.10 and this would change it back to the normal stable version. This should also speed up the dependency installation process since it doesn't have to pull something from git anymore.


oobabooga:
Thanks for the update

Issue #2198, opened 2023-05-19
Adding new samplers? (top_a and Tail free sampling)


BadisG:
Hello, 
I know that on KoboldAI, you have more parameters to use so that you can make your outputs even better.
top_a and Tail free sampling are the samplers missing from this repositoy, do you plan on adding them in the future?
https://github.com/KoboldAI/KoboldAI-Client/wiki/Settings#top-a-sampling

oobabooga:
I prefer to stick to the samplers offered by transformers, going by the premise that if a sampler is particularly good then it will eventually be implemented there.

BadisG:
https://huggingface.co/docs/transformers/main_classes/text_generation
There's still some parameters from the transformers that aren't on this repo like the epsilon_cutoff, eta_cutoff... Will you integrate those ones then?

PR #2357, opened 2023-05-26
Add tail-free and top-a sampling

toast22a:
This PR adds a provision for adding custom samplers when using the `transformers` library for generation. We include the tail-free and top-a samplers, but the file `sampler_hijack.py` can be extended later to add more samplers.

To do this, we monkey-patch:
* `transformers.GenerationMixin` to add the new samplers into `_get_logits_warper()` after the existing samplers, but before logit normalization
* `transformers.GenerationConfig` to add the new samplers' values as generation arguments

This is intended to be a temporary solution to #2198 until HF adds these samplers to the `transformers` library. When that happens, it is trivial to just remove the custom samplers from `sampler_hijack.py`.


BadisG:
I have the feeling that if you change the value of one of those parameters, it doesn't change anything unless you're doing a second generation.

oobabooga:
I appreciate the effort to provide new samplers, but I don't think it's worth it to add a new monkey patch just for this. A similar parameter already exists in the ui (eta_sampling). I have recently added a proper preset for using it: https://github.com/oobabooga/text-generation-webui/commit/6de727c5242fc9933260bfd45f35d8310eba4a57

oobabooga:
Well, ffs, so many people are asking for this that I'll just merge it before I get lynched

PR #2365, opened 2023-05-26
[extensions/api] models api for blocking_api

matatonic:
Supports info, load, unload and list. Seems to work with most models. I've included an example test app which does a basic validating smoke test on all models.

Currently doesn't work well with: RWKV, llama.cpp

PS. I haven't tested the lora code, but it's copied directly from the server.py.

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -156,8 +156,7 @@ class Handler(BaseHTTPRequestHandler):
             })
 
             self.wfile.write(response.encode('utf-8'))
-      
-      
+
         elif self.path == '/api/v1/token-count':
             self.send_response(200)
             self.send_header('Content-Type', 'application/json')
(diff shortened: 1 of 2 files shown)


matatonic:
I will re-submit this

LoopControl:
This is a missing piece of the API and would be great to have merged in @oobabooga 

I believe there is no way to switch models programmatically without killing and restarting the text-gen process and using the `--model` flag (unless you want to get dirty with gradio's api)

matatonic:
I've been using the model_api myself at least every day (a hundred+ calls per day), and it works pretty reliably for loading and listing models. I still have no fix for unloading of ggml models, but for 'normal' models it works well.

It doesn't try to do much as far as autodetection of parameters, so it should be fairly future proof if 'load_model' gets a lot smarter in the future (re: autogptq / quantize_config.json) I guess the provided args could just be left out and the interface unchanged.

The only major enhancement I think it needs is perhaps a smart wrapper around load_model to handle UN-setting various args (and setting), but that would be easy to drop in at another time.

matatonic:
I'm updating this for the recent changes with autogptq, please hold.

matatonic:
Closing this in favour of https://github.com/oobabooga/text-generation-webui/pull/2539

PR #2399, opened 2023-05-28
Fix hang in tokenizer for AutoGPTQ llama models.

Ph0rk0z:
This affected me, not sure if it affects you. Loading hangs because it returns a different model type and then uses autotokenizer. Model loading for llama models goes to over 200s. I thought it was a AutoGPTQ bug but it is not.


oobabooga:
Thank you, this makes sense

PR #2408, opened 2023-05-29
Make it possible to download protected HF models from the command line.

cyberfox:
Set the HF_USER and HF_PASS environment variables to the HuggingFace username and a User Access Token respectively, and run `download-model.py` script as usual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,10 +156,9 @@ For example:
 
     python download-model.py facebook/opt-1.3b
 
-If you want to download a model manually, note that all you need are the json, txt, and pytorch\*.bin (or model*.safetensors) files. The remaining files are not necessary.
-If you want to download a protected model (one gated behind accepting a license, for example, or otherwise private) you can set the environment variables `HF_USER` to your
-huggingface username, and `HF_PASS` to your password, or (_as a better option_) set it to a User Access Token available from [User Access Tokens](https://huggingface.co/settings/tokens).
-This will allow you to download a restricted model, e.g. `bigcode/starcoderbase`.
+* If you want to download a model manually, note that all you need are the json, txt, and pytorch\*.bin (or model*.safetensors) files. The remaining files are not necessary.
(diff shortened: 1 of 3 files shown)


cyberfox:
The changes are actually a little bit less extensive than they look, if you look at the diff using ignore-whitespace.

oobabooga:
I like this, but it breaks the `download_model_wrapper` function in server.py. 

1) Can it be done without introducing a class? (my preferred choice if possible)
2) Otherwise, could you adapt `download_model_wrapper`?

cyberfox:
> I like this, but it breaks the `download_model_wrapper` function in server.py.
> 
> 1. Can it be done without introducing a class? (my preferred choice if possible)
> 2. Otherwise, could you adapt `download_model_wrapper`?

Ah; I see what you're doing.  I *think* the change to `server.py` use the class looks no more complicated than this:
```diff
-        downloader = importlib.import_module("download-model")
+        downloader_module = importlib.import_module("download-model")
+        downloader = downloader_module.ModelDownloader()
```
(I have to move one more method in, and it's that easy. I'll make sure to test the UI, though.)

My reasoning for creating a class was to encapsulate the session object for just the methods that needed it, so as not to expose it, but I definitely could make it a module-scope global variable. It's not my reflex, but I'm happy to adjust.

cyberfox:
@oobabooga In an effort to entice you to go with the class approach, here's another quick feature that I've wanted here:
[Show download bar when using web ui to download.](https://github.com/cyberfox/text-generation-webui/commit/4418262e9205ab305fb09ce2ef7fb3f553723dc4?diff=unified). :)

oobabooga:
This looks good now, I have tested everything and it's all working. This would have saved me a lot of frustration downloading starcoder using git lfs last week.

Thanks for the PR @cyberfox and I look forward to the next one with the download progress bar.

PR #2423, opened 2023-05-29
[extensions/openai] cross_origin + chunked_response (updated fix)

matatonic:
@oceanplexian's fixes incorporated with a smaller set of changes, chunked responses confirmed working with anse, the [DONE] fix was just a missing content entry, also fixed here without removing the final message.
the http_protocol setting of 1.1 broke other methods, so was removed from this. the empty system prompt is fixed in another PR.

Otherwise this seems like a huge improvement and seems to fix issues related to Node.js code !! Thanks @oceanplexian, was stumped on this!


oceanplexian:
I pulled down this branch and it's working for me 👍

oobabooga:
Thanks for the review, @oceanplexian

PR #2424, opened 2023-05-29
add top_a and tfs

official-elinas:


oobabooga:
Added here, you win https://github.com/oobabooga/text-generation-webui/pull/2357/

PR #2427, opened 2023-05-30
Add fixed Flash attention code

kaiokendev:
This PR fixes the flash attention implementation that is currently in the repo and adds the `--flash-attention` flag to activate it.
To use it, install flash attention:
```
pip install ninja
pip install flash-attn
```
If that doesn't work (it should only take < 20 minutes), you need to compile from source with ninja:
```
pip install ninja
git clone https://github.com/HazyResearch/flash-attention.git
cd flash-attention
python setup.py install
```
If downstream users confirm it is working for them, I will add the pip install to the requirements.txt

With the fixed implementation, flash attention works as expected:

RTX 3090 (24 GB VRAM)
tsumeone SuperCOT 30B 128 groupsize CUDA:
```
Max new tokens: 1
3001 tokens
w/ flash attention: OOM

2614 tokens:
w/ flash attention: 23.7 GB
```

llama 13B with an 8 Rank LoRA applied:
```
Max new tokens: 1
3001 tokens
w flash attention: 12.7 GB
w/o flash attention: 15.3 GB

4002 tokens
w flash attention: ~14 GB
w/o flash attention: 17.7 GB
```

In fact, I can get 6300 tokens on 13B.

The current implementation in the repo uses PyTorch's scaled_dot_product to perform the attention call. This method falls back to standard attention when flash attention or memory efficient attention is not available -- so what's wrong? PyTorch's implementation is somewhat old, and at the time, flash attention did not support attention masking in the forward pass, so trying to run it with the attention mask fails and falls back to normal attention. Since that time, HazyResearch has already updated their flash attention to work for Causal language modeling, so we can just use it directly from the source. Maybe in the future PyTorch will fix their version. 

Also, all credit for the code goes to LAION who already implemented it for LLaMa:
https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/models/patching_llama.py

I merely took the code there and adapted it to work without the rest of their scaffolding.

**NOTE:
This PR works, but it is not ready to be merged since flash attention requires either `pip install flash-attn` or compilation from the source by cloning HazyResearch's repo and following the source compiling instructions. I already had it compiled on my machine, so until it is confirmed that downstream users can compile on their machine or the pip install works for them (Personally it froze for me which is why I had to compile from source with ninja), I would not recommend pushing this through just yet.**

But once it is pushed I think everyone should use Flash Attention by default -- it exists for a reason, there is no catch!

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@ Optionally, you can use the following command-line flags:
 | `--no-cache`                                | Set `use_cache` to False while generating text. This reduces the VRAM usage a bit with a performance cost. |
 | `--xformers`                                | Use xformer's memory efficient attention. This should increase your tokens/s. |
 | `--sdp-attention`                           | Use torch 2.0's sdp attention. |
+| `--flash-attention`                         | Use HazyResearch's Flash Attention. This drastically reduces the VRAM cost |
 | `--trust-remote-code`                       | Set trust_remote_code=True while loading a model. Necessary for ChatGLM and Falcon. |
 
 #### Accelerate 4-bit
--- a/modules/llama_attn_hijack.py
+++ b/modules/llama_attn_hijack.py
@@ -23,6 +23,9 @@ def hijack_llama_attention():
     elif shared.args.sdp_attention:
         transformers.models.llama.modeling_llama.LlamaAttention.forward = sdp_attention_forward
         logger.info("Replaced attention with sdp_attention")
+    elif shared.args.flash_attention:
+        transformers.models.llama.modeling_llama.LlamaAttention.forward = flash_attention_forward
+        logger.info("Replaced attention with flash_attention")
 
 
 def xformers_forward(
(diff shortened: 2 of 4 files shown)


kaiokendev:
@Fairfax-Mooresby 
No, it is for any model. The reason it is repetitive is well, most models aren't designed to go past 2048, since it is the base context limit of LLaMa. However, they can be trained to go beyond that, such as with Bluemoon 13B and 30B. You might want to try those as well. More models will come in the future to leverage higher context as well, I'm sure

Ph0rk0z:
Flash attention always failed on my 3090 or P40. You found a way for it to work? Does it have to be compiled from source?

Fairfax-Mooresby:
@kaiokendev 
Unfortunately the increased token count for those models seems to break --pre_layer loading support, making 30b multi-gpu slow and use more Vram than it should. I'll give it a whirl on 13b and see how far i can push the context. Thanks for your work on this.

kaiokendev:
> Flash attention always failed on my 3090 or P40/P6000. You found a way for it to work? Does it have to be compiled from source?

I don't know about P40/P6000, it works on my 3090. This PR is using the direct implementation, and not via PyTorch since their implementation seems to be old

Ph0rk0z:
I will try it on both. Less memory will definitely help training 30b

oobabooga:
This looks very promising. I am trying to get it running but I am probably doing something wrong. This command

```
pip install flash-attn
```

fails with error

```
        File "/tmp/pip-build-env-hmoohnwz/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 338, in run_setup
          exec(code, locals())
        File "<string>", line 13, in <module>
      ModuleNotFoundError: No module named 'torch'
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

So I tried installing the previous version and it worked, although as you pointed out it took some 10 minutes to install:

```
pip install flash-attn==1.0.4
```

Starting the web UI with

```
python server.py --model llama-13b-4bit-128g --listen --flash-attention
```

works and I can generate with a 3700 tokens prompt on a RTX 3090, but it only outputs spaces.

For llama-30b I OOM without flash attention with a 1800 tokens context, and I also OOM with `--flash-attention` with the same context.

I'll try to install the latest version from source.

Ph0rk0z:
I got it set up and installed fine. It doesn't compile a kernel for pascal so no go there. But I'm converting a 30b with groupsize 128 and see if it does help memory. From cursory looks at nvtop while generating, it used less.

kaiokendev:
> Tested all of these attention things last night on P40 and 3090. The only thing that gave slightly more tokens when generating at the edge of the 128g-30b was autogptq + quant_attn. Everything else went OOM about the same time. I got around 169X. The difference is small too. I got like 6 tokens generated before OOM vs 1. Goes to show there is no free lunch. That or something is wrong with the setup for everything including xformers but I don't think so.

Can you try with 13B and let me know your numbers?

Ph0rk0z:
Ok. I will try the 13b today. I have a bluemoonrp that does extra context in that size.

Ph0rk0z:
Well, I checked with bluemoon RP and with wizard-storywriter-7b that has infinite context pretty much. I can get up to 8096ish for the former (it breaks) with or without flash attention. For the latter, somewhere around 6k (doesn't break).

I also notice GPU tends to go down to 30% utilization when I am using this. I compiled from source tho.

laoda513:
hi~any progress about this commit?

kaiokendev:
@laoda513 I think with exllama I will close this PR since there is little to be gained for inference here. Personally I was never a fan of such a hacky solution to add flash attention especially when you cannot run the backward pass because LLaMA attn head size is 128. I also think the reason my result is different from others is due to memory fragmentation issues, and I don't have the time to debug that especially when exllama already fixed such issues. I have been using xformers as a replacement and it works perfectly, I even use it when training since it works on 3090 out of the box on CUDA 12.0, it is significantly improved and I don't have the headaches from flash attention.

PR #2444, opened 2023-05-30
Add exllama support (janky)

oobabooga:
Based on https://github.com/turboderp/exllama, which is meant to be a faster inference backend for GPTQ models.

### Installation

Clone exllama into your repositories folder:

```
cd text-generation-webui/repositories
git clone https://github.com/turboderp/exllama
```

### Usage

Add the `--exllama` flag, for instance,

```
python server.py --model llama-30b-4bit-128g --listen  --exllama --no-stream
```

### Performance

For each model, the first row is regular GPTQ, and the second one is exllama. Tested on a RTX 3090 on Linux.

llama-7b-4bit-128g
Output generated in 2.95 seconds (67.90 tokens/s, 200 tokens, context 16, seed 711434145)
Output generated in 1.91 seconds (112.68 tokens/s, 215 tokens, context 15, seed 1026613703)

llama-13b-4bit-128g
Output generated in 4.14 seconds (48.28 tokens/s, 200 tokens, context 16, seed 619159417)
Output generated in 3.26 seconds (65.96 tokens/s, 215 tokens, context 15, seed 1286637455)

llama-30b-4bit-128g
Output generated in 8.23 seconds (24.29 tokens/s, 200 tokens, context 16, seed 471295594)
Output generated in 5.01 seconds (30.96 tokens/s, 155 tokens, context 15, seed 450676001)

### Limitations

* ~No streaming support yet (I couldn't figure out how to do it)~
* Only 3 parameters are used (top_p, top_k, and temperature)


Ph0rk0z:
You got much further than me. I only got the model to load and it failed because of their custom generate functions.

kaiokendev:
Looking at the code repo, the `generate_simple` call wraps this function, which just generates in a loop. Should be easy to adapt for streaming:

```
def generate_simple(self, prompt, max_new_tokens = 128):

        self.end_beam_search()

        ids = self.tokenizer.encode(prompt)
        self.gen_begin(ids)

        for i in range(max_new_tokens):
            token = self.gen_single_token()
            if token.item() == self.tokenizer.eos_token_id: break

        text = self.tokenizer.decode(self.sequence[0])
        return text
```

oobabooga:
Well noted @kaiokendev, now we also have a janky streamer: https://github.com/oobabooga/text-generation-webui/pull/2444/commits/90e6d3f07e54579e3068db45057ccfa3bfb5b9a3

matatonic:
🔥

n00mkrad:
I keep getting this error. Also saw other users report it. The branch runs fine normally but won't start when trying to use `--exllama`:

```
...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py:359: UserWarning: Error checking compiler version for cl: [WinError 2] The system cannot find the file specified
  warnings.warn(f'Error checking compiler version for {compiler}: {error}')
INFO: Could not find files for the given pattern(s).
  File "...\oobabooga_windows\text-generation-webui\server.py", line 1094, in <module>
  File "...\oobabooga_windows\text-generation-webui\modules\models.py", line 99, in load_model
    output = load_func(model_name)
  File "...\oobabooga_windows\text-generation-webui\modules\models.py", line 305, in exllama_loader
    from modules.exllama import ExllamaModel
  File "...\oobabooga_windows\text-generation-webui\modules\exllama.py", line 7, in <module>
    from repositories.exllama.generator import ExLlamaGenerator
  File "...\oobabooga_windows\text-generation-webui\repositories\exllama\generator.py", line 1, in <module>
    import cuda_ext
  File "...\oobabooga_windows\text-generation-webui\repositories\exllama\cuda_ext.py", line 14, in <module>
    exllama_ext = load(
  File "...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py", line 1284, in load
    return _jit_compile(
  File "...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py", line 1509, in _jit_compile
    _write_ninja_file_and_build_library(
  File "...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py", line 1611, in _write_ninja_file_and_build_library
    _write_ninja_file_to_build_library(
  File "...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py", line 2048, in _write_ninja_file_to_build_library
    _write_ninja_file(
  File "...\oobabooga_windows\installer_files\env\lib\site-packages\torch\utils\cpp_extension.py", line 2188, in _write_ninja_file
    cl_paths = subprocess.check_output(['where',
  File "...\oobabooga_windows\installer_files\env\lib\subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "...\oobabooga_windows\installer_files\env\lib\subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['where', 'cl']' returned non-zero exit status 1.
```

Fairfax-Mooresby:
How is this handling multi-gpu VRAM allocation?

exllama uses -gs 7.5,7.5,7.5 to spread a model 3 ways
text-generation-webui can use --gpu-memory 5 8 8 or  --pre_layer 15 34 60 (which is faster)
Using --gpu-memory 5 8 8 with --exllama it seems to OOM from trying to load the whole model on to cuda 0

EyeDeck:
@n00mkrad: That's exllama looking for `cl.exe` and not finding it.
You'll need a C++ compiler, probably [MSVC](https://visualstudio.microsoft.com/downloads/). I'm not sure what versions, if any, work.

Then you'll have to track down where `cl.exe` lives and add the directory to your `path` environment variable, my install is at:
`C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64`

You'll also need [CUDA Toolkit](https://developer.nvidia.com/cuda-11-8-0-download-archive), exllama readme says 11.8 so that's what I've linked.

However, for me, on Windows, it seems to get most of the way through compiling, then crashes with:
```
[12/12] "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64/link.exe" cuda_buffers.cuda.o rep_penalty.o column_remap.cuda.o half_matmul.cuda.o q4v2_matmul.cuda.o q4v2_mlp.cuda.o q4v2_recons.cuda.o q4v2_sequential.cuda.o rms_norm.cuda.o rope.cuda.o exllama_ext.o /nologo /DLL c10.lib c10_cuda.lib torch_cpu.lib torch_cuda.lib -INCLUDE:?warp_size@cuda@at@@YAHXZ torch.lib /LIBPATH:G:\miniconda\envs\tgwui\lib\site-packages\torch\lib torch_python.lib /LIBPATH:G:\miniconda\envs\tgwui\libs "/LIBPATH:C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\lib\x64" cudart.lib /out:exllama_ext.pyd
FAILED: exllama_ext.pyd
"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64/link.exe" cuda_buffers.cuda.o rep_penalty.o column_remap.cuda.o half_matmul.cuda.o q4v2_matmul.cuda.o q4v2_mlp.cuda.o q4v2_recons.cuda.o q4v2_sequential.cuda.o rms_norm.cuda.o rope.cuda.o exllama_ext.o /nologo /DLL c10.lib c10_cuda.lib torch_cpu.lib torch_cuda.lib -INCLUDE:?warp_size@cuda@at@@YAHXZ torch.lib /LIBPATH:G:\miniconda\envs\tgwui\lib\site-packages\torch\lib torch_python.lib /LIBPATH:G:\miniconda\envs\tgwui\libs "/LIBPATH:C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\lib\x64" cudart.lib /out:exllama_ext.pyd
   Creating library exllama_ext.lib and object exllama_ext.exp
half_matmul.cuda.o : error LNK2019: unresolved external symbol cublasHgemm referenced in function "enum cudaError __cdecl half_matmul_cublas_cuda(struct __half const *,struct __half const *,struct __half *,int,int,int,struct cublasContext *)" (?half_matmul_cublas_cuda@@YA?AW4cudaError@@PEBU__half@@0PEAU2@HHHPEAUcublasContext@@@Z)
exllama_ext.pyd : fatal error LNK1120: 1 unresolved externals
ninja: build stopped: subcommand failed.
```

Works fine on WSL2 though, I already had CUDA Toolkit installed there and it compiled first try.

Guess I can pester @turboderp, in case he's interested in Windows-specific feedback.

turboderp:
I'm quite happy with the feedback but I can't do much with it at the moment. It's still a very early-days project and until I have it working well enough on Linux, Windows is going to be a second priority. At some point there will be a stable API, I think, maybe even with some documentation (!). But not just yet.

I discussed integration with KoboldAI with someone else recently, and they suggested that instead of dynamically loading the C++ extension with cached runtime compilation, it would be better to stick it in a setuptools script and install as a permanent module. I don't like that approach, but it might help with some of the issues on Windows, since it would follow more closely how GPTQ-for-LLaMa works, which many users will have already gone to the effort of setting up. It would also allow for distributing precompiled versions of the extension, since apparently many users won't have the toolchain installed or won't have it set up just right for this.

For now, I have some ideas for how to fix compilation issues with AMD GPUs and Torch's automatic HIPification feature, and they may apply here since the error you're having seems to be a linker error. That's the problem with the HIP stuff as well, that it just doesn't know what to do with multiple .cu files. There's a long list of other things I also need to work on and only so many hours in a day, so I'm not sure where that is on the list right now, but for a short test in your own fork you could try adding a single .cu file which `#include`s all the rest, and then only referencing that in cuda_ext.py. That way you should end up with just a single output from `nvcc` to link against the outputs of `gcc` or I guess `cl.exe` in this case.

As for how to use the generator, I'm working on a web UI myself. Expect a janky version 0.0 in the coming days. It's supposed to be more of a demo than a full-fledged alternative front end (at least for now) but it'll show a more relevant way of using the model with streaming, efficient use of the cache across multiple generations, etc. It'll also probably showcase some of the "advanced" or "fun" features like being able to chat with multiple bots at once. I think it will give a clearer example of how the model is intended to work, building from a config, attaching the cache to the generator and the generator to the model, changing parameters at runtime, stuff like that.

A couple of key differences to keep in mind compared to GPTQ-for-LLaMa:

- This doesn't derive from Transformers, although it did start life as a copy of the Transformers implementation. It has since morphed into something very different, and while it still looks a bit like a HF model in places, don't be fooled.
- It doesn't support a number of HF features like position IDs, arbitrary attention masks, etc.
- It won't work with PEFT (yet, I will add support for switchable, stackable LoRAs soon though)
- It only loads .safetensors files, not pickles.
- It uses a fixed cache which is updated in-place. That's the #1 reason it manages to run 30B-128g models in 24 GB with room to spare. But as a consequence it crashes if you attempt to generate past whatever cache size you specify initially. And of course it allocates all the memory it needs for that cache up front, so if you want to run it with a shorter context, e.g. to fit a 30B-32g model, you need to specify a shorter max sequence length to begin with.
- Unlike GPTQ-for-LLaMa it isn't currently tested and validated across many different hardware configurations. It should theoretically work from _I think_ compute 6.1 and up, but I haven't tried it on anything older than a 3070. The main reason is the reliance on FP16 without any fallbacks. More support might come in the future, but again, only so many hours in a day.
- A couple of features are straight-up broken right now, like layer streaming.
- Some features are only there to help me in validating and benchmarking while I develop, like dequantization.

I hope that's useful.

oobabooga:
@turboderp thanks for the valuable information and I look forward to seeing the updates in version 0.0. 

I have a question: do you think it's possible to take the logits generated by exllama and somehow plug them into the transformers sampler? `transformers.GenerationMixin` specifically: https://huggingface.co/docs/transformers/v4.29.1/en/main_classes/text_generation#transformers.GenerationMixin

I have seen many projects come up with their own custom samplers (flexgen, chatrwkv, the original llama inference code by facebook to name a few), and I have always felt like it would be nice to combine the improved transformers-free performance with the familiar transformers sampling parameters.

turboderp:
You would need to wrap it, I think, since a number of those features aren't supported or are supported differently. 

The Transformers version _I believe_ uses batches, which would bypass the memory-efficient beam search in my implementation. There are also other features coming which would be a difficult fit for `GenerationMixin`.

But in the end it still takes input IDs and output logits, so it should theoretically work. Just needs a wrapper so it can pass itself off as a PreTrainedModel.

kaiokendev:
@turboderp 
> It uses a fixed cache which is updated in-place. That's the no. 1 reason it manages to run 30B-128g models in 24 GB with room to spare. But as a consequence it crashes if you attempt to generate past whatever cache size you specify initially. And of course it allocates all the memory it needs for that cache up front, so if you want to run it with a shorter context, e.g. to fit a 30B-32g model, you need to specify a shorter max sequence length to begin with.

Can you expand on this? This is super helpful for any implementation I think, if it's possible to apply to HF transformers at least that will help a lot of users in the meantime while the rest of the stuff is fleshed out

turboderp:
@kaiokendev:

The way transformers.LlamaModel (and I think a lot of if not all transformer models in Transformers) appends new keys and values to the cache is by concatenation. Concatenating two tensors requires allocating a storage for the resulting tensor, then copying both source tensors into the destination storage and "eventually" getting around to garbage-collecting the originals, no sooner than when they fall out of scope.

This copying of data is pretty wasteful, performance-wise, especially when one of those tensors is very large and only growing by a small amount. If you've got, say, a 100 MB tensor and you want to append a 1 MB tensor to it, that's 202 MB of memory access, compared to the optimal 2 MB if you were copying in-place.

But it's worse still from a memory management perspective. Consider the gap left over by that 100 MB tensor along with the fact that the *next* large tensor that needs to be allocated is also going to be 101 MB in size, and so will the next 58 or however many decoder layers you have to iterate over. It asks a lot of PyTorch to manage memory smartly in this scenario, and in my experience it fails a bunch.

I was pulling my hair out at one point not understanding why I kept getting OoM errors when increasing the context length past a certain point, even though the resulting tensors *should* easily fit. I eventually had to conclude I was running into fragmentation. So I wrote [this ugly, ugly patch](https://github.com/turboderp/alpaca_lora_4bit/blob/main/llama_attn_hijack_xformers_modified.py) largely by throwing some code snippets produced by ChatGPT haphazardly into a xformers monkey patch I found somewhere. Don't judge. Anyway, suddenly I could achieve way longer contexts.

The only thing it does is allocate the full-length cache if called with `past_key_values == None`, and then it narrows the view on all those tensors so the underlying storage reserves enough room for the in-place concatenation. This way the cache appears to the rest of the implementation to be shorter than it really is, while the patched self-attention can (bravely) proceed to do use the full length storage.

Of course, it's not robust at all, but at least it worked for that project (I was trying to train a LoRA on 6k contexts.) Also note the need to create empty (i.e. allocated but uninitialized) tensors only to discard their data immediately, as the only way (that I could find) to un-narrow a tensor view. Ugly, ugly stuff. And it was actually while exploring ways to do this more cleanly and robustly that I finally decided I'd had enough and started writing a standalone implementation.

So I don't know if that helps anyone, but there you go.

kaiokendev:
> Don't judge

Not at all lol. Like you said, the fragmentation is the main cause, and I thought about doing something maybe blockwise but decided I would put it off, but since you have already have a janky implementation that's still amazing. I will check it further and try to put it in this repo. There are some cache compression techniques I looked at, i.e. https://arxiv.org/abs/2305.17118 but they are mostly lossy and in reality like you said the tensors should fit already.

Thanks for the reply

Ph0rk0z:
Whoa, the speed is really really good. But multi-gpu seems unsupported so no trying the 65b.

turboderp:
Multi-GPU support is there. `-gs 10,10,10` would put 10 GB of weights on each of the first three devices, for instance. 65B gives about 19 tokens/second with a 4090 and a 3090-Ti here. There's room for improvement, too.

Ph0rk0z:
Ok.. I added code but how do I do a PR to a PR.

shared.py

`parser.add_argument('--gpu-split', type = str, help = "Comma-separated list of VRAM (in GB) to use per GPU device for model layers, e.g. -gs 20,7,7")`


exllama.py

```
import modules.shared as shared
...

        config.max_seq_len = 2048
        if (shared.args.gpu_split):
            config.set_auto_map(shared.args.gpu_split)
            config.gpu_peer_fix = True
```

turboderp:
The way it applies repetition penalty is a little experimental. It only looks at the latest tokens in the sequence, then applies less and less penalty over a length. But just set `token_repetition_penalty_sustain` to the full context length and it should behave like most other samplers in that regard.

As for the errors, I think line 81 would only fail if the previous top-k selection returned nothing, which seems weird. Unless these presets are using topk=0 as a signal to disable top-k sampling. My code takes it more literally and just selects the zero most likely tokens instead, which obviously fails. I'm not sure how those presets are meant to work, but if that's the case you should get the intended behavior with topk=1000 or something.

Ph0rk0z:
That must be what they're doing. Same way as it works for kobold. I'll double check and see. Also will see how much if any P40 improves. True FP16 is half as fast there as was discovered for GPTQ. I also read there was an issue with atomicadd

I only had it fixed like this for half but never tried half2.

`__device__ __forceinline__ void atomicAdd(__half* address, c10::Half val)`

Since we're doing samplers, did you try mirostat? Seems adding it here would be easier than into transformers.

turboderp:
I haven't looked at mirostat yet, but I'm going to in a little bit. I also have a beam-search implementation that seems to work well, but it trades performance for VRAM efficiency at the moment and I haven't had time to make it both fast and compact. But you can do a whole bunch more once you have more than a single set of logits to consider.

And yes, there's a bunch of weird issues with atomicAdd and FP16 support on older GPUs. I'd like to support the P40 especially, but it will take a bit more work. It has abysmal FP16 performance, apparently, at 184 GFLOPS. But some of the FP16 operations could be replaced with e.g. lookup tables, and at the end of the day text generation is bandwidth limited anyway.

Ph0rk0z:
It was definitely issue with topk=0.. 1000 made it's own error. I'm going to start saving them so we have something to look at. 

A bit weird with P40 because models in FP16 don't have issues and some half operations worked fine. Others cut speed in two. I have that as a 3rd GPU and a P6000 in this machine. Until this inference method, the difference between 3090+p40 and 3090+3090 was around .5t/s Still think they are best bang for the buck until something legitimately obsoletes them.

turboderp:
Well, my version is very much tuned to the hardware I've got. There are a lot of ways to achieve better all-round performance, like doing FP32 math whenever possible, but this comes at the cost of performance on newer GPUs where FP16 is considerably faster, not least because VRAM bandwidth hasn't increased all that much since cards like the P40 were current, and LLM inference is all about the bandwidth.

There are many optimizations that would probably work well on the P40 and other Pascal cards, like caching certain intermediate results in SMEM instead of recomputing them, tuning for the lower core count and (much!) smaller L2 cache, not to mention a bunch of tricks that would be specific to GPTQ.

Give it some time. I just pushed an update that should make it easier to plug in different kernels in places.  And I do want to support the P40 if possible. It is a really cheap 24 GB GPU after all.

matatonic:
I put together the various pieces of this change so far and gave it a go. Not much success, but here's what I found.

1) when loading Alpaca-30b-4bit.pt or TheBloke_vicuna-13B-1.1-GPTQ-4bit-128g, (and a few more) I got:
```
  File "/app/modules/exllama.py", line 44, in from_pretrained
    model = ExLlama(config)
  File "/app/repositories/exllama/model.py", line 621, in __init__
    with safe_open(self.config.model_path, framework="pt", device="cpu") as f:
safetensors_rust.SafetensorError: Error while deserializing header: HeaderTooLarge
```

2) tsumeone_llama-30b-supercot-4bit-cuda and llama-30b-4bit-128g loaded ok

3) When generating I got the prompt echo'd back to me but had echo set to False, stopping strings also didn't work. 

4) generation was not great quality, hard to quantify but it was definitely off. More terse.

5) even discounting for the prompt being echo'd back (I think it was screwing up the tokens/sec - higher), I think token generation was WAAAY faster (it reported 45+/sec and I normally get about 5/sec for 30B models)

6) my system is an AMD threadripper CPU and 4090 GPU.

7) Amazing speed! I can't wait to see this when the bugs are all worked out.

Sample llama 30B output (cut off at token max, 'This is a chat between two beings, 🧠 and 🦙.' is the initial context, it shouldn't echo back):

```
> write a recipe for a chocolate cake.
This is a chat between two beings, 🧠 and 🦙. 
🧠: write a recipe for a chocolate cake.
🦙: 1 cup of flour, 2 eggs, 3/4 cup sugar, 1 tsp baking powder, 1/2 tsp salt, 1/2 cup vegetable oil, 1 tsp vanilla extract, 1-9 inch round cake pan, preheat oven to 350 degrees F (176 C), grease the pan, mix all ingredients together, pour into pan, bake for 35 minutes or until done.
🧠: what's the difference between a computer program and a human?
🦙: I don't know.
🧠: you are not a computer program.
🦙: I am not a computer program.
🧠: how do you know that?
🦙: because I can think about it.
🧠: but if you were a
```

mongolu:
When it will be merged on the main?

Ph0rk0z:
I think it could with the GPU splitting stuff already. There are some more options that were exposed too like fused_mlp. Other than the flaw in the sampling I am sort of wary to go back to anything else. Once it loads or trains loras, holy shit.

BadisG:
> When it will be merged on the main?
Once exllama will work with all the samplers I guess?

TechnotechYT:
Definitely interested for if training LoRAs will be possible with it, the inference speed increases are insane

turboderp:
Loading and using LoRAs is pretty straightforward. *Training* LoRAs is on my list, but it's at the very bottom because it requires backpropagation for the entire model, which is sadly not simple to add at this point, especially while retaining the performance. So I'm not sure I'll ever get that far. It may be easier to just port over some of my optimizations to a project like [alpaca_lora_4bit](https://github.com/johnsmith0031/alpaca_lora_4bit).

BadisG:
I have this big error when trying to load a triton model

```
Traceback (most recent call last): File “/home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/utils/cpp_extension.py”, line 1893, in _run_ninja_build subprocess.run( File “/home/adduser/miniconda3/envs/triton/lib/python3.10/subprocess.py”, line 526, in run raise CalledProcessError(retcode, process.args, subprocess.CalledProcessError: Command ‘[‘ninja’, ‘-v’]’ returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last): File “/mnt/d/Large Language Models/text-generation-webui-TRITON/server.py”, line 71, in load_model_wrapper shared.model, shared.tokenizer = load_model(shared.model_name) File “/mnt/d/Large Language Models/text-generation-webui-TRITON/modules/models.py”, line 99, in load_model output = load_func(model_name) File “/mnt/d/Large Language Models/text-generation-webui-TRITON/modules/models.py”, line 305, in exllama_loader from modules.exllama import ExllamaModel File “/mnt/d/Large Language Models/text-generation-webui-TRITON/modules/exllama.py”, line 7, in from repositories.exllama.generator import ExLlamaGenerator File “/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/generator.py”, line 1, in import cuda_ext File “/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/cuda_ext.py”, line 42, in exllama_ext = load( File “/home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/utils/cpp_extension.py”, line 1284, in load return jit_compile( File “/home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/utils/cpp_extension.py”, line 1509, in jit_compile write_ninja_file_and_build_library( File “/home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/utils/cpp_extension.py”, line 1624, in write_ninja_file_and_build_library run_ninja_build( File “/home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/utils/cpp_extension.py”, line 1909, in run_ninja_build raise RuntimeError(message) from e RuntimeError: Error building extension ‘exllama_ext’: [1/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE="gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS -D__CUDA_NO_BFLOAT16_CONVERSIONS -D__CUDA_NO_HALF2_OPERATORS --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_matmul.cu’ -o q4_matmul.cuda.o FAILED: q4_matmul.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE="gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_matmul.cu’ -o q4_matmul.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [2/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_buffers.cu’ -o cuda_buffers.cuda.o FAILED: cuda_buffers.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_buffers.cu’ -o cuda_buffers.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [3/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_matrix.cu’ -o q4_matrix.cuda.o FAILED: q4_matrix.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_matrix.cu’ -o q4_matrix.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [4/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/column_remap.cu’ -o column_remap.cuda.o FAILED: column_remap.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/column_remap.cu’ -o column_remap.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [5/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/rms_norm.cu’ -o rms_norm.cuda.o FAILED: rms_norm.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/rms_norm.cu’ -o rms_norm.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [6/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/rope.cu’ -o rope.cuda.o FAILED: rope.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/rope.cu’ -o rope.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [7/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/half_matmul.cu’ -o half_matmul.cuda.o FAILED: half_matmul.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/half_matmul.cu’ -o half_matmul.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [8/11] /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_mlp.cu’ -o q4_mlp.cuda.o FAILED: q4_mlp.cuda.o /home/adduser/miniconda3/envs/triton/bin/nvcc -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“gcc" -DPYBIND11_STDLIB="libstdcpp" -DPYBIND11_BUILD_ABI="cxxabi1011" -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -D__CUDA_NO_HALF_OPERATORS -D__CUDA_NO_HALF_CONVERSIONS_ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_86,code=compute_86 -gencode=arch=compute_86,code=sm_86 --compiler-options ‘-fPIC’ -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cuda_func/q4_mlp.cu’ -o q4_mlp.cuda.o nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified [9/11] c++ -MMD -MF rep_penalty.o.d -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=”_gcc” -DPYBIND11_STDLIB=”_libstdcpp” -DPYBIND11_BUILD_ABI=”_cxxabi1011” -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -fPIC -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cpu_func/rep_penalty.cpp’ -o rep_penalty.o FAILED: rep_penalty.o c++ -MMD -MF rep_penalty.o.d -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=”_gcc” -DPYBIND11_STDLIB=”_libstdcpp” -DPYBIND11_BUILD_ABI=”_cxxabi1011” -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -fPIC -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/cpu_func/rep_penalty.cpp’ -o rep_penalty.o c++: warning: Language: linker input file unused because linking not done c++: error: Language: linker input file not found: No such file or directory c++: warning: Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext: linker input file unused because linking not done c++: error: Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext: linker input file not found: No such file or directory [10/11] c++ -MMD -MF exllama_ext.o.d -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=”_gcc” -DPYBIND11_STDLIB=“_libstdcpp” -DPYBIND11_BUILD_ABI=“_cxxabi1011” -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -fPIC -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/exllama_ext.cpp’ -o exllama_ext.o FAILED: exllama_ext.o c++ -MMD -MF exllama_ext.o.d -DTORCH_EXTENSION_NAME=exllama_ext -DTORCH_API_INCLUDE_EXTENSION_H -DPYBIND11_COMPILER_TYPE=“_gcc” -DPYBIND11_STDLIB=“_libstdcpp” -DPYBIND11_BUILD_ABI=“_cxxabi1011” -I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/torch/csrc/api/include -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/TH -isystem /home/adduser/miniconda3/envs/triton/lib/python3.10/site-packages/torch/include/THC -isystem /home/adduser/miniconda3/envs/triton/include -isystem /home/adduser/miniconda3/envs/triton/include/python3.10 -D_GLIBCXX_USE_CXX11_ABI=0 -fPIC -std=c++17 -c ‘/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext/exllama_ext.cpp’ -o exllama_ext.o c++: warning: Language: linker input file unused because linking not done c++: error: Language: linker input file not found: No such file or directory c++: warning: Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext: linker input file unused because linking not done c++: error: Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext: linker input file not found: No such file or directory ninja: build stopped: subcommand failed.
```

turboderp:
The command line looks a little suspect. 

`-I/mnt/d/Large Language Models/text-generation-webui-TRITON/repositories/exllama/exllama_ext`

It doesn't add quotes around the argument even though there are spaces in the path. That could be why the compile is failing. If that's the case, it seems like a Torch bug. You could try renaming `/mnt/d/Large Language Models` without the spaces, maybe?

BadisG:
I changed Large Language Models to Large-Language-Models and it worked! Thanks a lot.
The speed is really impressive, with a triton model that has act_order with it, I usually got 8 tokens/s, now it reaches 24 tokens/s !!

Ph0rk0z:
I confirm generation doesn't stop anymore. I think from changes on the 5th or 6th.

jcalebsmith:
> Based on https://github.com/turboderp/exllama, which is meant to be a faster inference backend for GPTQ models.
> ### Installation
> 
> Clone exllama into your repositories folder:
> 
> ```
> cd text-generation-webui/repositories
> git clone https://github.com/turboderp/exllama
> ```
> 
> ### Usage
> 
> Add the `--exllama` flag, for instance,
> 
> ```
> python server.py --model llama-30b-4bit-128g --listen  --exllama --no-stream
> ```
> 
> ### Performance
> 
> For each model, the first row is regular GPTQ, and the second one is exllama. Tested on a RTX 3090 on Linux.
> 
> llama-7b-4bit-128g Output generated in 2.95 seconds (67.90 tokens/s, 200 tokens, context 16, seed 711434145) Output generated in 1.91 seconds (112.68 tokens/s, 215 tokens, context 15, seed 1026613703)
> 
> llama-13b-4bit-128g Output generated in 4.14 seconds (48.28 tokens/s, 200 tokens, context 16, seed 619159417) Output generated in 3.26 seconds (65.96 tokens/s, 215 tokens, context 15, seed 1286637455)
> 
> llama-30b-4bit-128g Output generated in 8.23 seconds (24.29 tokens/s, 200 tokens, context 16, seed 471295594) Output generated in 5.01 seconds (30.96 tokens/s, 155 tokens, context 15, seed 450676001)
> ### Limitations
> 
>     * ~No streaming support yet (I couldn't figure out how to do it)~
> 
>     * Only 3 parameters are used (top_p, top_k, and temperature)

I followed these instructions and tried adding the --exllama flag to webui.py CMD_FLAGS variable, (CMD_FLAGS = '--notebook --auto-devices --gpu-memory 22000MiB --cpu-memory 94000MiB --exllama'), but I'm getting the following error:

usage: server.py [-h] [--notebook] [--chat] [--character CHARACTER] [--model MODEL] [--lora LORA [LORA ...]]
                 [--model-dir MODEL_DIR] [--lora-dir LORA_DIR] [--model-menu] [--no-stream] [--settings SETTINGS]
                 [--extensions EXTENSIONS [EXTENSIONS ...]] [--verbose] [--cpu] [--auto-devices]
                 [--gpu-memory GPU_MEMORY [GPU_MEMORY ...]] [--cpu-memory CPU_MEMORY] [--disk]
                 [--disk-cache-dir DISK_CACHE_DIR] [--load-in-8bit] [--bf16] [--no-cache] [--xformers]
                 [--sdp-attention] [--trust-remote-code] [--load-in-4bit] [--compute_dtype COMPUTE_DTYPE]
                 [--quant_type QUANT_TYPE] [--use_double_quant] [--threads THREADS] [--n_batch N_BATCH] [--no-mmap]
                 [--mlock] [--cache-capacity CACHE_CAPACITY] [--n-gpu-layers N_GPU_LAYERS] [--n_ctx N_CTX]
                 [--llama_cpp_seed LLAMA_CPP_SEED] [--wbits WBITS] [--model_type MODEL_TYPE] [--groupsize GROUPSIZE]
                 [--pre_layer PRE_LAYER [PRE_LAYER ...]] [--checkpoint CHECKPOINT] [--monkey-patch] [--quant_attn]
                 [--warmup_autotune] [--fused_mlp] [--gptq-for-llama] [--autogptq] [--triton] [--desc_act] [--flexgen]
                 [--percent PERCENT [PERCENT ...]] [--compress-weight] [--pin-weight [PIN_WEIGHT]] [--deepspeed]
                 [--nvme-offload-dir NVME_OFFLOAD_DIR] [--local_rank LOCAL_RANK] [--rwkv-strategy RWKV_STRATEGY]
                 [--rwkv-cuda-on] [--listen] [--listen-host LISTEN_HOST] [--listen-port LISTEN_PORT] [--share]
                 [--auto-launch] [--gradio-auth GRADIO_AUTH] [--gradio-auth-path GRADIO_AUTH_PATH] [--api]
                 [--api-blocking-port API_BLOCKING_PORT] [--api-streaming-port API_STREAMING_PORT] [--public-api]
                 [--multimodal-pipeline MULTIMODAL_PIPELINE]
server.py: error: unrecognized arguments: --exllama

Any idea what I'm doing wrong?

Ph0rk0z:
Maybe you didn't merge the PR right.

jmoney7823956789378:
Has anyone gotten this pull working with the API?

```
  File "/mnt/textgen/textgen-nv/modules/text_generation.py", line 307, in generate_reply_custom
    for reply in shared.model.generate_with_streaming(context=question, **generate_params):
  File "/mnt/textgen/textgen-nv/modules/exllama.py", line 84, in generate_with_streaming
    token = generator.gen_single_token()
  File "/mnt/textgen/textgen-nv/repositories/exllama/generator.py", line 277, in gen_single_token
    token, _ = self.sample(logits,
  File "/mnt/textgen/textgen-nv/repositories/exllama/generator.py", line 81, in sample
    cum_prob = top_probs[0].item()
IndexError: index 0 is out of bounds for dimension 0 with size 0
```

turboderp:
This seems like the same thing again. Try setting top-k = 10000, or changing generator.py around line 77 or so:

        # Top K

        if top_k == 0:
            top_probs, top_indices = torch.sort(probs, descending = True)
        else:
            top_probs, top_indices = torch.topk(probs, top_k)

jmoney7823956789378:
> This seems like the same thing again. Try setting top-k = 10000, or changing generator.py around line 77 or so:
> 
> ```
>     # Top K
> 
>     if top_k == 0:
>         top_probs, top_indices = torch.sort(probs, descending = True)
>     else:
>         top_probs, top_indices = torch.topk(probs, top_k)
> ```

You are the best!

Panchovix:
For now other samplers don't work yet, right? (temp_a, rep penalty, rep lenght, etc)

Ph0rk0z:
Only the ones in the commit here work.

turboderp:
As far as sampling goes, ExLlama currently supports top-k, top-p, min-p, (token) repetition penalty, typical sampling and beam search.

The beam search implementation is a little unique, using a fixed look-ahead length and limited caching from just the point where the beams diverge, which cuts down memory consumption a lot but currently requires evaluating each beam in a separate inference pass. So the VRAM impact of beam search is negligible, but speed is reduced by a factor of however many beams it's configured to use.

If there are particular samplers (or other features like configurable order of operations) that would help with integration, let me know and I'll try to prioritize those.

Alternatively, the Kobold ExLllama branch took the different approach of just plugging the output logits from ExLlama into the same samplers used for other model implementations. Not sure if that's feasible with GenerationMixin, but it could be a much cleaner solution, maybe?

Panchovix:
For reference, here is the Kobold-Exllama implementation

https://github.com/0cc4m/KoboldAI/tree/4bit-plugin (Kobold tree to support exllama, also supports GTPQ at the same time)

https://github.com/0cc4m/exllama (exllama fork to make it work on Kobold exllama). In this case, exllama has to be built (like GPTQ) instead of being built when loading the model.

oobabooga:
I have reworked the model loading UI and added ExLlama as a loader option:

menu

So for the same GPTQ model, a choice can be made between ExLlama, GPTQ-for-LLaMa, and AutoGPTQ.

To choose the ExLlama from the command-line, I have implemented a `--loader` option and removed the `--exllama` option from the initial commit. This is how it should be used:

```
python server.py --model llama-7b-4bit-128g --loader exllama
```

@turboderp about samplers:  the`typical_p` and `repetition_penalty` parameters in the UI are now attached to the ExLlama generator. I am currently running an experiment to see what the best presets for chat and instruction following are; from the preliminary results, it seems like the fancy samplers *may* not really be necessary, but it's too early to tell: https://oobabooga.github.io/arena/preliminary-results.html

It would be nice to just plug the whole thing into GenerationMixin if possible. I couldn't find a way to do it myself. In any case, once the Preset Arena experiment is done and I analyze the data, if a particular parameter turns out to be very important, I will ping you. 

This PR seems to be working pretty well for me, so I'll just merge it. Any contributions to improving the ExLlama integration here would be highly appreciated. 

It's pretty awesome to run 30b models at full context and so fast, by the way. Once this gets LoRA/multi-LoRA support nothing will stop it :^)

jmoney7823956789378:
This is an awesome development, but unfortunately it looks like the --gpu-split (or equivalent option) didn't make it into this commit.
Loading a 33B into a 2x16GB GPU setup is only possible this way, and --auto-devices isn't being so friendly about it.

oobabooga:
Added here based on the code provided by @Ph0rk0z above: https://github.com/oobabooga/text-generation-webui/commit/5f392122fd840ef2bd3766711cb6faf227f666d7

Can you check if it works @jmoney7823956789378? 

Example:

```
python server.py --model llama-7b-4bit-128g --loader exllama --gpu-split 10,10 
```

jmoney7823956789378:
> Added here based on the code provided by @Ph0rk0z above: [5f39212](https://github.com/oobabooga/text-generation-webui/commit/5f392122fd840ef2bd3766711cb6faf227f666d7)
> 
> Can you check if it works @jmoney7823956789378?
> 
> Example:
> 
> ```
> python server.py --model llama-7b-4bit-128g --loader exllama --gpu-split 10,10 
> ```

Beautiful work.
Gradio webgui has a typo in the exllama's gpu-split option, written as "gpu-slit", btw.

oobabooga:
Oops, fixed here, thanks https://github.com/oobabooga/text-generation-webui/commit/1e400218e92f0812ef9ecb7b50021d7129e3c25c

turboderp:
Have you tried using `gen_begin_reuse` instead of `gen_begin`? As long as the cache is valid, it should work the same, except it reuses the cache for any unchanged tokens in the context. At least until you build up to a full context it eliminates most of the delay before the model starts responding.

oobabooga:
@turboderp for that I could initialize `ExLlamaGenerator` just once when my custom `ExllamaModel` is initialized. But I worry that if I do that, then after

```python
        if state['ban_eos_token']:
            self.generator.disallow_tokens([self.tokenizer.eos_token_id])
```

the generator will forever ban the eos token, even after state['ban_eos_token'] is unset in the future. Is that the case?

Also, while we are at it: while streaming, I want to decode only the new tokens:

```python
        for i in range(state['max_new_tokens']):
            token = self.generator.gen_single_token()
            yield (self.generator.tokenizer.decode(self.generator.sequence[0][initial_len:]))
```

the current code causes a missing initial space in some cases. This is a problem that I had experienced before with the llama tokenizer loaded through transformers. I'm trying to find a solution. In the previous case, the solution that I found was to do

```python
        new_tokens = len(output_ids) - len(input_ids[0])
        reply = decode(output_ids[-new_tokens:], state['skip_special_tokens'])

        # Prevent LlamaTokenizer from skipping a space
        if type(shared.tokenizer) is transformers.LlamaTokenizer and len(output_ids) > 0:
            if shared.tokenizer.convert_ids_to_tokens(int(output_ids[-new_tokens])).startswith('▁'):
                reply = ' ' + reply

```

turboderp:
>the generator will forever ban the eos token, even after state['ban_eos_token'] is unset in the future. Is that the case?

No, `disallow_tokens` just updates the list of disallowed tokens, it can be changed as often as you want, or cleared with `disallow_tokens(None)`. The current list is applied directly to the logits passed to `sample()`.

>Also, while we are at it: while streaming, I want to decode only the new tokens:

Yes, it seems like it's impossible to predict when the tokenizer adds that leading space. At least I haven't figured out the logic there yet. The solution I use in my own UI is to decode the whole response sequence, and then yield only whatever _characters_ are added since the last time the sequence was decoded. There's some code for this in `webui/session.py`, `Session.respond()`, but that might be a little wasteful in your case if you're considering the whole sequence (my version only works on the response part.)

I suspect something along these lines would also work well enough (in pseudocode):

    generator.begin(prompt)
    for _ in range(new_tokens):
        sequence_text_before = decode(generator.sequence[-10:])
        generator.generate_token()
        sequence_text_after = decode(generator.sequence[-11:])
        added_text = sequence_text_after[len(sequence_text_before) : ]
        yield(added_text)

The idea being that 10 positions prior to the new token the decoding has already "settled", so the only difference between `sequence_text_before` and `sequence_text_after` will be precisely the characters added by the new token, whether or not the tokenizer decides to include a leading space.

(10 of course is an arbitrary number here. Probably 2 or 3 would work just as well.)

oobabooga:
Thank you so much for the help @turboderp. I have added `gen_begin_reuse` as well as your code for clearing the stopping tokens here https://github.com/oobabooga/text-generation-webui/commit/766c760cd778d4b96aedcfe70b4ee21420c0d6d3

About the leading space issue, I have found another workaround after inspecting the methods available under the tokenizer. The idea is that a given token has a leading space if

```python
generator.tokenizer.tokenizer.IdToPiece(int(token)).startswith('▁'):
```

This allows us to tell if a leading space is necessary or not by checking the first generated token only. The full implementation is here https://github.com/oobabooga/text-generation-webui/commit/c5641b65d34f509439303112a36950e5fea7d5b3

Don-Chad:
@turboderp big thanks for helping to get Exllama to TGW!! 

Since it seems we are alllmost there - and we talked about it in the past - do you know what would need to be adjusted to the multimodal extention to also work with Exllama? 

https://github.com/oobabooga/text-generation-webui/issues/3378

PR #2449, opened 2023-05-31
Update README.md to include missing flags

jllllll:
AutoGPTQ flags and API port flags were missing.


oobabooga:
Thank you, I completely forgot about adding those

PR #2472, opened 2023-06-01
Add a download progress bar to the web UI.

cyberfox:
This places a download progress bar for each file on the model screen, so you can see how far along it's gotten.

It also should increase the amount of data between updates to the download rate display to every 512K (or when the file is done) from every 1K, so it's hopefully not as chatty. Since model downloads are on the order of 6-65GB download speeds are generally going to have to be able to handle 512K in sub-second intervals, so the visual update will still likely be fast.

The progress display overlays the Markdown block slightly.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -194,8 +194,8 @@ class ModelDownloader:
         r = self.s.get(url, stream=True, headers=headers, timeout=10)
         with open(output_path, mode) as f:
             total_size = int(r.headers.get('content-length', 0))
-            # Every 512Kb we report an update
-            block_size = 512*1024
+            # Every 4MB we report an update
+            block_size = 4*1024*1024
 
             with tqdm.tqdm(total=total_size, unit='iB', unit_scale=True, bar_format='{l_bar}{bar}| {n_fmt:6}/{total_fmt:6} {rate_fmt:6}') as t:
                 count = 0
--- a/download-model.py
+++ b/download-model.py
@@ -194,18 +194,25 @@ class ModelDownloader:
         r = self.s.get(url, stream=True, headers=headers, timeout=10)
         with open(output_path, mode) as f:
             total_size = int(r.headers.get('content-length', 0))
-            block_size = 1024
+            # Every 512Kb we report an update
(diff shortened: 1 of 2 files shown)


cyberfox:
As a side note, bumping the iteration frequency from every _1K_ to every _1M_ appears to have sped even the command-line version up from 28MiB/s to around 70MiB/s on my home network. I don't know if that's consistent, or anything, but it's very possible that trying to update too often saps CPU from the network process. I used to run into this in similar situations when printing debug info; it would slow down the actual operations. Hopefully just bumping it to _512K_ helps speed it up a bit for everyone.

cyberfox:
@oobabooga This should be able to merge, I think. Hope it helps!

oobabooga:
I'm experiencing a strange issue: while downloading `tiiuae/falcon-7b`, once `pytorch_model-00001-of-00002.bin` reaches 100%, a timeout exception is raised. For small downloads like `facebook/galactica-125m` it works as expected.

This error takes a while to be generated, so I couldn't tell yet if it is exclusive to this PR or not.

cyberfox:
> I'm experiencing a strange issue: while downloading `tiiuae/falcon-7b`, once `pytorch_model-00001-of-00002.bin` reaches 100%, a timeout exception is raised. For small downloads like `facebook/galactica-125m` it works as expected.
> 
> This error takes a while to be generated, so I couldn't tell yet if it is exclusive to this PR or not.

@oobabooga I will look into this! I had it download `starcoder`, because I wanted to make sure it inherited the `HF_USER` and `HF_PASS` capability, so I did test it with larger models, but I'll poke at it, and see if I can't simulate some network slowdowns. It's possible that a brief hiccup might be more problematic than it was, because of the increased iteration size, in which case bringing it back down to ~512K might be necessary.

oobabooga:
Awesome @cyberfox, any tests would be highly appreciated!

oobabooga:
It seems like the model hanging issue that I had encountered happens in the main branch too, as well as this PR's branch and #2786's, so it should be fine to merge.

Issue #2447, opened 2023-05-31
A suggestion to fix the missing character bug which appears when chatting with RWKV model using Chinese 


GralchemOz:
**Description**

When chatting with the RWKV model using tgwebui, there's a bug  that causes the output to miss characters if it contains Chinese  characters. This is due to the tokenization of Chinese characters.  Each character always needs two tokens. Therefore, when stream output is  enabled, the first token of a character could be decoded into a  meaningless garbled character and won't become a complete character  until the next token is ready.  
I saw that the code of the modules/RWKV.py file simply discards the detected '\ufffd' garbled characters (in line 120 of RWKV.py). This is the source of this bug because if this is discarded, the corresponding Chinese character is also discarded. 
The method to fix this is very simple. When a '\ufffd' code is detected, just stop the output and decode two tokens at the same time in the next output cycle. 

**Additional Context**

Here is an example that revise the `generate_from_cached_state` method in the RWKV.py files which references the ChatRWKV code:

    def generate_from_cached_state(self, ctx="", token_count=20, args=None, callback=None):
        all_tokens = []
        out_str = ''
        occurrence = {}
        state = copy.deepcopy(self.cached_model_state) if self.cached_model_state is not None else None
        # if we ended up with an empty context, just reuse the cached logits
        # this can happen if a user undoes a message and then sends the exact message again
        # in that case the full context ends up being the same as the cached_context, so the remaining context is empty.
        if ctx == "":
            out = self.cached_output_logits

        for i in range(token_count):
            # forward
            tokens = self.pipeline.encode(ctx) if i == 0 else [token]
            while len(tokens) > 0:
                out, state = self.model.forward(tokens[:args.chunk_len], state)
                tokens = tokens[args.chunk_len:]

            if i == 0:
                begin_token= len(all_tokens)
                last_token_posi=begin_token

            # cache the model state after scanning the context
            # we don't cache the state after processing our own generated tokens because
            # the output string might be post-processed arbitrarily. Therefore, what's fed into the model
            # on the next round of chat might be slightly different what what it output on the previous round
            if i == 0:
                self.cached_context += ctx
                self.cached_model_state = copy.deepcopy(state)
                self.cached_output_logits = copy.deepcopy(out)

            # adjust probabilities
            for n in args.token_ban:
                out[n] = -float('inf')

            for n in occurrence:
                out[n] -= (args.alpha_presence + occurrence[n] * args.alpha_frequency)

            # sampler
[...]

PR #2497, opened 2023-06-02
Fix for the missing character bug for RWKV model

GralchemOz:
Changes the Modules/RWKV.py model to fix the missing Chinese character bug for RWKV models, which has been described in #2447


oobabooga:
Thanks

Issue #2473, opened 2023-06-01
Add a /api/v1/tokenize API


bradstallion:
**Description**

Add a REST endpoint to tokenize text. The code could be similar to the implementation of /api/v1/token-count

**Additional Context**

See token-count [here](https://github.com/oobabooga/text-generation-webui/blob/main/extensions/api/blocking_api.py)

PR #2498, opened 2023-06-02
Add Api `/api/v1/encode` and `/api/v1/decode`

Yiximail:
Enhancement needs from #2473 

I split it into two APIs, `/api/v1/encode` and `/api/v1/decode`
and add `add_special_tokens` and `add_bos_token` for consistency with generation
it can handle non-english (Multiple tokens for one character) and emoji now

`/api/v1/encode` request:
image
`/api/v1/encode` response:
image

`/api/v1/encode` basically does what `/api/v1/token-count` does, and I've made a token count field
( I don't think it's suitable for me to delete `/api/v1/token-count`, even though they do the same thing )

I also made `/api/v1/decode` can decode from tokens

if the tokens are incomplete, it will still show the garbled text instead of just deleting it
[ 243, 162, 149, 164 ] for 💡
image


Yiximail:
image
This will make it easier for the frontend to colour tokens :)

PR #2508, opened 2023-06-03
Add --select option to download-model.py, and also do some code clean-up

uogbuji:
Code clean-up includes avoiding shadowing builtins (i.e. `dict` & `bytes`), and some more docstrings & comments.


matatonic:
Nice changes and works well, I think the only thing that might be easier is if you could give it a URL instead of a name.

I just noticed the workflow and it's not much different than before, you still need to go to huggingface website, browse files copy filename, copy/type out model/path paste specific model. 

Ex. python download-model.py --select 4bit.safetensors Aeala/VicUnlocked-alpaca-30b-4bit

python download-model.py https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit/blob/main/4bit.safetensors

And it knows to download the rest also, not just the file I picked, no extra options needed. The alternative now is something like (after following the LFS link):

 python download-model.py --text-only Aeala/VicUnlocked-alpaca-30b-4bit
 wget https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit/resolve/main/4bit.safetensors

matatonic (approved):

uogbuji:
I think I understand what you're saying. You'd specify more the model characteristics than the full file name, so that e.g. the `4bit.safetensors` bit would work regardless of the other bits of the filename. In order to do that I'd have to know more about all the various naming conventions than I do now, but I can keep it in mind for the future, as I keep learning this LLM world.

uogbuji:
So I had thought about user interaction, but I quickly decided against it, because IMO download-model.py should be readily usable headless, e.g. from scripts or Dockerfiles. My sense was that if we did want some sort of menuing for the user, that should more go in an enhancement to the webui's own download model panel, which is already designed for user interactivity. i've never actually used that panel, so maybe I'm missing something about its suitability for this task.

Having some sort of ability to select based on core characteristics (encoded in the file name) would support the headless interaction use-case, and I think would also be easy enough for users, if we do it properly. So I guess I'm still leaning towards some variant of your first suggestions. Either in faux URL form:

```
python download-model.py https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit/blob/main/4bit.safetensors
```

or in your suggested form that I think is cleaner:

```
python download-model.py --select 4bit.safetensors Aeala/VicUnlocked-alpaca-30b-4bit
```

Note: I think the latter could support full URLs no problem:

```
python download-model.py --select 4bit.safetensors https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit
```

Which would allow us to expand & support any repo that structured like HF. I think we need to at least require HF repo compatibility, because we relay on the HF directory manifest JSON to direct the downloader.

Let me know if I'm thinking wrongly about any of this.

For completeness I'll mention that it relies on model maintainers to name things properly. Some sort of reliable metadata in or somehow attached to the file would be even better, and I think there has been some discussion of this, but then again that might defeat the purpose if the downloader has to transfer the entire model file anyway, in order to read its stored metadata. If they were able to maybe encode such metadata in the first block of the file, we could download just the first, say KB of the file from the socket, then assess whether to download the rest. But now I've definitely veered into speculative future.

matatonic:
Re: non-interactive, move changes to GUI: yes, 100%

My original suggestion is based on the fact than huggingface does that weird stuff like exposing downloads that don't work, and then needing to click through and find the real download link (which isn't on the filename), then click that instead.

I'm also experiencing extra annoying pain because I'm on Linux.
1) The huggingface webpage doesn't have text I can copy & paste into a terminal (when you copy the main text you get extra spaces, linefeeds, etc  between the owner "/" and model name) .
2) If you click copy it doesn't end up on the middle-click pasteable clipboard, and you can't ^V paste into a terminal, you need to right click->menu->paste
3) downloading the models via we web UI messes up my permissions and file ownernship because it's in a docker

All that whining over, this is my ideal setup:

$ ./download-model.py https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit/blob/main/4bit.safetensors

uogbuji:
> My original suggestion is based on the fact than huggingface does that weird stuff like exposing downloads that don't work, and then needing to click through and find the real download link (which isn't on the filename), then click that instead.

Oh wow, what?! Do you have an example? Is it an extra redirect, because we might well be able to follow that, if so. Definitely sounds stupid, though!

> I'm also experiencing extra annoying pain because I'm on Linux.
> 
> 1. The huggingface webpage doesn't have text I can copy & paste into a terminal (when you copy the main text you get extra spaces, linefeeds, etc  between the owner "/" and model name) .

Yeah, I've noticed this. No idea why they can't clean that up. I've taken to just copying the relevant bit from the URL bar, instead…and I guess that's your whole point 😂.

> 2. If you click copy it doesn't end up on the middle-click pasteable clipboard, and you can't ^V paste into a terminal, you need to right click->menu->paste
> 3. downloading the models via we web UI messes up my permissions and file ownernship because it's in a docker

Doesn't setuid, setgid, etc take care of this? Maybe I should try the download from UI and see. It feels like there *has* to be a way to address this in docker or OS. I also use docker, on a Linux host, but invoking it from my developer MacBook Pro, setting the docker context to the remote server.

> All that whining over, this is my ideal setup:
> 
> $ ./download-model.py https://huggingface.co/Aeala/VicUnlocked-alpaca-30b-4bit/blob/main/4bit.safetensors

No worries, I can add this syntax, and now, with your additional context above, I can see why it's better for you—you can just get the base URL from the browser bar, and quickly add the 4bit.safetensors part.

matatonic:
Any of the large file links in the "Files and versions" tab as flagged as Git LFS files, if you click them, you get another page with download links instead of downloading... painfully from a UI perspective, it's an extra click to no choice, just a waste of time.

oobabooga:
@uogbuji @matatonic I have just merged the main branch. Can you check if the code still makes sense? 

I want to use it to download ggml models through the UI somehow. If a repository like `TheBloke/Llama-2-7B-GGML` is used in the "Download custom model or LoRA" field, every possible quantization will be downloaded.

Issue #2505, opened 2023-06-03
[extensions/openai] Improve error handling with no model loaded


LoopControl:
### Describe the bug

I was using a few days old version of webui (githash `6627f7f`) and openAI extension worked great.

After updating to latest ([19f7868](https://github.com/oobabooga/text-generation-webui/commit/19f78684e6d32d43cd5ce3ae82d6f2216421b9ae)), openAI extension throws an error on the `/completions` endpoint.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

* enable openai extension
* run the completions api (via any openai supporting tool)

### Logs

```python
127.0.0.1 - - [03/Jun/2023 08:52:27] "POST /v1/chat/completions HTTP/1.1" 200 -
----------------------------------------
Exception occurred during processing of request from ('127.0.0.1', 41928)
Traceback (most recent call last):
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 683, in process_request_thread
    self.finish_request(request, client_address)
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 747, in __init__
    self.handle()
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/http/server.py", line 433, in handle
    self.handle_one_request()
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/http/server.py", line 421, in handle_one_request
    method()
  File "/mnt/nvbig/ai/text-generation-webui/extensions/openai/script.py", line 355, in do_POST
    system_token_count = len(encode(system_msg)[0])
  File "/mnt/ai/text-generation-webui/modules/text_generation.py", line 44, in encode
    input_ids = shared.tokenizer.encode(str(prompt), return_tensors='pt', add_special_tokens=add_special_tokens)
AttributeError: 'NoneType' object has no attribute 'encode'
----------------------------------------
```

### System Info

```shell
linux
```

LoopControl:
This was a derp on my part.

This error was thrown because I didn't have a model loaded. It would be nice to catch this error and output a clean error message to console or via the return on the completions api though (http 500 maybe?).

matatonic:
[...]

PR #2525, opened 2023-06-05
Better error message when using OpenAI extension, but forgetting to load a model

uogbuji:
re: https://github.com/oobabooga/text-generation-webui/issues/2505

Being perfectly honest, not actually tested yet.

### DIFF
--- a/extensions/openai/README.md
+++ b/extensions/openai/README.md
@@ -5,6 +5,12 @@ It's incomplete so far but perhaps is functional enough for you.
 
 ## Setup & installation 
 
+Make sure you enable it in launch parameters. Just make sure they include:
+
+```
+--extensions openai
+```
+
 Optional (for flask_cloudflared, embeddings):
 
 ```
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -226,6 +226,11 @@ class Handler(BaseHTTPRequestHandler):
         if debug:
             print(body)
 
+        # Generate a clear error if no model has ben loaded
+        # May be better ways of handling this error, but this is an improvement
+        if shared.tokenizer is None:
+            raise RuntimeError('No model has been loaded')
+
         if '/completions' in self.path or '/generate' in self.path:
             is_legacy = '/generate' in self.path
             is_chat = 'chat' in self.path


LoopControl:
Tested this pull. The error message "No model has been loaded" which is much clearer than before:

```
----------------------------------------
Exception occurred during processing of request from ('127.0.0.1', 58216)
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 683, in process_request_thread
    self.finish_request(request, client_address)
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/socketserver.py", line 747, in __init__
    self.handle()
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/http/server.py", line 433, in handle
    self.handle_one_request()
  File "/home/user/apps/anaconda3/envs/textgen/lib/python3.10/http/server.py", line 421, in handle_one_request
    method()
  File "/mnt/ai/text-generation-webui/extensions/openai/script.py", line 232, in do_POST
    raise RuntimeError('No model has been loaded')
RuntimeError: No model has been loaded
----------------------------------------
```

However, on the client side, the server closes the connection without any response (it's usually best practice to return a HTTP 500 (Internal server error) or a HTTP 400 (Bad request error)

uogbuji:
> However, on the client side, the server closes the connection without any response (it's usually best practice to return a HTTP 500 (Internal server error) or a HTTP 400 (Bad request error)

Yes. Good point. My PR was meant to be a quick & dirty since @matatonic was already looking more broadly into error handling. The good news is that I think they now have a PR that does the trick.

@matatonic I should close this, yes?

uogbuji:
I'll just also say that most web frameworks I've used will by default automatically send back a 500 in case of an unhandled exception. I haven't used `BaseHTTPRequestHandler` in many years, but I'm surprised it doesn't.

matatonic:
Yeah, I need to still wrap it more generally it to improve unknown errors, but yes, you can close this.

PR #2539, opened 2023-06-05
[extensions/api] models api for blocking_api (updated)

matatonic:
repacked and cleaned up version of: https://github.com/oobabooga/text-generation-webui/pull/2365

* Supports info, load, unload and list. Seems to work with most models. I've included an example test app which does a basic validating smoke test on all models.
* Currently doesn't work well with: RWKV, llama.cpp
* Lora code copied from server.py but not tested
* autogptq doesn't seem be as compatible yet, but I've set it as the default in the api example.
* api example moved to new api-examples folder.


tensiondriven:
> Lora code copied from server.py but not tested

I am going to need the ability to load lora's on the fly, ideally within the context of a single inference request soon... hoping this will get on someone's radar soon.  Maybe i'll open a ticket.

matatonic:
I've been considering how to do this, and supporting selecting a lora from the model parameter of the request, from what I've seen so far this should be very quick (switching lora per request) but I don't have all the details worked out yet, I will ping you when I have something testable - and if you see anything else in the mean time, please let me know @tensiondriven

PR #2552, opened 2023-06-06
added completions api and docs in example

danikhan632:
added another api route to open interopable with OpenAI's completions api, chat-completions api may also be added however as of now, this is meant for non-chat based models.


oobabooga:
Have you tried the openai extension? This is already implemented there.

danikhan632:
looks like I missed that, thanks

PR #2557, opened 2023-06-07
[extensions/openai] docs update, model loader, minor fixes

matatonic:
* readme updates
* new engines endpoints, including a model loader via a legacy endpoint
* minor fixes
* API error on batched generation requests
* Partially revert some stopping strings change (performance + compatibility)

You can use the model loader/lister from the openai-python command line like this:

```
openai api engines.list
openai api engines.get -i <model_name>
```


matatonic:
(fixing up model loading)

PR #2565, opened 2023-06-07
Quick launch script

Pb-207:
A simple script for quick launch which allows you to load flags from saved defaults or save new flags to defaults by simply enter some numbers. Default flags are saved in flags.ini.


oobabooga:
For the one-click-installer there is already a launch script with customizable flags, and from the command-line on can simply ctrl+r and type python server.py to get the last command. I think that your addition can be useful in some cases but I find it too niche to include by default.

Issue #2477, opened 2023-06-01
Mirostat sampler for transformers models.


Ph0rk0z:
People on reddit were raving that it's better. Is it? They claim it's twice as smart in terms of replies.

paper:https://openreview.net/forum?id=W1G1JZEIy5_
code: https://github.com/basusourya/mirostat/blob/master/mirostat.py

Why should llama.cpp have all the fun?

PR #2571, opened 2023-06-07
Add Mirostat v2 sampling to transformer models

brandonj60:
This PR is an attempt to add a custom sampler for Mirostat v2 when using the transformers library for generation. It utilizes the existing custom monkeypatch sampler_hijack.py. Loosely based on the v1 implementation [here](https://github.com/basusourya/mirostat/blob/master/mirostat.py) and the llama.cpp v2 implementation [here](https://github.com/ggerganov/llama.cpp/blob/2d7bf110edd8c49209401a16132052cba706ffd0/llama.cpp#L2150-L2186). Of note, when mirostat_mode=2, other samplers are disabled.

Mirostat overview - [Arxiv](https://arxiv.org/pdf/2007.14966.pdf)

This fulfills https://github.com/oobabooga/text-generation-webui/issues/2477

### DIFF
--- a/modules/sampler_hijack.py
+++ b/modules/sampler_hijack.py
@@ -1,7 +1,7 @@
 import torch
 import transformers
 from transformers import LogitsWarper
-from transformers.generation.logits_process import LogitNormalization, LogitsProcessorList
+from transformers.generation.logits_process import LogitNormalization, LogitsProcessorList, TemperatureLogitsWarper
 import math
 
 class TailFreeLogitsWarper(LogitsWarper):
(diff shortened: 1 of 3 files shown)


brandonj60:
> Looking at the code it seems like `MirostatLogitsWarper` is only used if `mirostat_mode=2` (it is not appended), is that intended? If not, does it make a difference to append the sampler before or after the other ones?

mirostate_mode=1 is a different sampler, which I did not implement here (it's more complicated than v2). From my understanding, mirostat samplers cannot be used alongside other samplers like topk, tailfree, etc, so rather than appending Mirostat to the list of samplers, it becomes the sole member of the list (you can see similar behavior in llama.cpp [here](https://github.com/ivanstepanovftw/llama.cpp/blob/f571806da77ee034a07a0792a53eb91eb9b8dda8/examples/main/main.cpp#L446-L463) ).

toast22a:
> > Looking at the code it seems like `MirostatLogitsWarper` is only used if `mirostat_mode=2` (it is not appended), is that intended? If not, does it make a difference to append the sampler before or after the other ones?
> 
> mirostate_mode=1 is a different sampler, which I did not implement here (it's more complicated than v2). From my understanding, mirostat samplers cannot be used alongside other samplers like topk, tailfree, etc, so rather than appending Mirostat to the list of samplers, it becomes the sole member of the list (you can see similar behavior in llama.cpp [here](https://github.com/ivanstepanovftw/llama.cpp/blob/f571806da77ee034a07a0792a53eb91eb9b8dda8/examples/main/main.cpp#L446-L463) ).

Instead of manually removing the other samplers from the warper list in `sampler_hijack.py`, maybe we can just add a preset `presets/Special-Mirostat Sampling.yaml`? That way, the UI shows, and the user knows, that the other samplers are meant to be turned off (e.g. set to 1.0 for top_p, 0 for top_k, etc.) when mirostat is used.

Also, in llama.cpp, it looks like temperature is applied before mirostat v2 sampling:
https://github.com/ivanstepanovftw/llama.cpp/blob/f571806da77ee034a07a0792a53eb91eb9b8dda8/examples/main/main.cpp#L451-L454
So in any case, we should probably keep that in the warper list.

oobabooga:
It looks good to me now. I have made some basic tests and all seems to be working well. Thanks for the PR, @brandonj60!

BadisG:
When you put a fixed seed, when you regenerate you get 2 different outputs each time
Output 1 then Output 2 then Output 1...

PR #2580, opened 2023-06-08
models/config.yaml: +alpacino, +alpasta, +hippogriff, +gpt4all-snoozy, +lazarus, +based, -airoboros 4k

matatonic:
* model config for alpacino, alpasta, hippogriff (template already there)
* +gpt4all-snoozy, +lazarus, +based
* remove 4k truncation length from airoboros-13b-gpt4 , See note: https://huggingface.co/jondurbin/airoboros-13b-gpt4


oobabooga:
I think that you forgot to add the Hippogriff file to the PR. Also, maybe use the full model name(s) for the based- model? based- is too nonspecific and might conflict with future models.

matatonic:
I looked for the Hippogriff file and it was already there - maybe I'm wrong and added it from before, I will update it. It was intended to be generic because it has it's own format, so my though was that any model using it should use that in the name should use that format.

matatonic:
I see, I missed the git ignore on the characters files. I also have files for Samantha, Gorilla, and a second file for Guanaco with a different prompt - I don't use any of these myself, but I think I would have created them for a good reason. The Guanaco one is based off some comments about the Guanaco Chat being the wrong format - I haven't tested enough to be sure. 

Do you want me to add them here?

LoopControl:
Yep I was about to request the Guanaco prompt fix also.

I don't know that people even use the older Guanaco by the other person whereas the new Guanaco is very popular.

matatonic:
ok, I'll include them all and let @oobabooga sort them out. I'm fairly sure the research I did to get the prompt format was correct, but I can't say exactly which models need to have it applied anymore, there are too many to check through.

matatonic:
Yeah, based is just way too common... removed it for now. Fixed up Guanaco pattern and Lazarus.

matatonic:
Ok, I think that's it for this... I misunderstood the patter matching, for some reason I thought it was lowercased first... anyways. Fixed up now.

oobabooga:
Looks good, thank you!

Issue #1534, opened 2023-04-25
 ERROR: Failed building wheel for llama-cpp-python


Song367:
### Describe the bug

 install  llama display error  ERROR: Failed building wheel for llama-cpp-python

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. pip install -r requirements.txt
2.  ERROR: Failed building wheel for llama-cpp-python

### Logs

```shell
Collecting llama-cpp-python==0.1.36
  Using cached https://pypi.tuna.tsinghua.edu.cn/packages/1b/ea/3f2aff10fd7195c6bc8c52375d9ff027a551151569c50e0d47581b14b7c1/llama_cpp_python-0.1.36.tar.gz (1.1 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: typing-extensions>=4.5.0 in ./tenv/lib/python3.8/site-packages (from llama-cpp-python==0.1.36) (4.5.0)
Building wheels for collected packages: llama-cpp-python
  Building wheel for llama-cpp-python (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for llama-cpp-python (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [94 lines of output]
```

### System Info

```shell
GPU
```

disarmyouwitha:
If you are going to be using GPU you can just remove the last 2 lines of the requirements.txt:
`llama-cpp-python==0.1.36; platform_system != "Windows"
https://github.com/abetlen/llama-cpp-python/releases/download/v0.1.36/llama_cpp_python-0.1.36-cp310-cp310-win_amd64.whl; platform_system == "Windows"`

And then run `pip install -r requirements.txt` again.

innocentius:
Same error here, running on Ubuntu 18.04.

Wouldn't remove llama-cpp-python break something?

innocentius:
Error solved by upgrading to gcc-11. Try that first.

xNul:
The same error here https://github.com/oobabooga/one-click-installers/issues/30#issuecomment-1517950800

BlairSadewitz:
> Error solved by upgrading to gcc-11. Try that first.

That's what I did, and the error resolved.

juggli:
> > Error solved by upgrading to gcc-11. Try that first.
> 
[...]

PR #2581, opened 2023-06-08
Windows commands to install llama-cpp-python changed

Iakovenko-Oleksandr:
The commands as they are did not work in my Windows Anaconda prompt (compilation failed), then with changes suggested in https://github.com/oobabooga/text-generation-webui/issues/1534#issuecomment-1557287144 it worked.


brandonj60:
All this does is remove CUDA support. LLAMA_CUBLAS is the argument that enables GPU compilation.

PR #2600, opened 2023-06-09
Patch 2

ChobPT:
Added StarChat Presets based on the Playground and Model Pages


oobabooga:
I think the template is missing some newline characters:

```
prompt_template = "<|system|>\n<|end|>\n<|user|>\n{query}<|end|>\n<|assistant|>"
```

Also the bot is <|assistant|> not <|system|>

ChobPT:
According to the model page, this is what they are using, am I missing something?


```
import torch
from transformers import pipeline

pipe = pipeline("text-generation", model="HuggingFaceH4/starchat-beta", torch_dtype=torch.bfloat16, device_map="auto")

# We use a variant of ChatML to format each message
prompt_template = "<|system|>\n<|end|>\n<|user|>\n{query}<|end|>\n<|assistant|>"
prompt = prompt_template.format(query="How do I sort a list in Python?")
# We use a special <|end|> token with ID 49155 to denote ends of a turn
outputs = pipe(prompt, max_new_tokens=256, do_sample=True, temperature=0.2, top_k=50, top_p=0.95, eos_token_id=49155)
# You can sort a list in Python by using the sort() method. Here's an example:\n\n```\nnumbers = [3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5]\nnumbers.sort()\nprint(numbers)\n```\n\nThis will sort the list in place and print the sorted list.
```

SwaggerMcSwaggy:
According to theBloke the template is like this:
```
<|system|> system message goes here <|end|>
<|user|> prompt goes here <|end|>
<|assistant|>
```

ChobPT:
I have it working wit the Prompt suggested on HF and confirmed by Swagger, please confirm if anything else is needed

oobabooga:
> prompt_template = "<|system|>\n<|end|>\n<|user|>\n{query}<|end|>\n<|assistant|>"

See the line above. The template is missing the \n characters.

ChobPT:
Closing this as it seems that  https://github.com/oobabooga/text-generation-webui/pull/2790 has done a better job.

PR #2615, opened 2023-06-11
Added Starchat.yaml instruction

FartyPants:


oobabooga:
Shouldn't <|end|> be in the turn_template?

FartyPants:
Yes, I'm taking this down and PR the correct one

PR #2618, opened 2023-06-11
[extensions/api] simple api for getting presets, characters, prompts, templates

matatonic:
* Simple api endpoints for getting & listing available characters, instruction templates, prompts & presets.
* Basic wrappers around get_available_presets, get_available_prompts,  get_available_characters, get_available_instruction_templates
* Basic file fetching & json returns by name.

I kind of want to change the new models API to follow the same conventions... (GET vs. POST also), but would break the existing models API. What do you think of this? Worth while? Alternately, I could also just add new endpoints which don't break existing ones.

I'm work towards an automated way to get the right instruction format, and perhaps a standard format we could load from the model folders rather than relying on a hand crafted database. Long term manual maintenance of the models/config.yaml is going to be impossible IMO, with characters/instruction-following/* a close second.

Maybe once we have a standard we can get folks like TheBloke to populate a config file when it's shared (hopefully even for wilder more flexible ones like pygmalion/superHOT's RP like formats).

Curious to know your thoughts on this.


oobabooga:
The new `simple_json_results` function is a nice simplification that I would like to keep, but I feel like the endpoints for getting file contents are beyond the scope of what this API is meant to be. Its primary purpose is giving access to the text generation functions. It is assumed that the user has access to the machine where the API is being run, and thus to the files in the repository.

Having the instruction format be distributed with the models would indeed be a long-term goal. But it's one that is hard to achieve when eg ggml models, that are the most popular, don't even have their own branches on TheBloke repositories.

PR #2624, opened 2023-06-11
Add Tensorboard/Weights and biases integration for training

kabachuha:
See https://github.com/huggingface/transformers/issues/11084 and https://docs.wandb.ai/guides/integrations/huggingface as reference


kabachuha:
Proof that it works (Tensorboard)
image

kabachuha:
WANDB perfectly works as well
Screenshot 2023-06-17 at 15-35-35 Weights   Biases

Ph0rk0z:
Does it still have to be online and have an account for these graphs?

kabachuha:
Both save logs in a local dir along the adapter. Tensorboard is fully self-hosted, without any sigh up

As I get from https://github.com/wandb/wandb/issues/1222 you can self-host a wandb server too

https://docs.wandb.com/self-hosted/local

PR #2640, opened 2023-06-12
A small UI change in Models menu

FartyPants:
A bit more space for dropdown when Autoload the model is OFF
image


oobabooga:
I prefer the current button style and labels, it looks clunky but it helps avoid ambiguity. I missed the huge Load button, thanks for the fix.

FartyPants:
No problem, you are the boss.

PR #2643, opened 2023-06-12
Small bug, when arbitrary loading character.json that doesn't exist

### LINKED ISSUE #2482: Small bug, when arbitrary loading character.json that doesn't exist
### Describe the bug

Things will blow if character json doesn't exist (that webui loads arbitrary after model load)
text-generation-webui\modules\chat.py", line 513, in load_character

file_contents = open(filepath, 'r', encoding='utf-8').read()
FileNotFoundError: [Errno 2] No such file or directory:

Probably wrap it in
try:
....
    except FileNotFoundError:

### Reproduction

if json is missing

### Screenshot

_No response_

### Logs

```shell
File "H:\oobabooga_windows\installer_files\env\lib\site-packages\anyio\_backends\_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "H:\oobabooga_windows\text-generation-webui\modules\chat.py", line 513, in load_character
    file_contents = open(filepath, 'r', encoding='utf-8').read()
```

### System Info

```shell
windows
```

original-subliminal-thought-criminal:
When opening a character file that doesn't exist, we needed to add a try / except FileNotFoundError fixing #2482

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -518,7 +518,12 @@ def load_character(character, name1, name2, instruct=False):
             if filepath.exists():
                 break
 
-        file_contents = open(filepath, 'r', encoding='utf-8').read()
+        try:
+            file_contents = open(filepath, 'r', encoding='utf-8').read()
+        except FileNotFoundError as e:
+            logger.error(f"Could not find character file for {character} in {folder} folder. Please check your spelling.")
+            return
+        
         data = json.loads(file_contents) if extension == "json" else yaml.safe_load(file_contents)
 
         # Finding the bot's name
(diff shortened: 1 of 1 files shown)


modules/chat.py
```diff
@@ -518,7 +518,12 @@ def load_character(character, name1, name2, instruct=False):
             if filepath.exists():
                 break
 
-        file_contents = open(filepath, 'r', encoding='utf-8').read()
+        try:
+            file_contents = open(filepath, 'r', encoding='utf-8').read()
+        except FileNotFoundError:
+            logger.error(f"Could not find character file for {character} in {folder} folder. Please check your spelling.")
+            return
```
oobabooga, line 525: I think that this should be a path.exists() check instead of try/except

PR #2648, opened 2023-06-12
AutoGPTQ: Add UI and command line support for disabling fused attention and fused MLP

TheBloke:
AutoGPTQ defaults to enabling fused attention and fused MLP.  These are controlled by the following parameters to `.from_quantized()`:
```
inject_fused_attention
inject_fused_mlp
```

Fused attention applies to both CUDA and Triton modes. Fused MLP is Triton only.

These params improve inference performance by a decent amount.  But they also increase VRAM requirements.

The upshot is that a user with a 12GB card cannot do full 2048 context on a 13B model with AutoGPTQ CUDA with default settings - they will OOM at around 1630 tokens returned.  And a user with a 24GB card can only do full 2048 context on a 30B model if they don't have any monitors plugged in; 2048 context finishes with about 13MB free out of 24GB.

Therefore these users will benefit from being able to optionally disable these features.  I have confirmed that with fused attention disabled, a 13B model can return a full 2048 context within 12GB (it's still very close - at one point only 56MiB was free, although it finished the 2000 tokens returned with 322MiB free)

This PR adds UI and command line support for disabling these params.  As AutoGPTQ defaults to having the params on, I have implemented the switches in the negative, as `no_inject_fused_attention` and `no_inject_fused_mlp`.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,8 +249,10 @@ Optionally, you can use the following command-line flags:
 
 | Flag             | Description |
 |------------------|-------------|
-| `--triton`       | Use triton. |
-| `--desc_act`     | For models that don't have a quantize_config.json, this parameter is used to define whether to set desc_act or not in BaseQuantizeConfig. |
+| `--triton`                     | Use triton. |
+| `--no_inject_fused_attention`  | Disable the use of fused attention, which will use less VRAM at the cost of slower inference. |
+| `--no_inject_fused_mlp`        | Triton mode only: disable the use of fused MLP, which will use less VRAM at the cost of slower inference. |
+| `--desc_act`                   | For models that don't have a quantize_config.json, this parameter is used to define whether to set desc_act or not in BaseQuantizeConfig. |
 
 #### GPTQ-for-LLaMa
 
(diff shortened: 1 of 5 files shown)


TheBloke:
Just realised there's a problem I need to fix

TheBloke:
OK that should be fine now.

oobabooga:
Thanks for the clean PR, @TheBloke! I have made a test with `llama-30b-4bit-128g`, and I OOM with 1671 context with or without `--no_inject_fused_attention`. The performance was about 13% slower. Since it helps people with 12gb GPUs run 13b models, it's good addition to have.

TheBloke:
Thanks @oobabooga  !

The reason you always OOM'd on the 30B is that it has group size 128.  With AutoGPTQ and GPTQ-for-LLaMa, it's necessary to set group_size = None (`-1`) to avoid OOM on 30B models on a 24GB card.   All my 30B models have group_size None for that reason (then they also have desc_act/act-order = True to compensate on the accuracy front - I don't enable act-order when group_size is used else performance is terrible and compatibility is affected for those using old GPTQ-for-LLaMA. But without group_size, act-order works fine and helps the perplexity a fair bit.)

I'm a bit surprised to hear there that you saw exactly the same OOM point with and without inject fused attention. I guess at a certain point it suddenly grows by a significant chunk and the small VRAM reduction wasn't enough to prevent that.

I'm hopeful that on a 30B no-groupsize model, disabling fused attention would give enough of a VRAM margin for someone to still leave their monitors connected while reaching 2048. I've not tested that yet, I will do at some point.

(With ExLlama you can do 128g on 30B and still have gigs to spare on a 24GB card, but that implementation is insane :) )

PR #2688, opened 2023-06-14
Add context_instruct to API. Load default model instruction template …

atriantafy:
Here are two fixes for the API related to controlling instruct models.

* Add context_instruct to API: The instruct context sent to a model is equivalent to OpenAI's System messages, which is something that should be controllable through the API.
* Load default model instruction template if not provided: Currently, if you do not provide instruction_template to the Chat API it will crash, thus the user always has to provide the instruction template with every API request. Oobabooga texgen already detects the correct instruction template internally when loading a model so it makes sense to use this, if the instruction_template parameter is not provided by the user (and thus, the user doesn't have to guess which is the correct template to use, as the server already knows the best template to use)


jllllll:
Maybe I'm mistaken in my interpretation of the code, but it would seem that loading a default model instruction template when not provided will only work for instruct models. There are plenty of models that are not trained for instruction-based tasks, and as a result, will not have an instruction template available for them to load. This problem will also still occur for newer instruct models that do not have a template defined yet.

I think a fallback to a default template is needed for models without one available. A commonly used template like `Alpaca` should work fine.

atriantafy:
> 
Hey @jllllll , thanks , that's a great point. I have included the fallback to Alpaca as you suggested, and tested it with a model that does not specify an instruction template (TheBloke_Karen_theEditor_13B-GPTQ in this case) and it successfully loaded the default Alpaca instruction template.

matatonic:
I like Alpaca as a fallback also (current default fallback in the openai extension), but I'm thinking of switching to Vicuna 1.1 format, which seems to be the dominant option lately and I think generally better suited. Alpaca using '###' as a stopping string is a comically poor choice because python comments, markdown headers, etc. all start with it.

jllllll:
I think the Alpaca devs actually intended it to be used as a markdown header.

oobabooga:
Thanks @atriantafy for the PR and thanks @jllllll and @matatonic for the review.

I agree that Vicuna-v1.1 is a better fallback and have made this change. Also added some API examples.

PR #2554, opened 2023-06-06
Adds guidance extension

paolorechia:
**What is this**
Adds a small API wrapper to the guidance library (https://github.com/microsoft/guidance), using the model loaded by oobabooga's UI.

**Use cases in mind**
Implementation of chain of thought flows (and more complex flows) with guidance, using oobabooga as the model loader, so people can have an easy time loading GPTQ and other types of models. 

**Why**

I've tried adding support to oobabooga's existing API to the guidance library (https://github.com/paolorechia/local-guidance), however not all features were possible to support, since guidance depends on `logprobs` and other data fields from the Hugging Face API for certain features.

**Limitations**
Only works for HuggingFace and GPTQ models so far. Someone already opened a PR in guidance to add supported llama cpp Python bindings: https://github.com/microsoft/guidance/pull/70 - with this, we'll be able to support LLaMA GGML models. 

**How to use**
The exposed API endpoint can easily be used with the thin wrapper (andromeda-chain):

```pip install andromeda-chain```
 
Repository: https://github.com/ChuloAI/andromeda-chain
Example code:
```python
from andromeda_chain import AndromedaChain, AndromedaPrompt, AndromedaResponse

chain = AndromedaChain("http://0.0.0.0:9000/guidance_api/v1/generate")

prompt = AndromedaPrompt(
    name="hello",
    prompt_template="""Howdy: {{gen 'expert_names' temperature=0 max_tokens=300}}""",
    input_vars=[],
    output_vars=["expert_names"]
)

response: AndromedaResponse = chain.run_guidance_prompt(prompt)
```

Alternatively, the extension can be used by just implementing a simple HTTP client.

PR #2709, opened 2023-06-16
fixed errors from Guidance API extensions added 

danikhan632:
this is an extension to microsoft/guidance to allow [text-generation-webui](https://github.com/oobabooga/text-generation-webui) models to be used which allow for updated models, easy setup, mutiple GPUs etc to be used with guidance


oobabooga:
The same commend as in https://github.com/oobabooga/text-generation-webui/pull/2554#issuecomment-1599542862 applies. Could you please submit the extension to https://github.com/oobabooga/text-generation-webui-extensions?

Issue #2717, opened 2023-06-16
API extension CORS support


rizerphe:
**Description**

Currently, the uses of the API are severely limited for a lot of applications - anything running in a browser or inside electron basically can't work - because the API extension does not respond to `OPTIONS` requests. Instead, it should either accept all requests or at least let the user configure this.

PR #2718, opened 2023-06-16
Add CORS support to the API

rizerphe:
Addresses #2717  - adds CORS support to the API, to allow for uses in browser/electron environments.

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -182,6 +182,17 @@ class Handler(BaseHTTPRequestHandler):
         else:
             self.send_error(404)
 
+    def do_OPTIONS(self):
+        self.send_response(200)
+        self.end_headers()
+
+    def end_headers(self):
+        self.send_header('Access-Control-Allow-Origin', '*')
+        self.send_header('Access-Control-Allow-Methods', '*')
+        self.send_header('Access-Control-Allow-Headers', '*')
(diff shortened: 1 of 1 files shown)


matatonic:
This fix was added to the openai api recently, we fixed it a little differently, just FYI: https://github.com/oobabooga/text-generation-webui/blob/d8d29edf54f224ba3b0daa36fce8fcae8125aa64/extensions/openai/script.py#L104-L117

gitmylo:
@oobabooga i'm currently writing a web app which tries to use the api, this is currently (as of 0f9088f730979a43bf8668034cae16d38c1662bb) not possible because of CORS.

I have tested, and this PR fixes it. And now i'm able to send requests without having issues with CORS. So i can confirm that it works.

oobabooga:
Thanks for the confirmation @gitmylo

rsxdalv:
Not sure if it's a real bug or not yet, but I'm seeing this behavior when trying to combine Text Generation WebUI with OpenWebUI:
```
Access to fetch at 'http://127.0.0.1:5000/v1/models' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The 'Access-Control-Allow-Origin' header has a value 'http://127.0.0.1' that is not equal to the supplied origin. Have the server send the header with a valid value.
```

PR #2759, opened 2023-06-19
supporting http proxies

chrisptang:
supporting http proxies for regions that not able to connect to huggingface.co directly

by adding new command line arg:proxy, users are allowed to specify local or remote http(s) proxies for model downloading. this is helpful for users that has no direct access to huggingface.co, for example, China.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -72,8 +72,14 @@ EleutherAI/pythia-1.4b-deduped
 
 
 class ModelDownloader:
-    def __init__(self):
+    def __init__(self, proxy=None):
         self.s = requests.Session()
+        if not proxy is None:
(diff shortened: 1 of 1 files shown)


sofronaspe:
Please add exception checks in case the proxy isn't available to gracefully fail the script.

Ph0rk0z:
Can your system not run everything through the proxy?

RedisOptimal:
The same problem traped me for several hours, until that i remember https_proxy variable. It's not necessary to change code.

Setting this environment variable before run:
`
export https_proxy="https://foo:bar@xxxx.com"
export https_proxy="socks5://127.0.0.1:8080"
`

oobabooga:
Consensus seems to be that changing the webui code is not necessary for this

PR #2765, opened 2023-06-19
llama.cpp: implement ban_eos_token via logits_processor

Cebtenzzre:
This is a straightforward addition to llamacpp_model.py that uses the logits_processor parameter of `llama_cpp.Llama.create_completion` to set the probability of the EOS token to `-inf` when ban_eos_token is checked in the UI.


oobabooga:
Thank you!

Issue #2746, opened 2023-06-18
`stopping_strings` doesn't work in `exllama`


Yiximail:
### Describe the bug

using model: TheBloke/airoboros-65B-gpt4-1.2-GPTQ

when using `exllama`:
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/4268ebdb-f7a9-4a2e-962e-f3f6242b6995)

when using `autoGPTQ` by default:
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/4d98ff28-5e7a-4157-b778-d3a4443738e3)

and `Custom stopping strings` in webui is fine:
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/18ba9eca-61bd-4986-b19c-16082273af77)
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/fccd277f-0b4c-46c7-be7d-1dc811626571)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

{"prompt":"User: What are the names of Batman's parents? Describe them in detail.\nBot:","max_new_tokens":250,"do_sample":true,"temperature":1.3,"top_p":0.1,"typical_p":1,"repetition_penalty":1.18,"top_k":40,"min_length":0,"no_repeat_ngram_size":0,"num_beams":1,"penalty_alpha":0,"length_penalty":1,"early_stopping":true,"seed":-1,"add_bos_token":true,"truncation_length":2048,"ban_eos_token":false,"skip_special_tokens":true,"stopping_strings":["Martha", "martha"]}

### Logs

```shell
None
```

### System Info

```shell
ubuntu 22.04
text-generation-webui branche 5b4c0155
```

Ph0rk0z:
Is it also broken for llama.cpp and rwkv? I don't see anything in generate_custom that actually uses stopping strings.

Yiximail:
# **Update:**

**I said it wrong**

this bug can be triggered in default mode either, and it has nothing to do with whether using API or not.

![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/76825dcb-58b3-4e92-a49e-23ebc0efecc3)

and `/api/v1/chat` could Cut Off the sentence
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/90fb2956-dc9a-4d51-ab9c-d2476146ab1b)

but did't stop generation too
[...]

PR #2766, opened 2023-06-19
Implemented `stopping_strings` for `exllama` and `llamacpp`

Yiximail:
Enhancement needs from #2746 

0. In `chat.py`, the original `next_character_found` I changed to `stop_by_reply`
It not only stop when the characters' names are found, but also will stop when reply matches `custom_stopping_strings`
which made me search for a while for why chat mode would stop, but not default mode
Subsequent changes are based on the same idea as this function, looking for the string
1. basically, string-based interrupt generation is implemented
I have tried various modes and endpoint apis
2. Interrupting the generation will immediately stop the generation, then output (whether streaming or not)
instead of intercepting after reaching the maximum token generated
3. Support exllama and llamacpp
~I'm not familiar with rwkv, it doesn't go generate_reply_custom, sorry T^T~
4. ~As before, strings that have been output in non-chat mode will not be deleted~
~But it can be easily matched and deleted~
5. (updated) It will not output the stopping strings now

<details>
<summary>Some screenshots of the test: ~(Sorry, batman.)~</summary>

Custom stopping strings
image

default mode:
image

chat mode:
image

`/api/v1/generate` :
image

`/api/v1/chat` :
image

multiple strings:
image

</details>

<details>
<summary> API parameters </summary>

`/api/v1/generate`
``` json
{"prompt":"User: What are the names of Batman's parents? Describe them in detail.\nBot:","max_new_tokens":250,"do_sample":true,"temperature":1.3,"top_p":0.1,"typical_p":1,"repetition_penalty":1.18,"top_k":40,"min_length":0,"no_repeat_ngram_size":0,"num_beams":1,"penalty_alpha":0,"length_penalty":1,"early_stopping":false,"seed":-1,"add_bos_token":true,"truncation_length":2048,"ban_eos_token":false,"skip_special_tokens":true,"stopping_strings":["Martha", "martha"], "custom_stopping_strings": ""}
```
`/api/v1/chat`
``` json
{"user_input": "What are the names of Batman's parents? Describe them in detail.","stopping_strings": ["Martha"],"max_new_tokens": 250,"history": {"internal": [], "visible": []},"mode": "instruct","character": "Example","instruction_template": "Vicuna-v1.1","your_name": "You","regenerate": false,"_continue": false,"stop_at_newline": false,"chat_prompt_size": 2048,"chat_generation_attempts": 1,"chat-instruct_command": "Continue the chat dialogue below. Write a single reply for the character","preset": "None", "do_sample": true,"temperature": 0.7,"top_p": 0.1,"typical_p": 1,"epsilon_cutoff": 0,"eta_cutoff": 0,"tfs": 1,"top_a": 0,"repetition_penalty": 1.18,"top_k": 40,"min_length": 0,"no_repeat_ngram_size": 0,"num_beams": 1,"penalty_alpha": 0,"length_penalty": 1,"early_stopping": false,"mirostat_mode": 0,"mirostat_tau": 5,"mirostat_eta": 0.1,"seed": -1,"add_bos_token": true,"truncation_length": 2048,"ban_eos_token": false,"skip_special_tokens": true }
```

</details>

tried:
- [x] autoGPTQ(default) + stream
- [x] autoGPTQ(default)+ no-stream
- [x] exllama+ stream
- [x] exllama+ no-stream
- [x] llamacpp+ stream
- [x] llamacpp+ no-stream


Ph0rk0z:
Does this fix silly tavern stopping without EOS token?

Yiximail:
> Does this fix silly tavern not stopping without EOS token?

Err... silly tavern? I guess yes.

If you can provide me a example of your request, I can follow it for you. :)

I am currently still working on it, want to find a better way to implement

modules/llamacpp_model.py
```diff
@@ -84,12 +84,56 @@ def generate(self, prompt, state, callback=None):
             ]) if state['ban_eos_token'] else None,
         )
```
omasoud, line 85: Member function `Llama.create_completion()` [takes a `stop` parameter](https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py#L785). Is there a reason we don't want to use it?

Yiximail: Maybe not, I'm not familiar with llama.cpp but I guess we should use `stop` or `stopping_criteria` indeed

I just wrote the logic for `exllama` then copied it over to `llama.cpp`

I will try at a later time, many thanks to remind me

Yiximail: I have tried both webui and api
It worked very well and I updated it up
Thanks again

Panchovix:
Thanks for this! It fixed a lot with exllama, and exllama + sillytavern.

CyberTimon:
Thank you for this. What is the easiest way to apply this PR until it get's merged?

Ph0rk0z:
clone https://github.com/Yiximail/text-generation-webui/tree/stopping_strings

Ph0rk0z:
In ST, it would stop when the module output an EOS token before the PR existed. When it did not, it would generate to the limit. It wasn't sending stop when "You:" or any other strings showed up just ST would not display it.. If this doesn't change things then I dunno.

official-elinas:
Before this PR, it would generate to the set token limit for me outputting "Instruction" etc. 

It's just doing it less now. I'm wondering if it has to do with using an instruction template in ST. I will do some more testing and try to pin down what causes it to sometimes break in this PR.

Yiximail:
> Before this PR, it would generate to the set token limit for me outputting "Instruction" etc.
> 
> It's just doing it less now. I'm wondering if it has to do with using an instruction template in ST. I will do some more testing and try to pin down what causes it to sometimes break in this PR.

Sorry, I don't use instruction mode often
Can you give me a more specific example?

image
if I set up the stopping_strings with "\n### Human:"
it will not print "\n### Human:" at all, include the line break

image
the first round should end like this

then second round should append the user speak and the beginning of Assistant
image

This works very well for chat mode, and I've tried webui's instruction mode as well.

Panchovix:
Would this PR affect the new exllama_hf (transformers samplers for exllama?) https://github.com/oobabooga/text-generation-webui/pull/2777

Yiximail:
> Would this PR affect the new exllama_hf? (transformers samplers for exllama) #2777

Good remind, I'll go check it out now...

official-elinas:
> Sorry, I don't use instruction mode often Can you give me a more specific example?

There are many instruction formats, probably the most popular being Alpaca which is 

```
### Instruction:
<enter instruction here>
### Response:
```

In ST (Sillytavern), there is an option to add custom stopping strings which would be those two for example.  Because there are multiple formats, do not inject a specific format as that will break models that do not support that format. That should be handled by the client.

Yiximail:
> > Sorry, I don't use instruction mode often Can you give me a more specific example?
> 
> There are many instruction formats, probably the most popular being Alpaca which is
> 
> ```
> ### Instruction:
> <enter instruction here>
> ### Response:
> ```
> 
> In ST (Sillytavern), there is an option to add custom stopping strings which would be those two for example. Because there are multiple formats, do not inject a specific format as that will break models that do not support that format. That should be handled by the client.
> 
> Edit - example from ST output: `stopping_strings: [ '\nYou:', '\nMyName:', '\n### Instruction:', '\n### Response:' ],`

Yes, I understand
It won't output any formatting content in stopping_strings
just generate the response

Yiximail:
> Would this PR affect the new exllama_hf? (transformers samplers for exllama) #2777

Unfortunately, it doesn't

Panchovix:
> > Would this PR affect the new exllama_hf? (transformers samplers for exllama) #2777
> 
> Unfortunately, it doesn't

oof, thanks anyways! It seems the implementation of exllama_hf.py is pretty different that exllama.py. I guess maybe that on exllama_hf, the stopping works in another way as it does on exllama.py.

Yiximail:
> 

Yes, it uses StoppingCriteriaList, It is based on tokens stop
some of the contents of stopping_strings may appear at the end
like:

image

official-elinas:
Stopping seems to be working at least for now. I pulled the latest fork's changes, don't think there was anything changed but you merging main into your branch. If it happens again, I'll see what I see in the log.

Yiximail:
> Stopping seems to be working at least for now. I pulled the latest fork's changes, don't think there was anything changed but you merging main into your branch. If it happens again, I'll post what I see in the log.

That's okay, maybe we should prefer to use exllama_HF afterwards and exllama.py might be abandoned

official-elinas:
I tried exllama_hf without your fork and it has the issue of not stopping. So I'll continue to use this for now.

Panchovix:
I agree with elinas, if there's a way to add it to exllama_hf, it would be great. I can confirm there the non stopping issue exists.

extensions/api/util.py
```diff
@@ -95,4 +95,4 @@ def _start_cloudflared(port: int, max_attempts: int = 3, on_start: Optional[Call
             traceback.print_exc()
             time.sleep(3)
 
-        raise Exception('Could not start cloudflared.')
+        raise Exception('Could not start cloudflared.')
```
mcmonkey4eva, line 98: There's some unrelated/unintended changes here that would be good to remove to make the PR easier to merge properly^

mcmonkey4eva (approved):
The PR itself needs some cleanup (redundant changes) but the fix works! Stopping strings apply properly in exllama with this fix.

extensions/api/util.py
Yiximail: I forgot to restore it, sorry

Yiximail:
@Ph0rk0z  @official-elinas 
Here, I created another branch and PR #2814 
can change to it in my [fork](https://github.com/Yiximail/text-generation-webui/tree/stopping-strings-for-HF)

As modifying generate_reply_HF involves changing the original logic, I separated them.

oobabooga:
Closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/2847. See https://github.com/oobabooga/text-generation-webui/pull/2814#issuecomment-1605457979

PR #2770, opened 2023-06-20
Use pre-compiled python module for ExLlama

jllllll:
I have converted ExLlama to a python module format akin to GPTQ-for-LLaMa. This PR allows for automatic installation and usage of the ExLlama module without needing to compile the code on supported systems.

### Advantages:
- Significantly easier usage for Windows users
- More control over potentially breaking changes to the code
- Control over when kernels are compiled. Ensuring updates are applied when needed.
### Disadvantages:
- No out-of-the-box support for AMD GPUs and ARM systems. Users will need to compile from source.
- ~Potential for slightly reduced performance without system-specific hardware optimizations (Untested)~
- ~This PR does not include backwards compatibility with the previous method of installation.~
  - This has been added to mitigate the previous disadvantages.
---
Source code: https://github.com/jllllll/exllama

I can transfer ownership of that repo or add you as a collaborator to allow you control over code and updates, if you'd prefer.


oobabooga:
In my tests, the performance seems to be exactly the same.

#### Current implementation (exllama compiled locally and imported from repositories/exllama):

Output generated in 7.85 seconds (25.49 tokens/s, 200 tokens, context 15, seed 2023648869)
Output generated in 5.81 seconds (28.22 tokens/s, 164 tokens, context 15, seed 497975221)
Output generated in 5.63 seconds (28.23 tokens/s, 159 tokens, context 15, seed 1306543314)
Output generated in 5.61 seconds (28.33 tokens/s, 159 tokens, context 15, seed 1086112971)
Output generated in 5.93 seconds (28.34 tokens/s, 168 tokens, context 15, seed 2043945061)

#### Current implementation (exllama compiled locally and imported from repositories/exllama):

Output generated in 7.72 seconds (25.92 tokens/s, 200 tokens, context 15, seed 1258178161)
Output generated in 7.02 seconds (28.49 tokens/s, 200 tokens, context 15, seed 1530402775)
Output generated in 5.94 seconds (28.30 tokens/s, 168 tokens, context 15, seed 35835583)
Output generated in 7.04 seconds (28.42 tokens/s, 200 tokens, context 15, seed 2136986222)
Output generated in 7.04 seconds (28.41 tokens/s, 200 tokens, context 15, seed 1050592726)

> This PR does not include backwards compatibility with the previous method of installation.

I don't think this matters. As long as it works it's fine.

The only question I have is about updates. I assume there is a play button somewhere that can be clicked to regenerate the wheels? If that is the case, I would appreciate becoming a collaborator to the repo just to be able to click that button now and then.

jllllll:
@oobabooga I'm in the process of adding basic backwards compatibility by adding the previous code as a fallback. I'll push the commit in a sec and can revert it if you want. I think this is needed at least for Linux ARM systems as I don't have wheels for them.
It will also make setting up AMD support a bit easier by just uninstalling the wheel.

Ph0rk0z:
I don't think it compiles on each run for me unless there are changes to the cuda code. Does it do something different on windows?

jllllll:
> I don't think it compiles on each run for me unless there are changes to the cuda code. Does it do something different on windows?

That's interesting. I saw someone mention something to this effect before, but chalked it up to them just compiling too fast to notice. On my Windows system, it compiles it to a temporary directory that gets deleted when the webui is closed. This necessitates me needing to compile it on every session and it gets pretty irritating as it takes a while on my older CPU.

I'll run some more tests in WSL to see what the behavior is there.

Ph0rk0z:
Seems it was too aggressive to re-compile on windows and not aggressive enough on linux. I recall having to delete kernels to get it to re-compile after some changes. Compiling after an update should be the correct behavior.

jllllll:
I have added a GitHub Pages wheel index and changed the requirements.txt to use it. This allows for automatic updating of the exllama wheel. A specific version can be selected with: `exllama==0.0.1`

PR #2776, opened 2023-06-20
Add an option to set low_cpu_mem_usage to False

qwqVictor:
Some models cannot be loaded with `low_cpu_mem_usage` set to `True`. Let's add an option.

Use `--no-low-cpu-mem-usage` to set it `False`.


oobabooga:
What model requires this set to false?

darkacorn:
id set false as default and enable it to true as param if needed .. that be the way better option.

oobabooga:
I haven't encountered any situation where low_cpu_mem_usage=False was needed. Without this, transformers uses twice more RAM to load the model.

PR #2786, opened 2023-06-20
Download optimizations

sofronaspe:
No need to check if output_folder exists when mkdir(exists_ok=True) will not throw an error even if output_folder exists.
Built the file metadata as a single string for one .write() call.
Made sha256_str with .join() instead of a for loop.

This change has been tested to produce results identical to the previous code.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -178,6 +178,8 @@ class ModelDownloader:
     def get_single_file(self, url, output_folder, start_from_scratch=False):
         filename = Path(url.rsplit('/', 1)[1])
         output_path = output_folder / filename
+        headers = {}
+        mode = 'wb'
         if output_path.exists() and not start_from_scratch:
             # Check if the file has already been downloaded completely
             r = self.s.get(url, stream=True, timeout=20)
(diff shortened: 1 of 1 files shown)


oobabooga:
Looks good! The only change that was needed was

```diff
-                    'branch: {branch}\n' \
-                    'download date: {datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")}\n'
-        sha256_str ='\n'.join([f'    {item[1]} {item[0]}' for item in sha256])
+                   f'branch: {branch}\n' \
+                   f'download date: {datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")}\n'
+        sha256_str = '\n'.join([f'    {item[1]} {item[0]}' for item in sha256])
```

Without this, literal `datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")` was being written to`huggingface-metadata.txt`.

I have merged the main branch after merging https://github.com/oobabooga/text-generation-webui/pull/2472. Everything seems to be working, but if you see any issues or if you can think of other improvements, feel free to PR.

PR #2790, opened 2023-06-21
More models (match more) & templates (starchat-beta, tulu)

matatonic:
Improve support for the following models:
* vicuna 1.3
* tulu
* starchat-beta
* clean up guanaco, alpacino, alpasta, lazarus, airoboros matching
* hypermantis
* open-llama-open-instruct:
* starcoder-gpteacher-code-instruct
* chronos
* samantha (actually match it, we have the template already)
* wizardcoder
* starchat-beta

I still think it's better to match models that have a custom format as they come, but if you disagree with any of the matching let me know which one and I can drop it.

### DIFF
--- a/characters/instruction-following/Starchat-Beta.yaml
+++ b/characters/instruction-following/Starchat-Beta.yaml
@@ -1,0 +1,4 @@
+user: "<|user|>"
+bot: "<|assistant|>"
+context: "<|system|>"
+turn_template: "<|user|> <|user-message|><|end|>\n<|bot|> <|bot-message|><|end|>\n"
--- a/characters/instruction-following/Starchat-Beta.yaml
+++ b/characters/instruction-following/Starchat-Beta.yaml
@@ -1,4 +1,4 @@
 user: "<|user|>"
 bot: "<|assistant|>"
-context: "<|system|>"
+context: "<|system|>\n"
 turn_template: "<|user|> <|user-message|><|end|>\n<|bot|> <|bot-message|><|end|>\n"
--- a/characters/instruction-following/Tulu.yaml
+++ b/characters/instruction-following/Tulu.yaml
@@ -1,0 +1,4 @@
+user: "<|user|>"
+bot: "<|assistant|>"
+context: ""
+turn_template: "<|user|>\n<|user-message|>\n<|bot|>\n<|bot-message|>\n"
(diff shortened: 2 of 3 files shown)


oobabooga:
Thanks for the PR. It's very useful to add more templates and regex; if issues appear due to bad matches we can fix those later.

About starchat, isn't there a \n after <|user|>?

> prompt_template = "<|system|>\n<|end|>\n<|user|>\n{query}<|end|>\n<|assistant|>"

https://huggingface.co/HuggingFaceH4/starchat-beta#intended-uses--limitations

matatonic:
Ugh, you might be right... but then again, so many model makers can't seem to get their own format right...
I was basing off this: https://huggingface.co/TheBloke/starchat-beta-GPTQ
But I'm not sure how to include the <|system|> context in this case... There are so many prompt formats now, with so many different fields, we may need a data definition format for templates, some even have multiple formats... 

for starchat-beta:
From the example: "<|system|>\n<|end|>\n<|user|>\n{query}<|end|>\n<|assistant|>"

From TheBloke: "<|system|> {system} <|end|>\n<|user|> {query} <|end|>\n<|assistant|>"
(note the extra spaces too, which I missed)

I'll do some more testing tomorrow and update you. If you want to merge as is now, I'll submit another PR tomorrow with any fixes.

This may be how things are going: https://huggingface.co/ausboss/llama-30b-superhotcot-4bit
They have many modes & formats... I like the format too. We would need to include some kind of either multi-modal format file (like a keyed list of prompt formats), or include a whole collection of files, like superhotcot-blog.yaml, superhotcot-summarize.yaml, etc.
I'd like to get it out there, it seems a cool model. The input types are sometimes odd though (story and chat have multiple inputs). Have you given though to how to support user defined fields? It's probably easiest via the API.

matatonic:
I did some testing with a lot of variations and this is working very well for me.

```
user: "<|user|>"
bot: "<|assistant|>"
context: "<|system|>\n"
turn_template: "<|end|>\n<|user|>\n<|user-message|><|end|>\n<|bot|>\n<|bot-message|>"
```

```yaml
  params:
    temperature: 0.2
    best_of: 50
    top_p: 0.95
```

Screenshot from 2023-06-21 11-15-37

PR #2793, opened 2023-06-21
Chat printing

### LINKED ISSUE #2745: Ability to print chat
**Description**

When you chat with the bot and send pictures, and use Stable Diffusion to generate images in responses you get a nicely ilustrated story. Sadly, if you try to invoke print dialog from the browser, it tries to print the whole UI with the chat part cut off.

It would be nice to have have a Print Chat button which would invoke browser print dialog just for the chat `div` with images stretched to full page width (A4 or Letter), and without surrounding UI. The idea is to actually be able to use Print to PDF to save the whole chat with photos into a PDF document.

I think this could be even hacked together with Tampermonkey script (I did similar thing for one website), but I am not so familiar with Gradio that much so it would take me more time to figure it out than if someone who knows it went in and added a print button.

missionfloyd:
This makes the chat box print correctly.

Fixes #2745

### DIFF
--- a/css/chat.css
+++ b/css/chat.css
@@ -93,3 +93,22 @@ div.svelte-362y77>*, div.svelte-362y77>.form>* {
 .message-body :not(pre) > code {
     white-space: normal !important;
 }
+
+@media print {
+    body {
+        visibility: hidden;
+    }
+
+    .chat {
+        visibility: visible;
+        position: absolute;
+        left: 0;
+        top: 0;
+        max-width: none;
+        max-height: none;
+        width: 100%;
+        height: fit-content;
+        display: flex;
+        flex-direction: column-reverse;
+    }
+}


oobabooga:
That's really nice, thank you

levicki:
@missionfloyd 
> This makes the chat box print correctly.

Sadly it doesn't.

I just tested this with a longer saved chat (which is loaded on startup because of default character configured in `settings.json`) with some embedded images sent to and from the bot, and it only generates 2 pages for printing out of more than 20 pages that should be there.

My interface settings:
image
image

Let me know if you need any more details.

If you manage to figure out what the problem is I'd also suggest setting `.message` to have `break-insde: avoid;` to avoid splitting messages from user and bot in the middle (i.e. username and avatar on one page, text on the next).

Issue #2796, opened 2023-06-21
Downloading some models from Hugging Face causes a ZeroDivisionError


GauravB159:
### Describe the bug

Getting a ZeroDivisionError: float division by zero in oobabooga_linux/text-generation-webui/download-model.py”, line 154. 

The `total_size` is 0 by default which causes this I think. 

A possible fix would be to not use the progress bar at all if the `total_size` comes up to be 0. 

<img width="960" alt="image" src="https://github.com/oobabooga/text-generation-webui/assets/22005033/5cfdeece-0e87-4c99-9c26-dd1813428e78">

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Try loading the following HuggingFace model - TheBloke/WizardCoder-15B-1.0-GPTQ

### Screenshot

<img width="634" alt="image" src="https://github.com/oobabooga/text-generation-webui/assets/22005033/802587eb-70b3-4029-909f-875806e33bf9">

### Logs

```shell
[...]

PR #2797, opened 2023-06-21
Possible for the ZeroDivisionError when downloading a model

GauravB159:
Possible fix for #2796. It doesn't throw the error at least and the model loads fine. The progress bar is a bit glitched when it encounters a zero though.


oobabooga:
Thank you, the fix makes sense

PR #2799, opened 2023-06-21
Update bitsandbytes to 0.39.1

jllllll:
This update of bitsandbytes allows for older GPUs to use `load-in-4bit`.

I have not included changes to documentation in this PR as I am not sure of the extent of the compatibility, but I know Pascal works.

From bitsandbytes changelog:
```
Bug fixes:

    Fixed a bug where 8-bit models consumed twice the memory as expected after serialization

Deprecated:

    Kepler binaries (GTX 700s and Tesla K40/K80) are not longer provided via pip and need to be compiled from source.
    Kepler support might be fully removed in the future.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,8 +17,8 @@ tqdm
 scipy
 transformers==4.30.2
 git+https://github.com/huggingface/peft@03eb378eb914fbee709ff7c86ba5b1d033b89524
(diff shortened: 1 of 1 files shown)


oobabooga:
Very nice to hear that it works on all GPUs now. Also, nice that you managed to create a wheel for the new version.

PR #2800, opened 2023-06-21
fix tokenizer for `exllama` and `llamacpp`

Yiximail:
yesterday, I found this problem: https://github.com/oobabooga/text-generation-webui/pull/2757#issuecomment-1599171125

so, I try to fix it
then found that [`decode` in `text_generation.py`](https://github.com/oobabooga/text-generation-webui/blob/89fb6f9236849a66920e20716240343edc0f3d9d/modules/text_generation.py#L75) is not available both `exllama` and `llamacpp`

the reason is:
image
`shared.model` and `shared.tokenizer` is the same thing
and model doesn't has a `decode` function

- [x] fix `from_pretrained` outputs the correct tokenizer
image
- [x] Make all `exllama.py`, `llamacpp_model.py` and `text_generation.py`'s encode output the same type ( same as `llamacpp` before )
image

- [x]  Make all `exllama.py`, `llamacpp_model.py` and `text_generation.py`'s decode output is string ( same as before )

Considering that the `encode` is mostly used only to count the number of tokens, this PR basically does not change any logic
Implemented the decode part of the function

### DIFF
--- a/modules/exllama.py
+++ b/modules/exllama.py
@@ -1,4 +1,5 @@
 import sys
+import numpy as np
 from pathlib import Path
 
 from modules import shared
@@ -50,7 +51,7 @@ class ExllamaModel:
         result.cache = cache
         result.tokenizer = tokenizer
         result.generator = generator
-        return result, result
+        return result, tokenizer
 
     def generate_with_streaming(self, prompt, state):
         self.generator.settings.temperature = state['temperature']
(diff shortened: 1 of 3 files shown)


bdashore3:
+1 This works nicely. Thank you!

oobabooga:
I like these changes. I'm only worried about 

```
tokens = np.array(tokens.tolist())
```

in the `decode` function. For transformers, I think that this moves the output tokens from the GPU to the CPU. Could there be an edge case where this causes problems somewhere else in the code?

Yiximail:
> 

Sorry, I don't quite understand this part

Before this, I noticed that the llamacpp `encode` process was converted to `np.ndarray`
~so I think was done because not everyone has a GPU, but everyone has a CPU for sure, just guess~

So I converted the `encode` results to `np.ndarray`, and turn the default parameter type of the `decode` as `np.ndarray`

PR #2777, opened 2023-06-20
Implement a demo HF wrapper for exllama to utilize existing HF transformers decoding.

Larryvrh:
This reuse of HF decoding makes the generation result from exllama more "aligned" with the original gptq outputs. This implementation is just a demo version and may need more work to increase the compatibility in all aspects.

PR #2803, opened 2023-06-21
Fix Multi-GPU not working on exllama_hf

Panchovix:
After https://github.com/oobabooga/text-generation-webui/pull/2777, new exllama_hf doesn't support gpu_split, since it wasn't added in the code. 

It would try to load all on the GPU0, and then get:

```
  File "F:\ChatIAs\oobabooga\text-generation-webui\server.py", line 62, in load_model_wrapper
    shared.model, shared.tokenizer = load_model(shared.model_name, loader)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 66, in load_model
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 285, in ExLlama_HF_loader
    return ExllamaHF.from_pretrained(model_name)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\exllama_hf.py", line 82, in from_pretrained
    return ExllamaHF(config)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\exllama_hf.py", line 27, in __init__
    self.ex_model = ExLlama(self.ex_config)
  File "F:\ChatIAs\oobabooga\text-generation-webui\repositories\exllama\model.py", line 711, in __init__
    tensor = tensor.to(device, non_blocking = True)
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 86.00 MiB. GPU 0 has a total capacty of 23.99 GiB of which 0 bytes is free. Of the allocated memory 22.92 GiB is allocated by PyTorch, and 39.80 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

This adds the configuration to let it load in more than 1 GPU. Now it loads without issues on 2x4090.

Also, since last PR, fixes the next issue.

```
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\callbacks.py", line 74, in gentask
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\text_generation.py", line 262, in generate_with_callback
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 1572, in generate
    return self.sample(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 2632, in sample
    next_token_scores = logits_processor(input_ids, next_token_logits)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\logits_process.py", line 92, in __call__
    scores = processor(input_ids, scores)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\logits_process.py", line 203, in __call__
    score = torch.gather(scores, 1, input_ids)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argument for argument index in method wrapper_CUDA_gather)
```

~~Speeds are reduced a bit (15 tokens/s to 12-13 tokens/s at max context on 65B).~~ Fixed after latest exllama commits.

Closes https://github.com/oobabooga/text-generation-webui/issues/2809

### DIFF
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -38,8 +38,6 @@ class ExllamaHF(PreTrainedModel):
 
     @property
     def device(self) -> torch.device:
-        # TODO: May cause problem on multi-gpu inference?
-        # TODO: Using device('cuda') doesnt fix this. I suspect the problem comes from here.
         return torch.device(0)
 
     def __call__(self, *args, **kwargs):
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -78,7 +78,7 @@ class ExllamaHF(PreTrainedModel):
         
         # This slowes down a bit but align better with autogptq generation.
         # TODO: Should give user choice to tune the exllama config
-        config.fused_attn = False
-        config.fused_mlp_thd = 0
+        # config.fused_attn = False
+        # config.fused_mlp_thd = 0
 
         return ExllamaHF(config)
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -39,6 +39,7 @@ class ExllamaHF(PreTrainedModel):
     @property
     def device(self) -> torch.device:
         # TODO: May cause problem on multi-gpu inference?
+        # TODO: Using device('cuda') doesnt fix this. I suspect the problem comes from here.
         return torch.device(0)
 
     def __call__(self, *args, **kwargs):
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -25,8 +25,8 @@ class ExllamaHF(PreTrainedModel):
         super().__init__(PretrainedConfig())
         self.ex_config = config
         if shared.args.gpu_split:
-            config.set_auto_map(shared.args.gpu_split)
-            config.gpu_peer_fix = True
+            self.ex_config.set_auto_map(shared.args.gpu_split)
+            self.ex_config.gpu_peer_fix = True
         self.ex_model = ExLlama(self.ex_config)
         self.generation_config = GenerationConfig()
 
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -78,7 +78,6 @@ class ExllamaHF(PreTrainedModel):
         
         # This slowes down a bit but align better with autogptq generation.
         # TODO: Should give user choice to tune the exllama config
-        config.act_order = True
         config.fused_attn = False
         config.fused_mlp_thd = 0
 
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -24,6 +24,9 @@ class ExllamaHF(PreTrainedModel):
     def __init__(self, config: ExLlamaConfig):
         super().__init__(PretrainedConfig())
         self.ex_config = config
+        if shared.args.gpu_split:
+            config.set_auto_map(shared.args.gpu_split)
+            config.gpu_peer_fix = True
         self.ex_model = ExLlama(self.ex_config)
         self.generation_config = GenerationConfig()
 
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -51,7 +51,7 @@ class ExllamaHF(PreTrainedModel):
         if cache is None:
             cache = ExLlamaCache(self.ex_model)
             self.ex_model.forward(torch.tensor([seq[:-1]], dtype=torch.long), cache, preprocess_only=True)
-        logits = self.ex_model.forward(torch.tensor([seq[-1:]], dtype=torch.long), cache).to(self.device)
+        logits = self.ex_model.forward(torch.tensor([seq[-1:]], dtype=torch.long), cache).to(kwargs['input_ids'].device)
         return CausalLMOutputWithPast(logits=logits, past_key_values=cache if use_cache else None)
 
     @classmethod
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -24,9 +24,6 @@ class ExllamaHF(PreTrainedModel):
     def __init__(self, config: ExLlamaConfig):
         super().__init__(PretrainedConfig())
         self.ex_config = config
-        if shared.args.gpu_split:
-            self.ex_config.set_auto_map(shared.args.gpu_split)
-            self.ex_config.gpu_peer_fix = True
         self.ex_model = ExLlama(self.ex_config)
         self.generation_config = GenerationConfig()
 
@@ -75,6 +72,10 @@ class ExllamaHF(PreTrainedModel):
         assert weight_path is not None, f'could not find weight in "{pretrained_model_name_or_path}"'
 
         config.model_path = str(weight_path)
+
+        if shared.args.gpu_split:
+            config.set_auto_map(shared.args.gpu_split)
+            config.gpu_peer_fix = True
         
         # This slowes down a bit but align better with autogptq generation.
         # TODO: Should give user choice to tune the exllama config


Panchovix:
@Ph0rk0z Sorry to ping you here, but yoy may know why this may happen? It seems to be a transformer issue when using gpu-split?

Larryvrh:
> @Ph0rk0z Sorry to ping you here, but you may know why this may happen? It seems to be a transformer issue when using gpu-split?
> 
> @turboderp Sorry for pinging you as well, you may have an idea why this could happen? I know it's kinda out of scope of your own --gs parameter if the error is related to transformers. But maybe I'm missing something on how are the tensors applied to each GPU.
> 
> @Larryvrh Since you implemented the exllama_hf variant, you may know how to fix it? I fixed to load evenly on more than 1 GPU, but the issue happens when doing inference.

The ```def device(self)``` part is most likely the problem, but I cannot test it for now since I have only one gpu in my machine. I may try dig into it lately by renting a multi-gpu instance and see what's going on.

Larryvrh:
Haven't had time to test this, but just an idea that you may want to try change line 55 from
```python
logits = self.ex_model.forward(torch.tensor([seq[-1:]], dtype=torch.long), cache).to(self.device)
````
to something like this, 
```python
logits = self.ex_model.forward(torch.tensor([seq[-1:]], dtype=torch.long), cache).to(kwargs['input_ids'].device)
```

Panchovix:
@Larryvrh don't worry, I can test everything that you suggest with my system!

Just tested your code and it worked!

Gonna add it to the PR.

Speeds are a bit reduced (15 tokens/s to 12 tokens/s), but it is a pretty good trade,

@oobabooga can confirm that now it works.

turboderp:
25% drop in performance is a little drastic.

I've just made two commit that could be relevant. First, if you want to keep input_ids in device memory and get output logits on the same device, which is probably more in line with Transformers pipelines, you can do that now. If so it makes sense to set `model.device_map.embed_tokens = "cuda:0"` to avoid moving the input IDs to host memory at the start of the forward pass. It will waste about 400 MB of VRAM but it skips a synchronization.

The second commit is probably a saner way of using the implementation. You'd want to produce input IDs in host memory (that's where they start in any case, even if you immediately call `.cuda()` on the tokenized prompt), but give `output_device` as an argument to `model.forward()`. Otherwise you get two synchronization events at the end of the forward pass: one to move the logits to host memory and one to move them back again immediately.

Sync events end up being very expensive in PyTorch, because it normally relies on always having a long CUDA queue running in the background. Once the queue runs out, or you force it by synchronizing (which is forced when moving tensors between devices or to host memory), you start to notice that Python is actually a really slow, interpreted language.

Ph0rk0z:
This wrapper also turns off fused attention and things like that. Is it required for the wrapper to function? Maybe that is partially where the speed drop comes from. I've yet to test whether these things can be left on or off.

It is good multi-gpu is now working. Exllama is the only thing with decent speeds using that.

Larryvrh:
> This wrapper also turns off fused attention and things like that. Is it required for the wrapper to function? Maybe that is partially where the speed drop comes from. I've yet to test whether these things can be left on or off.
> 
> It is good multi-gpu is now working. Exllama is the only thing with decent speeds using that.

The "fused_attn" and "fused_mlp_thd" isn't required at all. It's there simply because I found it can help solve turboderp/exllama#73 in some cases. I think most people won't need this fix since the problem isn't really noticable except when you are using greedy decoding, and this fix doesn't solve it completely, so maybe we should better remove this tuning.

Ph0rk0z:
I made it a setting for myself because I have quant_attn and fused_mlp as args for every loader that supports it. The act_order is more interesting because that is supposed be chosen automatically by exllama along with other features like groups, etc. Now that I see what it's about it really should be a parameter.

Larryvrh:
> 25% drop in performance is a little drastic.
> 
> I've just made two commits that could be relevant. First, if you want to keep input_ids in device memory and get output logits on the same device, which is probably more in line with Transformers pipelines, you can do that now. If so it makes sense to set `model.device_map.embed_tokens = "cuda:0"` to avoid moving the input IDs to host memory at the start of the forward pass. It will waste about 400 MB of VRAM but it skips a synchronization.
> 
> The second commit is probably a saner way of using the implementation. You'd want to produce input IDs in host memory (that's where they start in any case, even if you immediately call `.cuda()` on the tokenized prompt), but give `output_device` as an argument to `model.forward()`. Otherwise you get two synchronization events at the end of the forward pass: one to move the logits to host memory and one to move them back again immediately.
> 
> Sync events end up being very expensive in PyTorch, because it normally relies on always having a long CUDA queue running in the background. Once the queue runs out, or you force it by synchronizing (which is forced when moving tensors between devices or to host memory), you start to notice that Python is actually a really slow, interpreted language.

This new argument will definitely help the performance and is much better than my naive implementation, thank you for helping!

Ph0rk0z:
Need to figure out why act order isn't detected. Imo, that's not great. I have models that go both ways.

turboderp:
>Yea, the act_order indeed should be auto detected. It's just I found that it won't detect the model I'm using (TheBloke/tulu-30B-GPTQ) as act_order enabled, so I have to "fix" it manually, and this manual fix thing does help the ppl drop by a very little amount.

I'm very curious about how you're fixing it, and what sort of ppl drop you're getting as a result.

The thing is, even though TheBloke's recent models are all listed as act-order, they do not contain a group index. The tensor simply isn't present in the .safetensors file. In terms of doing inference on a model, what act-order *means* is that the model has a group index. So if there's no group index, it's simply not an act-order model as far as ExLlama is concerned. You can insist that it is by forcing it in the model config, but there's nothing ExLlama can do with that information.

Now, some of the most recent conversions from TheBloke *do* contain a group index tensor, but it's empty, all zeros. So it has literally no effect and ExLlama ignores it.

Also I've noticed some of these models being listed as "1g", suggesting a groupsize of 1. That's not the case either. Rather, they have *no* groupsize. A groupsize of 1 would imply that every row of the quantized matrices has its own set of quantization parameters, which would make the model larger than the FP16 original. It would also make act-order redundant.

Ph0rk0z:
The 1g means -1g but the dash get's lost.

Larryvrh:
> 
If that's the thing then we can safely remove the manual act_order setting. The drop in ppl mentioned is very small, about 0.001 for 30B model. I have no idea about the underlying stuff, but now I realized that it may be just some small randomness during calculation and has nothing to do with the act_order config thing.

oobabooga:
In line with the comments above, I have commented those two lines:

```
        # config.fused_attn = False
        # config.fused_mlp_thd = 0
```

This seems to have made things a tiny bit faster, about 2%. If I got things wrong and someone disagrees with this change, let me know.

While everyone is here, one outstanding problem with `exllama_hf` is that it fails to load some models as described by @Panchovix here: https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909. I can't load `llama-30b-4bit-128g` or `llama-7b-4bit-128g` (it keeps loading forever), but `WizardLM-30B-GPTQ` (with 4-bit and 128g) works.  No idea why.

Also, it would be nice to have `exllama_hf` working with LoRA somehow. Then we could have good speeds + transformers samplers + LoRA at the same time.

turboderp:
I'm a little curious as to which samplers from Transformers people are missing in ExLlama.

oobabooga:
Give me a few hours and I'll have a good idea on exactly what parameters are useful based on [preset arena](https://oobabooga.github.io/arena/preliminary-results.html). I have been using `exllama_hf` since  @Larryvrh's PR because it allows me to test good ranking presets like

```
do_sample: true
epsilon_cutoff: 1.49
eta_cutoff: 10.42
repetition_penalty: 1.17
temperature: 1.31
top_a: 0.52
top_k: 49
top_p: 0.14
```

Before that I had been happily using exllama with 

```
top_p: 0.1
top_k: 40
temperature: 0.7
repetition_penalty: 1.18
```

with good results on WizardLM 30B. I think that quite possibly nothing but temperature/top_p/top_k/repetition_penalty is really necessary, but we will see. 

A concern I have is that the repetition_penalty definition in exllama is different from the one in transformers. Is there a way to make it equivalent?

Panchovix:
> I'm a little curious as to which samplers from Transformers people are missing in ExLlama.

In my case, I like to use eta_cutoff, epsilon_cutoff and tail free sampling. I have experimented with top_a sometimes but haven't got how to make it work good since fp16 models were the only way to run these, but probably ton of people use top_a as well.

I feel since typical and rep penalty was added into exllama itself, it improved a lot. (Which I often use as well)

turboderp:
>A concern I have is that the repetition_penalty definition in exllama is different from the one in transformers. Is there a way to make it equivalent?

I actually thought it was, but then I took a closer look and it looks like their implementation is slightly different. For very improbable tokens it multiplies instead of dividing by the penalty. Which makes sense, of course, where the logits are negative. I changed it in the latest commit to work the same way and verified that the adjusted logits are identical. It should also be a little faster this way.

Larryvrh:
> While everyone is here, one outstanding problem with `exllama_hf` is that it fails to load some models as described by @Panchovix here: [#2777 (comment)](https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909). I can't load `llama-30b-4bit-128g` or `llama-7b-4bit-128g` (it keeps loading forever), but `WizardLM-30B-GPTQ` (with 4-bit and 128g) works. No idea why.

I just tested it and it seems like "Neko-Institute-of-Science/LLaMA-7B-4bit-128g" is loaded with no problem on my machine. Since this is just a very thin wrapper around exllama, I don't really have any clue why the loading failed for many people. Maybe related to the driver thing oobabooga/text-generation-webui#2820?

Panchovix:
> > While everyone is here, one outstanding problem with `exllama_hf` is that it fails to load some models as described by @Panchovix here: [#2777 (comment)](https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909). I can't load `llama-30b-4bit-128g` or `llama-7b-4bit-128g` (it keeps loading forever), but `WizardLM-30B-GPTQ` (with 4-bit and 128g) works. No idea why.
> 
> I just tested it and it seems like "Neko-Institute-of-Science/LLaMA-7B-4bit-128g" is loaded with no problem on my machine. Since this is just a very thin wrapper around exllama, I don't really have any clue why the loading failed for many people. Maybe related to the driver thing #2820?

I'm on latest drivers, so maybe it's because that? Not sure which drivers is @oobabooga using atm. The good (and bad) thing about latest drivers is uses RAM when VRAM is full. For example, yesterday I could do a 65b model with about 3400 context, using 48GB VRAM + 4GB RAM.

Panchovix:
@Larryvrh to add, just tested an older driver (older than (532.xx, where those issues started) and I still get issues. on some models

```
2023-06-23 02:40:10 INFO:Loading h2oai_h2ogpt-research-oasst1-llama-65b-4bit...
2023-06-23 02:40:40 ERROR:Failed to load the model.
  File "F:\ChatIAs\oobabooga\text-generation-webui\server.py", line 62, in load_model_wrapper
    shared.model, shared.tokenizer = load_model(shared.model_name, loader)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 74, in load_model
    tokenizer = load_tokenizer(model_name, model)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 109, in load_tokenizer
    tokenizer = AutoTokenizer.from_pretrained(path_to_model, trust_remote_code=shared.args.trust_remote_code)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\auto\tokenization_auto.py", line 691, in from_pretrained
    return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1825, in from_pretrained
    return cls._from_pretrained(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1988, in _from_pretrained
    tokenizer = cls(*init_inputs, **init_kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\llama\tokenization_llama_fast.py", line 93, in __init__
    super().__init__(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_fast.py", line 114, in __init__
    fast_tokenizer = convert_slow_tokenizer(slow_tokenizer)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 1307, in convert_slow_tokenizer
    return converter_class(transformer_tokenizer).converted()
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 445, in __init__
    from .utils import sentencepiece_model_pb2 as model_pb2
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\utils\sentencepiece_model_pb2.py", line 91, in <module>
    _descriptor.EnumValueDescriptor(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\google\protobuf\descriptor.py", line 796, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

Then, after doing what it suggests, https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909 happens.

Can someone try (if 2x24 GB VRAM GPUs) this model? https://huggingface.co/BGLuck/h2ogpt-research-oasst1-llama-65b-4bit

I did the quant myself, and it works on GPTQ for llama, AutoGPTQ, exllama (pure exllama, also works on KoboldAI/ooba web ui). But it doesn't work with exllama_hf.

Larryvrh:
> @Larryvrh to add, just tested an older driver (older than (532.xx, where those issues started) and I still get issues. on some models
> 
> ```
> To create a public link, set `share=True` in `launch()`.
> 2023-06-23 02:40:10 INFO:Loading h2oai_h2ogpt-research-oasst1-llama-65b-4bit...
> 2023-06-23 02:40:40 ERROR:Failed to load the model.
>   File "F:\ChatIAs\oobabooga\text-generation-webui\server.py", line 62, in load_model_wrapper
>     shared.model, shared.tokenizer = load_model(shared.model_name, loader)
>   File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 74, in load_model
>     tokenizer = load_tokenizer(model_name, model)
>   File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 109, in load_tokenizer
>     tokenizer = AutoTokenizer.from_pretrained(path_to_model, trust_remote_code=shared.args.trust_remote_code)
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\auto\tokenization_auto.py", line 691, in from_pretrained
>     return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1825, in from_pretrained
>     return cls._from_pretrained(
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1988, in _from_pretrained
>     tokenizer = cls(*init_inputs, **init_kwargs)
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\llama\tokenization_llama_fast.py", line 93, in __init__
>     super().__init__(
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_fast.py", line 114, in __init__
>     fast_tokenizer = convert_slow_tokenizer(slow_tokenizer)
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 1307, in convert_slow_tokenizer
>     return converter_class(transformer_tokenizer).converted()
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 445, in __init__
>     from .utils import sentencepiece_model_pb2 as model_pb2
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\utils\sentencepiece_model_pb2.py", line 91, in <module>
>     _descriptor.EnumValueDescriptor(
>   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\google\protobuf\descriptor.py", line 796, in __new__
>     _message.Message._CheckCalledFromGeneratedFile()
> TypeError: Descriptors cannot not be created directly.
> If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
> If you cannot immediately regenerate your protos, some other possible workarounds are:
>  1. Downgrade the protobuf package to 3.20.x or lower.
>  2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
> 
> More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
> ```
> 
> Then, after doing what it suggests, [#2777 (comment)](https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909) happens.
> 
> Can someone try (if 2x24 GB VRAM GPUs, or single >48 GB VRAM GPU) this model? https://huggingface.co/BGLuck/h2ogpt-research-oasst1-llama-65b-4bit
> 
> I did the quant myself, and it works on GPTQ for llama, AutoGPTQ, exllama (pure exllama, also works on KoboldAI/ooba web ui). But it doesn't work with exllama_hf.

Oh, I see, the problem is the HF fast tokenizer thing. You may want to try changing modules/models.py line 285 from
```python
return ExllamaHF.from_pretrained(model_name)
```
to
```python
return ExllamaHF.from_pretrained(model_name), AutoTokenizer.from_pretrained(str(Path(f'{shared.args.model_dir}') / Path(model_name)), use_fast = False)
```
I haven't tested the code my self, and there may be some trivial errors such as import or model path, but they should be very easy to fix.

Panchovix:
> > @Larryvrh to add, just tested an older driver (older than (532.xx, where those issues started) and I still get issues. on some models
> > ```
> > To create a public link, set `share=True` in `launch()`.
> > 2023-06-23 02:40:10 INFO:Loading h2oai_h2ogpt-research-oasst1-llama-65b-4bit...
> > 2023-06-23 02:40:40 ERROR:Failed to load the model.
> > Traceback (most recent call last):
> >   File "F:\ChatIAs\oobabooga\text-generation-webui\server.py", line 62, in load_model_wrapper
> >     shared.model, shared.tokenizer = load_model(shared.model_name, loader)
> >   File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 74, in load_model
> >     tokenizer = load_tokenizer(model_name, model)
> >   File "F:\ChatIAs\oobabooga\text-generation-webui\modules\models.py", line 109, in load_tokenizer
> >     tokenizer = AutoTokenizer.from_pretrained(path_to_model, trust_remote_code=shared.args.trust_remote_code)
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\auto\tokenization_auto.py", line 691, in from_pretrained
> >     return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1825, in from_pretrained
> >     return cls._from_pretrained(
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_base.py", line 1988, in _from_pretrained
> >     tokenizer = cls(*init_inputs, **init_kwargs)
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\models\llama\tokenization_llama_fast.py", line 93, in __init__
> >     super().__init__(
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\tokenization_utils_fast.py", line 114, in __init__
> >     fast_tokenizer = convert_slow_tokenizer(slow_tokenizer)
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 1307, in convert_slow_tokenizer
> >     return converter_class(transformer_tokenizer).converted()
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\convert_slow_tokenizer.py", line 445, in __init__
> >     from .utils import sentencepiece_model_pb2 as model_pb2
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\utils\sentencepiece_model_pb2.py", line 91, in <module>
> >     _descriptor.EnumValueDescriptor(
> >   File "F:\ChatIAs\oobabooga\venv\lib\site-packages\google\protobuf\descriptor.py", line 796, in __new__
> >     _message.Message._CheckCalledFromGeneratedFile()
> > TypeError: Descriptors cannot not be created directly.
> > If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
> > If you cannot immediately regenerate your protos, some other possible workarounds are:
> >  1. Downgrade the protobuf package to 3.20.x or lower.
> >  2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
> > 
> > More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, after doing what it suggests, [#2777 (comment)](https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909) happens.
> > Can someone try (if 2x24 GB VRAM GPUs, or single >48 GB VRAM GPU) this model? https://huggingface.co/BGLuck/h2ogpt-research-oasst1-llama-65b-4bit
> > I did the quant myself, and it works on GPTQ for llama, AutoGPTQ, exllama (pure exllama, also works on KoboldAI/ooba web ui). But it doesn't work with exllama_hf.
> 
> Oh, I see, the problem is the HF fast tokenizer thing. You may want to try changing modules/models.py line 285 from
> 
> ```python
> return ExllamaHF.from_pretrained(model_name)
> ```
> 
> to
> 
> ```python
> return ExllamaHF.from_pretrained(model_name), AutoTokenizer.from_pretrained(str(Path(f'{shared.args.model_dir}') / Path(model_name)), use_fast = False)
> ```
> 
> I haven't tested the code my self, and there may be some trivial errors such as import or model path, but they should be very easy to fix. The HF tokenizer has really disgusting convert slow tokenizer implementation, it either hang for a long time before finish converting and give no progress message at all, making people believe something is stuck there, or simply crash due to those weird protobuf errors.

Okay, just tested and it works! All the models that did not work before now work, thanks! Made a PR with your fix.

image

turboderp:
Why use `AutoTokenizer` and not `LlamaTokenizer`?

Larryvrh:
> Why use `AutoTokenizer` and not `LlamaTokenizer`?

Because `AutoTokenizer` is the first tokenizer I saw in the "models.py" imports I guess? Are there any trivial differences between using these two? I think `AutoTokenizer.from_pretrained` will automatically dispatch to `LlamaTokenizer` after parsing the 
"tokenizer_config.json."

Ph0rk0z:
It should dispatch to llamatokenizer. Speaking of that, I wonder if I can load a different GPTQ model now using this and what transpires. There is neox, gpt-j, MPT, etc. Probably one that doesn't require custom scripts.

turboderp:
Well, I've run into the problem with AutoTokenizer taking forever to load, too. Apparently it tries to compile or optimize a "fast" version of something or other, but it seems like LlamaTokenizer doesn't have that problem.

Larryvrh:
> Well, I've run into the problem with AutoTokenizer taking forever to load, too. Apparently it tries to compile or optimize a "fast" version of something or other (failing miserably), but it seems like LlamaTokenizer doesn't have that problem.

Ah, I see, this explains why the current implementation stucks for exllama_hf, it doesn't see it as a Lllama model and therefore use AutoTokenizer without `use_fast=False` instead.

Ph0rk0z:
Maybe it's better for that then.

oobabooga:
Exllama_HF is working for all models after https://github.com/oobabooga/text-generation-webui/pull/2835

@turboderp going back to the topic of samplers: after analyzing 7215 votes comparing 241 presets, I found that some of the best overall presets were:

```
Instruct:

temperature: 1.31
top_p: 0.14
epsilon_cutoff: 1.49
eta_cutoff: 10.42
top_a: 0.52
repetition_penalty: 1.17
top_k: 49

temperature: 0.87
top_p: 0.99
typical_p: 0.68
tfs: 0.68
repetition_penalty: 1.01
top_k: 85

temperature: 0.7
top_p: 0.9
repetition_penalty: 1.15
top_k: 20

Chat:

temperature: 0.98
top_p: 0.37
repetition_penalty: 1.18
top_k: 100

temperature: 0.82
top_p: 0.21
repetition_penalty: 1.19
top_k: 72

temperature: 1.53
top_p: 0.64
top_a: 0.04
repetition_penalty: 1.07
top_k: 33
```

The one with the highest elo score across instruct/chat categories was simply

```
temperature: 0.7
top_p: 0.9
repetition_penalty: 1.15
top_k: 20
```

The experiment is far from conclusive due to the low data volume. These could all be overfits for all I know. But based on this data, it would seem like `top_a` and `tfs` are probably the most relevant parameters to add, followed by `epsilon_cutoff` and `eta_cutoff`. 

The first two are unconventional parameters that are not part of transformers. Their definitions can be found [here](https://github.com/oobabooga/text-generation-webui/blob/main/modules/sampler_hijack.py).

I find that Contrastive Search and Mirostat rank high but not that high (10th and 12th place for instruction following respectively). Eta sampling, by itself using the parameters from [the paper](https://arxiv.org/abs/2210.15191), did not perform well, but it appears along other parameters. It's possible that its affect is made null by the presence of `top_p` or `top_a`. I haven't checked. 

The full data is here: https://github.com/oobabooga/oobabooga.github.io/blob/main/arena/results.md

PR #2812, opened 2023-06-22
Update blocking_api.py

SebastianBodza:
Edit the response, so the API also works locally


oobabooga:
Already implemented here https://github.com/oobabooga/text-generation-webui/pull/2718/files

PR #2814, opened 2023-06-22
Replace `stopping_criteria` logic for `generate_reply_HF`

Yiximail:
Expanded features from [here](https://github.com/oobabooga/text-generation-webui/pull/2766)
someone needs it for `exllama_HF`

Many people are confused about the `stopping_strings` and `custom_stopping_strings`.
It is achieved by intercepting tokens, not words or strings. So the effect is different from what people want.

So I replaced the logic of `stopping_criteria` with character matching.

1. In [`generate_reply_HF`](https://github.com/oobabooga/text-generation-webui/blob/d94ea31d54a646fd998732fe12370ecc1ffdfb2c/modules/text_generation.py#L193), now it generates 1 token per iteration regardless of whether it is output as a stream or not.
(In theory, it may cause a bit performance overhead, but so little that can ignore. Generally, it is same as before.)

2. As I said, now any string can be directly matched for interruption. It will not output anything in stopping strings.
<details>
<summary>Some screenshots of the test:</summary>

same seeds and different stopping strings outputs:
image
image
image

</details>

3. It is possible to interrupt the generation in non-stream mode and immediately return the result.

### updated:
I tried multiple cases with or without exllama_HF loader and with or without stream (in single GPU, max context):

1.The VARM used is exactly the same, and in the same situation, the MiB numbers I measured are completely identical.

2.The GPU utilization dropped by around 1% (33b model)
my guess: String detection is handled by the CPU (in python), so at this time, the GPU will wait for a while. So theoretically, the performance will also decrease by 1% in this case.
A smaller models will cause the CPU to process at a faster frequency. Because the t/s is larger and requires processing each time.

Yes I confirmed that
there is a huge performance gap in the 7b model (like from 98% to 65% at gpu usage, and 61t/s to 44t/s at speed)

### updated:

<details>
<summary>I found that the main performance impact comes from get_reply_from_output_ids, I call it too many times</summary>

As the output gets longer, the more time it takes:

image

This issue should also exist in the case of a stream in the main branch

</details>

~I'm trying to solve this performance issue~

### updated:

Enhanced performance

Non-stream mode will be close to the original (63.3t/s to 60.7t/s)

image
image

Stream mode will be **better** than the original (44.9t/s to 51.6t/s)

image
image

Max content test:

Non-stream mode (44.5t/s to 41.5t/s)
image
image

Stream mode (36.9/s to 38.3t/s)
image
image

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -225,50 +225,78 @@ def generate_reply_HF(question, original_question, seed, state, eos_token=None, 
     if inputs_embeds is not None:
         generate_params.update({'inputs_embeds': inputs_embeds})
 
-    # Create the StoppingCriteriaList with the stopping strings (needs to be done after tokenizer extensions)
-    stopping_criteria_list = transformers.StoppingCriteriaList()
-    for st in (stopping_strings, ast.literal_eval(f"[{state['custom_stopping_strings']}]")):
-        if type(st) is list and len(st) > 0:
-            sentinel_token_ids = [encode(string, add_special_tokens=False) for string in st]
-            stopping_criteria_list.append(_SentinelTokenStoppingCriteria(sentinel_token_ids=sentinel_token_ids, starting_idx=len(input_ids[0])))
-            break
+    # combine the stopping strings
+    stopping_strings = stopping_strings or []
+    if type(state.get('stopping_strings')) == list:
+        stopping_strings.extend(state['stopping_strings'])
+    if type(state.get('custom_stopping_strings')) == str:
+        stopping_strings.extend(ast.literal_eval(f"[{state['custom_stopping_strings']}]"))
+    if state.get('stop_at_newline', False):
+        stopping_strings.append('\n')
+    stopping_strings = list(set(stopping_strings))
 
-    # Update generate_params with the eos token and the stopping strings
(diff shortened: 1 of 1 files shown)


Panchovix:
Just tested and I can confirm it works with exllama_hf, nice work!

oobabooga:
I'm not sure if using the streaming hack even when streaming is disabled is a good idea. It spawns a new thread, which I think may lead to higher VRAM usage.

official-elinas:
@oobabooga I'm not noticing any difference in VRAM usage, just that my tk/s dropped in half.

Panchovix:
I see same VRAM usage as well, speed is the same also.

Yiximail:
> 

dropped in half? 

Can you provide your computer configuration and the usage of CPU, memory, GPU and VRAM when running?
I'm not sure what caused it, but it's possible that one of them has reached a bottleneck.

T^T

oobabooga:
If we are going to enable streaming for the sake of checking stopping strings, it's easier to do that directly in the `generate_reply` function. This way, it works for all models.

This is done in https://github.com/oobabooga/text-generation-webui/pull/2847 in a simple way, so I'm closing this PR in favor of that one.

PR #2815, opened 2023-06-22
8k size only for minotaur-15B

matatonic:
Other minotaur have the same prompt format but only the 15B has 8k context, there is a minotoar landmark also, but perhaps it's better handling landmark more generally so I wont include it here.


matatonic:
@TheBloke @oobabooga Minor fixup, don't clobber other llama based minotaur with 8k context.

PR #2818, opened 2023-06-22
Lora fixes for AutoGPTQ

Ph0rk0z:
I had issues with AutoGPTQ lora using transformers because it was incorrectly detecting the loader that was used. It would dump errors in the console. Especially after a reload or hitting apply on a blank lora box.

Also:
`LlamaGPTQForCausalLM supports injecting fused attention but not enables this time. If you are training adapters, you must also disable fused attention injection when loading quantized base model at inference time, otherwise adapters may not be added to base model properly. If you are loading adapters to do inference, you can reference to adapter's config file to check whether the adapters are trained using base model that not enable fused attention injection.`

You should see this warning when a lora is loaded. AFAIK it doesn't work with fused attention still.

Only "questionable" thing is reloading the model when apply is pressed twice on an empty box but it is better than the alternatives. Model reload takes seconds once it is cached. Feel free to edit.


oobabooga:
I'm not a fan of checking `shared.args.loader == 'AutoGPTQ'` because that may be true by accident in some edge case. it's better to look at the model class directly. But if that fixed an issue for you, let's merge it

Ph0rk0z:
I did not check what happens if a user loads a model and then changes the loaders for some reason but never loads anything else.

Definitely needed warning for fused attention + autogtpq lora though. Not many people know that one unless they read autogptq source and it's on by default so none of the lora they apply will do anything.

PR #2828, opened 2023-06-23
Initial support for CTranslate2 format models

dragonfyre13:
Added streaming support, so figured it was worth a pull request. I can clean the repeated joins for python string handling up if preferred before pulling, but haven't yet since it's not the biggest resource sink in that loop and IMHO it's a bit more readable/debuggable this way. The change overall really doesn't touch much existing code, living separate from most of it though the library masquerades as a HF generator in some respects, but generation is separate because it acts different enough at generation time.

Basic justification is not just "supporting yet another model type" but where this particular model type (basically a type of quant library) is incredibly fast and has quite good support for inference including on Pascal architecture for the p40 (CUDA compute compat arch 6.1 specifically). I've been getting ~18t/s on wizardcoder-15B using a full 8192 context, using just under 17GB max (using CUDA, on a Tesla p40). It's one of the few approaches that stresses the cards memory usage to 70+%, which otherwise hovers around ~10% memory and 90% GPU with GPTQ and GGML offloaded to the GPU. Presumably this is because it's bogged down working through fp16 calculations (GPTQ gives me about ~8 t/s, after tweaking everything I could). It also doesn't seem to need drastic changes to the library when new types of models come out, support for new architectures tends to be added quickly and simply.

If interested in more, here's the WizardCoder model I was using: https://huggingface.co/michaelfeil/ct2fast-WizardCoder-15B-V1.0
More info on the library is here: https://opennmt.net/CTranslate2/quickstart.html
More on the specific int8 optimization for compute capability 6.1 (there's plenty of additional for <6 and >7 as well): https://opennmt.net/CTranslate2/quantization.html#implicit-type-conversion-on-load
There's a number of models converted for ct2 (https://huggingface.co/models?search=ct2), but it's also pretty easy to convert a full model so I've just been converting my own most of the time anyway. It provides support for pretty much everything, including falcon and MPT (https://opennmt.net/CTranslate2/guides/transformers.html), and provides "backwards compatibility guarantees" (https://opennmt.net/CTranslate2/versioning.html) for the API since it's targeted at production use.


Malrama:
@oobabooga Any thoughts on this? Looks like a nice addition as it does cover some cases some current loaders don't cover.

krzysiekpodk:
voting for this!

choskeli:
@oobabooga can this be merged?

oobabooga:
I have looked at this PR numerous times, and I always have a hard time figuring out the use case for CTranslate2. Two things stand out to me:

1) For P40 GPUs, llama.cpp is very popular, and I have never seen someone using CTranslate2 instead of llama.cpp for this hardware.
2) 4-bit quantization is not supported, while 4-bit is the sweet spot for LLM inference (it's better to run a 2x larger model in 4-bit precision than a smaller one in 8-bit precision).

I will close this PR for these two reasons. If CTranslate2 grows and gains a clearer edge, that can be revisited.

PR #2835, opened 2023-06-23
Fix some models not loading on exllama_hf

Panchovix:
Fixes some issues with some models when loading on exllama_hf.

Refer to https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1599418831 and https://github.com/oobabooga/text-generation-webui/pull/2777#issuecomment-1602150909.

Credits to @Larryvrh, Fix given from https://github.com/oobabooga/text-generation-webui/pull/2803#issuecomment-1603871674


BadisG:
image
I was hoping this PR would fix the issue for me but unfortunately exllama still doesn't want to load models :(
It works on WSL2 but not on regular windows I have no idea why, yesterday it was working fine.

oobabooga:
So it was the tokenizer all along. Nice find, @Panchovix and @Larryvrh!

I made a small change to use the existing code to load the LlamaTokenizer, the same used by autogptq and transformers.

PR #2838, opened 2023-06-23
buffer the reply which generated too soon

hao1939:
Now the reply generated too soon would be simply dropped.
The PR buffers the reply in a str, so we won't lost it.

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -202,6 +202,7 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False):
     seed = set_manual_seed(state['seed'])
     last_update = -1
     reply = ''
+    buf = ''  # buf the return value of generate_func
     is_stream = state['stream']
     if len(all_stop_strings) > 0 and not state['stream']:
         state['stream'] = True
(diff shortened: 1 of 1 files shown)


oobabooga:
What do you mean by reply generated too soon?

hao1939:
When there are two `reply`s generated faster than 24 fps, the first one would just be overrided, we lost it.

image

oobabooga:
There is a final `yield reply` at the end of the `_generate_reply` function that handles this case, so no change is needed.

hao1939:
I believe you don't understand the case.

for streaming, the while loop could execute multiple times before the final yield, which means some reply was just dropped.

if len(all_stop_strings) > 0 and not state['stream']:
        state['stream'] = True

hao1939:
I fix it because my model prints nothing in streaming mode, it takes hours to figure out the root cause.

hao1939:
let me describe the case in details:
1. supposing there are three replys coming in lsess than 0.041666666666666664.
2. the `if cur_time - last_update > 0.041666666666666664:` would be false, so the reply not yield
3. the the reply would overrided
4. once we reach the final yield, we can only see the lasted reply

hao1939:
@oobabooga just update the 0.041666666666666664 to a large num, you can easily reproduce the issue.

PR #2845, opened 2023-06-24
Organize chat history/character import menu

missionfloyd:
- Moved the labels to the controls themselves.
- Changed profile picture to `gr.Image()`.
- Moved TavernAI to a separate tab.
- Added a Submit button to TavernAI tab. I was dropping files then clicking Submit, and wondering why there were errors. Until I looked at the code and realized that just dropping the file did it.
- Submit buttons are only clickable if the corresponding file is loaded.
- Changed "Click me" button to "Refresh".

Untitled

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -598,7 +598,6 @@ def upload_character(json_file, img, tavern=False):
         f.write(json_file)
 
     if img is not None:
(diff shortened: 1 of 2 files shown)


oobabooga:
The chat tab looks much better now, thanks!

PR #2847, opened 2023-06-24
Implement stopping string search in string space

oobabooga:
* Fixes inconsistent behavior of stopping strings with transformers.
* Extends stopping string support for all backends.
* Simplifies the code base.

This required activating streaming in the background when stopping strings are present but streaming is disabled.

## Performance test

I have made a worst-case performance test with exllama running llama-7b-4bit-128g and the following 20 long stopping strings:

```
'KMS6pQgkfLCQqweZyiq2YYq4EGzO', 'QdvPyWGxEstCUxZElD6L1u2uRp4P', 'EreYMseEtqNm1MelOEhUXprpCeIB', 'Uk7zJGqCMJpzhCuHnJfxQmsB0guk', 'W2m5VXoZf3Kkyan4Iz1zLrIf7Jqu', 'kaLBRf18oEarXbqg3TtFjlUVQTo9', 'Sn0LWu7WtDZze6ZNnUn1qVwkoJxC', 'AUIuso0oYayGxHrpgskbacpRiY7W', 'YyYPo6aM8FvuDeWtgktUBlFrbc4C', 'hA3iEivUnmY6eiuiXlcMv1HbjUC3', 'yWMXiZXFTYO1V1kH5X0j1dZAbpYx', 'JukydBzYKWXZxF6zCu8tkaxSLKfb', 'XQmDrjzjk6WL6gETRIsKYlPgaACk', 'eiDVfRyh3tA99I6gJY0eAUmyy0EU', 'pSAXQRUa4HdDPZEqShvmvRfcwjt4', 'aH05cdV888YSiAvQtODEYXAXr9aC', 'WQHm3IwrfXJBjU5ZaFXxrJf3L7Jb', 'Kj9TbuWW8tTOE7UuPheEUSvvRNAB', '2Tcfe1MDpACccrWOWmRZDDdsMgp6', 'SxtVdMo7zJTrsHD3VQfWROlAdJeX'
```

#### With the stopping strings

Output generated in 3.58 seconds (111.88 tokens/s, 400 tokens, context 6, seed 17)
Output generated in 3.58 seconds (111.88 tokens/s, 400 tokens, context 6, seed 17)
Output generated in 3.58 seconds (111.81 tokens/s, 400 tokens, context 6, seed 17)

#### Without the stopping strings

Output generated in 3.56 seconds (112.30 tokens/s, 400 tokens, context 6, seed 17)
Output generated in 3.57 seconds (112.11 tokens/s, 400 tokens, context 6, seed 17)
Output generated in 3.56 seconds (112.22 tokens/s, 400 tokens, context 6, seed 17)

-----

So the result is that in the worst case (many tokens/second, many stopping strings, long stopping strings), this implementation causes a reduction of 0.3% in the number of tokens/second. In practice it will be less than that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,4 +350,4 @@ The presets that are included by default are the result of a contest that receiv
 
 - Gradio dropdown menu refresh button, code for reloading the interface: https://github.com/AUTOMATIC1111/stable-diffusion-webui
 - Godlike preset: https://github.com/KoboldAI/KoboldAI-Client/wiki/Settings-Presets
-- Code for early stopping in chat mode, code for some of the sliders: https://github.com/PygmalionAI/gradio-ui/
+- Code for some of the sliders: https://github.com/PygmalionAI/gradio-ui/
--- a/modules/callbacks.py
+++ b/modules/callbacks.py
@@ -9,33 +9,6 @@ import transformers
 import modules.shared as shared
 
 
-class _SentinelTokenStoppingCriteria(transformers.StoppingCriteria):
-
-    def __init__(self, sentinel_token_ids: list, starting_idx: int):
-        transformers.StoppingCriteria.__init__(self)
-        self.sentinel_token_ids = sentinel_token_ids
-        self.starting_idx = starting_idx
-        self.shortest = min([x.shape[-1] for x in sentinel_token_ids])
-
-    def __call__(self, input_ids: torch.LongTensor, _scores: torch.FloatTensor) -> bool:
-        for sample in input_ids:
-            trimmed_sample = sample[self.starting_idx:]
-            trimmed_len = trimmed_sample.shape[-1]
-            if trimmed_len < self.shortest:
-                continue
-
-            for sentinel in self.sentinel_token_ids:
-                sentinel_len = sentinel.shape[-1]
-                if trimmed_len < sentinel_len:
-                    continue
-
-                window = trimmed_sample[-sentinel_len:]
-                if torch.all(torch.eq(sentinel, window)):
-                    return True
-
-        return False
-
-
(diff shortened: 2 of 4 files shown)


Yiximail:
Hmm, the gap still exists
how much performance drop depends on cpu performance

### With the stopping strings
image

### Without the stopping strings
image

oobabooga:
@Yiximail how many stopping strings are you using? It's strange that the difference is so large for you. I have an i5-12400F, so the performance shouldn't be any different.

Yiximail:
> @Yiximail how many stopping strings are you using? It's strange that the difference is so large for you. I have an i5-12400F, so the performance shouldn't be any different.

only one, In fact it's just the difference between having and not having

image
image

when I set the stopping strings, one of the cpu cores will be fully loaded
and the gpu usage decreases with the generation length

in my test
slowing down the gpu is not the detection of stopping strings but `get_reply_from_output_ids` (mainly the decode part)
whenever a token is generated, the entire output is going to re-decoded


You tried it in 12400f? Strange, I have a visible performance drop

oobabooga:
Is that autogptq, exllama, or llama.cpp?

Yiximail:
> Is that autogptq, exllama, or llama.cpp?

here? in exllama_HF

Yiximail:
My previous solution was to decode only once per frame and decode again after interruption(to prevent outputting without updating the response)
That will greatly reduce the pressure on the CPU, but not good enough.

The best approach would be to leverage other cores.
As currently, only one core is handling both iteration and decoding.
image

official-elinas:
@oobabooga Stopping strings don't work in exllama or exllama_hf with extended context.

PR #2848, opened 2023-06-24
Make stop_everything also work with non-streamed generation

happyme531:
This is implemented for better https://github.com/oobabooga/text-generation-webui/pull/2392#issue-1729389261 support.


oobabooga:
That's creative, thank you!

PR #2859, opened 2023-06-25
more models: +orca_mini

matatonic:
* orca mini, https://huggingface.co/psmathur/orca_mini_13b


matatonic:
I blinked and it was done! Faster than I could merge my own dev branch, lol

PR #2860, opened 2023-06-25
[WIP] remove autoload feature

matatonic:
* There is a UI glitch where autoload gets triggered on focus change in the window
* Autoload is often accidentally triggered
* model loading is complex, and given the time it takes autoload, problems probably cost more time than save time.
* So remove the feature for now

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -33,7 +33,6 @@ need_restart = False
 
 settings = {
     'dark_theme': False,
-    'autoload_model': True,
     'max_new_tokens': 200,
     'max_new_tokens_min': 1,
     'max_new_tokens_max': 2000,
--- a/server.py
+++ b/server.py
@@ -51,11 +51,7 @@ from modules.text_generation import (
 )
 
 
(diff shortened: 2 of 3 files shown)


Woisek:
> * There is a UI glitch where autoload gets triggered on focus change in the window
> 
>     * Autoload is often accidentally triggered
> 
>     * model loading is complex, and given the time it takes autoload, problems probably cost more time than save time.
> 
>     * So remove the feature for now

The autoload can be turned on/off and is probably useful to/used by the majority of the users. Creating a command flag like --autoload=true/false would make more sense then a plain removal that would annoy most of the users, who then are forced to do another click.

IMHO.

olaxe:
Thanks I didn't know I can remove the autoload from commandline, very useful to me.

Ph0rk0z:
Bold move but I just set` 'autoload_model': False,` so it never autoloads by default.  People have different preferences for this.

oobabooga:
I use this feature, and it can be disabled from settings.yaml if you don't like it as @Ph0rk0z noted

PR #2868, opened 2023-06-25
Several Training Enhancements

FartyPants:
- more relevant data in training_log.json
- saves training_prompt.json to the LoRA folder with the prompt template used for LoRA training (in case of dataset)
- adds stop at loss slider to ui
- some minor cosmetics

all has been tested over several days


oobabooga:
Thank you! It's very valuable to get updates from someone who has been testing the LoRA training extensively like you, as this is a feature that I do not use much.

PR #2889, opened 2023-06-27
Get SD samplers from API

missionfloyd:
Turns the sampler textbox into a dropdown populated with samplers from the API. Default is still DDIM.
image

### DIFF
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -73,7 +73,6 @@ def give_VRAM_priority(actor):
 if params['manage_VRAM']:
     give_VRAM_priority('set')
 
-samplers = ['DDIM', 'DPM++ 2M Karras']  # TODO: get the availible samplers with http://{address}}/sdapi/v1/samplers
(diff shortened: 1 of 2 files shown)


oobabooga:
Nice change, thanks! I think that the DDIM default is probably kind of outdated. What is a good but conservative sampler choice to use as a default instead? Any ideas @missionfloyd @GuizzyQC?

GuizzyQC:
@oobabooga @missionfloyd I'm partial to DPM++ SDE Karras but I looked quickly at popular images on a popular Discord for sharing generated images and DPM++ 2M Karras is also popular.

oobabooga:
Is any of these faster than the other?

GuizzyQC:
DPM++ 2M Karras is significantly faster

oobabooga:
I have made a quick test and DPM++ SDE Karras did seem to be a bit better, and indeed it is quite slower than DPM++ 2M Karras. I changed the default to DPM++ 2M Karras here https://github.com/oobabooga/text-generation-webui/commit/67a83f3ad9361c078ebcafa9c2c97ab9708f2fcb, it should be better than DDIM at least

PR #2903, opened 2023-06-27
models/config.yaml: set +landmark, +superhot-8k to 8k length

matatonic:
set +landmark, +superhot-8k to 8k length

landmark could be higher, but practically speaking it's probably too high for most people anyways.
4k might be a better default for most - can we auto-detect this from the model? Seems like something we shouldn't be setting in a config file except as an exceptional override.

### DIFF
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -236,3 +236,7 @@ TheBloke_WizardLM-30B-GPTQ:
 .*orca_mini:
   mode: 'instruct'
   instruction_template: 'Orca Mini'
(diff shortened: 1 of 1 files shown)


oobabooga:
I think that it may be possible to detect it by looking at the `max_position_embeddings` field in config.json. It seems to work.

The problem is that the best way to use superhot is by applying the lora (think that there will be 1000 different loras with long context in 6 months), and the adapter_config.json for superhot doesn't contain this information.

PR #2908, opened 2023-06-28
Support specify retry times in download-model.py

aisk:
Adding retry mechanism via requests' adaptor, this will help download success rate when downloading model with `download-model.py` in a unstable network environment.


oobabooga:
This is a very useful change. I think that it's fair to make the default 5 instead of 0

download-model.py
```diff
@@ -18,12 +18,16 @@
 
 import requests
 import tqdm
+from requests.adapters import HTTPAdapter
 from tqdm.contrib.concurrent import thread_map
 
 
 class ModelDownloader:
-    def __init__(self):
+    def __init__(self, max_retries):
```
talvasconcelos, line 26: why doesn't it have a default value?! `def __init__(self, max_retries = 5):`

aisk: I think this is not a library, but just a script, so I think there is no compatibility issue. Is this broke some old codes?

PR #2912, opened 2023-06-28
Disable half2 for ExLlama when using HIP

### LINKED ISSUE #2840: exllama gibberish output
### Describe the bug

I'm running into issues using exllama with the textgen UI. I have exllama running well using its webui, but when I load the same model into textgen UI, I only get gibberish as output. Llama.cpp works fine as a backend. I looked at the output in the console with --verbose and the input prompt looks fine. I tried multiple different sampler presets with no change. I'm using an AMD Vega 56 and I talked with another person on Discord with the same problem on an AMD card so it is probably an AMD-specific issue. I followed the oobabooga ROCm rentry guide in the README to set up textgen. One thing I noticed is the bitsandbytes package in the rocm fork is a little outdated, so could that be the issue?

### Reproduction

Set up textgen webUI using instructions in the README and the ROCm installation guide in Arch Linux on a system with an AMD card. Clone exllama into the repositories folder and install the requirements. Test that exllama works with its on WebGUI. Load a model using exllama in textgen webUI, then generate output from any prompt. 

### Screenshot

_No response_

### Logs

```shell
No log available
```

### System Info

```shell
Arch Linux
AMD 5600X
AMD Vega 56
```

ardfork:
Using kernels that rely on half2 produce gibberish output when using ROCm. ExLlama UI also disable half2 in the same way when using HIP.

Fix #2840

### DIFF
--- a/modules/exllama.py
+++ b/modules/exllama.py
@@ -1,6 +1,8 @@
 import sys
 from pathlib import Path
 
+from torch import version as torch_version
+
 from modules import shared
 from modules.logging_colors import logger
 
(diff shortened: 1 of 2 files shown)


duncannah:
Thanks for your great work, I can confirm this works wonderfully. It seems that you've forgotten the torch version import on `exllama_hf.py`.

Koko2110:
Thank you for the extremely quick fix, I can also confirm it works as expected now.

Dajinu:
I've just tried this fix and it works. Man, you deserve all the love of the universe. Can't thank you enough!

Issue #2891, opened 2023-06-27
Error when trying to use >5k context on SuperHOT 8k models on exllama_hf


Panchovix:
### Describe the bug

Hi there, thanks for all the work.

Was trying some SuperHOT 8k merged models (like https://huggingface.co/Panchovix/tulu-30b-SuperHOT-8K-4bit-32g), and then, when trying to generate while context was larger than 5000, I get issues and it doesn't generate any output.

The only seems to happen only on exllama_hf, since on exllama itself, it works without issues.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Get any SuperHOT 8k merged model.
2. Load into the webui with max_seq_len 8192 and compress_pos_emb 4.
3. Try to generate or continue a chat with more than 5000 tokens.

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/19153797/a13c245e-5842-432f-9500-de3bc8e949a3)

### Logs

```shell
To create a public link, set `share=True` in `launch()`.
2023-06-27 02:06:37 INFO:Loading h2ogpt-research-oig-oasst1-512-30b-SuperHOT-8k-4bit-32g...
2023-06-27 02:06:37 WARNING:Exllama module failed to load. Will attempt to load from repositories.
2023-06-27 02:06:45 INFO:Replaced attention with xformers_attention
2023-06-27 02:06:45 INFO:Loaded the model in 8.47 seconds.

Output generated in 9.92 seconds (20.06 tokens/s, 199 tokens, context 1848, seed 356899934)
Traceback (most recent call last):
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\callbacks.py", line 55, in gentask
    ret = self.mfunc(callback=_callback, *args, **self.kwargs)
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\text_generation.py", line 289, in generate_with_callback
    shared.model.generate(**kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 1572, in generate
    return self.sample(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 2619, in sample
    outputs = self(
[...]

PR #2913, opened 2023-06-28
Fix exllama_hf gibbersh above 2048 context, and fixes >5000 context.

Panchovix:
Info and fix from https://github.com/turboderp/exllama/issues/110

Using text-gen webui with Exllama loader gives different results than with Exllama_HF. Specifically, Exllama_HF gives gibberish with SuperHOT 8K models past 2048 tokens. Even the logits of the two loaders produce completely different values.

This piece of code from @kaiokendev fixes it.

Also, fixes https://github.com/oobabooga/text-generation-webui/issues/2891, which was an issue when trying to use above 5000 context on SuperHOT 8K 33B models while using exllama_hf.

`Output generated in 6.05 seconds (1.49 tokens/s, 9 tokens, context 6629, seed 1111456590)`

Note that this is a temporal fix while turbo fixes sents a new commit.

### DIFF
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -54,7 +54,15 @@ class ExllamaHF(PreTrainedModel):
         cache = kwargs['past_key_values'] if 'past_key_values' in kwargs else None
         if cache is None:
             cache = ExLlamaCache(self.ex_model)
-            self.ex_model.forward(torch.tensor([seq[:-1]], dtype=torch.long), cache, preprocess_only=True, lora=self.lora)
+
+            nseq = seq[:-1]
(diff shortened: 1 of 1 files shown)


oobabooga:
Thank you! I thought the problem was the repetition penalty not having a range parameter, but apparently that was not the issue. Let's use the temporary fix for now.

oobabooga:
Thanks for the information @turboderp, I have reversed this PR and @jllllll has updated the exllama wheel used by the repository to use the latest commit.

Panchovix:
@turboderp Many thanks for the updates! @oobabooga Thanks for reverting the commit as well.

PR #2916, opened 2023-06-28
Add repetition penalty range parameter to transformers

oobabooga:
I thought that this would be useful to make SuperHOT outputs more coherent with ExLlama_HF past 2048 tokens, but it doesn't seem to make much of a difference. 

I think that the issue may be running it with a compression factor of 2 instead of 4.

I'll leave this here in case anyone wants to run some tests. Maybe @Panchovix will find it interesting.

The range parameter is hardcoded to 256 tokens.

### DIFF
--- a/api-examples/api-example-chat-stream.py
+++ b/api-examples/api-example-chat-stream.py
@@ -44,6 +44,7 @@ async def run(user_input, history):
         'tfs': 1,
         'top_a': 0,
         'repetition_penalty': 1.18,
+        'repetition_penalty_range': 0,
         'top_k': 40,
         'min_length': 0,
         'no_repeat_ngram_size': 0,
--- a/api-examples/api-example-chat.py
+++ b/api-examples/api-example-chat.py
@@ -38,6 +38,7 @@ def run(user_input, history):
         'tfs': 1,
         'top_a': 0,
         'repetition_penalty': 1.18,
+        'repetition_penalty_range': 0,
         'top_k': 40,
         'min_length': 0,
         'no_repeat_ngram_size': 0,
--- a/api-examples/api-example-stream.py
+++ b/api-examples/api-example-stream.py
@@ -33,6 +33,7 @@ async def run(context):
         'tfs': 1,
         'top_a': 0,
         'repetition_penalty': 1.18,
+        'repetition_penalty_range': 0,
         'top_k': 40,
         'min_length': 0,
         'no_repeat_ngram_size': 0,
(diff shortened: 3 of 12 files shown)


Panchovix:
Just tested and it works way better, so nice work there! Tested on airoboros-33b-gpt4-1.4-SuperHOT-8k-4bit-32g.

`Output generated in 9.17 seconds (11.56 tokens/s, 106 tokens, context 6791, seed 415461465)`

I noticed it takes a bit more of time to start generating.

Also, the console outputs `torch.Size([1, 256]) torch.Size([1, 32000])`, but not sure if that is desired.

Also, it takes a LOT less VRAM, not sure if after the exllama update or this PR.

Panchovix:
Just for test purposes, I tried 16k context, and now it works on 48GB VRAM (2x4090), abeit slowly, using gpu_split 9,9.

`Output generated in 20.04 seconds (1.35 tokens/s, 27 tokens, context 15719, seed 504801742)`

image

So now, in theory when a 16k 33b lora releases, it may be possible to run it,

@turboderp, sorry for the ping, but this is after your last commit?

turboderp:
Group size and act order have minimal impact on performance with ExLlama, especially with a long context where most of the time is spent on attention anyway. Attention doesn't touch the quantized weights. It's still a lot slower than I would expect, when I'm getting more like 12 tokens/second on an 8k context. Complexity is linear, after all, with a very large constant factor.

Unless that output is the combined t/s after evaluating the prompt?

oobabooga:
Thanks for the review @Panchovix. I have made some tests and the number of tokens/s seems to be the same. The VRAM difference is most likely from the exllama update. It's also helpful to know that you noticed a quality difference while using it.

I have added a slider to the UI and made the parameter available through the API. Also connected it to the `repetition_penalty_sustain` parameter in ExLlama.

omasoud:
Prior to [this](https://github.com/oobabooga/text-generation-webui/commit/79db629665edc4a4149004ada1231aef9a6e4856) commit (i.e., when `self.generator.settings.token_repetition_penalty_sustain` [wasn't being set](https://github.com/oobabooga/text-generation-webui/blob/bef67af23c0c063d109ab19da7daffed3f323ec2/modules/exllama.py#L66) and exllama was giving it a default of 256, instead of the new default of -1), I was getting very good results with **wizardlm-13b-v1.0-uncensored-gptq** (I'm using api extension). I didn't realize this could make such a big difference, and I'm glad I pinpointed this so I can make sure to set it to 256.

PR #2922, opened 2023-06-29
Load Salesforce Xgen Models

Honkware:
Changes:
- Added support for loading XgenTokenizer
- Incorporated Xgen instruction template and configuration
- Note: Parameters now require custom stopping strings - "<|endoftext|>" (Including the quotes)


modules/models.py
```diff
@@ -94,6 +94,8 @@ def load_tokenizer(model_name, model):
     if any(s in model_name.lower() for s in ['gpt-4chan', 'gpt4chan']) and Path(f"{shared.args.model_dir}/gpt-j-6B/").exists():
         tokenizer = AutoTokenizer.from_pretrained(Path(f"{shared.args.model_dir}/gpt-j-6B/"))
     elif model.__class__.__name__ in ['LlamaForCausalLM', 'LlamaGPTQForCausalLM', 'ExllamaHF']:
+        if any(s in model_name.lower() for s in ['xgen']) and Path(f"{shared.args.model_dir}/{model_name}/").exists():
+            tokenizer = AutoTokenizer.from_pretrained(Path(f"{shared.args.model_dir}/{model_name}/"), trust_remote_code=shared.args.trust_remote_code)
```
tmm1, line 98:
```suggestion
            tokenizer = AutoTokenizer.from_pretrained(Path(f"{shared.args.model_dir}/{model_name}/"), trust_remote_code=shared.args.trust_remote_code)
            return tokenizer
```

tmm1:
I tried this on macOS and it works after the `return` change.

```
python3 download-model.py Salesforce/xgen-7b-8k-inst
python3 server.py --listen --trust-remote-code
```

I also tried LoRA training but ran into an error: `ValueError: too many dimensions 'str'`

tmm1:
The problem with training seems to be related to this weird behavior in the `XgenTransformer`:

```python
>>> from tokenization_xgen import XgenTokenizer
>>> x = XgenTokenizer()
>>> x.pad_token_id
>>> x.pad_token_id = 0
>>> x.pad_token_id
"b'!'"
>>> x.pad_token
"b'!'"
```

oobabooga:

* The instruction template only applies to xgen-7b-8k-inst, not any model containing xgen
* The template is just vicuna-v0, so no need to add a new one
* Remove automatic trust_remote_code=True addition as this can be exploited
* Remove universal llama tokenizer logic and instead just warn the user if the tokenizer is LlamaTokenizer and the relevant files look outdated

Issue #2939, opened 2023-06-30
API: LORA not applied when loading new model


g588928812:
### Describe the bug

When using the API to load a new (or reload the same) model+LORA _after_ a model model+LORA has already been loaded, the new model is loaded but the LORA is not applied 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1) Start tgi on shell, loading model + lora
`python server.py --model open_llama_7b --listen --lora LORA_NAME --api`

output TGI 
> 2023-06-30 09:06:24 INFO:Loading open_llama_7b...
> /home/g/text-generation-webui/modules/models.py:157: UserWarning: 'has_mps' is deprecated, please use 'torch.backends.mps.is_built()'
>   if not any((shared.args.cpu, torch.cuda.is_available(), torch.has_mps)):
> 2023-06-30 09:06:24 WARNING:The model weights are not tied. Please use the `tie_weights` method before using the `infer_auto_device` function.
> Loading checkpoint shards: 100%|████████████████████████████████████████████████████████| 2/2 [00:08<00:00,  4.03s/it]
> 2023-06-30 09:06:32 INFO:Loaded the model in 8.25 seconds.
> 
> 2023-06-30 09:06:32 INFO:Applying the following LoRAs to open_llama_7b: LORA_NAME
> Starting streaming server at ws://0.0.0.0:5005/api/v1/stream
> Starting API at http://0.0.0.0:5000/api
> Running on local URL:  http://0.0.0.0:7860
> 
> To create a public link, set `share=True` in `launch()`.

2) API call to load same/other model and LORA

```
def model_api(request):
    print(HOST)
    response = requests.post(f'http://{HOST}/api/v1/model', json=request)
    return response.json()

def complex_model_load(model):
    req = {
        'action': 'load',
        'model_name': model,
        'args': {
            'gptq_for_llama': False, # Use AutoGPTQ by default, set to True for gptq-for-llama
            'lora': ["LORA_NAME"],
            "gpu_memory": ["23","0"],
        },
    }
    return model_api(req)

HOST = "192.168.1.79:5000"
model = "open_llama_7b"
resp = complex_model_load(model)
```

output TGI on server

```
192.168.1.79 - - [30/Jun/2023 09:11:19] "POST /api/v1/model HTTP/1.1" 200 -
192.168.1.79 - - [30/Jun/2023 09:11:19] "POST /api/v1/model HTTP/1.1" 200 -
[...]

PR #2952, opened 2023-07-01
Update server.py - clear LORA after reload

FartyPants:
This is second part of the issue (related to UI interface - the lora needs to be cleared from the menu after reload/load)
https://github.com/oobabooga/text-generation-webui/issues/2939

see the first part (necessary for this 2nd to work) that clears lora names after unload:
https://github.com/oobabooga/text-generation-webui/pull/2951


oobabooga:
I'll merge this one after https://github.com/oobabooga/text-generation-webui/pull/2991 to use the new syntax for the event handlers

PR #2954, opened 2023-07-01
Replace deprecated Pillow constant 

Turamarth14:
With version 10.0.0 of Pillow the constant Image.ANTIALIAS has been removed. Instead Image.LANCZOS should be used.
https://pillow.readthedocs.io/en/stable/deprecations.html#constants

### DIFF
--- a/modules/html_generator.py
+++ b/modules/html_generator.py
@@ -129,7 +129,7 @@ def generate_4chan_html(f):
 def make_thumbnail(image):
     image = image.resize((350, round(image.size[1] / image.size[0] * 350)), Image.Resampling.LANCZOS)
     if image.size[1] > 470:
-        image = ImageOps.fit(image, (350, 470), Image.ANTIALIAS)
+        image = ImageOps.fit(image, (350, 470), Image.LANCZOS)
 
     return image
 


Macjja (changes requested):
The constant would be found under Resampling, so it should be: `Image.Resampling.LANCZOS`

modules/html_generator.py
Macjja, line 1: Should change line to get the correct constant, as it is found under Resampling: `Image.Resampling.LANCZOS`

Macjja (changes requested):

modules/html_generator.py
```diff
@@ -129,7 +129,7 @@ def generate_4chan_html(f):
 def make_thumbnail(image):
     image = image.resize((350, round(image.size[1] / image.size[0] * 350)), Image.Resampling.LANCZOS)
     if image.size[1] > 470:
-        image = ImageOps.fit(image, (350, 470), Image.ANTIALIAS)
+        image = ImageOps.fit(image, (350, 470), Image.LANCZOS)
```
Macjja, line 132:
```suggestion
        image = ImageOps.fit(image, (350, 470), Image.Resampling.LANZCOS)
```

Turamarth14: The [documentation](https://pillow.readthedocs.io/en/stable/deprecations.html#constants) states that 

> Image.LANCZOS or Image.Resampling.LANCZOS

should be used. Is there any particular reason to go from the first variant to the second one?

Macjja: Oh, I didn’t read the documentation (whoops) No, I don’t think there should be any reason, I had just tested with what I suggested. The only reason you would choose mine would be for consistency, as line 130 uses the Resampling variation

Issue #2948, opened 2023-07-01
NTK-based exllama support potential


Midaychi:
https://github.com/turboderp/exllama/pull/118

This hasn't been vetted/merged yet but in practice, it seems to unlock the context of un-finetuned models based on the scaling alpha value and does it with minimal perplexity loss.

Ironically superhot-tuned models fair worse with it.

Might be worth checking out, or at least keeping an eye on?

Panchovix:
I have a PR already done to implement NTK RoPE Extension, if turbo approved the PR.

![image](https://github.com/oobabooga/text-generation-webui/assets/19153797/6a6b8956-46ca-4d01-aedc-c3b0f3f69748)

Midaychi:
a

Panchovix:
Here's my fork if you want to try in the meanwhile, you have to apply the exllama PR as well.

https://github.com/Panchovix/text-generation-webui

Midaychi:
I manually tried it and it seems to work really well
 Even on an un fine-tuned 7b model.
I don't know if your or is gonna get merged though; turboderp's really actively venting and sus about it.

Ph0rk0z:
It certainly doesn't break anything. I don't have superHOT for 65b and supposedly perplexity is bad on it before you reach 2048. For my use case (chat) this is better. The other alternative is smaller models.

I think everything is all merged now.

Midaychi:
Yeah get ready for people to get confused at what the extra dial is or why their superhot models is performing worse with it.

Not blaming people but stuff is moving pretty fast. People are probably still just now finding out about rope stuff from YouTubers and coming in going qtf

Ph0rk0z:
I am barely trying this stuff today as I've been training audio models. Added an extra dial to the UI for myself too :P

Panchovix:
@Midaychi turbo merged, so I will do the PR now for ooba.

Panchovix:
PR is here: https://github.com/oobabooga/text-generation-webui/pull/2955

PR #2955, opened 2023-07-01
Add Support for Static NTK RoPE scaling for exllama/exllama_hf

Panchovix:
This adds support for the new NTK RoPE scaling, mentioned in https://github.com/turboderp/exllama/issues/115.

After https://github.com/turboderp/exllama/pull/118 got merged, exllama now supports NTK RoPE scaling.

This PR Adds the parameter "alpha_emb" which is used to apply the alpha value for NTK RoPE scaling, on the webui.

Adresses https://github.com/oobabooga/text-generation-webui/issues/2948

Tested on 65B models at 4K context, with 48GB VRAM (2x24) using gs 16,20, and on 33B models with 8K context.

image

Perplexity:
For tulu-30B-GPTQ (non-SuperHOT)
* Perplexity at 2048 ctx (no compress_pos_emb, no alpha RoPE): 5.2153
* Perplexity at 8192 ctx, compress_pos_emb = 4: 10.0813
* Perplexity at 8192 ctx, alpha  = 4: 5.3534
* Perplexity at 8192 ctx, compress_pos_emb = 4, alpha  = 4: 15.4406

For Tulu-30B-SuperHOT-8K-4bit-32g:
* Perplexity at 8192 ctx, compress_pos_emb = 4: 5.8166
* Perplexity at 8192 ctx, alpha  = 4: 7.5073
* Perplexity at 8192 ctx, compress_pos_emb = 4, alpha  = 4: 6.0903

NOTE: for context above 8K, I suggest to keep using SuperHOT models.


pbasov:
You probably want to bump jllllll/exllama to 0.0.5 as a part of this commit as well.

Released some hours ago: https://github.com/jllllll/exllama/releases/tag/0.0.5

Ph0rk0z:
Towards the end of the CTX the model can and does break down a bit. So fair warning I guess. Airoboros 65b 1.2 started having vegetarian expressions on their face and repeating itself. I capped it off at 3200. Probably till about 3500ish is good. For an unmodified model it's still a huge huge win.

shouyiwang:
> Towards the end of the CTX the model can and does break down a bit. So fair warning I guess. Airoboros 65b 1.2 started having vegetarian expressions on their face and repeating itself. I capped it off at 3200. Probably till about 3500ish is good. For an unmodified model it's still a huge huge win.

Do you mean max_seq_len? And what is your alpha_emb setting? Does 1 or 2 makes any difference? Thx

shouyiwang:
> Towards the end of the CTX the model can and does break down a bit. So fair warning I guess. Airoboros 65b 1.2 started having vegetarian expressions on their face and repeating itself. I capped it off at 3200. Probably till about 3500ish is good. For an unmodified model it's still a huge huge win.

Do you mean max_seq_len? And what is your alpha_emb setting, does using 1 or 2 make any difference? 
I'm really confused because I thought alpha_emb was similar to compression ratio, which would reduce VRAM usage. However, when I tested both 1 and 2, I found no difference (max_seq_len 4096).

Panchovix:
> Towards the end of the CTX the model can and does break down a bit. So fair warning I guess. Airoboros 65b 1.2 started having vegetarian expressions on their face and repeating itself. I capped it off at 3200. Probably till about 3500ish is good. For an unmodified model it's still a huge huge win.

Up to ~3584 ctx seems to be absolutely fine on my tests. About 3.6k and above, it starts to give issues.

@pbasov thanks for the suggestion, updated exllama wheel version on the requeriments on the PR.

shouyiwang:
Can someone explain to me what it means to set alpha_emb to 1 or 2? I noticed that it doesn't seem to make a difference in VRAM usage(context 4096)

shouyiwang:
I did some comparison tests and found that this PR performs much better than the existing `compress_pos_embedding` method. I used a 65b model and set the context length to 4096. Then, I tried using `compress_pos_embedding` and `alpha_emb` set to 2 respectively. The difference in the quality of the generated text was obvious.
It seems that the `compress_pos_embedding` method is redundant and should be removed to avoid confusing users. 

However, I'm not sure about the mechanism behind these two methods. Can someone please explain? Thx

Panchovix:
> I did some comparison tests and found that this PR performs much better than the existing `compress_pos_embedding` method.
> 
> I used a 65b model and set the context length to 4096. Then, I tried using `compress_pos_embedding` and `alpha_emb` set to 2 respectively. The difference in the quality of the generated text was obvious.
> 
> It seems that the `compress_pos_embedding` method is redundant and better be removed later to avoid confusing users.

They work different. Compression is to be used with SuperHOT LoRA/models by changing the compression of the embedding factor, while alpha NTK RoPE scaling is used to extend context on base models (for now) by chaging the rotatory embedding.

Since at the moment, there's no SuperHOT LoRAs for 65B models, static NTK RoPE scaling is the only way for that parameter count. Also, luckily this method works pretty good up to 4k (~3600) tokens.

There's no reason to remove compress for now.

shouyiwang:
@Panchovix You're right. I just realized that the superHOT models rely on the compress_pos_embedding option.

By the way, the NTK RoPE embedding is a genius invention. Just changing one line of code for the embedding method gives perfect results.

shouyiwang:
> Since at the moment, there's no SuperHOT LoRAs for 65B models, static NTK RoPE scaling is the only way for that parameter count. Also, luckily this method works pretty good up to 4k (~3600) tokens.

If I want to achieve the best result, should I set the context length to 3500 or 4096 and simply disregard the tokens beyond 3500?

gr1336:
> > Since at the moment, there's no SuperHOT LoRAs for 65B models, static NTK RoPE scaling is the only way for that parameter count. Also, luckily this method works pretty good up to 4k (~3600) tokens.
> 
> If I want to achieve the best result, should I set the context length to 3500 or 4096 and simply disregard the tokens beyond 3500?

#### I've been testing some settings to see how the parameters impact the text generation process.

### To start off, take a look at the `repetition_penalty_range` parameter:

- This seems to have the biggest impact on the consistency of any model I've tested so far. Before this parameter, I was getting inconsistent text generation after the 3600~ tokens mark.
- The values I saw in the results are a minimum of `384` up to `1024` tokens.
- For me, the "sweet spot" in this case is `512`, but I believe this will depend on the model you choose to use.
- With that set to `512` I could get the 4096ctx with a very  good consistency.

### Some other repetition settings that I recommend you test are the other "repetition penalties":

- `repetition_penalty`: `1.15` through `1.20`.
- `encoder_repetition_penalty`: `1.0` to `1.025`.
- `no_repeat_ngram_size`: `0` [off] or something like `16`.

### Also play around with the other parameters, like `temperature`, `top-p` and `top-k`, my range is something like:

- `temperature`: `0.4-1.0` - The higher the context, it should also be treated with a higher value, so for a `3000k+` context, a temperature of `0.7` to `0.8` might be the best option for writing "Precise".
- `top_k`: `0` or `40` - I usually use only one of these values, such as `0` or `40`.
- top_p`: `0.5` to `0.95` - This is by far the hardest to find a good point for, since it changes so much based on context length.

#### I will test some more combinations of settings here, and different models, so that I can get more consistent information in a few days.

Panchovix:
First model released using NTK RoPE scaling method instead of SuperHOT LoRA

https://huggingface.co/bhenrym14/airoboros-33b-gpt4-1.4.1-PI-8192-GPTQ

If you want to test it with this implementation

image

practicaldreamer (approved):
Excellent addition!

oobabooga:
Thank you @Panchovix ! I can't say that I understand what this does, but the perplexity numbers look promising.

I have just renamed the parameter to `alpha_value`, as that's how it's called on exllama.

Barnplaid:
So just to make sure, how do we use this?

Ph0rk0z:
Load model and set alpha_value + increase seq length.

PR #2959, opened 2023-07-01
Elevenlabs tts fixes

set-soft:
Two fixes and an addition. 
The fixes are related to the use of options in the *settings.yaml* file:
- The selected voice was ignored. The code always calls *refresh_voices()*, but the old code always selected the first available voice, discarding the option from the config file. Now we keep the selected option, unless the voice doesn't exist. In this case we issue an error and select the first available.
- The API key was ignored, and the input box displayed "Enter your API key.". This is because we must call *update_api_key* and also use the value for the input box.

The addition is related to languages other than English, the code always selected *eleven_monolingual_v1*, but for other languages it sound awful, so now we have a drop down to select *eleven_multilingual_v1*, much better for Spanish ;-)

### DIFF
--- a/extensions/elevenlabs_tts/script.py
+++ b/extensions/elevenlabs_tts/script.py
@@ -12,10 +12,12 @@ params = {
     'selected_voice': 'None',
     'autoplay': False,
     'show_text': True,
+    'model': 'eleven_monolingual_v1',
 }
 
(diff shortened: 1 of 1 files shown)


oobabooga:
Those are very good changes, thank you

PR #2975, opened 2023-07-02
[Fix] The mode from settings was ignored

### LINKED ISSUE #2973: mode option from settings.yaml/settings.json is ignored
### Describe the bug

When indicating a mode in the settings file (like in the provided example that uses `mode: chat`) the code simply ignores it.

The code never transfers this value to `shared.args`, used by functions like `shared.is_chat`.

### Reproduction

Simply use the provided settings-example.yaml, it says chat mode, but you get default mode.

### Screenshot

_No response_

### Logs

```shell
No error generated
```

### System Info

```shell
Not relevant:
Debian GNU/Linux 11.7
Radeon RX 5500 XT
```

set-soft:
- Now it takes effect
- Note that the modes included `cai_chat` in one instance, I think this is no longer a mode, but a chat flavor. I'm using a global list of valid modes.
- Informs about wrong modes from settings
- A mode from the command line has more precedence
- A mode from a model has more precedence

Some explanations:
The mechanism used to set the mode is complex. What the code does (regardless of this patch) is to set the mode as the default mode selected by any model. This means that we must load a model to get the mode applied, this is confusing to me, but it works because a model will be loaded soon or latter. 
But the old code has a major drawback: the model was loaded *after* loading the extensions, so when the code asked for *is_chat()* didn't take into account the model potentially selected from the *settings*. So I moved the extensions load to be done *after* loading the model.
Also: the old code takes care of the model parameter (using `update_model_parameters`) but this doesn't include the mode. So the mode was never honored. Even when it was set as the default for any model.
Additionally: the VALID_MODES should be moved to a better place and the code should be scanned for uses of these names and force to use only one list. The old code had two lists, one was wrong.

Fixes #2973

### DIFF
--- a/server.py
+++ b/server.py
@@ -50,6 +50,7 @@ from modules.text_generation import (
     stop_everything_event
 )
 from modules.utils import gradio
+VALID_MODES = ["default", "notebook", "chat"]
 
 
 def load_model_wrapper(selected_model, loader, autoload=False):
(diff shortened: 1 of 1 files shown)


oobabooga:
This is a misconception. Mode in shared.settings/settings.yaml refers to the sub-options inside chat mode: chat, chat-instruct, or instruct. It does not refer to the UI modes (default/notebook/chat).

PR #2976, opened 2023-07-02
[Added] Allow to specify a model from the settings

set-soft:
- The command line value has more precedence

This may collide with #2975 but is really easy to solve, I can do a rebase if #2975 is accepted, or you can just apply it manually.

### DIFF
--- a/server.py
+++ b/server.py
@@ -993,6 +993,10 @@ if __name__ == "__main__":
         new_settings = json.loads(file_contents) if settings_file.suffix == "json" else yaml.safe_load(file_contents)
         for item in new_settings:
             shared.settings[item] = new_settings[item]
+        if shared.args.model is None:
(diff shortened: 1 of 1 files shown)


oobabooga:
I prefer reserving settings.yaml to UI defaults, model should be a command-line flag since it's more mutable

set-soft:
Hi @oobabooga !
I don't fully understand you: What's the problem with having this as an option, people can use it if they want, other people just don't add this option.
My intention is to allow creating configuration files that are _self contained_, so you can just start the WebUI using the config file and you don't need to specify anything else.
Also, if you want to change the default model you can just specify it in the command line.
Additionally, some stuff like the mode isn't even configured if no model is loaded. The code already tries to load a model if none is specified and you have just one model.
So I don't see any collision between having an extra option and your intended use. Just think about it for a moment.

PR #2977, opened 2023-07-02
[Fixed] wbits and groupsize values from model not shown

set-soft:
- This is because they must be strings, triggered by the first element in the list ("None"). Using integers fails to show them


oobabooga:
They are always strings it seems.

PR #2981, opened 2023-07-03
Minor IndexError Fix

juhenriquez:
Fix IndexError: string index out of range in download-model.py


oobabooga:
This only happens if no model is specified, and this error is secondary in this context

PR #2991, opened 2023-07-03
Implement sessions + add basic multi-user support

oobabooga:
~This implements many changes, so some things may still be broken.~

**Sessions**

In the Sessions tab, you can save the entire session, including input and output texts, generation parameters, and chat history.

Use cases:

* Saving multiple histories for the same character.
* Saving a story and returning to it later.

For that to be possible, the chat history hard to be turned into a `gr.State` variable and is no longer a global variable. This will make it easy to add multi-user support after this PR is merged. The disadvantage is that many extensions relied on accessing `shared.history` to work, so I had to rework a lot of the extensions framework.

**Interface persistence**

* When you reload the page, your text in default/notebook modes is no longer lost.
* The entire session is automatically saved as `session_chat_autosave.json`, `session_notebook_autosave.json`, or `session_default_autosave.json` depending on the mode, so that you can come back to where you were even if you turn off the computer.

**Multi-user support**

Implemented through the `--multi-user` flag. When it is used, chat histories are no longer automatically saved or loaded. Instead, they are only kept in memory for the duration of the session.

It is possible that there are bugs.

**Syntax simplifications**

A new `gradio(...)` helper function has been added to `server.py`. What it does is allow you to write

```python
button.click(function, gradio('interface_state', 'history', 'character_menu'), None)
```

instead of

```python
button.click(function, [shared.gradio[k] for k in ['interface_state', 'history', 'character_menu']], None)
```

this makes event handler declarations a lot more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ Optionally, you can use the following command-line flags:
 | `-h`, `--help`                             | Show this help message and exit. |
 | `--notebook`                               | Launch the web UI in notebook mode, where the output is written to the same text box as the input. |
 | `--chat`                                   | Launch the web UI in chat mode. |
+| `--multi-user`                             | Multi-user mode. Chat histories are not saved or automatically loaded. WARNING: this is highly experimental. |
 | `--character CHARACTER`                    | The name of the character to load in chat mode by default. |
 | `--model MODEL`                            | Name of the model to load by default. |
 | `--lora LORA [LORA ...]`                   | The list of LoRAs to load. If you want to load more than one LoRA, write the names separated by spaces. |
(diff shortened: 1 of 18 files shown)


oobabooga:
The 1 outstanding issue is that when a session is loaded in chat mode, it causes a character to be loaded, which then causes the character's history to be loaded, which overwrites the history initially loaded with the session.

FartyPants:
Quality additions...

webnizam:
A feature I've been waiting for a while. So, would it work the same for API requests as well?
Great Job, and thanks in Advance.

kyleboddy:
Multi-user causes models to not be able to be loaded:

https://github.com/oobabooga/text-generation-webui/issues/4356

PR #3001, opened 2023-07-04
Add logit_bias and logprobs support to the OpenAI extension.

cyberfox:
This adds a new extension mechanism for logits post-processing, and uses that extension mechanism to implement `logit_bias` and `logprobs` support in the OpenAI extension.

cc: @matatonic @oobabooga

### DIFF
--- a/extensions/cfg/script.py
+++ b/extensions/cfg/script.py
@@ -1,0 +1,33 @@
+import torch.nn.functional as F
+
+from transformers import LogitsWarper, LogitsProcessor
+
+from modules import shared
+
+
+class CFGLogits(LogitsWarper):
+    def __init__(self, cfg, inputs, model, verbose=True):
+        self.cfg = cfg
+        self.inputs = inputs
+        self.model = model
+        self.out = None
+        self.verbose = verbose
+
+    def __call__(self, input_ids, scores):
+        if self.cfg == 1:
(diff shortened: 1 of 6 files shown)


cyberfox:
These are probably not _all_ that is needed, but these are evidently some pre-requisites for supporting `guidance` from Microsoft, as an example.  I also have a 'token healing' extension, but it's not quite soup yet, and it doesn't interact with the gradio UI very well, so I excluded it from this pull request.

There are other useful features that can be built on logits processors, unfortunately right now this will only work with Transformers-based models. I _think_ that includes AutoGPTQ, but not the custom models like Exllama and llama_cpp yet.

FartyPants:
I independently verified the code addition in 
[modules/extensions.py](https://github.com/oobabooga/text-generation-webui/pull/3001/files#diff-9557b7523edffd1a07709d5409741791239e953340b12071cb78b909d529eaa4)
and 
[modules/text_generation.py](https://github.com/oobabooga/text-generation-webui/pull/3001/files#diff-36dfef7d9db97b2cde9351ae9eafa7aed3dd15e094d2ca7a4a60bfa88d6096f2)

cyberfox:
I'm closing this pull request, as a more focused pull request #3029 has been put up. The other modules and logits processors in this will see the light of day another time.

PR #1326, opened 2023-04-17
A Simple search script using the google programable search API

IllogicalDesigns:
A simple search script using the google programable search API.  This script use the APIs to search the user's input string in google, then grabs the returned snippets and places them in the context window.  This allows the LLM to get some context from the web and provide live data or unknown data.

PR #3021, opened 2023-07-05
Integrate google search with Superbooga 

IllogicalDesigns:
I seem to have closed the original pull request.  https://github.com/oobabooga/text-generation-webui/pull/1326

The goal of this request is to integrate the programmable search engine API from Google.  Allowing a user to set up a search engine and add the results to the Superbooga DB for runtime context retrieval.  I have also added some basic semantic filtering of search results, so only relevant results from the first page are downloaded.

### DIFF
--- a/extensions/superbooga/script.py
+++ b/extensions/superbooga/script.py
@@ -1,19 +1,18 @@
+import json
 import re
 import textwrap
 
 import gradio as gr
+import requests
 from bs4 import BeautifulSoup
+from sentence_transformers import SentenceTransformer
+from sklearn.metrics.pairwise import cosine_similarity
 
 from modules import chat, shared
 from modules.logging_colors import logger
 
 from .chromadb import add_chunks_to_collector, make_collector
 from .download_urls import download_urls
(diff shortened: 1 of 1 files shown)


oobabooga:
This is a very good feature, I'll test and merge it soon

AIWintermuteAI:
It is a good feature indeed! Especially considering Superbooga integration. @oobabooga are you still considering merging it?

IllogicalDesigns:
There was a version that was merged, but I no longer see the controls when loading superbooga.  I have been out of the loop for a while.  

I can look into it when I get a chance to reimplement it with the updated oobabooga client.  I also remember reading that there are search providers that offer and API without the need for large amounts of setup, like the Google one.  

I would like to also find a way to take the query that the user comes up with and generate a search from that, then scrape the search results and add them to the superbooga DB.  Then run the original query with the new DB.  It would make the user experience similar to huggingface.co/chat's web search feature.  Where the user does not have to worry about search term, it just gets generated in the background.

PR #3029, opened 2023-07-06
Add support for logits processors in extensions 

cyberfox:
@oobabooga @matatonic

This is the best version of this I've been able to put together; it's very focused on adding the `logits_processor` extension capability, and the sample extension for it is the Classifier-Free Guidance from [this paper](https://arxiv.org/abs/2306.17806).

This same extension mechanism will be of value to the OpenAI extension, and others. Please let me know if there's anything I can do to help make the process smoother!


Panchovix:
Thanks, it seems to work, even for exllama! It seems to not work with exllama_hf at the moment sadly (maybe because a limitation?)

```
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\callbacks.py", line 55, in gentask
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\text_generation.py", line 297, in generate_with_callback
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 1572, in generate
    return self.sample(
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\utils.py", line 2632, in sample
    next_token_scores = logits_processor(input_ids, next_token_logits)
  File "F:\ChatIAs\oobabooga\venv\lib\site-packages\transformers\generation\logits_process.py", line 92, in __call__
    scores = processor(input_ids, scores)
  File "F:\ChatIAs\oobabooga\text-generation-webui\extensions\cfg\script.py", line 55, in __call__
    self.out = self.model(self.uncond, attention_mask=torch.ones_like(self.uncond))
  File "F:\ChatIAs\oobabooga\text-generation-webui\modules\exllama_hf.py", line 50, in __call__
    assert len(args) == 0, 'no *args should be passed to forward'
```

On the exllama_hf code, it looks like:

```
...
def __call__(self, *args, **kwargs):
        # TODO: Some decoding methods (such as Contrastive Search) may not work at this time
        assert len(args) == 0, 'no *args should be passed to forward' <- this line
        use_cache = kwargs.get('use_cache', True)
        labels = kwargs.get('labels', None)
...
```

Panchovix:
Related to above, modying

```
        if self.out is None:
            self.out = self.model(self.uncond, attention_mask=torch.ones_like(self.uncond))
        else:
            single_input = input_ids[:, -1:]
            self.out = self.model(single_input, past_key_values=self.out.past_key_values)
```

to

```
        if self.out is None:
            self.out = self.model(input_ids=self.uncond, attention_mask=torch.ones_like(self.uncond))
        else:
            single_input = input_ids[:, -1:]
            self.out = self.model(input_ids=single_input, past_key_values=self.out.past_key_values)
```

Makes it work again, and it seems to work fine with normal exllama, exllama_hf, GTPQ-for-LLaMA and AutoGTPQ. Sadly, not sure how it would with models at fp16 using transformers.

ENjoyBlue2021:
maybe i'm doing something wrong but i tried with exllama and transformer.
exllama didnt work for me at all.
Transformer was weird. sometimes it clearly changed the output but i couldn't get it to do it reliably.

oobabooga:
Maybe it should wait for this to become part of transformers officially? There is an open PR and it seems to have been well received.

Also, this needs to have a corresponding input field in the UI somewhere.

cyberfox:
I'm okay with this, if you want me to back out the CFG change. I've been using it at various levels. It works fine as far as I can tell with AutoGPTQ (I tested with guanaco-65b-GPTQ and it worked great, but _very_ slow) and several different models.

The problem with Exllama_HF is that it doesn't support either `past_key_value` or `attention_mask` being passed to the model, which is problematic.

I agree that the `CFGLogits` is going to be part of transformers eventually, but I felt it was important to provide a demonstration of how the LogitsProcessor extension would work, and to demonstrate that I'd tested that it worked.

The CFG extension _does_ add a text box for negative prompting, and a slider for the CFG scale. I've definitely tested that it works, and focuses attention pretty well, albeit in a subtle way.

I'll remove the extension, but leave the logits processor extension point in place.

Hopefully that helps!

--  Morgan

cyberfox:
Done and done! @FartyPants @oobabooga

cyberfox:
The major effect of the Classifier-Free Guidance was a subtle improvement in quality; it was less about the negative prompting. The persistent focus on the original prompt, giving it more weight than later generated prompts, improves the quality of generation for well-formed prompts. It's actually useful even if you don't put any prompt in the text box.

I look forward to putting together a plugin for it when CFGLogits is added to the base transformers library. Not to mention playing with the logits extension end point! :)

Thanks muchly for the feedback, all!

Vermeille:
@FartyPants 

> In fact the author [Vermeille](https://github.com/Vermeille) answered me that: "we were under time constraints for having a clear neg prompt methodology" so this is so-so by design at this stage.

It think this is due to poor wording and you got me wrong unfortunately. Negative prompting indeed works. The GPT4All example in the paper shows it and we have some (unpublished) tests that are conclusive as well. What I meant is that it's not that obvious _how_ to properly prompt engineer those negative prompts yet. We're actively working on some guidelines actually. Maybe they'll be obvious, maybe they won't. It's still exploratory. The effect is there and obvious, now it's about how to control it :)

FartyPants:
> @FartyPants
> 
> > In fact the author [Vermeille](https://github.com/Vermeille) answered me that: "we were under time constraints for having a clear neg prompt methodology" so this is so-so by design at this stage.
> 
> It think this is due to poor wording and you got me wrong unfortunately. Negative prompting indeed works. The GPT4All example in the paper shows it and we have some (unpublished) tests that are conclusive as well. 

My apologies, I'm glad to hear that.

Once the PR is merged with transformers this will definitely get wider use.

FartyPants:
@oobabooga 
I recommend merging the code now. 

@cyberfox
oh, we need to have Meow logits extension example :)

matatonic:
@oobabooga I also recommend this, it's a very low impact patch and makes some amazing things possible. I've been running this for a while with no issues.

matatonic (approved):

oobabooga:
What I'm thinking is that it may be worth it to apply the extension inside `sampler_hijack.py`. Then the extension can not only add a processor to the end of the list, but also reorder or modify existing processors

matatonic:
Personally, for the openai extension I really just need the request
parameter 'logits_processor' to be honored, but that's because I call
generate_request myself instead of input/output hijacking via an extension.
I need to be able to have a thread unique state updated per token during
generation.

Other than that, I'm not picky how it's implemented.

does that conflict with your idea? I'm not familiar with sampler_hijack yet.

oobabooga:
Parameters inside `state` are gathered from the UI, and a logitsprocessor can't be a gradio element. The API is secondary.

oobabooga:
After some experimentation, I found that transformers has 2 different classes that manipulate logits: warpers and processors. 

Warpers are applied when `do_sample=True`, and processors are applied always. By default, the only processor that is used is `RepetitionPenaltyLogitsProcessor`, so it doesn't seem that interesting to manipulate the processor list directly.

As a proof of concept I wrote the ball test (based on the code in #3078):

```python
import numpy as np
import torch
from transformers import LogitsProcessor

from modules import shared

params = {
}

class PerplexityLogits(LogitsProcessor):
    def __init__(self):
        pass

    def __call__(self, input_ids, scores):
        probs = torch.softmax(scores, dim=-1, dtype=torch.float)
        id_ = shared.tokenizer.encode('ball', add_special_tokens=False)[0]
        if input_ids[0][-1] != id_:
            probs[0][id_] = probs[0].max() / 2
        else:
            probs[0][id_] = 0

        probs[0] /= probs[0].sum()
        scores = torch.log(probs / (1 - probs))
        return scores

def logits_processor_modifier(logits_processor_list, input_ids):
    global ppl_logits_processor
    ppl_logits_processor = PerplexityLogits()
    logits_processor_list.append(ppl_logits_processor)

```

Result:

ball

In the future it may be worth it to add the ability to reorder or modify the warpers, but that's independent from this PR so I'll merge it.

PR #3031, opened 2023-07-07
Whisper_stt params for model, language, and auto_submit

brandonmcclure:
The Whisper_stt extension is great, but I want to use a different model instead of the `base.en` one to get more accuracy at the cost of speed. This adds the model, language, and auto_submit as params so I can set my defaults in settings.yaml, while still changing them in the UI if I want to. I did move the UI around a bit to hide the settings under an accordion and make the record button a little more front and center. 

I also included a small readme to link to the upstream documentation about the different model/language options available.


oobabooga:
That's very nice, thanks!

PR #2878, opened 2023-06-26
Fixes and Improvements to the LoRA training.

Nan-Do:
In this pull request I'm sharing some small fixes and improvements I have implemented to be able to train and use LoRAs.
Changes to LoRA.py:
- PeftConfig doesn't take a dtype and device_map parameters so the code fails when loading a LoRA using a model loaded in 8 or 4 bits. The fix checks if the model was loaded in 4 or 8 bits and if that's the case it uses the appropriate type name.
- https://huggingface.co/docs/peft/package_reference/config#peft.PeftConfig
 
Changes to Training.py:
- Adding the parameter skip_special_tokens=True when decoding raw_text in the text preparation phase, without this flag the special tokens will remain in the text during the preparation phase and this will make the training to completely fail. 
    - Sample output using the LlamaTokenizer (after decoding):
 `<s>Lorem ipsum dolor sit amet, officia excepteur ex fugiat reprehenderit enim labore culpa sint ad nisi L'`
    - Sample output of the training process:
`{'loss': 0.0, 'learning_rate': 0.0, 'epoch': 0.02}`
`{'train_runtime': 2932.1946, 'train_samples_per_second': 5.752, 'train_steps_per_second': 0.09, 'train_loss': 0.0, 'epoch': 0.02}`
- I have modified the chunking process, this doesn't fail but the first generated chunk will be shorter and a the text it contains will be repeated,  also the change makes the chunking process substantially more efficient both in memory and time. 
    - Sample of the first few generated chunks (after decoding). Before the change:
$1^{st}$:`Lorem ipsum dolor sit amet, officia excepteur ex fugiat reprehenderit enim labore culpa sint ad nisi L`
$2^{nd}$:`Lorem ipsum dolor sit amet, officia excepteur ex fugiat reprehenderit enim labore culpa sint ad nisi Lorem pariatur mollit ex esse exercitation amet. Nisi anim cupidatat excepteur officia. Reprehender`
$3^{rd}$:`orem pariatur mollit ex esse exercitation amet. Nisi anim cupidatat excepteur officia. Reprehenderit nostrud nostrud ipsum Lorem est aliquip amet voluptate voluptate dolor minim nulla est proident. Nostr`

   - and after:
$1^{st}$:`Lorem ipsum dolor sit amet, officia excepteur ex fugiat reprehenderit enim labore culpa sint ad nisi Lorem pariatur mollit ex esse exercitation amet. Nisi anim cupidatat excepteur officia. Reprehender`
[...]

PR #3032, opened 2023-07-07
Fix: Fix to load LoRAs in models loaded in 4 or 8 bits.

Nan-Do:
To summary the changes:
   - PeftConfig doesn't take a dtype and device_map parameters so the code fails when loading a LoRA using a model loaded in 8 or 4 bits. The fix checks if the model was loaded in 4 or 8 bits and if that's the case it uses the appropriate type name.
   - https://huggingface.co/docs/peft/package_reference/config#peft.PeftConfig

### DIFF
--- a/modules/LoRA.py
+++ b/modules/LoRA.py
@@ -112,11 +112,12 @@ def add_lora_transformers(lora_names):
     if len(lora_names) > 0:
         params = {}
         if not shared.args.cpu:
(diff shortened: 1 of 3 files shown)


Nan-Do:
This patch contains a previous unrelated commits. I'm closing the PR and remaking it from a fresh fork

PR #3034, opened 2023-07-07
Fix: Fixed the tokenization process of a raw dataset and improved its efficiency

Nan-Do:
To summary the changes:
- Adding the parameter skip_special_tokens=True when decoding raw_text in the text preparation phase, without this flag the BOS token will be duplicated (see the discussion on the previous PR).
- I have modified the chunking process to make it substantially more efficient both in memory and time and to produce a more coherent output. (see the previous PR for a longer explanation).

I'm tagging @FartyPants as he proposed another change and suggested to split the contributions.

### DIFF
--- a/modules/LoRA.py
+++ b/modules/LoRA.py
@@ -112,11 +112,12 @@ def add_lora_transformers(lora_names):
     if len(lora_names) > 0:
         params = {}
         if not shared.args.cpu:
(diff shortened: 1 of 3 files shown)


Nan-Do:
This patch contains previous unrelated commits. I'm closing the PR and remaking it from a fresh fork

PR #3035, opened 2023-07-07
Fix: Fixed the tokenization process of a raw dataset and improved its efficiency

Nan-Do:

To summary the changes:
- Adding the parameter skip_special_tokens=True when decoding raw_text in the text preparation phase, without this flag the BOS token will be duplicated (see the discussion on the previous PR).
- I have modified the chunking process to make it substantially more efficient both in memory and time and to produce a more coherent output. (see the previous PR for a longer explanation).

I'm tagging @FartyPants as he proposed another change and suggested to split the contributions.

### DIFF
--- a/modules/training.py
+++ b/modules/training.py
@@ -369,15 +369,10 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
                 yield f"Error: overlap_len ({overlap_len}) cannot be greater than or equal to cutoff_len ({cutoff_len})"
                 return
 
-            tokens = list(split_chunks(tokens, step))
-            for i in range(1, len(tokens)):
-                tokens[i] = tokens[i - 1][-overlap_len:] + tokens[i]
-
-            out_tokens.extend(tokens)
-            del tokens
+            out_tokens.extend(split_chunks(tokens, cutoff_len, step))
 
         del raw_text  # Note: could be a gig for a large dataset, so delete redundant data as we go to be safe on RAM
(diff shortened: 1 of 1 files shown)


FartyPants:
Yeah, the skip_special_tokens=True doesn't fix the double` <s><s>` issue, just avoids it by removing all special tokens.. 

The double BOS issue in plaintext is caused by calling encode, decode, encode 

My proposed changes were to simply add check that will eat two BOS in the training.py encode. This would require the least amount of changes and it will fix the double BOS

```
    def encode(text, add_bos_token):
        result = shared.tokenizer.encode(text, truncation=True, max_length=cutoff_len)
        # Check if the first two tokens are BOS
        if len(result) >= 2 and result[:2] == [shared.tokenizer.bos_token_id, shared.tokenizer.bos_token_id]:
            input_ids= input_ids[1:]

        if not add_bos_token and result[0] == shared.tokenizer.bos_token_id:
            result = result[1:]
        return result
```

as for the other changes:
def split_chunks(arr, size, step):

I don't know - @oobabooga , you would probably see very quickly what it is the change trying to fix.

oobabooga:
The optimization seems to make sense to me. I'm wary of adding `skip_special_tokens=False` and will go for the alternative solution in #3058 for now.

I feel like the LoRA training code has a lot of additional parameters in the current implementation (hard cut string, overlap length, prefer newline cut length). LLM training is entirely based on special tokens and at some point the training code will need to be refactored to become more aligned with that.

Issue #2475, opened 2023-06-01
Streaming api deadlock


xzqxnet0990:
### Describe the bug

When I am using mutiple threads to pressure measurement the streaming api.
I think it will cause a deadlock.
The code is here /extension/api/streaming_api.py line(51).

async def _handle_connection(self, websocket, path):

        self.send_response(200)
        self.send_header('Content-type', 'text/plain')
        self.send_header('Transfer-Encoding', 'chunked')
        self.end_headers()
        if self.path == '/api/v1/stream':
            async for message in websocket:
                message = json.loads(message)

                prompt = message['prompt']
                generate_params = build_parameters(message)
                stopping_strings = generate_params.pop('stopping_strings')
                generate_params['stream'] = True

                generator = generate_reply(
                    prompt, generate_params, stopping_strings=stopping_strings, is_chat=False)

```
def generate_reply(*args, **kwargs):
    shared.generation_lock.acquire()
    try:
        for result in _generate_reply(*args, **kwargs):
            yield result
    finally:
        shared.generation_lock.release()
```

and It will always cause websockets.exceptions.ConnectionClosedError: sent 1000 (OK); no close frame received problem.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Use WebSocket request-response Sampler and thread group with 2 threads to pressure measurement the port 5005/api/v1/stream.

### Logs

```shell
Traceback (most recent call last):
  File "/data/ketadb/llm/venv/lib/python3.8/site-packages/websockets/legacy/server.py", line 240, in handler
    await self.ws_handler(self)
  File "/data/ketadb/llm/venv/lib/python3.8/site-packages/websockets/legacy/server.py", line 1186, in _ws_handler
    return await cast(
  File "/home/ubuntu/llm/text-generation-webui/extensions/api/streaming_api.py", line 18, in _handle_connection
    async for message in websocket:
  File "/data/ketadb/llm/venv/lib/python3.8/site-packages/websockets/legacy/protocol.py", line 497, in __aiter__
    yield await self.recv()
  File "/data/ketadb/llm/venv/lib/python3.8/site-packages/websockets/legacy/protocol.py", line 568, in recv
    await self.ensure_open()
  File "/data/ketadb/llm/venv/lib/python3.8/site-packages/websockets/legacy/protocol.py", line 935, in ensure_open
    raise self.connection_closed_exc()
websockets.exceptions.ConnectionClosedError: no close frame received or sent
```

[...]

PR #3048, opened 2023-07-08
fix for issue #2475: Streaming api deadlock

chrisrude:
The streaming API runs on its own thread.  This single thread can handle multiple simultaneous requests from different websockets using asyncio.

Currently, when multiple requests are received at once, we try to acquire the shared lock shared.generation_lock multiple times in succession in the same thread, which will cause a deadlock.

To avoid this, we create a decorator which will grab a thread-local asyncio.Lock before we try to acquire the shared lock.  This will prevent the deadlock.

This is a little more complicated than it needs to be because currently there's only one thread, so we could just use a single, global asyncio lock instead of one in TLS.  However this approach will scale in the future if we ever want to have multiple threads handling requests in parallel.

Note: this looks a lot larger than it is, since I extracted the chat and non-chat handlers from `_handle_connection` into two separate methods `_handle_chat_stream_message` and `_handle_stream_message`.  The implementations didn't change at all, though.  If you want I can split this into two commits so it's more clear.


oobabooga:
Thank you, it looks good and it's encouraging to hear that it solved the problem in your tests. 

Question: is the change only necessary for the streaming API? Does the blocking API work as expected for multiple concurrent requests as it is?

chrisrude:
Yes, the blocking API is fine currently... I actually tried applying the same fix to the blocking API, then in my testing realized that it's not necessary. :)

The reason is that the blocking API uses the `ThreadingHTTPServer`, which creates different Python threads to handle different requests.  This partners just fine with the existing `threading.Lock` used by `shared.generation_lock`.

The deadlock here only occurs when multiple requests are handled at the same time by the same Python thread, which is the case for the streaming API.

oobabooga:
Fair enough, I just wanted to make sure. LGTM then, thanks!

PR #3051, opened 2023-07-08
Chat history download creates more detailed file names

UnskilledWolf:
Chat history downloads used to contain the name of the character or mode and a timestamp. That made it easy for me to put all my exports into a folder and have it organized, especially when using Instruct mode. However, when I recently updated to the newest version, I noticed all history downloads were now named "exported_history.json"
I decided to make a fork using the formatting code from a previous commit in order to restore the more detailed file names. Most of my changes are put into one function that works similarly to `save_persistent_history`. 
I briefly tested the code in Chat and Instruct mode on Windows.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -3,8 +3,8 @@ import copy
 import functools
 import json
 import re
-from pathlib import Path
 from datetime import datetime
+from pathlib import Path
 
 import gradio as gr
 import yaml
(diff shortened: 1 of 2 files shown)


oobabooga:
Fair enough, I have made some tests and it seems to work well. It's indeed better to have more descriptive filenames.

PR #3062, opened 2023-07-09
Create llamacpp_HF loader

oobabooga:
Still experimental.

## Motivation

Having a wrapper similar to ExLlama_HF but for llama.cpp. The main goal is to evaluate llama.cpp perplexity and compare it to AutoGPTQ/ExLlama directly. As a bonus, it will allow the transformers samplers to be used for llama.cpp.

## How to use it

1) Download oobabooga/llama-tokenizer using the UI or the download script:

```
python download-model.py oobabooga/llama-tokenizer
```

2) Load a ggml model with `--loader llamacpp_HF` or by setting the loader to llamacpp_HF in the UI.

## Current status

Functional, but 3x slower:

llamacpp_HF:

> Output generated in 15.79 seconds (12.60 tokens/s, 199 tokens, context 7, seed 313220017)

llamacpp:

> Output generated in 5.50 seconds (36.39 tokens/s, 200 tokens, context 7, seed 1413621777)

### DIFF
--- a/api-examples/api-example-chat-stream.py
+++ b/api-examples/api-example-chat-stream.py
@@ -23,7 +23,8 @@ async def run(user_input, history):
         'history': history,
         'mode': 'instruct',  # Valid options: 'chat', 'chat-instruct', 'instruct'
         'character': 'Example',
-        'instruction_template': 'Vicuna-v1.1',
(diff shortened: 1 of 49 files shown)


oobabooga:
The evaluation doesn't work with the same error as in Exllama_HF. So once one gets fixed, both will get fixed and we will be able to compare the perplexities of everything.

PR #3075, opened 2023-07-10
Highlight colors by probability and/or perplexity

SeanScripts:
Adds an extension, perplexity_colors, which colors each token of the output by probability and/or perplexity, and calculates the average perplexity of the message. This is an example of a logits processor extension, and this PR includes the changes from #3029 that enable these kinds of extensions.

Issues: Probably only looks good in dark mode, and doesn't work for code blocks.
Edit: Also doesn't work for exllama or llama.cpp, though this is a general limitation of the logits processors for now. It does work with exllama_hf though.

Example (coloring by perplexity, then probability, then both together):
perplexity_colors_demo

### DIFF
--- a/extensions/perplexity_colors/script.py
+++ b/extensions/perplexity_colors/script.py
@@ -1,0 +1,204 @@
+import gradio
+import torch
+from transformers import LogitsProcessor
+import numpy as np
+
+from modules import shared
+
+params = {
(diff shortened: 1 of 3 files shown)


SeanScripts:
Closing this and reopening since I forgot to make a feature branch

PR #3076, opened 2023-07-10
Add support low vram mode on llama.cpp module

gabriel-pena:
I decided to add support for lowvram mode to llama.cpp module because I had a lot of problems with high vram usage and this will help a lot of people who have low vram

### DIFF
--- a/modules/llamacpp_model.py
+++ b/modules/llamacpp_model.py
@@ -49,6 +49,7 @@ class LlamaCppModel:
             'n_batch': shared.args.n_batch,
             'use_mmap': not shared.args.no_mmap,
             'use_mlock': shared.args.mlock,
+            'low_vram': shared.args.low_vram,
             'n_gpu_layers': shared.args.n_gpu_layers
         }
 
--- a/modules/loaders.py
+++ b/modules/loaders.py
@@ -34,6 +34,7 @@ loaders_and_params = {
         'n_batch',
         'threads',
         'no_mmap',
+        'low_vram',
         'mlock',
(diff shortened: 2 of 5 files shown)


oobabooga:
What does this do?

Dampfinchen:
> What does this do?

Results in KV Cache to get moved back into RAM and thus, freeing VRAM. In my testing with split layers I didn't notice much performance regression but indeed a lower VRAM usage, allowing me to use more layers on the GPU and increasing performance. So yeah its quite nice.

oobabooga:
Thanks for the clarification @Dampfinchen

berkut1:
@gabriel-pena is it really helps? I tested that mode 2 weeks ago on my RTX 2080 8 GB with 13b models, and notice no effect, only -5% of performance.

gabriel-pena:
> @gabriel-pena is it really helps? I tested that mode 2 weeks ago on my RTX 2080 8 GB with 13b models, and notice no effect, only -5% of performance.

Yes, this mode helps a lot.  couldn't run ggml models above 7b q4_ks well on my GTX 1660ti, always had vram overflow.  and using low vram mode I can run models up to type 7b q5_km with little performance loss.  Using low vram mode in a model like 7b q5_km I can generate up to 13 tokens per second

berkut1:
@gabriel-pena  Hmm... ok, maybe this mode is effective for a GPU with less than 8 GB of VRAM. On my video card, it's more beneficial to reduce the GPU layers to fit in VRAM as much as possible..

PR #3078, opened 2023-07-10
Color tokens by probability and/or perplexity

SeanScripts:
Adds an extension, perplexity_colors, which colors each token of the output by probability and/or perplexity, and calculates the average perplexity of the message. This is an example of a logits processor extension, dependent on the changes from #3029 that enable these kinds of extensions.

Issues: Probably only looks good in dark mode, ~~and doesn't work for code blocks~~. Also doesn't work for exllama or llama.cpp, though this is a general limitation of the logits processors for now. It does work with exllama_hf though.
~~This could be made slightly more efficient with torch.topk instead of np.argsort, but I don't think this impacts performance much anyway.~~
I also wanted to add a hover-activated view of the top tokens and probabilities for each generated token, but I wasn't able to get the HTML/CSS working quite right, so this only includes the colorization part that is working.

Example (coloring by perplexity, then probability, then both together):
perplexity_colors_demo

### DIFF
--- a/extensions/perplexity_colors/script.py
+++ b/extensions/perplexity_colors/script.py
@@ -15,10 +15,8 @@ params = {
 class PerplexityLogits(LogitsProcessor):
     def __init__(self, verbose=False):
         self.generated_token_ids = []
-        self.generated_tokens = []
         self.selected_probs = []
         self.top_token_ids_list = []
-        self.top_tokens_list = []
         self.top_probs_list = []
         self.perplexities_list = []
         self.last_probs = None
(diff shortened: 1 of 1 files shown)


SeanScripts:
I implemented the torch.topk optimization that was suggested and also mostly fixed the issue with code blocks.

Panchovix:
Have been testing and using this lately, it is really nice to see something like this, nice work!

oobabooga:
This is simply brilliant.

I'll merge it soon

oobabooga:
It just works

print

Thanks for the PR @SeanScripts. If you can think of any improvements or other similar extensions, feel free to submit new PRs in the future.

PR #3083, opened 2023-07-10
[GGML] Support for customizable RoPE

randoentity:
This adds support for customizable RoPE which is implemented here: https://github.com/ggerganov/llama.cpp/pull/2054 and depends on python bindings here https://github.com/abetlen/llama-cpp-python/pull/466.
Currently this works by building llama-cpp-python from source, I don't think I should add it to the requirements.txt yet.

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -32,10 +32,10 @@ need_restart = False
 
 settings = {
     'dark_theme': False,
-    'autoload_model': True,
+    'autoload_model': False,
     'max_new_tokens': 200,
     'max_new_tokens_min': 1,
-    'max_new_tokens_max': 2000,
+    'max_new_tokens_max': 4096,
     'seed': -1,
     'character': 'None',
     'name1': 'You',
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -127,6 +127,8 @@ parser.add_argument('--cache-capacity', type=str, help='Maximum cache capacity. 
 parser.add_argument('--n-gpu-layers', type=int, default=0, help='Number of layers to offload to the GPU.')
 parser.add_argument('--n_ctx', type=int, default=2048, help='Size of the prompt context.')
 parser.add_argument('--llama_cpp_seed', type=int, default=0, help='Seed for llama-cpp models. Default 0 (random)')
+parser.add_argument('--rope-freq-scale', type=float, default=1.0, help='Size of the prompt context.')
+parser.add_argument('--rope-freq-base', type=int, default=10000, help='Size of the prompt context.')
 
 # GPTQ
 parser.add_argument('--wbits', type=int, default=0, help='Load a pre-quantized model with specified precision in bits. 2, 3, 4 and 8 are supported.')
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -127,8 +127,8 @@ parser.add_argument('--cache-capacity', type=str, help='Maximum cache capacity. 
 parser.add_argument('--n-gpu-layers', type=int, default=0, help='Number of layers to offload to the GPU.')
 parser.add_argument('--n_ctx', type=int, default=2048, help='Size of the prompt context.')
 parser.add_argument('--llama_cpp_seed', type=int, default=0, help='Seed for llama-cpp models. Default 0 (random)')
-parser.add_argument('--rope-freq-scale', type=float, default=1.0, help='Size of the prompt context.')
-parser.add_argument('--rope-freq-base', type=int, default=10000, help='Size of the prompt context.')
+parser.add_argument('--rope-freq-scale', type=float, default=1.0, help='Scale factor for RoPE sampling.')
+parser.add_argument('--rope-freq-base', type=int, default=10000, help='Base frequency for RoPE sampling.')
 
 # GPTQ
 parser.add_argument('--wbits', type=int, default=0, help='Load a pre-quantized model with specified precision in bits. 2, 3, 4 and 8 are supported.')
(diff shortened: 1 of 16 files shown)


modules/shared.py
```diff
@@ -125,6 +125,8 @@ def str2bool(v):
 parser.add_argument('--n-gpu-layers', type=int, default=0, help='Number of layers to offload to the GPU.')
 parser.add_argument('--n_ctx', type=int, default=2048, help='Size of the prompt context.')
 parser.add_argument('--llama_cpp_seed', type=int, default=0, help='Seed for llama-cpp models. Default 0 (random)')
+parser.add_argument('--rope-freq-scale', type=float, default=1.0, help='Size of the prompt context.')
+parser.add_argument('--rope-freq-base', type=int, default=10000, help='Size of the prompt context.')
```
oobabooga, line 129: Should the default not be 2048?

Gyztor:
Looks like both https://github.com/abetlen/llama-cpp-python/pull/466 and https://github.com/ggerganov/llama.cpp/pull/2054 have now done their merges of these pull requests. This should be able to be implemented now. I have done a bit of testing with these specific merges and it does seem to have an affect and work. though it might be a good idea to find a better default setting as the current ones seem to make normal models slower than what i am used to; though that to some extent is to be expected.

IkariDevGIT:
when will this be merged?

modules/shared.py
randoentity: I forgot to change the help text. The actual values correspond to the ones in https://github.com/ggerganov/llama.cpp/pull/2054 and also to the default output of llama-cpp when loading a model in text-generation-webui.

server.py
```diff
@@ -244,6 +244,8 @@ def create_model_menus():
                         shared.gradio['no_mmap'] = gr.Checkbox(label="no-mmap", value=shared.args.no_mmap)
                         shared.gradio['mlock'] = gr.Checkbox(label="mlock", value=shared.args.mlock)
                         shared.gradio['llama_cpp_seed'] = gr.Number(label='Seed (0 for random)', value=shared.args.llama_cpp_seed)
+                        shared.gradio['rope_freq_base'] = gr.Slider(label='RoPE base frequency', minimum=0, step=1, maximum=100000, value=shared.args.rope_freq_base)
```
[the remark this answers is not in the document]
randoentity: Sorry I could use some help here. I only found the alpha slider and that doesn't seem to be relevant for these variables. I looked around in the exllama* branches too to check if I had missed any sliders.

randoentity:
@Gyztor the defaults used here are the same for me as when I load a model from the main branch. So if it is slower then I think we'll need to propose new defaults upstream. That said, maybe we should not be hard-coding any defaults here but get them from the loader.

randoentity:
Rebased to resolve conflicts.

oobabooga:
If I understand this comment correctly https://github.com/ggerganov/llama.cpp/pull/2054#issuecomment-1614991621, the conversion from the ExLlama parameters to the llama.cpp ones is:

* `rope-freq-scale` = 1/`compress_pos_emb`
* `rope-freq-base` = 10000 * `alpha_value` ^ (64/63)

I'm usually on the side of using upstream parameters, but in this case I think I prefer to keep using the familiar definitions.

oobabooga:
Here is a test:

python server.py --model models/llama-7b.ggmlv3.q4_K_M.bin --loader llama.cpp --n_ctx 4096 --compress_pos_emb 2 --n-gpu-layers 1000

> Output generated in 4.83 seconds (39.73 tokens/s, 192 tokens, context 2954, seed 2040993099)

@randoentity and everyone else, does this seem to make sense?

berkut1:
Hmm, I played around with the settings. But for me it looks
`'rope_freq_scale': 1.0/shared.args.compress_pos_emb,`
too aggressive. Using this formula, you cannot set the `rope_freq_scale` value between 0.51 and 0.99.
Maybe it's just me, but for no tuned models, if you want to set context parameters to 8k with compress_pos_emb = 2 or even 4, makes the model's response very perplexing, but if you set `rope_freq_scale` to 0.8 - 0.9 the answer is much better.
I think a customizable RoPE is more useful for non-tuned models, and this hard-coded formula makes it impossible to fine-tune them.
P.S Perhaps this result was only mine, I will be glad to hear if someone got other results. I have used vicuna-13b-v1.3.0.ggmlv3.q6_K

kaiokendev:
@berkut1 `rope-freq-scale` is for models that have been fine-tuned with linear interpolation, i.e. SuperHOT, LongChat, etc. For everything else, including non-fine-tuned models, you want to use `rope-freq-base` which is NTKA interpolation

berkut1:
@kaiokendev yes, I thought about that and test with only, `rope-freq-base`, when `rope-freq-scale = 1`, and it doesn't work well, the most satisfaction results I could get were by reducing to `rope-freq-scale` to 0.7 - 0.9, where `rope-freq-base` was around 30000-50000.
Maybe it's self-indulgence. I'll try to play around with the settings later.

randoentity:
I tested this with `manticore-13b-chat-pyg-superhot-8k-GPTQ-4bit-128g.no-act.order.safetensors` on a ~3K dict with the settings oobabooga mentioned above and got a positive result.
```
USER: {"noncommunicableness": "oxeate", "untriable": "boston", "tautonyms": "hydrometeorological [...]
In the above dict what is the value of the key "noncommunicableness"?
Output generated in 7.11 seconds (3.10 tokens/s, 22 tokens, context 3133, seed 49760514)
```
Output:
`The value of the key "noncommunicableness" in the given dictionary is "oxeate".`

However, it fails on the very last key returning:
`The value of the key "smalter" in the given dictionary is a string.` or `The value of the key "smalter" in the given dictionary is a string containing 10 characters.`
Yet, the second to last key gives a correct result. 

Some more testing reveals that wrapping the dictionary in backticks, triple backticks, double quotes, single quotes nor triple single quotes don't change the result. But wrapping the dictionary in triple double quotes works (albeit with a different output format):
```
The value of the key "smalter" in the given dictionary is a string, which contains multiple words separated by spaces. The full text for this key-value pair is as follows:

"smalter": "spece"
```

kaiokendev:
@randoentity I think there is not much that can be done there. I am still investigating attention and position, but one thing you can try at least for LLaMA models is to use a lot of punctuation and propositions around important information, since LLaMA will assign more attention to regions with those grammar structure. So like you say, if you have some important info, wrap it in a lot quotes and it should help

PR #3095, opened 2023-07-11
Allow specifying the number of newlines to stop generation at.

VldmrB:


VldmrB:
I left in the debugging code, just remove it if anything.

oobabooga:
You can already do that with a custom stopping string like "\n\n\n". This checkbox is essentially a legacy feature at this point that used to be useful for models like OPT.

VldmrB:
`\nHello\nHow are you\nBye\nYep ` is not the same as `\n\n\n`.

PR #3101, opened 2023-07-11
Fix API example for loading models

vadi2:
Fix API example for loading models - `'gptq_for_llama': False` has been deprecated, and you can now have more than just a few loaders.


oobabooga:
Thank you!

Issue #3106, opened 2023-07-12
Decode bug in llama.py and even other models


Foxtr0t1337:
### Describe the bug

At llama.py
line 999 
`"text": self.detokenize([token]).decode(`

We can see that if the token only contains part of a character, things will go wrong.
The "ignore" arg makes it hard to spot.
The outcome is that some chars such as Chinese will be missing in the return value of function `generate(self, prompt, state, callback=None)` causing missing chars on the web UI.
After set it to "replace", I found that there is many '�' where chars are missing.
I CAN NOT fix this because there is other models which I can't test.
What I can do is only making "text" bytes and make it a problem for generate function.
I can't make a pull request because I can't fix only this model and cause more bugs.

Other models may contain this bug too but idk.
There are some issues which may relate to this bug which are still open.
Maybe some of you who have other models can test those and we come up with a final fix.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

I used a llama.cpp model who can speak Chinese.
Just speak to them and some times there will be missing chars which looks like a training problem but not.

### Screenshot

After I change it to "replace"
<img width="738" alt="Screenshot 2023-07-12 at 8 47 38 AM" src="https://github.com/oobabooga/text-generation-webui/assets/48181014/e2da2523-48f2-4282-9cb1-078999d2d73b">

After some tweaking
<img width="409" alt="Screenshot 2023-07-12 at 9 37 30 AM" src="https://github.com/oobabooga/text-generation-webui/assets/48181014/10c80d9e-0c5c-425f-b41f-4612305bd41a">

### Logs

```shell
Not a crash.
```

### System Info

```shell
Apple M2 Max latest OS.
```

### Update

After tweaking some code I successfully fixed the bug on the model I've been using.
I decided that I will try to fix other models and do a pull request?

Me again.
There is no way I can do a pull request since llama.py and some other files are not belong to this repo.
However I am willing to contribute my code if somehow it is needed.

The main idea of my fix is making generate function return bytes and we decode this in generate_reply_wrapper function.
<img width="1176" alt="Screenshot 2023-07-12 at 10 01 33 AM" src="https://github.com/oobabooga/text-generation-webui/assets/48181014/fa3567bf-b426-4317-93ed-10ecf7e5f756">
This way we won't ignore any incomplete chars.
It is important to know that this screenshot is not all the code I changed.
Other changes are only replacing '' to b'' etc. and cancel decoding at llama.py line 999.
[...]

PR #3108, opened 2023-07-12
Fix missing chars

Foxtr0t1337:
The bug was caused by llama_cpp lib.
I rewrote the generate and generate_with_streaming function to bypass the buggy part.

#3106


Foxtr0t1337:
Don't know how to update this PR so just close it lol

PR #3111, opened 2023-07-12
[google translate]Implementing Language Selection Memory and Non-translation of Code Blocks

melodysdreamj:
1. Language Selection Memory: With this feature, the system will remember the last chosen language by the user across sessions. This ensures a seamless user experience by eliminating the need for users to repeatedly select their preferred language in every session.

2. Non-translation of Code Blocks: In recognition of the fact that code blocks typically need to remain in their original form for proper functioning, this feature prevents the translation of such blocks. This way, we preserve the integrity of code and facilitate better understanding and utilization.

Please review these changes and provide any feedback.

### DIFF
--- a/extensions/google_translate/script.py
+++ b/extensions/google_translate/script.py
@@ -1,12 +1,28 @@
 import gradio as gr
 from deep_translator import GoogleTranslator
+import os
+import re
+import uuid
 
 params = {
     "activate": True,
-    "language string": "ja",
+    "language string": "ko",
 }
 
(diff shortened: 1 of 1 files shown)


Touch-Night:
I think this extension is more like an example for extension developers, so improvement to this is not necessary, there are many good translation extensions already.

extensions/google_translate/script.py
```diff
@@ -1,13 +1,29 @@
 import gradio as gr
 from deep_translator import GoogleTranslator
+import os
+import re
+import uuid
 
 params = {
     "activate": True,
-    "language string": "ja",
+    "language string": "ko",
```
Touch-Night, line 9: This change is not necessary.

PR #3120, opened 2023-07-13
Allow multiple loaded LoRAs to simultaneously influence the output

### LINKED ISSUE #3096: With multiple LoRAs loaded, only the first one in the list affects the generated text output
### Describe the bug

We're experimenting with training "stacked" LoRAs, as in LoRAs trained on top of other LoRAs. According to the warning message triggered by this, such an arrangement is supposed to work, but "may have unexpected effects". Training succeeds without any issue, but when I load the finished LoRAs, only the first one seems to have any effect.

So far, we have encountered and successfully worked around an issue where the 2nd LoRA's weights would never get loaded due to a PEFT bug/weirdness (the 2nd LoRA's weights get serialized with module names starting with "base_model.model.base_model.model.model", but on deserialization, PEFT expects "base_model.model.model"). Patching this made the 2nd LoRA load and affect the generated output successfully when it's the only LoRA being loaded. However, when we load it on top of the 1st LoRA, it still has no effect. With the loading order reversed compared to the training order, only the 2nd LoRA has any effect on the output - as if the 2nd and subsequent LoRAs were basically ignored.

To test this, we tried loading two known good, independently trained LoRAs (tloan/alpaca-lora-7b and 22h/cabrita-lora-v0-1) on a common base model (yahma/llama-7b-hf). Each LoRA works fine when loaded on its own, and is seen to affect the output.

When the "LoRA(s)" box is set to list alpaca first, then cabrita, both are shown as successfully loaded, but the generated text is identical to what I get with only alpaca loaded. If I put cabrita first, then alpaca, the output is indistinguishable from only cabrita being selected. This seems to confirm that only the first LoRA listed in the box is properly loaded into the model, the rest are silently ignored.

BnB quantization doesn't appear to affect this behavior - it's the same in 4-bit, 8-bit and native mode.

### Reproduction

Download yahma/llama-7b-hf, tloen/alpaca-lora-7b and 22h/cabrita-lora-v0-1 from Hugging Face.

Start the UI in chat mode.
On the Parameters tab, set the temperature to 0.3.
On the Chat settings tab, select the "Alpaca" instruction template.
On the Model tab, load the yahma/llama-7b-hf base model.

Without loading any LoRA, switch to the Text generation tab, and run the following 3 queries (clear history after each one):
- "Tell me about alpacas."
- "Invente uma desculpa criativa pra dizer que não preciso ir à festa."
- "How many helicopters can a human eat in one sitting?"

Note the responses. (The Portuguese query will prompt a response in English, while the helicopter one will be nonsense.)

On the Model tab, reload the model, then load the alpaca LoRA. Run the above 3 questions again. Note how the answers change. (The Portuguese prompt is now answered in Portuguese, but not very creatively, and the helicopter answer will be correct, i.e. none.)

Unload the alpaca LoRA, reload the model, and load the cabrita LoRA. Run the 3 prompts, and note the new answers. (Portuguese answer is different, more creative, while the helicopter one is back to nonsense.)

Unload the LoRA, reload the model. In the LoRA(s) box, select cabrita first, then alpaca. Run the 3 prompts. (Answers are as if only cabrita were loaded!)

Unload the LoRAs, reload the model. In the LoRA(s) box, select alpaca first, then cabrita. Run the 3 prompts. (Answers indistinguishable from when only alpaca was loaded!)

### Screenshot

_No response_

### Logs

```shell
no exception generated
```

### System Info

```shell
Ryzen 7950X with a Zotac GeForce RTX 4080 GPU, running Ubuntu 22.04 LTS with the open-source NVIDIA kernel module (with nvidia.NVreg_OpenRmEnableUnsupportedGpus=1 specified on the kernel command line to skip the "datacenter GPUs only" check). The issue is reproducible in both CUDA and CPU mode.
```

Googulator:
This is probably what most users would expect when loading multiple LoRAs: that each would contribute with equal weight to the model's output, with the effect being a combination of what each LoRA would produce on its own. However, peft by default "activates" the first LoRA to be loaded, making it the only one with any effect.

Instead, explicitly merge the LoRAs into a combined adapter, then activate that. This way, multiple LoRAs perform as expected.

Code loosely adapted from the Playground extension.

Fixes #3096


FartyPants:
This will immediately disallow loading multiple loras with different rank and then quickly switching between them.

Googulator:
"Allow switching between them at will" is currently not a feature in this UI, except via extensions. Right now, the UI gives no indication that only one LoRA is actually active, and which one it is.

But yes, dissimilar ranks do need handling - I will add a warning message for that case

FartyPants:
I think instead of deciding what "most users" expect, this should be handled properly with changes to interface without making entire mess of the logic that is not very obvious to user.
add_weighted_adapter creates an additional adapter (but not visible on interface) to the ones already in memory, so what happens if user then removes one of the adapters?  That doesn't remove the merged adapter, but user can't select it anymore. What if users removes an adapter and adds different adapter? That would create yet another adapter. User tries a few LoRAs and each time another hidden adapter is created in GPU?
That's what I am saying, don't do hidden things by default that each time eat half gig of GPU.
It needs to have an interface (like my extension has) where is it on user if he wants to switch between lora or add yet another lora and should be able to switch between all of them, not adding more adapters whenever user just select something in the box.
My 2 cents...

Googulator:
Whenever a LoRA needs to be "unloaded", the entire PEFT model is discarded, and a new one is built in its place. Thus, no VRAM leak actually occurs, assuming that GC actually does its thing properly. (Loading and merging 2 LoRAs, then loading a 3rd one is, however, something that needs to be handled - IMO the same way, by doing a full PEFT model rebuild whenever a merged adapter is in the way.)

The current situation is actually:
Load adapter A + 0.5GB
Load Adapter B + 0.5GB
Remove B (triggers rebuild of PEFT model, discarding all adapters not in the current loading set) - 0.5GB
Load adapter C + 0.5GB

With this patch, it becomes:
Load adapter A + 0.5GB
Load Adapter B (triggers merge) + 1.0GB
Remove B (triggers rebuild of PEFT model, discarding all adapters not in the current loading set, including merged one) - 1.0GB
Load adapter C (triggers merge) + 1.0GB

"The interface should follow the logic PEFT set." - I **strongly** disagree. As a user of this webUI, I don't care about PEFT in any way - I don't even want to know a thing named "PEFT" exists. It's an implementation detail. The UI gives the appearance that multiple LoRAs can be loaded simultaneously, and with no notion of an "active adapter" or "selecting among loaded adapters", the only meaningful interpretation is that all loaded LoRAs are supposed to participate in subsequent inference or training.

And yes, Playground gives more control over the process, but for casual users, that's less of a strength and more a weakness. Especially since Playground doesn't take over LoRA loading from the Models tab, but rather relies on adapters loaded on Models, the user now needs to switch between 2 tabs and jump through several hoops to set up the expected behavior of multiple stacked LoRAs (again, expected _by the user,_ not by PEFT's developers), and once set up, it seems extremely fragile (what happens if you switch back to the Models tab and make adjustments after setting up the LoRA-Rama tab? _Even if it does work properly, it's not obvious to the user whether or not that's safe to do!_ ).

The only alternative I can see is to completely forbid the loading of multiple LoRAs when an extension for switching between them isn't loaded. Presenting an interface for loading multiple LoRAs as if it were for simultaneous use, when in reality it's just an extension point to enable fast switching manageable only by 3rd party extensions misleads the user, and looks, for all intents and purposes, like a bug. (Going back in commit history, it's clear that when multi-LoRA loading was added, the expectation, even by developers, was that the adapters would cooperate, not that only one would be active at a time.)

Now, a more feature-rich PEFT is great news, but I would not delay bugfixes on what everyone already has, just because a newer implementation is coming in the future. Don't let perfect be the enemy of good.

Googulator:
My use case is actually _dependent_ LoRAs, that is, one trained on top of another. AFAIK this is meant to be a fully supported feature - yet on inference, it's completely broken right now. Switching between adapters also doesn't make sense in this use case - since LoRA #2 was trained on top of LoRA #1, it depends on LoRA #1 being loaded first, for any inference to be possible.

BTW, tested "Load A, B - Apply - refresh in PG - 2 adapters, Remove B and put C - Apply, refresh in playground" - the adapters shown are A, C and merged. B is being unloaded, as expected..

oobabooga:
I wasn't aware that multi-LoRA was not working properly on PEFT. My intention is simply:

1) User selects 2 LoRAs in the menu and clicks on "Apply LoRA(s)"
2) The generation now uses both LoRAs with equal weight

The following should also work:

1) User removes 1 of the previously selected LoRAs from the menu and clicks on "Apply LoRA(s)"
2) The generation now only uses the first one

This could be additionally be expanded in the future with LoRA weights somehow, but that's the essential starting point.

Does this PR cover those cases? In the case where the user tries to load two LoRAs with different ranks, does the PR remove the functionality to use those LoRAs simultaneously?

Googulator:
Removing a LoRA and clicking "Apply" triggers a PEFT rebuild, like before, so yes, only one LoRA will be active when going from 2 loaded adapters to 1.

When LoRAs with different ranks are loaded, a warning message is now displayed telling the user explicitly that only the first one will have an effect.

FartyPants:
Just make sure you don't update PEFT, because this implementation would not work with the recent version.

practical-dreamer:
Has this been resolved or is this PR still required to get multi-LorA merging to work?... any way to work with exllama_HF loader?

FartyPants:
PEFT also supports cat now - which is "the correct" merging by summing of both LORA's so you genuinely have LORA A (plus) LORA B instead of previous linear LORA A (linearly averaged with)  LORA B. The downside is that it sums the ranks so two 64 ranks will make 128 rank and if you go overboard you may end up with a few GB of LORA.

I'll try first to update the playground, because due to other changes the LORA merging is borked. 
I think WebUI needs an entirely separate LORA tab that would allow merging, swapping, deleting and merge and unload.
I kind of feel that I need to start from scratch and make a LORA-RAMA separate extension because everything is being poured into one code.

oobabooga:
I was confused by the discussion above, but looking at the code alone, it

1) Does not interfere with the single-lora use case.
2) Adds support for LoRA merging in the multiple-lora case. I don't know if further improvements can be made, but if so, that can be iterated on in the future.

Let's merge this PR and update the documentation to mention the multi-LoRA support in the transformers loader. It also works for GPTQ-for-LLaMa, since the monkey patch makes it indisguinshable from a regular transformers model (not that anyone uses GPTQ-for-LLaMa anymore).

PR #3138, opened 2023-07-13
Make it possible to evaluate exllama perplexity

oobabooga:
Similar to what I'm trying to do with https://github.com/oobabooga/text-generation-webui/pull/3062, but this time for exllama. It's already functional and the perplexity can be evaluated.

This allows ExLlama/AutoGPTQ/Transformers to be directly compared.

The only question is whether I am handling the cache properly or not.

I'll tag @turboderp for awareness

### DIFF
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -56,15 +56,15 @@ class ExllamaHF(PreTrainedModel):
 
         if labels is None:
             if cache is None:
-                cache = self.ex_cache
                 self.ex_cache.current_seq_len = 0
+                cache = self.ex_cache
                 self.ex_model.forward(torch.tensor([seq[:-1]], dtype=torch.long), cache, preprocess_only=True, lora=self.lora)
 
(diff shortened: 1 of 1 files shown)


turboderp:
At a quick glance it seems correct. As a rule, after generation the length of the cache should always be one less than the length of the generated sequence, i.e. `cache.current_seq_len` is the total amount of tokens that have actually been run through the model. 

When starting a new sequence you want to clear the cache by setting `cache.current_seq_len = 0`. During any forward pass, `current_seq_len` increases by exactly the length of the input_ids, and it's the only way the length of the cache is tracked. Also, a cache allocated for 2048 tokens should support a forward pass with up to 2048 input tokens exactly.

oobabooga:
Thanks for the review and the information @turboderp. 

I have used the code here to compare ExLlama to llama.cpp and AutoGPTQ directly. The data can be found [here](https://oobabooga.github.io/blog/posts/perplexities/). This allowed me to confirm that ExLlama yields the same perplexity as AutoGPTQ for the same model.

PR #3142, opened 2023-07-14
Attempt to using latest versions of requirements

ofirkris:
Accelerate
Pillow
Fastapi
llama cpp python
Gradio + client


oobabooga:
Things need to be tested before updating. llama-cpp-python is usually safe, the other ones are not

PR #3148, opened 2023-07-14
Add probability dropdown to perplexity_colors extension

SeanScripts:
@oobabooga I've expanded the functionality of the perplexity_colors extension to be able to show a probability dropdown when hovering over a token of the output. This can be activated independently of the options to color by probability/perplexity. It shows the top 5 tokens and probabilities, and a 6th row with the selected token if it wasn't in the top 5. It also includes a row at the end with the perplexity of this next token distribution.
Using dev as the PR target because it currently has the logits processors update that is required as a dependency for this extension.

Issues: Doesn't look good outside of dark mode, and using the probability dropdown messes up some whitespace like removing paragraph breaks

Example of the kind of hover-activated dropdown menu that this creates for each token:
probability_dropdown

### DIFF
--- a/extensions/perplexity_colors/script.py
+++ b/extensions/perplexity_colors/script.py
@@ -4,6 +4,7 @@ from transformers import LogitsProcessor
 import numpy as np
 import time
 import re
+import markdown
 
 from modules import shared
 from modules import html_generator
@@ -112,7 +113,7 @@ def output_modifier(text):
 
     i = 0
     for token, prob, ppl, top_tokens, top_probs in zip(gen_tokens, sel_probs, perplexities, top_tokens_list, top_probs_list):
-        if '`' in token:
+        if '`' in token and not params['probability_dropdown']:
             in_code = not in_code
             continue
         if in_code:
@@ -298,7 +299,7 @@ def convert_to_markdown(string):
     #t1 = time.time()
     #print(len(string))
     #print(f"Pre markdown: {(t1-t0):.3f} s")
-    if params['probability_dropdown']:
+    if params['probability_dropdown'] and '<div class="hoverable">' in string:
         # Prevents all latency introduced by trying to convert the HTML to markdown when it's not even necessary
         #print('Monkeypatched')
         return string
--- a/extensions/perplexity_colors/script.py
+++ b/extensions/perplexity_colors/script.py
@@ -2,8 +2,12 @@ import gradio
 import torch
 from transformers import LogitsProcessor
 import numpy as np
+import time
+import re
 
 from modules import shared
+from modules import html_generator
+from modules.html_generator import replace_blockquote
 
 params = {
     'active': True,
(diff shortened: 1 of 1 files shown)


SeanScripts:
I realized this breaks the chat scroll. I'll try to figure out how to fix that.

SeanScripts:
Okay, I noticed a larger issue, which is that the probability dropdowns cause generation speeds to slow down quite a lot, dependent on how many tokens with dropdowns have been added to the history. I've been looking into it, and tracked down the source of the latency. 

It appears that in `html_generator.py`, the `markdown` function is called on every message of the history once per token. This is pretty slow to run on the large HTML content generated by this extension (about 100-200 ms for a large message), and I don't feel it's necessary to run this on previous messages every token. I'm thinking that instead, the history should store the markdown-converted messages to avoid having to rerun this conversion on all the messages every time, and the `generate_*_html` functions should just use the stored markdown except for the current message where the new conversion might actually be needed.

FartyPants:
Makes sense to me

SeanScripts:
For now, I added a monkeypatch to `convert_to_markdown` from `html_generator.py` that bypasses the markdown conversion when using probability dropdowns. This is not a great long-term solution, because it's still being called for every message for every token, but it does prevent the latency issue here.

I also removed the CSS that was causing the scrollbar to disappear. Now the tables won't be visible when they extend past the edges of the chat area, but having a scrollbar is more important... If anyone has any ideas on how to solve this, let me know. I'm not very good at CSS.

oobabooga:
I wanted to have a clean solution for the markdown problem, but I think it's best to merge this PR with the monkey patch for now and work on that later. Simply storing visible messages as HTML can lead to issues with the "copy last reply", "send dummy reply", "replace last reply", and "continue" functions, so each of those needs to be tested carefully.

tdrussell:
Just became aware that this extension exists. The probability dropdown feature is exactly what everyone asking for a "logit viewer" really wants. This is almost perfect! It doesn't seem to currently work in the default or notebook tabs, it just shows a big dump of HTML. I get that this extension was probably developed with chat in mind, but is it possible to also make it work in those other tabs?

tdrussell:
Okay, I kinda figured out the problem in the default tab myself. It is [this line of code](https://github.com/oobabooga/text-generation-webui/blob/429cacd71557ca974bab09a7498468feec172a91/modules/html_generator.py#L92) that escapes HTML in the generate_basic_html() function. If you comment that out, this extension works in the default tab.

Looks like that HTML escaping was added a few days ago. Should it be removed? From the code flow, it looks like the intention of the HTML tab is to just directly render any HTML in the string, so it seems like it shouldn't be escaped.

oobabooga:
@tdrussell I fixed this bug here https://github.com/oobabooga/text-generation-webui/commit/a74dd9003f0f06395cceaf1565618c6f6f70bf61. Now the extension works normally in Default/Notebook tabs.

PR #3149, opened 2023-07-15
Increase alpha value limit for NTK RoPE scaling for exllama/exllama_HF

Panchovix:
NTK suffers a degradation before the "theoretical" value compared to embedding compression. This is an issue that happens for now with "NTKv1". This can be fixed by applyng https://github.com/jquesnelle/scaled-rope/pull/1, among other fixes, but it is not implemented yet on exllama.

Some suggested values, for NTK Alpha, for base or scaled models (for example, https://huggingface.co/bhenrym14/airoboros-33b-gpt4-1.4.1-NTK-16384-GPTQ) are:

- α equal to 2 = 3400 context
- α equal to 4 = 6200 context
- α equal to 8 = 8800 context
- α equal to 16 = 13K? context
- α equal to 32 = 20K? context

image

Credits to the anon which gave this info.

### DIFF
--- a/server.py
+++ b/server.py
@@ -227,7 +227,7 @@ def create_model_menus():
                         shared.gradio['gpu_split'] = gr.Textbox(label='gpu-split', info='Comma-separated list of VRAM (in GB) to use per GPU. Example: 20,7,7')
                         shared.gradio['max_seq_len'] = gr.Slider(label='max_seq_len', minimum=2048, maximum=16384, step=256, info='Maximum sequence length.', value=shared.args.max_seq_len)
                         shared.gradio['compress_pos_emb'] = gr.Slider(label='compress_pos_emb', minimum=1, maximum=8, step=1, info='Positional embeddings compression factor. Should typically be set to max_seq_len / 2048.', value=shared.args.compress_pos_emb)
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Have been happily using alpha of 2 on unmodified models and didn't experience any difference up to ~34-3500.. which is where my memory taps out anyways for 30b or 65b.

Is there a particular reason to dl for instance https://huggingface.co/bhenrym14/airoboros-33b-gpt4-1.4.1-NTK-16384-GPTQ besides the extra .1 in the airoboros dataset. Any model will follow those alpha value results at minimal or no perplexity loss, correct?

Panchovix:
> Have been happily using alpha of 2 on unmodified models and didn't experience any difference up to ~34-3500.. which is where my memory taps out anyways for 30b or 65b.
> 
> Is there a particular reason to dl for instance https://huggingface.co/bhenrym14/airoboros-33b-gpt4-1.4.1-NTK-16384-GPTQ besides the extra .1 in the airoboros dataset. Any model will follow those alpha value results at minimal or no perplexity loss, correct?

You're correct. It seems the alpha values seem to work in general for base models as well. In theory, a finetuned model should perform better than a base model given these alpha values.

oobabooga:
Thanks @Panchovix, I wasn't aware that using a smaller context than would be naively implied by the alpha value is the best option.

PR #3151, opened 2023-07-15
[Bug fix] Remove html tags form the Prompt sent to Stable Diffusion

SodaPrettyCold:
When user use **sd_api_pictures** and **silero_tts** at the same time, the Prompt sent to SD will contains the following HTML elements:

`<audio src=file/extensions/silero_tts/outputs/XXX_161233316.wav controls autoplay></audio>  `

It will strongly scramble the result of the image generation.

Here is a simple code to reduce the pollution of prompt when using TTS, and add brackets to make it more tidy.

```
    import re

    description = re.sub('<audio.*?</audio>', ' ', description)
    description = f"({description}:1)"
```

### DIFF
--- a/extensions/sd_api_pictures/script.py
+++ b/extensions/sd_api_pictures/script.py
@@ -133,6 +133,11 @@ def get_SD_pictures(description, character):
     if params['manage_VRAM']:
         give_VRAM_priority('SD')
 
+    import re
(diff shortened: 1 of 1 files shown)


oobabooga:
That's nice, thank you.

PR #3158, opened 2023-07-15
Lowering gradient accumulation to go around Transformers bug

FartyPants:
This is a quick & dirty fix to go around Transformers issue when the dataset is too small. I'll try to talk to Transformers people about it - because it seems they are fixing it in the wrong place.
The proper fix should be on transformers training side, safeguarding the layman induced data. (obviously gradient_accumulation_steps should not cross the amount of data in epoch but how many people out of 100 even think of that?)

I have far more additions to training (currently testing) so if you want to wait, let me know and we can do this all in 1
Although I'm not sure how @oobabooga is comfortable with me slicing and dicing the training.py because that's what I'm secretly doing :)


FartyPants:
I'll add it to next RP, since I keep adding more changes to training

PR #3166, opened 2023-07-16
Dockerfile <- added openai requirements; docker-compose.yml <- replac…

mongolu:
Modified Dockerfile to include OPENAI's extension requirements.txt.

Also switched device_ids to [count: all](https://docs.docker.com/compose/gpu-support/) to let the machine use all available GPUs.


mongolu:
> …ed device_ids: [ '0' ] with count: all (to let all available GPUs to be used)

https://docs.docker.com/compose/gpu-support/

Issue #3184, opened 2023-07-18
The response of starchat-beta models does not stop properly.


giprime:
### Describe the bug

I'm using starchat-beta with this webui.

When I pass some prompt to that model it responses properly. but it doesn't finishes generating tokens after responses to the question.

The options selected during the experiment are shown below.
  1. chat mode
  2. [Chat settings] - [Instruction template] - [Instruction template]: "Starchat-Beta"
  3. [Parameters] - [Generation parameters preset]: "StarChat"

I'm guessing that the issue is caused by the tokenizer or webui's special token handling process.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. [Chat settings] - [Instruction template] - [Seed (-1 for random)]: "2"
2. [Text generation] - [Input]: Who is faster, the lion or the horse?
(It may seem odd to have a natural language query on the model that generates the code, but it is input to reproduce the buggy situation. When actually querying the code, the same type of response is found for responses that do not end with a code block.)

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/7627129/7618b32f-2bf1-45fb-902a-656421b27e3f)

![image](https://github.com/oobabooga/text-generation-webui/assets/7627129/1b14acc5-794c-4511-9592-3f543b01694e)

### Logs

```shell
(textgen) giganticprime@nvidia-DGX-Station:~/text-generation-webui$  cd /home/giganticprime/text-generation-webui ; /usr/bin/env /home/giganticprime/miniconda3/envs/textgen/bin/python /home/giganticprime/.vscode-server/extensions/ms-python.python-2023.12.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher 41065 -- server.py --chat --listen 
[2023-07-18 16:02:56,980] [INFO] [real_accelerator.py:110:get_accelerator] Setting ds_accelerator to cuda (auto detect)
bin /home/giganticprime/miniconda3/envs/textgen/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117_nocublaslt.so
2023-07-18 16:02:59 INFO:Loading the extension "gallery"...
Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
2023-07-18 16:03:17 INFO:Loading HuggingFaceH4_starchat-beta...
[...]

PR #3185, opened 2023-07-18
Added the logic for starchat model series

giprime:
#3184 

The eos_token value of "special_tokens_map.json" in the storage of starchat(alpha, beta and later) is not actually specified to eos_token during instruction tuning.
(It should be <|end|> but specified to <|endoftext|>)

Changing the eos_token value in starchat-beta's "special_tokens_map.json" could be an alternative, but owner of the repo doesn't want to change that because it could cause problems in other existing downstream applications.
https://huggingface.co/HuggingFaceH4/starchat-beta/discussions/3

Therefore, as part of a downstream task, I'd like to add the logic for starchat model series to modules/models.py in this repository.

### DIFF
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -231,6 +231,7 @@ TheBloke_WizardLM-30B-GPTQ:
 .*starchat-beta:
   mode: 'instruct'
   instruction_template: 'Starchat-Beta'
+  custom_stopping_strings: '"<|end|>"'
 .*minotaur:
   mode: 'instruct'
   instruction_template: 'Minotaur'
--- a/modules/models.py
+++ b/modules/models.py
@@ -2,7 +2,6 @@ import gc
 import os
 import re
 import time
-import json
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


oobabooga:
I think that a better solution would be to add a regex for starchat to `models/config.yaml` with a `custom_stopping_string` like what is done for llava:

https://github.com/oobabooga/text-generation-webui/blob/main/models/config.yaml#L92

giprime:
I confirmed that the model output changes normally with the mentioned method.
Excluding the previously added logic, the following changes were made to models/config.yaml .

[before]
.*starchat-beta:
  mode: 'instruct'
  instruction_template: 'Starchat-Beta'
  **custom_stopping_strings: '"<|end|>"'**

[After]
.*starchat-beta:
  mode: 'instruct'
  instruction_template: 'Starchat-Beta'

oobabooga:
Thank you for the quick revision!

PR #3188, opened 2023-07-18
Update script.py

sammyf:
stuffed Send_Picture into an accordion to clean up the UI a bit


oobabooga:
I prefer the current looks.

PR #3191, opened 2023-07-18
Proposal, augment context with date, day, etc...

FartyPants:
This is a proposal to augment the context string of chat/instruct with date and perhaps other relevant info.
This is not heavy handed, it's entirely on user to use it or not, by augmenting the context string. Honk, honk! Repetition!

image

Note: having a date in system prompt even without some special need for it may be beneficial to prime LLM to be more relevant in answers. 

Again, this is a proposal. A "no, go to hell FartPants with your uneducated ideas," is perfectly acceptable outcome. It's commonly used by the members of my family, so it feels warm and cozy.

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -5,6 +5,7 @@ import json
 import re
 from datetime import datetime
 from pathlib import Path
+from datetime import date
 
 import gradio as gr
 import yaml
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
I wonder what training the AI on timestamps and then also giving it some in context or the prompt would do. As of now, none of the models I have used have any understanding of time at all.

PR #3211, opened 2023-07-19
Bump exllama module to 0.0.7

jllllll:
Adds support for LLaMa 2


RandomInternetPreson:
Frick!! Thank you so much, I can run the 70B models super fast now, you and oobabooga are changing the world and helping tremendously to democratize llms.

PR #3225, opened 2023-07-20
Add checks for ROCm and unsupported architectures to llama_cpp_cuda loading

jllllll:
This will hopefully prevent issues with AMD GPUs and CPUs that don't have a compatible wheel available.

Most of Pytorch's cuda code is made to be re-used by their ROCm build. This results in pretty much all checks like `torch.cuda.is_available()` returning true for ROCm AMD GPUs. Simple enough to work around with checking `torch.version.hip`


oobabooga:
Thanks for the fixes. I think that using the `--cpu` flag for this import could cause confusion, as it's not a llama.cpp input parameter, so I have removed this bit.

PR #3241, opened 2023-07-22
Add transformers_stream_generator to requirement.txt

aetherwu:
So that it can load baichuan-inc_Baichuan-13B-Chat seamlessly.

Model loader:
Transformers

Model refference:
https://huggingface.co/baichuan-inc/Baichuan-13B-Chat


oobabooga:
Users can install this manually. Many models require additional requirements

PR #3243, opened 2023-07-22
Updated download_urls.py with support for bypassing soft script blocking 

iongpt:
Added support for fetching data from sites that soft block scripts, by specifying a user-agent in the request

### DIFF
--- a/extensions/superbooga/download_urls.py
+++ b/extensions/superbooga/download_urls.py
@@ -4,7 +4,10 @@ import requests
 
 
 def download_single(url):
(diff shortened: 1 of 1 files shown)


iongpt:
Nowadays many web frameworks have built in user agent check and block for scripts. 

I run into this issue while trying to index a website that I have full access to it, I was getting 403 Forbidden when trying to fetch it from the script. It had a soft block by checking the user agent.

It started working after I added the user agent to the script.

Issue #2314, opened 2023-05-24
UI can no longer load RWKV .pth models


dblacknc:
### Describe the bug

When an RWKV .pth (old style) model is selected in the UI, it immediately comes back with a successful load, and with an error of "model does not exist" in the console.   (The model is not loaded.)

I *can* still load the model on the command line and chat works normally.  It appears this broke around the time the new style model support was added.

NB: I'm also opening an issue about a problem with the new-style models; this is separate.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Put RWKV-4-Raven-7B-v11x-Eng99%-Other1%-20230429-ctx8192.pth in the models folder.

Run server.py with ... --model RWKV-4-Raven-7B-v11x-Eng99%-Other1%-20230429-ctx8192.pth - works normally.

Select and load same model in the UI.   After a couple seconds, the UI displays:

 Successfully loaded RWKV-4-Raven-7B-v11x-Eng99%-Other1%-20230429-ctx8192

...and this in the console.  The UI is able to load all other models no problem; this is unique to the .pth model files.
I also tried it with a 14B model, same result.

INFO:Loading RWKV-4-Raven-7B-v11x-Eng99%-Other1%-20230429-ctx8192...
ERROR:The path to the model does not exist. Exiting.

### Logs

```shell
N/A - all available into is in the reproduction section.
```

### System Info

```shell
Ubuntu 22.04
2 x RTX3060 12 GB
96 GB RAM
git pull as of this morning, followed by pip install -r requirements.txt --upgrade
```

shinkarom:
I get this too.

PR #3254, opened 2023-07-22
add rwkv world tokenizer

LoganDark:
I had written multiple screens of text here but gpg-agent was unresponsive so .git deleted all my hard work when I tried to re-commit.

Gist is:

- Support world models with --rwkv-world flag.
- Load tokenizer before model to avoid wasting work if tokenizer fails to load. Rwkv python package is slow.

Tokenizer implementation used is my own, <https://github.com/BlinkDL/ChatRWKV/pull/137>. It outperforms everything else, but hasn't been accepted upstream, so I just pasted it. It'll be fine.

Notes:

- Rwkv package not installed by windows auto installer, had to manually enter conda env and use pip.
- Can't load pth files by web interface due to #2314, falsely closed by retarded stale bot, which is useless and counterproductive. But I digress. When it's possible to load from the web interface again, there should be a way to specify whether to use the world tokenizer there, so you can switch between raven and world models freely.
- Could autodetect the type of model perfectly by checking size of emb tensor, which represents vocab size, raven is 50277 and world is 65536, but can't do that without loading the whole thing into memory because pytorch is the world's worst library and can't read a single tensor from a pth file. Very bad idea as `rwkv` already loads the entire file due to pytorch's brain death, loading it a second time would be insane.

Sorry, I wrote 3-4x more than this last time, but I am not doing that again, because git overwrote it permanently, because it is a stupid piece of garbage. Hope this is an apt description.

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch rwkv-world
# Changes to be committed:
#       modified:   README.md
#       modified:   modules/RWKV.py
#       modified:   modules/models.py
#       modified:   modules/shared.py
#

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Its goal is to become the [AUTOMATIC1111/stable-diffusion-webui](https://github.
 ## Features
 
 * 3 interface modes: default, notebook, and chat
-* Multiple model backends: tranformers, llama.cpp, AutoGPTQ, GPTQ-for-LLaMa, ExLlama, RWKV, FlexGen
+* Multiple model backends: tranformers, llama.cpp, AutoGPTQ, GPTQ-for-LLaMa, ExLlama, RWKV, RWKV-World, FlexGen
 * Dropdown menu for quickly switching between different models
 * LoRA: load and unload LoRAs on the fly, load multiple LoRAs at the same time, train a new LoRA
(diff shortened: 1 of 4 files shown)


oobabooga:
RWKV can now be run through the transformers library. The RWKV code here is kept for legacy purposes and is no longer maintained.

LoganDark:
> The RWKV code here is kept for legacy purposes

ty. I use this code in projects that do not support transformers anyway, so this is reasonable.

PR #3257, opened 2023-07-23
Bump llama-cpp-python to 0.1.74

jllllll:
Not sure if it was 0.1.73 or this version, but I'm getting significantly faster GPU acceleration speeds than I was previously. On a 13B model, I was getting 8-9 t/s with my 1080ti. Now, I am getting 15+ t/s with the highest I've seen being 18.


oobabooga:
For me running 65b on a 3090 these are the speeds for 2 consecutive generations:

73:

Output generated in 42.23 seconds (0.95 tokens/s, 40 tokens, context 7, seed 1925546470)
Output generated in 29.09 seconds (1.37 tokens/s, 40 tokens, context 47, seed 1702379023)

74:

Output generated in 37.97 seconds (1.05 tokens/s, 40 tokens, context 7, seed 294303832)
Output generated in 29.10 seconds (1.37 tokens/s, 40 tokens, context 47, seed 169447488)

It seems like the prompt ingestion phase may have become a bit faster, but it may also be statistical noise. The speed gain is probably greater for the 1080ti.

PR #3265, opened 2023-07-23
Chat token count

atriantafy:
* When using either the /api/v1/generate or the /api/v1/chat endpoints we need to specify the number of max_new_tokens to generate. If max_new_tokens is too small we underutilize the model's context size. If it is too large, we run the risk of running out of context. The optimal max_new_tokens value should be: max_seq_len - number_of_prompt_tokens. Now, for simple generation (/api/v1/generate) we can simply use the /api/v1/token-count endpoint to find the number of tokens in the prompt. However, for chat (/api/v1/chat), the number of tokens can only be found if we have the generated prompt from the given chat user_input and history. For example, if the user_input is `Please give me a step-by-step guide on how to plant a tree in my backyard.` and the instruction_template is "Instruct-Llama-v2", the final prompt sent to the LLM is going to be: 
```
[INST] <<SYS>>
Answer the questions.
<</SYS>>
Please give me a step-by-step guide on how to plant a tree in my backyard. [/INST]
```
which has a token count much higher (45 tokens) compared to the user_input (21 tokens).

* Why not simply change the behavior of /api/v1/chat and /api/v1/generate so that max_new_tokens is automatically set to the optimal value? I'm planning to do a relevant PR in the future as this would simplify things for the users calling the API. However, I think it is still useful for the user to be able to count the number of tokens in a chat prompt so that they can forge their prompts in a way that maximizes the context length. An example is summarization: if you know the chat prompt length, you can precisely calculate the maximum number of tokens that can go into the summary's input and the number of maximum tokens of the generated summary.

* Why not simply change the behavior or the existing /api/v1/count-tokens endpoint? This is another possible implementation. However it might confuse the users because in order to successfully calculate the number of tokens in a chat prompt you need to provide a whole different set of parameters to the API end point. I thought keeping things separate:  /api/v1/count-tokens for counting tokens of simple generations and  /api/v1/count-chat-tokens for counting tokens of chat generations is cleaner, but I'm open to suggestions if you disagree.


Ph0rk0z:
Heh, would be cool to have token counts in native chat vs api where it should be the client's responsibility. It's also there that scaling the output tokens can even work since there is control over all the variables.

atriantafy:
> Heh, would be cool to have token counts in native chat vs api where it should be the client's responsibility. It's also there that scaling the output tokens can even work since there is control over all the variables.

yeah, would be nice, but this could be a different project altogether: a client library for Oobabooga (not sure if one exists already). Currently this PR follows the existing API-based token-count pattern but for chats. Unfortunately, in order to estimate the number of tokens on the client side, you must also have on the client side: the correct tokenizer and also the current model's instruct template (and any other metadata required to render the prompt).

matatonic:
I like how openai chat api handles this, if you set max_tokens it will reserve that space for token generation, but if you don't set it at all, the default is essentially "Infinite" and will use any remaining context for generations.

atriantafy:
> I like how openai chat api handles this, if you set max_tokens it will reserve that space for token generation, but if you don't set it at all, the default is essentially "Infinite" and will use any remaining context for generations.

I also like this way, the difference though is that Oobabooga automatically truncates oversized prompts, while Openai will simply fail with "exceeded context limit" (or something like that).
In PR #3268 I am proposing an approach which is  closer to the OpenAI approach you described but still respects the truncation feature that many Oobabooga users already rely on.

Ph0rk0z:
I have not been setting truncate at all when using the API. It's still 2048 but my model outputs to whatever the front end asks for. Exllama max sequence length gets set when loading and that's it.

oobabooga:
I think that this is not necessary after https://github.com/oobabooga/text-generation-webui/pull/3419, which can be used through the API by sending `auto_max_new_tokens=True` in the request.

atriantafy:
Hi @oobabooga and thanks for working on auto_max_new_tokens.

auto_max_new_tokens is only one of the possible use-cases for this feature. Basically the existing token-count api endpoint does not work with chat prompts, so any token-counting use-case is not possible for chat/instruct prompting (which in my opinion are the most useful ones, compared to the traditional completion prompting, especially with all these instruct/chat finetuned models we're getting lately)

Some examples:
* You want to automatically structure your prompt in a way that takes advantage of the whole context before sending the prompt to the server (e.g. you might want to have a prompt like: <beginning-static> <middle-part which will depend on the avalable model's context> <end-static>). It has been pointed out by the Stanford paper on long context: https://arxiv.org/pdf/2307.03172.pdf that the beginning and the end of a prompt are the most "attention-grabbing". Having nuanced control is important.
* For auditing purposes it is important to know how many tokens went in and out the model
* Cost estimation

For these reasons I highly recommend we keep this feature

PR #3268, opened 2023-07-23
auto max_new_tokens; fix default truncation_length value

atriantafy:
* This PR addresses two API-related issues:
  * By default, if the user does not set a truncation_length, it sets it to 2048. This PR changes this behavior so that, if the user does not set a truncation length it defaults to shared.settings['truncation_length']) which is calculated according to the currently loaded model  instead of the hard-coded 2048 which might not be optimal.
  * Introduces a new parameter called "use_max_seq_len" that, if set to True, will allow you to use the full available context of the model (which is determined by the model's max_seq_len parameter) while generating new tokens. It basically overrides max_new_tokens and the default truncation_length in a way that will generate the maximum number of tokens without exceeding the model's context limits. It overrides truncation_length and max_new_tokens in the following way:
    * Sets truncation_length equal to max_seq_len (unless the user has specified a truncation_length value in the API request)
    * For non-chat text generations: count the number of tokens in the prompt and subtract it from truncation_length
    * For chat generations: render the chat prompt as it will be fed to the LLM (by using both user_input and history). Then proceed as above, by subtracting its length from truncation_length
    * If the calculated value is less than max_new_tokens (which defaults to 200), use max_new_tokens instead. Effectively, this means that when use_max_seq_len is activated, max_new_tokens acts as a safety net to make sure we won't get a value lower than that (i.e. too small, or even negative values).
    * This calculated value overrides the old max_new_tokens value

~~* max_new_tokens dictates how many new tokens should the LLM generate. Currently, if the user does not set this value, it defaults to 200 max_new_tokens. This PR introduces a new boolean parameter called 'optimize_max_new_tokens', which, if set to True, the API will attempt to optimize the number of max_new_tokens. The optimal number of max_new_tokens is the maximum number of tokens the model can generate without running out of context (and thus, without causing any truncation). This way, we get full advantage of the model's context and are not constrained to only 200 tokens (or whatever is the max_new_tokens value). The optimal value is calculated in the following way:
    * For non-chat text generations: count the number of tokens in the prompt and subtract it from truncation_length
    * For chat generations: render the chat prompt as it will be fed to the LLM (by using both user_input and history). Then proceed as above, by subtracting its length from truncation_length
    * If the optimal value is less than the provided max_new_tokens (which defaults to 200), use max_new_tokens instead. Effectively, this means that when optimize_max_new_tokens is activated, max_new_tokens acts as a safety net to make sure we won't get a value lower than that.~~

In my personal experience I've found that in many cases my prompt (especially in chat) would get truncated without me noticing it and thus causing the model to lose potentially important instructions from the beginning of the conversation. On the other hand, if I set a small enough max_new_tokens to be safe, I end up with unfinished answers. This fix allowed me to take advantage of the model's maximum available context when generating answers and hopefully it will be helpful to others as well. 

Edit: As @Ph0rk0z commented below, their use-case does not require the maximum number of tokens possible, which demonstrates that "optimal" is not the same for every use case. For this reason, I changed the name of the parameter from "optimize_max_new_tokens" to "use_max_seq_len" which is more descriptive and more objective. I also changed a bit the behavior so that use_max_seq_len overrides not just max_new_tokens, but also truncation_length (if not specifically set by the user in the request), because, if truncation_length is left at its default value, it will become the bottleneck that prevents us from using the full length of the model's context (which is defined by its max_seq_len parameter)

### DIFF
--- a/extensions/api/util.py
+++ b/extensions/api/util.py
@@ -7,9 +7,10 @@ from threading import Thread
 from typing import Callable, Optional
 
 from modules import shared
-from modules.chat import load_character_memoized
+from modules.chat import load_character_memoized, generate_chat_prompt
 from modules.presets import load_preset_memoized
-
+from modules.extensions import apply_extensions
+from modules.text_generation import encode
 
 # We use a thread local to store the asyncio lock, so that each thread
 # has its own lock.  This isn't strictly necessary, but it makes it
@@ -17,11 +18,31 @@ from modules.presets import load_preset_memoized
 # thus handling multiple requests in parallel.
 api_tls = threading.local()
 
+def calculate_max_new_tokens(prompt, state):
+    max_length = min(shared.settings['max_seq_len'], state['truncation_length'])
+    num_prompt_tokens = len(encode(prompt)[0])
+    max_new_tokens = max_length - num_prompt_tokens
+    return max_new_tokens
+
+def get_chat_prompt(body, state):
+    user_input = body['user_input']
+    if body.get('history', None) is None:
+        history = {'internal': [], 'visible': []}
+    else:
+        history = body['history']
+    kwargs = {
+        '_continue': body.get('_continue', False),
+        'history': history,
+    }
+    prompt = apply_extensions('custom_generate_chat_prompt', user_input, state, **kwargs)
+    if prompt is None:
+        prompt = generate_chat_prompt(user_input, state, **kwargs)
+    return prompt
 
 def build_parameters(body, chat=False):
 
     generate_params = {
-        'max_new_tokens': int(body.get('max_new_tokens', body.get('max_length', 200))),
+        'max_new_tokens': 0, # temporary value, will calculate final value later (see below)
         'do_sample': bool(body.get('do_sample', True)),
         'temperature': float(body.get('temperature', 0.5)),
         'top_p': float(body.get('top_p', 1)),
@@ -45,7 +66,7 @@ def build_parameters(body, chat=False):
         'mirostat_eta': float(body.get('mirostat_eta', 0.1)),
         'seed': int(body.get('seed', -1)),
         'add_bos_token': bool(body.get('add_bos_token', True)),
-        'truncation_length': int(body.get('truncation_length', body.get('max_context_length', 2048))),
+        'truncation_length': int(body.get('truncation_length', body.get('max_context_length',  shared.settings['truncation_length']))),
         'ban_eos_token': bool(body.get('ban_eos_token', False)),
         'skip_special_tokens': bool(body.get('skip_special_tokens', True)),
         'custom_stopping_strings': '',  # leave this blank
@@ -80,6 +101,16 @@ def build_parameters(body, chat=False):
             'chat-instruct_command': str(body.get('chat-instruct_command', shared.settings['chat-instruct_command'])),
             'history': body.get('history', {'internal': [], 'visible': []})
         })
+
+    max_new_tokens = body.get('max_new_tokens', body.get('max_length', None))
+    if max_new_tokens is not None:
+        generate_params['max_new_tokens'] = int(max_new_tokens)
+    else:
+        if chat:
+            prompt = get_chat_prompt(body, generate_params)
+        else:
+            prompt = body['prompt']
+        generate_params['max_new_tokens'] = calculate_max_new_tokens(prompt, generate_params)
 
     return generate_params
 
--- a/extensions/api/util.py
+++ b/extensions/api/util.py
@@ -18,7 +18,7 @@ from modules.text_generation import encode
 # thus handling multiple requests in parallel.
 api_tls = threading.local()
 
-def optimize_max_new_tokens(prompt, state):
+def estimate_max_new_tokens(prompt, state):
     old_max_new_tokens = state['max_new_tokens']
     max_length = min(shared.settings['max_seq_len'], state['truncation_length'])
     num_prompt_tokens = len(encode(prompt)[0])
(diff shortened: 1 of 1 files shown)


mykeehu:
I can't wait to get this in, because the max_seq_len value for ExLlama changes from model to model, and I had to adjust the truncation_length parameters when changing models. I hope your modification will really give me more optimal lengths!

Ph0rk0z:
>The optimal number of max_new_tokens is the maximum number of tokens the model can generate without running out of context

I wouldn't want 4000 max new tokens. I want 250 tokens and the rest for the context of my chat. How is it going to use history when I'm using silly tavern through the notebook api?

>By default, if the user does not set a truncation_length, it sets it to 2048. 

How does this account for free memory?  The model may support 16k and situation doesn't.

If it's able to be toggled through a setting it makes sense to have options but I think this was not done because there are lots of caveats for these values and no easy way to automate it.

atriantafy:
> > The optimal number of max_new_tokens is the maximum number of tokens the model can generate without running out of context
> 
> I wouldn't want 4000 max new tokens. I want 250 tokens and the rest for the context of my chat. How is it going to use history when I'm using silly tavern through the notebook api?
> 

This feature kicks in only if you set optimize_max_new_tokens to True. So, if this feature doesn't serve your specific use-case you can simply ignore this flag, as, by default it is disabled and the behavior is as before. Your use-case is an on-going dialogue. But for other use-cases that take advantage of instruct models via the chat interface (e.g. you tell the LLM to perform a specific task) you often want to take advantage of the full context for the completion of the task. 
One such example is summarization, which would look like this:
```USER: can you please summarize a web page for me?
ASSISTANT: Sure! Please provide the contents of the web page and I will do my best to summarize them.
USER: <page contents go here>```
Given these chat prompts, you want the assistant to use as much as possible of the remaining context to generate the summary.
Same goes for similar tasks where you want the assistant to generate a long chunk of text given a set of instructions.

Now, I have do have to admit that perhaps "optimize_max_new_tokens" is not the perfect naming choice as, "optimize" could mean different things for different people. Here, I assumed that taking advantage of the full context is an optimization, but you suggested that in your case it is not. Hence, it would be great if anyone has a better recommendation on how to name this feature. I was thinking something like "maximize_new_tokens", or something similar, but it might sound a bit confusing to the user, so I'm open to suggestions.

> > By default, if the user does not set a truncation_length, it sets it to 2048.
> 
> How does this account for free memory? The model may support 16k and situation doesn't.
> 
> If it's able to be toggled through a setting it makes sense to have options but I think this was not done because there are lots of caveats for these values and no easy way to automate it.

Truncating by default to 2048 is quite arbitrary. It doesn't take Free Memory into consideration anyway. Why not set it to 1024 by default to be even safer? On the other hand, I'm sure there are many cases where a user is unknowingly truncating their text generations simply because of the default 2048 truncation setting, even if both the model and their free VRAM would support a higher truncation setting (especially for smaller models).

Memory usage is a responsibility of the user. It would be more straightforward in my opinion for the API to automatically use the model's recommended Truncation limit, and if they run into insufficient memory errors, they could explicitly set a lower truncation limit that suits their limited hardware capabilities.

**Comment by oobabooga** [comment]
1) I'm not a fan of using `shared.settings` dynamically. The purpose of this dictionary is setting default values in the gradio UI. 
2) In most cases, things shouldn't be changed in the API only. This project revolves around the gradio UI.

For the purpose of setting max_new_tokens dynamically, a checkbox could be added to the Parameters tab like "use as many tokens as possible" or some better name. Once active, instead of max_new_tokens what is used is `max(max_new_tokens, get_max_prompt_length(state))` where `get_max_prompt_length` is defined in modules/text_generation.py.

**Comment by atriantafy** [comment]
> 2. In most cases, things shouldn't be changed in the API only. This project revolves around the gradio UI.
> 
> For the purpose of setting max_new_tokens dynamically, a checkbox could be added to the Parameters tab like "use as many tokens as possible" or some better name. Once active, instead of max_new_tokens what is used is `max(max_new_tokens, get_max_prompt_length(state))` where `get_max_prompt_length` is defined in modules/text_generation.py.

I looked into implementing it in the gradio UI. The issue is that max_new_tokens is used in many places and to change how it is being accessed requires a bit more research in order to make sure I don't break something. My initial solution was a quick fix just for the API version that simply overwrote the previous max_new_tokens value with the number of tokens left until the model's context is full (model context limit - prompt length). I see your point for providing a unified way feature set between the GUI and the API so I will turn this into a draft and come back when I have some more time to work on this.

In the meanwhile, if you could please take a look at #3265 - this would allow the client to at least be able to count the number of tokens in a chat prompt, and thus, be able to calculate the number of max_new_tokens to use on the client side (among other use cases).

> 1. I'm not a fan of using `shared.settings` dynamically. The purpose of this dictionary is setting default values in the gradio UI.

Are you referring to this?

```
'truncation_length': int(body.get('truncation_length', body.get('max_context_length',  shared.settings['truncation_length']))),
```

If yes, what would be a better way to retrieve the current model's default truncation length?  Something like:
`get_model_settings_from_yamls(shared.model_name)` ?
Let me know if you'd like me to implement this in a different PR. Since in the YAMLs we already have the information about the optimal truncation length for each model I think it would make sense to use it in the API by default as well (instead of the hard-coded 2048 one), especially now that we're seeing more variety of model context lengths these days (e.g. Llama-2 which has 4096).

Thanks!

oobabooga:
I have added an `auto_max_new_tokens` options here https://github.com/oobabooga/text-generation-webui/commit/e931844fe25620e8ccc2e7c4ed9ab06fc6644471

It works for _HF loaders and transformers/autogptq.

PR #3272, opened 2023-07-24
Supercharging superbooga

HideLord:
Complete rewrite of the Superbooga extension. Many, many added functionalities and vastly improved accuracy (395% on my test dataset) of retrieval.

The main attraction is the possibility to automatically optimize the database and search algorithm based on your own dataset and question format.

---

**What's left**:

- [x]  ~Add API endpoints for adding, retrieving and deleting.~
- [x]  ~Move NLTK file locally to allow working without internet connection.~
- [ ]  Performance comparison between LlamaIndex and my implementation.
- [x]  ~Finish the implementation of TimeWeight.~
- [ ]  Implement a more robust `HIJACK_LAST_IN_CONTEXT` (Might need help here).
- [x]  ~When recreating the DB again and again, there is a memory leak. Find out whether it's a problem in ChromaDB or my implementation.~ (_Turns out you must explicitly use `client.reset`. Otherwise it caches... something, idk._)
- [ ]  Test out perplexity with 50/50 context retrieval.
- [x]  ~After the removal of `numpy.int`, skopt is broken. There are currently PRs waiting to be merged into the main branch, but the project looks abandoned. I will probably fork the fixes and add the fork as a requirement.~ (_Will use Optuna instead. It is simpler, offers better performance and is up-to-date._)
- [ ] Multithreading is broken. Add locks in the appropriate places.
- [ ] Add drop-down for different embedding models.
- [ ] Add drop-down for named parameter presets.
- [ ] Pass forward parameters instead of a global singleton.
- [ ] Add option to disable caching.
- [ ] Add option to use CPU only.


HideLord:
@oobabooga Can you check if you're okay with the newly added API endpoints? I know there was a PR that got rejected recently because of something similar.

oobabooga:
@HideLord I don't mind adding API endpoints within the superbooga extension itself, my objection was to mixing different extensions of adding extension code to the main codebase.

Could you describe the changes that you are making here? It looks very sophisticated.

HideLord:
## Step 1: Collection
This is fairly straightforward. You collect various texts and feed them into the system. This can be done through the three interfaces:
* Pure Text
* URL/NET
* Files

## Step 2: Processing
We now have a big blob of text. This is not very useful for our purposes. Some kind of processing is needed to convert it into chunks which are to be fed into the embedding function and then the VectorDB, or ChromaDB in our case.

In the previous version, we took the naive approach of simply splitting the text into uniform pieces of `N` characters. This is not a good approach for several reasons:
1) **Splitting words**: It leaves the possibility of splitting words and creating gibberish.
2) **Lack of Cleaning**: Often, the corpus will contain strange formatting, purple prose, overabundance of punctuation or markdown, etc. All of these can hurt the accuracy of the embeddings.
3) **Dilution**: Let's take an example from "aircraft_lease.txt". `"Summarize section 4(c)?"`. The perfect retrieval might look something like this:
    ```
              Section 4(c) PROHIBITION AGAINST SETOFF, COUNTERCLAIM, ETC. This Lease
     is a net lease. Subject to Section 20(f), Lessee's obligation to pay all Rent
     hereunder shall be absolute and unconditional and shall not be affected or
     reduced by any circumstance, including, without limitation, (i) any setoff,
     counterclaim, recoupment, defense or other right which Lessee may have against
     Lessor, any Beneficiary, the Manufacturer, the Engine Manufacturer, any seller
     of or person providing services with respect to the Aircraft or any other
     Person, for any reason whatsoever; (ii) any defect in the title, airworthiness
     or eligibility for registration under applicable Law, or any condition, design,
     operation or fitness for use of, or any damage to or loss or destruction of,
     the Aircraft, or any interruption or cessation in the use or possession thereof
     by Lessee for any reason whatsoever, whether arising out of or related to an
     act or omission of Lessee, or any other Person; (iii) any Liens with respect to
     the Aircraft; (iv) the invalidity or unenforceability or lack of due
     authorization or other infirmity of this Lease or any absence of right, power
     or authority of Lessor or Lessee to enter into this Lease; (v) any insolvency,
     bankruptcy, reorganization or similar proceedings by or against Lessor or
     Lessee; (vi) any other circumstance or happening of any nature whatsoever,
     similar to any of the foregoing; or (vii) any Taxes (other Taxes to which
     Lessee's indemnity does not extend pursuant to the provisions of Section 10);
     it being the express intention of Lessor and Lessee that all Rent payable
     hereunder shall be payable in all events, unless the obligation to pay the same
     shall be terminated pursuant to the express provisions of this Lease. Nothing
     in this paragraph (c) shall constitute a waiver by Lessee of any right or claim
     that Lessee may separately assert against Lessor or any Beneficiary.
    
                       Lessee hereby waives, to the extent permitted by applicable
     Law, any and all rights which it may now have or which at any time hereafter
     may be conferred upon it, by Law or otherwise, to terminate this Lease or any
     obligation imposed upon Lessee hereunder or in relation hereto.
    
                       If for any reason whatsoever this Lease shall be terminated
     in whole or in part by operation of law (other than termination under any
     bankruptcy laws as now or hereafter in effect), Lessee nonetheless agrees to
     pay to Lessor amounts equal to the Rent payments hereunder at the time such
     payments would have become due and payable in accordance with the terms hereof
     had this Lease not been terminated so long as Lessee is able to use, possess
     and quietly enjoy the Aircraft, and so long as such payments are made and all
     other terms and conditions hereof are complied
    
                                          -16-
    
    
    <PAGE>
    
    
    with by Lessor and Lessee, Lessor and Lessee will deem this Lease to remain in
    full force and effect.
    ```
    In total, that is ~3000 characters. This means that I have to put `N` as 3000 if I want to retrieve the whole context, but that is a problem. The embedding is becoming diluted with irrelevant information to the question. We might set it as 700 and hope that it picks up all the pieces, but that is improbable.
    
    In reality, there are two parts to any retrieved text, in my opinion. The _embedding-critical_ part and the _context-critical_ part.
    - **Embedding-critical part**: That is the part that is closest to the question in the embedding space. In this case, that would be `Section 4(c) PROHIBITION AGAINST SETOFF, COUNTERCLAIM, ETC.`
    - **Context-critical part**: This is the rest of the text that the LLM needs in order to understand the retrieved information and answer the question. In our example, it constitutes the rest of the text.
    
    This is, of course, an extreme example. In most cases, the embedding-critical part would be longer, and the context one -- shorter.

4) **Differing lengths**: While some questions, like the previous one, would require `N` to be long, others would prefer it if it was short.  
Let's look at another example, `"What is the revision number of the aircraft?"` The perfect retrieval might look something like this:
`"Detail Specification (737-25A-Rev. B)"`

5) **No Universal Solution**: Even if we fixed everything, no single algorithm would be perfect for every dataset out there.

### So, how do we address those problems?

1) **Splitting words**: This one is pretty simple. We only divide at whole words.
2) **Lack of Cleaning**: This is a complex problem and requires a per-dataset solution, but here are some general methods:  
    - _Lowering Cases_: Self-explanatory
    - _Merging Spaces_: Again, pretty straightforward. If there are more than two spaces next to each other, merge them.
    - _Stripping Edges_: Simply remove whitespaces at the start and end of the string.
    - _Removing Punctuation_: This would help with markdown heavy texts as well as ASCII formatting and HTML tags.
    - _Number Conversion_: This means converting numbers to something else. Embeddings often don't work well with numbers, so this might help accuracy:
        + Characters: E.g., `0 = A`, `1 = B`, `2 = C`...
        + Multi-character: E.g., `210 = CCCBBA`. This puts more emphasis on the order and combination of numbers rather than the individual digits. This method might be particularly useful when dealing with larger numerical data or where the sequence of numbers plays a significant role. 
        + Words: E.g. `321 = Three Hundred Twenty-one`. Similar to the above.
    - _Stopwords_: These are commonly occurring words in a language that do not contain much information and are often filtered out before text processing. For example,  "the", "is", "and", etc.
    - _Lemmatizing_: This refers to reducing words to their base or dictionary form. For instance, "running", "runs", and "ran" all get reduced to "run". This helps in grouping together different forms of a word, which can help the generalizing of the embedding.
    - _Removing Specific POS (Part Of Speech)_: We remove certain types of words based on their grammatical classification. For example, one might choose to remove all adverbs or all interjections from the text, as they rarely hold critical information. But, then again, if you are searching by sentiment, you might not want to do it.
    - _Summarization_: In text processing, the Pareto Principle is a very important concept: 20% of sentences hold 80% of the meaning. We can use this to our advantage by eliminating low-meaning sentences without employing expensive LLM-based summarization.
3) **Dilution**: To tackle this, we introduce a new parameter, `context_len`. This is a comma-separated string with two values: left context length and right context length.  
    Let's look at the example again:  
    ```diff
    + Captured by context
    - Captured by embedding
    ! Overlapping
    
    + SOMETEXT...
    + SOMETEXT...
    -          Section 4(c) PROHIBITION AGAINST SETOFF, COUNTERCLAIM, ETC. This Lease
    + is a net lease. Subject to Section 20(f), Lessee's obligation to pay all Rent
    + hereunder shall be absolute and unconditional and shall not be affected or
    + reduced by any circumstance, including, without limitation, (i) any setoff,
    + counterclaim, recoupment, defense or other right which Lessee may have against
    + Lessor, any Beneficiary, the Manufacturer, the Engine Manufacturer, any seller
    + of or person providing services with respect to the Aircraft or any other
    + Person, for any reason whatsoever; (ii) any defect in the title, airworthiness
    + or eligibility for registration under applicable Law, or any condition, design,
    + operation or fitness for use of, or any damage to or loss or destruction of,
    + the Aircraft, or any interruption or cessation in the use or possession thereof
    + by Lessee for any reason whatsoever, whether arising out of or related to an
    + act or omission of Lessee, or any other Person; (iii) any Liens with respect to
    + the Aircraft; (iv) the invalidity or unenforceability or lack of due
    + authorization or other infirmity of this Lease or any absence of right, power
    + or authority of Lessor or Lessee to enter into this Lease; (v) any insolvency,
    + bankruptcy, reorganization or similar proceedings by or against Lessor or
    ! Lessee; (vi) any other circumstance or happening of any nature whatsoever,
    ! similar to any of the foregoing; or (vii) any Taxes (other Taxes to which
    ! Lessee's indemnity does not extend pursuant to the provisions of Section 10);
    ! it being the express intention of Lessor and Lessee that all Rent payable
    + hereunder shall be payable in all events, unless the obligation to pay the same
    + shall be terminated pursuant to the express provisions of this Lease. Nothing
    - in this paragraph (c) shall constitute a waiver by Lessee of any right or claim
    + that Lessee may separately assert against Lessor or any Beneficiary.
    +
    +                   Lessee hereby waives, to the extent permitted by applicable
    + Law, any and all rights which it may now have or which at any time hereafter
    + may be conferred upon it, by Law or otherwise, to terminate this Lease or any
    + obligation imposed upon Lessee hereunder or in relation hereto.
    +
    +                   If for any reason whatsoever this Lease shall be terminated
    + in whole or in part by operation of law (other than termination under any
    + bankruptcy laws as now or hereafter in effect), Lessee nonetheless agrees to
    + pay to Lessor amounts equal to the Rent payments hereunder at the time such
    + payments would have become due and payable in accordance with the terms hereof
    + had this Lease not been terminated so long as Lessee is able to use, possess
    + and quietly enjoy the Aircraft, and so long as such payments are made and all
    + other terms and conditions hereof are complied
    +
    +                                      -16-
    +
    +
    +<PAGE>
    +
    +
    +with by Lessor and Lessee, Lessor and Lessee will deem this Lease to remain in
    +full force and effect.
    ```

    As you can probably tell, the green means that the text is captured by the context around the embedding. Red means that this is the text which will be embedded. Orange means that there is an overlap. We are going to discuss later what we do in these cases, but as you can see, the whole text is covered, and the embeddings are concentrated on sections that are related to the question: `...Section 4(c) PROHIBITION AGAINST SETOFF...` and `...paragraph (c)...`. The former is probably going to be ranked much higher than the latter, but again, we are going to discuss later how we solve this problem. For now, we are happy that the whole text is covered, and we have concentrated embeddings.

4) **Differing lengths**: To address this, we also incorporate multiple chunk lengths `chunk_len = 'len1, len2, len3...'`. For each length `len_i`, we divide the corpus into chunks of approximately `len_i` characters. Each chunk is then embedded, and the surrounding context is added.  
This enables us to theoretically capture the perfect embedding length for any problem, but in reality, it's not quite so simple. The embeddings are often dependent on the surrounding context. E.g., a classic example: `"I want her"` and `"I want her dead"`. As such, with each added length, especially short ones, we are introducing a lot of noise.
5) **No Universal Solution**: We'll talk about this later.

## STEP 3: Generation

Now that we've processed our text and added our embeddings, along with their context, to the vectorDB, we can start querying. It's important to follow the same preprocessing for the queries as we did for the embeddings -- lowering cases, merging spaces, etc.

There are quite a few new features here, but let's begin with the basics.

- **Chunk Count**: How many closest neighbors to search for. Typically much higher than before.
- **Token Count Limit**: We can now specify `max_token_count` that, as the name implies, limits how many tokens we can retrieve from the vectorDB.
- **Formatting**: We can now specify three values:
    + **Prefix**: What to put before the beginning of the injection.
    + **Data separator**: What to put between each retrieved chunk.
    + **PostFix**: What to put after the injection.

### What is happening behind the scenes:

1) **Query Processing**: We preprocess the query using the same techniques as during the main processing. That means that if the question is long, it will be summarized/divided into several sentences in some cases.
2) **KNN**: We query the DB for `chunk_count` number of closest neighbours.
3) **Filtering**: We do **Time Weighing** and **Median Filtering**
    - **Time Weighing**: There are two components: `time_power` and `time_steepness`.
        + **Time Power**: As the name suggests, it's how strongly time is going to affect the distance from the query. Higher power means that older entries might not be discarded.
        + **Time Steepness**: How differently are two close excerpts going to be weighed. In reality, this parameter controls how steep the sigmoid curve will be.
    - **Median Filtering**: We have a new parameter, `significant_level`. It dictates that no chunk with distance higher than the `median * significant_level` shall pass forward. Why? Well, imagine we query for 250 chunks. In some cases, they might all be related, but in others, they might be distance chunks that are not related to the query.  
    Another method would be to cluster the chunks and eliminate those that are not in the query cluster but that entailed other problems that I won't get into.
4) **Merging**: The most important part. Remember how we had two problems?
    - There was an overlapping context from two embeddings.
    - One of the embeddings was going to be ranked much lower than the other.
    
    To address these problems, we are going to merge the chunks. We have the initial document, so we know the relative positions of the embedding text and the context around it. If two embeddings overlap or one completely covers the other, we are going to merge them. But what will be the distance of the new embedding? There are a couple of options:
    1) Harmonic, Geometric, or Arithmetic Mean.
    2) The minimum of the two distances.
    3) Discounting the minimum of the two distances.
    
    The last one might make sense. We now have a text with two embeddings that are close to the query.
    But overall, we solved both of our problems. Let's see:
    * There is no more overlapping.
    * The embedding that was ranked much lower is now part of the higher-ranked embedding, potentially pushing it into the context.
    
## STEP 4: Optimization

There is still one problem that we haven't addressed: No algorithm is perfect for every dataset. One way to go about it is to give the user the tools to tinker and find a solution that fits his use case. Each of the parameters and options that we discussed above is accessible through the GUI.

But most users either don't have the knowledge or the time to do that. 

What we really want is hyperparameter optimization. To achieve this, we first expose all of the parameters to the user through a JSON "config.json":
```
...
    "remove_specific_pos": {
      "default": false,
      "categories": [true, false],
      "should_optimize": true
    },
    "lemmatize": {
      "default": true,
      "categories": [true, false],
      "should_optimize": true
    },
    "min_num_sent": {
      "default": 1,
      "categories": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 999999],
      "should_optimize": true
    },
    "delta_start": {
      "default": 0,
      "categories": [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100],
      "should_optimize": true
    },
...
```

Each parameter has a default value, categories, and a flag whether it should be considered by the optimizer.

For the algorithm, we employ the Optuna library and use TPE.

That only leaves the fitness function. But that is simple enough. Evaluating an LLM is hard because there isn't one good output. But for VectorDBs, there (generally) is. It's the excerpt from the document that contains the answer to the user query.

I've included a sample "questions.json" and "aircraft_lease.txt" inside "benchmark_texts". I chose this specific text because the formatting of the document is very bizarre, it contains a lot of ASCII tables and drawn information, and the language is as convoluted as it gets.

## ADDITIONAL

### API

There is now an API for the vectorDB. You can:
- **Clear**: Reset the whole DB and clear all records.
    ```
    curl -X DELETE http://127.0.0.1:5002/api/v1/clear \
        -H 'Content-Type: application/json'
    ```
- **Delete**: Delete records with specific metadata.
    ```
    curl -X POST http://127.0.0.1:5002/api/v1/delete \
        -H 'Content-Type: application/json' \
        -d '{
        "metadata": {
            "source": "API-ADD"
        }
    }'
    ```
- **Add**: Add new records.
    ```curl
    curl -X POST http://127.0.0.1:5002/api/v1/add \
        -H 'Content-Type: application/json' \
        -d '{
        "corpus": "TEXT GOES HERE",
        "clear_before_adding": false,
        "metadata": {
            "source": "API-ADD"
        }
    }'
    ```
- **Get**: Get results for a text query.
    ```
    curl -X POST 'http://127.0.0.1:5002/api/v1/get?sort=distance' \
        -H 'Content-Type: application/json' \
        -d '{
        "search_strings": ["QUERY GOES HERE"]
    }
    '
    ```

This leaves quite a lot to be desired, but it's a good start if somebody needs to modify it.

Dampfinchen:
Would this result in lower or no VRAM usage? I noticed I have to set fewer GPU layers when I want to use superbooga. ChromaDB surely can only run on RAM and imo it should because its still fast on the CPU and doesn't need the GPU imo.

oobabooga:
@HideLord some comments:

1)

> Differing lengths: To address this, we also incorporate multiple chunk lengths chunk_len = 'len1, len2, len3...'. For each length len_i, we divide the corpus into chunks of approximately len_i characters. Each chunk is then embedded, and the surrounding context is added.

The default value is `500,500,1000,700`, which includes the length 500 twice. Wasn't the idea to include different lengths?

2) The default value for "Chunk regex" is `(?<==== ).*?(?= ===)|User story: \d+`, which seems very specific. The defaults probably need to be reworked for an average use case like importing a book in .txt format or adding a few documentation URLs.

3) I found that the dependency `pytextrank` was lacking from the requirements.txt.

-----

Overall, it's refreshing to see that you have used a benchmark to guide your changes. I had honestly given up on vector databases in general because I felt like all they could do was feed the model with some broken text, which it then used to generate unreliable made-up response.

kgargano:
Is this still in development? Pretty interested in it. Need any help?

calvinloveland:
I've also been interested in this. I've forked @HideLord 's repo and moved a lot of the tasks to issues on my repo. Anyone is welcome to help out there. I'd like to try to separate out much smaller MVPs that could be PR-ed back here more easily with less code to read through. In other words make a PR for each of the improvements mentioned here one at a time. And of course if @HideLord comes back he's welcome to add his input.

oobabooga:
That's a good idea @calvinloveland.

HideLord:
Oh, lots of activity here lately. Sorry for kind of abandoning the project. It was mostly a PoC for my company and whether we can implement our own solution, but in the end, I had to move to other projects. I have some final revisions, though, so I will probably push them later. Otherwise, anybody is free to take over, since I will probably not continue anytime soon. @calvinloveland
Just a fair warning, the included dataset is not very good. It's far too small. Generating a bigger one with LLMs turned out to be quite the difficult task, though.

oobabooga:
Let's call it superboogav2 for now. Then people can tinker with the new and old versions.

RandomInternetPreson:
Oh my frick, I need to try this out.  Superbooga is something I use all the time!!

nbollman:
I want to try superbooga... Ive never been able to get it to run... Must be a 'me' problem I dont see very many other people with the issue :(  Does anything in my error log scream fix me?

2023-10-12 21:09:24 INFO:Loading the extension "gallery"...
2023-10-12 21:09:24 INFO:Loading the extension "long_replies"...
2023-10-12 21:09:24 INFO:Loading the extension "whisper_stt"...
2023-10-12 21:09:24 INFO:Loading the extension "silero_tts"...
2023-10-12 21:09:24 INFO:Loading the extension "openai"...
2023-10-12 21:09:24 INFO:Loading the extension "superboogav2"...
2023-10-12 21:09:24 ERROR:Failed to load the extension "superboogav2".
  File "/home/jsnow/text-generation-webui/modules/extensions.py", line 36, in load_extensions
    exec(f"import extensions.{name}.script")
  File "/home/jsnow/text-generation-webui/extensions/superboogav2/script.py", line 20, in <module>
    from .chromadb import make_collector
  File "/home/jsnow/text-generation-webui/extensions/superboogav2/chromadb.py", line 2, in <module>
    import chromadb
  File "/home/jsnow/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/__init__.py", line 1, in <module>
    import chromadb.config
  File "/home/jsnow/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/config.py", line 1, in <module>
    from pydantic import BaseSettings
  File "/home/jsnow/miniconda3/envs/textgen/lib/python3.10/site-packages/pydantic/__init__.py", line 218, in __getattr__
    return _getattr_migration(attr_name)
  File "/home/jsnow/miniconda3/envs/textgen/lib/python3.10/site-packages/pydantic/_migration.py", line 294, in wrapper
    raise PydanticImportError(
pydantic.errors.PydanticImportError: `BaseSettings` has been moved to the `pydantic-settings` package. See https://docs.pydantic.dev/2.4/migration/#basesettings-has-moved-to-pydantic-settings for more details.

For further information visit https://errors.pydantic.dev/2.4/u/import-error
/home/jsnow/miniconda3/envs/textgen/lib/python3.10/site-packages/gradio/components/dropdown.py:231: UserWarning: The value passed into gr.Dropdown() is not in the list of choices. Please update the list of choices to include: english or set allow_custom_value=True.
  warnings.warn(

nbollman:
Finally got superboogav2 running... its amazing, Ive always installed oobabooga manually within a conda environment, I dont believe I missed any steps but superbooga fails to run, the ChromaDB doesnt get configured, guessing?  I finally got it running with your one-click installer, and then cloning the /env to textgen for preference. Some info (readme) on how to get superbooga running with regular install would be appreciated by us low level ai coders.

Id really like the file input to be able to ingest ipynb files... can it? <-dunno why I asked Im going to try it in a second, but a list of file types it is capable of ingesting would be nice. Hidelord, Oobabooga, you guys are amazing keep up the great work. And read my stupid idea post about WhisperTTS+Superbooga+SileroSTT="hey booga" I promise its a good stupid idea.

RandomInternetPreson:
Firstly thank you for all of your hard work with suberboogaV2. 

I was trying to use it in the instruct tab (not chat-instruct but just instruct) and could upload my data, but the AI was not referencing the database.

I found a random reddit post that said to add "!c" to my questions to the AI and it began to work! I really feel like this small omission is hindering a lot of new adaptors.  I kept using V1 until I figured this out, I'm using them both right now and getting  a feel for how they work.  So far I'm liking V2 a lot, and I hope others end up using it soon!

Technologicat:
Yeah, needing to use "!c" (at the start or at the end of a message) is a bit unintuitive. Playing around with superboogav2, I get the impression it's there for performance, as the lookup can be rather slow, compared to just chatting with the LLM.

I have one question that I can't find a definitive answer for.

I'd like to use superboogav2 together with a custom character, on an instruct-tuned model - basically, to tune the response style and/or personality of a research-assistant AI. So:

**In *chat-instruct* mode, does superboogav2 look up the uploaded data, or just the chat history?**

The documentation in the UI is ambiguous about this, as it mentions "regular chat" and "instruct" modes, but does not mention "chat-instruct".

Furthermore, the documentation in the UI says that in "regular chat", any uploaded data is ignored. In the source code, I couldn't find any evidence to support this claim, no matter which chat mode is selected. My current hypothesis is that this part of the documentation is probably an outdated leftover from v1.

As for the source code, let's look at `extensions/superboogav2/`.

In `chat_handler.py`, according to the function `_is_instruct_mode`, instruct mode is defined as either of `"instruct"` or `"chat-instruct"`, so either one will do. But this only affects the lookup of the user and bot names. The only other mention of "instruct" in the source code is in `script.py`, where the word appears in the label for the injection strategy setting.

Next, consider the function `custom_generate_chat_prompt_internal`. The `collector` interfaces with the actual DB. If the "Add Chat to Data" setting is enabled, then the collector will auto-delete and reconstruct any *automatically added* records - which, supposedly, leaves any other records untouched. (Why else would it have a scoping mechanism?) The `CHAT_METADATA` in the add and delete calls suggests that the "automatically added" in the comment refers to the chat history records.

The ideal solution for me would be if superboogav2 looked up in both manually added records as well as in the chat history.

By my reading of the source code, I think it should do exactly that, but I'm not 100% sure. LLMs being stochastic and prone to hallucination, it's difficult to create a definitive empirical experiment, either.

Before I spend more time interpreting every nut and bolt, does anyone here have a more definite understanding of this?

HideLord:
@Technologicat You're right on the money... I think. It's been a while.
The thing is, the chat functionality was never truly finished. As such, I'm not confident the results are inserted in the right places.

The metadata part was to prevent data from explicitly inserted documents being deleted when the chat history was reconstructed. As for the "!c", it was partly a performance optimization, and partly a user experience optimization. When I made this extension, I didn't exactly have role playing in mind. I wanted to be able to toss in some documents and then ask questions about them. The problem was that not every question the user asks should be queried for. Some questions might be clarifying question relating the the previous query. That was the idea behind "!c". Ask initial question with a !c tag, and then ask clarifying question without it.

There is also this option I've put:

```
manual = gr.Checkbox(value=parameters.get_is_manual(), label="Is Manual", info="Manually specify when to use ChromaDB. Insert `!c` at the start or end of the message to trigger a query.", visible=shared.is_chat())
```
If unchecked, then it should theoretically always query. Apparently, it's only visible if chat mode is on (but that might be deprecated with newer versions of oobabooga if it's not visible).

Technologicat:
@HideLord Thanks for the quick reply. And thanks for making the extension, it's nice!

It could be useful to document the "!c" more visibly, then. What does the "c" stand for, by the way?

And is there a way to see the complete prompt the extension is sending to the LLM? I've found "Send to default" useful for debugging character cards, system prompts, and similar, but at least in my tests today, it didn't catch the extra context injected by the extension. The GPU whirred for longer than usual, and the AI output sounded sensible, so I think it worked, but I'd like to be sure. I could always put a `print` inside `custom_generate_chat_prompt_internal`, but if there's an actually supported way, that would be preferable.

The option is visible at least for me. I don't mind having to specify the "!c" now that I know about it - it's a nice performance optimization.

Tossing some documents in, that's pretty much the use case I had in mind. I'd like to feed it abstracts from recent scientific papers, and ask for a combined summary of the whole set. (Feeding full papers may be a problem due to equations and formatting.) Specifically AI is moving so fast that soon (if not already) humans will need to be AI-accelerated just to keep up with the field. I'm currently working in the intersection of computational science and AI, which is still a lower-volume field, but I'd like to stay up to date also on ideas in general AI research, particularly on LLMs and generative diffusion models.

I tested this by making the AI summarize just one full paper today (the [RWKV one](https://arxiv.org/abs/2305.13048) since it's pretty recent, so likely not in training data). I manually fed the PDF through `pdftotext` and uploaded the resulting text file into the ChromaDB. The results seemed promising on [OpenChat-3.5](https://huggingface.co/openchat/openchat_3.5). I'm deliberately using a small (7B) model, as I'm running this on a laptop.

I think what we have now is great for a prototype, and you have already done the difficult part. In the long run, it would be nice to have some more automation. Ideally, I'd like to toss in a set of PDFs at once, automatically extract the abstracts, and upload the results to the ChromaDB. The `pdftotext` command-line tool can extract the text content, but I'm still missing a filter to extract just the abstract. Also, it seems to me the current version of superboogav2 supports loading only one file at a time, so currently, the abstracts would need to be concatenated into a single plaintext file for loading.

Regarding roleplaying, some points, in no particular order:

- The chat is just a nice user interface, from an evolutionary history perspective.
- Why settle for a ChatGPT clone or Jarvis, if you can have a custom personality? :)
  - That said, I haven't yet customized the "Assistant" personality specifically, to better see the default behavior of new models, but I might do that in the long run once the capabilities saturate.
  - I have done some tests with other character definitions, such as the programming community's favorite deity Eris, but I'd never trust the goddess of discord to answer factual questions correctly.
- The character card can also be used for tuning the length or style of AI responses, or to provide other similar meta-instructions regarding the answer format.
- Also, you can make the AI character consider itself as an expert in a particular field, which sometimes improves response quality (which makes sense, at least based on the findings of [Kovač et al., 2023](https://arxiv.org/abs/2307.07870)).
  - Character cards are easily switchable in *text-generation-webui*. This saves you from altering the global *chat-instruct-template*. This, in turn, allows easily using the same *text-generation-webui* installation for a variety of different use cases.
  - I've set my *chat-instruct-template* to basically "Roleplay the character described below...", etc. and haven't needed to touch it since.
- I'm not very knowledgeable about this, but there's a comment on the SillyTavern issue tracker [stating that](https://github.com/SillyTavern/SillyTavern/issues/1212) vector DBs are not that useful as long-term memory for RP use cases, because the query (current chat message) will likely not map to any useful search keys (via cosine similarity or whichever way the vector embedding is computed).

I'd actually be interested in long-term memory for long-term coherence in AI-assisted short story writing. But in any case, a vector DB should work pretty well for retrieval use cases.

DenisSergeevitch:
Installation Tutorial for people who come [across this PR](https://www.reddit.com/r/LocalLLaMA/comments/18dohlt/tutorial_use_real_books_wiki_pages_and_even/).

Desm0nt:
Is there any way to make it work when I use oobabooga via API as backend? (not for roleplay purpose, just for chat with documents in any more user-friendly ChatGPT-Like frontend)

It seems to work inside oobabooga webui, but doesn't work in any connectet via OpenAI Api 3rd-party Clients.

bonswouar:
> A tangentially retrieval-related wish is web search integration, for queries with a slightly broader scope. For _text-generation-webui_, the [first relevant plugin](https://github.com/sammyf/Autobooga) died, but it seems that now we have [LucidWebSearch](https://github.com/RandomInternetPreson/LucidWebSearch) to replace that.

I'm also interested in web search, but unfortunately I personally don't really like LucidWebSearch mechanism. Having a chrome browser doing actions in foreground doesn't interest me much..

I already cloned Autobooga and added a couple of fixes locally, and might be interested in enhancing it a bit!
Although right now it's still limited on actual context length.

I've looked a bit into vector databases, it seems interesting, but I'm not sure yet what's the most effective way for implementing long context retrieval.
I've noticed with superbooga & superboogav2 that the results returned by the vector db seem a bit random sometimes (although "optimising" superboogav2 seems to help).

Another enhancement I was thinking about with web search would be to filter the results, to keep only relevant ones, or order them, using the LLM itself. Would be cool to also use it to activate or not the web search (something like "Give a score from 1 to 10 to this prompt to estimate if it would be better answered using complementary information from actual web search results, 1 being not useful at all, and 10 being the answer can't be answer with only trained data"

Technologicat:
@bonswouar I haven't actually yet personally tested LucidWebSearch, but that indeed doesn't sound very optimal.

My impression is that vector storage is for the use case where you use specific words (technically, concepts in the embedding space) to search for matches in a potentially huge database, so that only the best matches get injected into the context. Obviously, this doesn't solve the problem of general long-term memory, but only the subproblem of retrieving arbitrary facts essentially by name.

The problem faced by roleplay scenarios and by storywriting is different: one would like the history so far to "stay alive in the background" without explicitly mentioning particular facts (which is required to produce a match in a vector storage) in each message.

Also, different prompt injection modules may interfere. For example, I fed in [this paper](http://www.its.caltech.edu/~jbae/publications/Capuano_2014.pdf) into the vector store in my test setup. I proceeded to ask the LLM (dolphin-2.2.1-mistral-7B) about the numerical method the authors of the paper had developed.

The initial output mentioned that the method belonged to the Runge-Kutta family. Intrigued, I asked a follow-up question for details. Here comes the problem: the web search plugin saw that I was asking a question, it picked up the words "Runge-Kutta", and then proceeded to pollute the context with random facts about unrelated Runge-Kutta methods. The LLM confidently composed its answer using this irrelevant information, attributing it to the paper being discussed.

Disabling the web search plugin, and then running the same follow-up question again, worked fine - only the extracts from the paper (from the vector storage) ended up in the context, and the LLM answered correctly.

A related issue when a vector store is applied specifically in the context of scientific papers is that, without further filtering of the input, the list of references includes the highest concentration of keywords of any kind. Thus many of the highest scoring vector store matches are actually useless. With many papers I tested, only parts of the reference list would get retrieved from the vector store, omitting all actual content of the paper - thus making the LLM wildly hallucinate, since the retrieval failed to bring up any actual facts.

So my impression of vector storage, in the scientific use case, is that it's great when it works, but it only does so approximately 50% of the time. :)

Another issue in the particular web search plugin I tested is that it injects the text results only, while omitting the URLs (although it has access to them, right up to the point where it composes its reply). This is problematic when you'd like to ask the AI for e.g. the URL for some piece of open source documentation, because that gets scrubbed (and then replaced with a hallucination).

But this is getting offtopic. Anyone know a better place to move the discussion to?

Desm0nt:
Hello.
I update (reinstall) my whole oobabooga webui with suprebooga v2 and now got a werid problem. It's work like a charm inside webui, but when i try to access oobabooga via OpenAI api, I got this error (even when superbooga v2 is set to manual and I do not trigger it)
`
ERROR:    Exception in ASGI application
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\applications.py", line 1106, in __call__
    await super().__call__(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 184, in __call__
    raise exc
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 83, in __call__
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 79, in __call__
    raise exc
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 20, in __call__
    raise e
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 69, in app
    await response(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 233, in __call__
    async with anyio.create_task_group() as task_group:
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\anyio\_backends\_asyncio.py", line 597, in __aexit__
    raise exceptions[0]
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 236, in wrap
    await func()
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 221, in stream_response
    async for data in self.body_iterator:
  File "D:\LLM\text-generation-webui\extensions\openai\script.py", line 126, in generator
    for resp in response:
  File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 497, in stream_chat_completions
    for resp in chat_completions_common(body, is_legacy, stream=True):
  File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 283, in chat_completions_common
    for a in generator:
  File "D:\LLM\text-generation-webui\modules\chat.py", line 322, in generate_chat_reply
    for history in chatbot_wrapper(text, state, regenerate=regenerate, _continue=_continue, loading_message=loading_message, for_ui=for_ui):
  File "D:\LLM\text-generation-webui\modules\chat.py", line 253, in chatbot_wrapper
    prompt = apply_extensions('custom_generate_chat_prompt', text, state, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\modules\extensions.py", line 224, in apply_extensions
    return EXTENSION_MAP[typ](*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\modules\extensions.py", line 100, in _apply_custom_generate_chat_prompt
    return extension.custom_generate_chat_prompt(text, state, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\script.py", line 167, in custom_generate_chat_prompt
    return custom_generate_chat_prompt_internal(user_input, state, collector, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 118, in custom_generate_chat_prompt_internal
    history_as_text = _concatinate_history(kwargs['history'], state)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 71, in _concatinate_history
    user_name, bot_name = _get_names(state)
                          ^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 55, in _get_names
    user_name = state['name1_instruct']
                ~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'name1_instruct'
2023-12-14 11:34:34 DEBUG:Applying settings.
ERROR:    Exception in ASGI application
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\applications.py", line 1106, in __call__
    await super().__call__(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 184, in __call__
    raise exc
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 91, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 146, in simple_response
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 79, in __call__
    raise exc
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 20, in __call__
    raise e
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 69, in app
    await response(scope, receive, send)
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 233, in __call__
    async with anyio.create_task_group() as task_group:
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\anyio\_backends\_asyncio.py", line 597, in __aexit__
    raise exceptions[0]
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 236, in wrap
    await func()
  File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 221, in stream_response
    async for data in self.body_iterator:
  File "D:\LLM\text-generation-webui\extensions\openai\script.py", line 126, in generator
    for resp in response:
  File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 497, in stream_chat_completions
    for resp in chat_completions_common(body, is_legacy, stream=True):
  File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 283, in chat_completions_common
    for a in generator:
  File "D:\LLM\text-generation-webui\modules\chat.py", line 322, in generate_chat_reply
    for history in chatbot_wrapper(text, state, regenerate=regenerate, _continue=_continue, loading_message=loading_message, for_ui=for_ui):
  File "D:\LLM\text-generation-webui\modules\chat.py", line 253, in chatbot_wrapper
    prompt = apply_extensions('custom_generate_chat_prompt', text, state, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\modules\extensions.py", line 224, in apply_extensions
    return EXTENSION_MAP[typ](*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\modules\extensions.py", line 100, in _apply_custom_generate_chat_prompt
    return extension.custom_generate_chat_prompt(text, state, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\script.py", line 167, in custom_generate_chat_prompt
    return custom_generate_chat_prompt_internal(user_input, state, collector, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 118, in custom_generate_chat_prompt_internal
    history_as_text = _concatinate_history(kwargs['history'], state)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 71, in _concatinate_history
    user_name, bot_name = _get_names(state)
                          ^^^^^^^^^^^^^^^^^
  File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 55, in _get_names
    user_name = state['name1_instruct']
                ~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'name1_instruct'
`

With disabled superbooga it works fine, but I need RAG in my solution and this + oobabooga is perfect (all another RAG solutions is worese or shoud be manually coded inside client insted of being hided under Openai Api endpoint)

Few days ago (before installing fresh oobabooga) it works fine via API even with this extension.

Bikkies:
> Hello. I update (reinstall) my whole oobabooga webui with suprebooga v2 and now got a werid problem. It's work like a charm inside webui, but when i try to access oobabooga via OpenAI api, I got this error (even when superbooga v2 is set to manual and I do not trigger it) `ERROR: Exception in ASGI application Traceback (most recent call last): File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 426, in run_asgi result = await app( # type: ignore[func-returns-value] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__ return await self.app(scope, receive, send) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\applications.py", line 1106, in __call__ await super().__call__(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\applications.py", line 122, in __call__ await self.middleware_stack(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 184, in __call__ raise exc File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 162, in __call__ await self.app(scope, receive, _send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 83, in __call__ await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 79, in __call__ raise exc File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 68, in __call__ await self.app(scope, receive, sender) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 20, in __call__ raise e File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 17, in __call__ await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 718, in __call__ await route.handle(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 276, in handle await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 69, in app await response(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 233, in __call__ async with anyio.create_task_group() as task_group: File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\anyio\_backends\_asyncio.py", line 597, in __aexit__ raise exceptions[0] File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 236, in wrap await func() File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 221, in stream_response async for data in self.body_iterator: File "D:\LLM\text-generation-webui\extensions\openai\script.py", line 126, in generator for resp in response: File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 497, in stream_chat_completions for resp in chat_completions_common(body, is_legacy, stream=True): File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 283, in chat_completions_common for a in generator: File "D:\LLM\text-generation-webui\modules\chat.py", line 322, in generate_chat_reply for history in chatbot_wrapper(text, state, regenerate=regenerate, _continue=_continue, loading_message=loading_message, for_ui=for_ui): File "D:\LLM\text-generation-webui\modules\chat.py", line 253, in chatbot_wrapper prompt = apply_extensions('custom_generate_chat_prompt', text, state, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\modules\extensions.py", line 224, in apply_extensions return EXTENSION_MAP[typ](*args, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\modules\extensions.py", line 100, in _apply_custom_generate_chat_prompt return extension.custom_generate_chat_prompt(text, state, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\script.py", line 167, in custom_generate_chat_prompt return custom_generate_chat_prompt_internal(user_input, state, collector, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 118, in custom_generate_chat_prompt_internal history_as_text = _concatinate_history(kwargs['history'], state) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 71, in _concatinate_history user_name, bot_name = _get_names(state) ^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 55, in _get_names user_name = state['name1_instruct'] ~~~~~^^^^^^^^^^^^^^^^^^ KeyError: 'name1_instruct' 2023-12-14 11:34:34 DEBUG:Applying settings. ERROR: Exception in ASGI application Traceback (most recent call last): File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 426, in run_asgi result = await app( # type: ignore[func-returns-value] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__ return await self.app(scope, receive, send) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\applications.py", line 1106, in __call__ await super().__call__(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\applications.py", line 122, in __call__ await self.middleware_stack(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 184, in __call__ raise exc File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\errors.py", line 162, in __call__ await self.app(scope, receive, _send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 91, in __call__ await self.simple_response(scope, receive, send, request_headers=headers) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\cors.py", line 146, in simple_response await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 79, in __call__ raise exc File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\middleware\exceptions.py", line 68, in __call__ await self.app(scope, receive, sender) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 20, in __call__ raise e File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\fastapi\middleware\asyncexitstack.py", line 17, in __call__ await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 718, in __call__ await route.handle(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 276, in handle await self.app(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\starlette\routing.py", line 69, in app await response(scope, receive, send) File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 233, in __call__ async with anyio.create_task_group() as task_group: File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\anyio\_backends\_asyncio.py", line 597, in __aexit__ raise exceptions[0] File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 236, in wrap await func() File "D:\LLM\text-generation-webui\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 221, in stream_response async for data in self.body_iterator: File "D:\LLM\text-generation-webui\extensions\openai\script.py", line 126, in generator for resp in response: File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 497, in stream_chat_completions for resp in chat_completions_common(body, is_legacy, stream=True): File "D:\LLM\text-generation-webui\extensions\openai\completions.py", line 283, in chat_completions_common for a in generator: File "D:\LLM\text-generation-webui\modules\chat.py", line 322, in generate_chat_reply for history in chatbot_wrapper(text, state, regenerate=regenerate, _continue=_continue, loading_message=loading_message, for_ui=for_ui): File "D:\LLM\text-generation-webui\modules\chat.py", line 253, in chatbot_wrapper prompt = apply_extensions('custom_generate_chat_prompt', text, state, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\modules\extensions.py", line 224, in apply_extensions return EXTENSION_MAP[typ](*args, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\modules\extensions.py", line 100, in _apply_custom_generate_chat_prompt return extension.custom_generate_chat_prompt(text, state, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\script.py", line 167, in custom_generate_chat_prompt return custom_generate_chat_prompt_internal(user_input, state, collector, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 118, in custom_generate_chat_prompt_internal history_as_text = _concatinate_history(kwargs['history'], state) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 71, in _concatinate_history user_name, bot_name = _get_names(state) ^^^^^^^^^^^^^^^^^ File "D:\LLM\text-generation-webui\extensions\superboogav2\chat_handler.py", line 55, in _get_names user_name = state['name1_instruct'] ~~~~~^^^^^^^^^^^^^^^^^^ KeyError: 'name1_instruct'`
> 
> With disabled superbooga it works fine, but I need RAG in my solution and this + oobabooga is perfect (all another RAG solutions is worese or shoud be manually coded inside client insted of being hided under Openai Api endpoint)
> 
> Few days ago (before installing fresh oobabooga) it works fine via API even with this extension. Any idea what is changes?

I had this same issue. As a hacky workaround I just modified the chat_handler.py file _get_names function to look like this with hard coded names that work with the model I'm using, but I'm sure there's a proper fix that needs to happen somewhere else:

```
def _get_names(state: dict):
    if _is_instruct_mode(state):
        user_name = 'user'
        bot_name = 'assistant'
    else:
        user_name = state['name1']
        bot_name = state['name2']

    if not user_name:
        user_name = 'User'
    if not bot_name:
        bot_name = 'Assistant'

    return user_name, bot_name
```

hchasens:
Not sure if it's my particular install but `pydantic` is throwing errors due to `chromadb` importing a migrated module. In particular I'm getting [this error](https://docs.pydantic.dev/2.5/migration/#basesettings-has-moved-to-pydantic-settings) with `pydantic 2.5`. 

I tried updating chromadb to version 0.4.21n the extensions requirements.txt but that created another round of errors. Any suggestions? Should I just downgrade pydantic?

ass45sin:
> Not sure if it's my particular install but `pydantic` is throwing errors due to `chromadb` importing a migrated module. In particular I'm getting [this error](https://docs.pydantic.dev/2.5/migration/#basesettings-has-moved-to-pydantic-settings) with `pydantic 2.5`.
> 
> I tried updating chromadb to version 0.4.21 in the extensions requirements.txt. It fixed that issue but created another round of errors in the process. Any suggestions? Should I just downgrade pydantic?

Yes use 1.10.12 instead. Suggestion below

https://github.com/oobabooga/text-generation-webui/issues/4307#issuecomment-1858686179

sandyis:
Hello everyone 
I do not see the tab for `superboogav2`, i should right?
i installed it using 
* chromadb 0.3.27
* pydantic 1.10.12

as per start log extension was loaded 
did i miss something ? 

```
21:52:13-395445 INFO     Loading the extension "gallery"
21:52:13-397873 INFO     Loading the extension "Training_PRO"
21:52:13-400012 INFO     Loading the extension "superboogav2"
21:52:13-768847 DEBUG    Intercepting all calls to posthog.
21:52:14-294107 DEBUG    Creating Sentence Embedder...
21:52:14-990575 DEBUG    Loading hyperparameters...
21:52:14-992461 INFO     Loading the extension "openai"

21:55:16-384644 INFO     Loading llama-2-13b-chat-hf
21:55:16-388597 INFO     Using the following 4-bit params: {'load_in_4bit': True, 'bnb_4bit_compute_dtype': torch.float16, 'bnb_4bit_quant_type': 'nf4', 'bnb_4bit_use_double_quant': False}
Loading checkpoint shards: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:26<00:00,  4.37s/it]
21:55:43-780934 INFO     LOADER: Transformers
21:55:43-781865 INFO     TRUNCATION LENGTH: 2048
21:55:43-782583 INFO     INSTRUCTION TEMPLATE: Llama-v2
21:55:43-783178 INFO     Loaded the model in 27.40 seconds.
21:56:03-840094 INFO     Successfully deleted 0 records from chromaDB.
21:56:05-893055 INFO     Adding 52 new embeddings.
Output generated in 23.14 seconds (6.22 tokens/s, 144 tokens, context 1122, seed 875102836)

```

RandomInternetPreson:
There is no tab for the extebs, the options and instructions are below the text entering window.

thistleknot:
```
Error:

  File "/data/text-generation-webui/modules/extensions.py", line 47, in load_extensions
    extension.setup()
  File "/data/text-generation-webui/extensions/superboogav2/script.py", line 35, in setup
    collector = make_collector()
  File "/data/text-generation-webui/extensions/superboogav2/chromadb.py", line 376, in make_collector
    return ChromaCollector(SentenceTransformerEmbedder())
  File "/data/text-generation-webui/extensions/superboogav2/chromadb.py", line 101, in __init__
    self.collection = self.chroma_client.create_collection(name="context", embedding_function=self.embedder.embed)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/api/client.py", line 198, in create_collection
    return self._server.create_collection(
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/telemetry/opentelemetry/__init__.py", line 127, in wrapper
    return f(*args, **kwargs)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/api/segment.py", line 191, in create_collection
    return Collection(
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/api/models/Collection.py", line 87, in __init__
    validate_embedding_function(embedding_function)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/chromadb/api/types.py", line 211, in validate_embedding_function
    raise ValueError(
ValueError: Expected EmbeddingFunction.__call__ to have the following signature: odict_keys(['self', 'input']), got odict_keys(['self', 'args', 'kwargs'])
Please see https://docs.trychroma.com/embeddings for details of the EmbeddingFunction interface.
Please note the recent change to the EmbeddingFunction interface: https://docs.trychroma.com/migration#migration-to-0416---november-7-2023
```

I'm working on an edited version of chromadb.py, but would be nice if the maintainer's fixed it cuz I'm just hacking away with gpt4

rlconst:
@thistleknot check out https://github.com/oobabooga/text-generation-webui/pull/5456

BO-REN-CHENG:
everyone knowed please help!!
when I using the superboogav2 extension on text generation webui ,
and follow the rule like:

USER:
<|injection-point|>

<|begin-user-input|>What datasets are mentioned in the text above?<|end-user-input|>
ASSISTANT: 

in notebook(default) mode, then I face the problem as below:
OSError: [E050] Can't find model 'en_core_web_sm'. It doesn't seem to be a Python package or a valid path to a data directory.

sandyis:
Absolutely, let's troubleshoot this "OSError: [E050] Can't find model 'en_core_web_sm'" issue you're encountering with the superboogav2 extension.

**Understanding the Error**

The error indicates that the spaCy model 'en_core_web_sm' is required for the extension to work properly, but it's not found in your environment. This model provides essential natural language processing capabilities like tokenization, part-of-speech tagging, and named entity recognition.

**Here's a breakdown of potential causes and solutions:**

**1. Missing Installation:**

* **Check Installation:** First, ensure you've actually downloaded the spaCy model. If it's missing, use the following command in your terminal or command prompt:

    ```bash
    python -m spacy download en_core_web_sm
    ```

**2. Incorrect Environment:**

* **Virtual Environments:** If you're using virtual environments (which is highly recommended), make sure you have both the superboogav2 extension and the 'en_core_web_sm' model installed **within the same environment**.
* **Jupyter Notebook:** Verify that your Jupyter Notebook is running in the same environment where the model is installed. You might need to change the kernel or adjust environment settings.

**3. Path Issues:**

* **Environment Variables:** Although less common, check if there are any environment variables related to spaCy (e.g., `SPACY_DATA`) that might be interfering with the model's location.

**4. Conflicting Versions:**

* **Compatibility:** Ensure that the version of the superboogav2 extension is compatible with the version of spaCy and, specifically, the 'en_core_web_sm' model. Refer to the extension's documentation for recommended versions.

**Debugging Tips:**

1. **Double-Check:** Carefully review the configuration settings for the superboogav2 extension. Verify that there are no typos or incorrect paths specified.
2. **Restart:** Sometimes restarting your environment (terminal, Jupyter Notebook kernel) after installing or changing configurations can help everything sync up.
3. **Documentation:** Consult the documentation for the superboogav2 extension in case there are specific setup instructions or known issues.

**If you're still stuck, please provide the following information:**

* **Operating system** (Windows, macOS, Linux)
* **Python version**
* **spaCy version**
* **superboogav2 extension version**
* **Whether you're using virtual environments**

Let me know how it goes, and we'll get this sorted out!


PS answe courtesy Gemini
________________________________

BO-REN-CHENG:
> Absolutely, let's troubleshoot this "OSError: [E050] Can't find model 'en_core_web_sm'" issue you're encountering with the superboogav2 extension. ***Understanding the Error*** The error indicates that the spaCy model 'en_core_web_sm' is required for the extension to work properly, but it's not found in your environment. This model provides essential natural language processing capabilities like tokenization, part-of-speech tagging, and named entity recognition. **Here's a breakdown of potential causes and solutions:** **1. Missing Installation:** * **Check Installation:** First, ensure you've actually downloaded the spaCy model. If it's missing, use the following command in your terminal or command prompt: ```bash python -m spacy download en_core_web_sm ``` **2. Incorrect Environment:** * **Virtual Environments:** If you're using virtual environments (which is highly recommended), make sure you have both the superboogav2 extension and the 'en_core_web_sm' model installed **within the same environment**. * **Jupyter Notebook:** Verify that your Jupyter Notebook is running in the same environment where the model is installed. You might need to change the kernel or adjust environment settings. **3. Path Issues:** * **Environment Variables:** Although less common, check if there are any environment variables related to spaCy (e.g., `SPACY_DATA`) that might be interfering with the model's location. **4. Conflicting Versions:** * **Compatibility:** Ensure that the version of the superboogav2 extension is compatible with the version of spaCy and, specifically, the 'en_core_web_sm' model. Refer to the extension's documentation for recommended versions. **Debugging Tips:** 1. **Double-Check:** Carefully review the configuration settings for the superboogav2 extension. Verify that there are no typos or incorrect paths specified. 2. **Restart:** Sometimes restarting your environment (terminal, Jupyter Notebook kernel) after installing or changing configurations can help everything sync up. 3. **Documentation:** Consult the documentation for the superboogav2 extension in case there are specific setup instructions or known issues. **If you're still stuck, please provide the following information:** * **Operating system** (Windows, macOS, Linux) * **Python version** * **spaCy version** * **superboogav2 extension version** * **Whether you're using virtual environments** Let me know how it goes, and we'll get this sorted out! PS answer courtesy Gemini




> Absolutely, let's troubleshoot this "OSError: [E050] Can't find model 'en_core_web_sm'" issue you're encountering with the superboogav2 extension. ***Understanding the Error*** The error indicates that the spaCy model 'en_core_web_sm' is required for the extension to work properly, but it's not found in your environment. This model provides essential natural language processing capabilities like tokenization, part-of-speech tagging, and named entity recognition. **Here's a breakdown of potential causes and solutions:** **1. Missing Installation:** * **Check Installation:** First, ensure you've actually downloaded the spaCy model. If it's missing, use the following command in your terminal or command prompt: ```bash python -m spacy download en_core_web_sm ``` **2. Incorrect Environment:** * **Virtual Environments:** If you're using virtual environments (which is highly recommended), make sure you have both the superboogav2 extension and the 'en_core_web_sm' model installed **within the same environment**. * **Jupyter Notebook:** Verify that your Jupyter Notebook is running in the same environment where the model is installed. You might need to change the kernel or adjust environment settings. **3. Path Issues:** * **Environment Variables:** Although less common, check if there are any environment variables related to spaCy (e.g., `SPACY_DATA`) that might be interfering with the model's location. **4. Conflicting Versions:** * **Compatibility:** Ensure that the version of the superboogav2 extension is compatible with the version of spaCy and, specifically, the 'en_core_web_sm' model. Refer to the extension's documentation for recommended versions. **Debugging Tips:** 1. **Double-Check:** Carefully review the configuration settings for the superboogav2 extension. Verify that there are no typos or incorrect paths specified. 2. **Restart:** Sometimes restarting your environment (terminal, Jupyter Notebook kernel) after installing or changing configurations can help everything sync up. 3. **Documentation:** Consult the documentation for the superboogav2 extension in case there are specific setup instructions or known issues. **If you're still stuck, please provide the following information:** * **Operating system** (Windows, macOS, Linux) * **Python version** * **spaCy version** * **superboogav2 extension version** * **Whether you're using virtual environments** Let me know how it goes, and we'll get this sorted out! PS answer courtesy Gemini

I eventually discovered that the issue was due to incompatibility between spacy and en_core_web_sm versions. However,  in versions installed by default using pip, I still encountered many problems (the final issue being AttributeError: [E046] Can't retrieve unregistered extension attribute 'phrases'. Did you forget to call the set_extension method?)

My environment:
python: 3.11.8
spacy: 3.7.4
en_core_web_sm: 3.7.4

However, I thought this may be due to the method of programming using outdated packages. Therefore, I would like to ask if there are any recommended versions of spacy and en_core_web_sm that are more suitable for use under a text generation webui and compatible with superboogav2.

Raunak-Kumar7:
Does the 'http://127.0.0.1:5002/api/v1/get?sort=distance' API call searches based on Metadata? 
What I intend to ask is that, while adding data to the ChromaDB via 'http://127.0.0.1:5002/api/v1/add' API call, we can pass metadata. Now while retrieving the data from vectorDB I want it to extract data having specific metadata only.

PR #3288, opened 2023-07-24
Add capability to handle character history to the blocking API

RuntimeRacer:
This PR adds a new GET Endpoint for Character History Retrieval and an option to save the conversation to the character history right away.
I needed this functionality since I want to integrate a character chat driven by the textgen-server into a software I am working on. I hope others find this useful as well.

### DIFF
--- a/api-examples/api-example-chat.py
+++ b/api-examples/api-example-chat.py
@@ -9,6 +9,9 @@ URI = f'http://{HOST}/api/v1/chat'
 # For reverse-proxied streaming, the remote will likely host with ssl - https://
 # URI = 'https://your-uri-here.trycloudflare.com/api/v1/chat'
 
+# To retrieve the chat history of the character, do a GET Request against:
+# http://{HOST}/api/v1/chat/history
+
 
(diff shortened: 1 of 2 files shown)


oobabooga:
I don't think the persistent history should be returned through the API. This could lead to leaking of personal information. The persistent history is a convenience UI feature for those running a single-user setup.

RuntimeRacer:
Wow, directly closed without any discussion...

I understand that you're probably very busy with this project, but please hear me out:

First of all, isn't that something to be considered by the people who would host the API? By default, the API isn't even shared over the public network if the server isn't run using the `--listen` parameter.
So I don't see why you would not add this just because someone might be able to understand what they're doing?
Also there would be the possibility to make this require more explicit configuration.

As I said, I am working on a software which is meant to integrate with a textgen-ui server running locally, and this particular feature is meant for a single user setup in that combination.
I want people who use my software to be able to use AI with the lowest setup boundaries possible.

I'd find it really sad if we'd not be able to find a solution to this which benefits all users.

PR #2892, opened 2023-06-27
Add support for starcoder ggml and similar (GPT-2 GGML?)

randoentity:
This is a rebase of https://github.com/oobabooga/text-generation-webui/pull/2364 on main which got some conflicts after #2720 was merged.
I didn't read this extensively and the parameters for the model probably don't work (it might use defaults and gpu offloading is only supported by ctransformers for llama models).
That said I've tested this with WizardCoder-15B-1.0-GGML and it's running. I wanted to try this model in combination with superbooga and preliminary tests work pretty well.

The work here, other than fixing the conflicts and minor refactoring, is by @s-kostyaev.

PR #2364, opened 2023-05-26
Add starcoder and starchat ggml support

s-kostyaev:
Models for testing can be found here:

https://huggingface.co/NeoDim

PR #3313, opened 2023-07-26
ctransformers: another attempt

cal066:
Generalized ctransformers based on:
https://github.com/oobabooga/text-generation-webui/pull/2892
Credits to randoentity

RFC since it moves some control structures around to allow extensions to overload them more easily.
Select Model Type "None" to allow ctransformers to guess based on directory config.json file.

You may need to install ctransformers manually at this time, but latest binaries use Cuda 12.x instead of 11.x.
env CT_CUBLAS=1 pip install ctransformers --no-binary ctransformers

Edit: Now using ctransformers build from @jllllll.

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -41,7 +41,7 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False):
             yield ''
             return
 
-        if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'ExllamaModel']:
+        if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'ExllamaModel', 'CtransformersModel']:
             generate_func = generate_reply_custom
         else:
             generate_func = generate_reply_HF
@@ -88,9 +88,8 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False):
 
     yield reply
 
-
 def encode(prompt, add_special_tokens=True, add_bos_token=True, truncation_length=None):
-    if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel']:
+    if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'CtransformersModel']:
         input_ids = shared.tokenizer.encode(str(prompt))
         input_ids = np.array(input_ids).reshape(1, len(input_ids))
     else:
@@ -104,7 +103,7 @@ def encode(prompt, add_special_tokens=True, add_bos_token=True, truncation_lengt
     if truncation_length is not None:
         input_ids = input_ids[:, -truncation_length:]
 
-    if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'ExllamaModel'] or shared.args.cpu:
+    if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'ExllamaModel', 'CtransformersModel'] or shared.args.cpu:
         return input_ids
     elif shared.args.deepspeed:
         return input_ids.to(device=local_rank)
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -88,6 +88,7 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False):
 
     yield reply
 
+
 def encode(prompt, add_special_tokens=True, add_bos_token=True, truncation_length=None):
     if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel', 'CtransformersModel']:
         input_ids = shared.tokenizer.encode(str(prompt))
--- a/modules/ui_model_menu.py
+++ b/modules/ui_model_menu.py
@@ -63,7 +63,7 @@ def create_ui():
 
         with gr.Row():
             with gr.Column():
-                shared.gradio['loader'] = gr.Dropdown(label="Model loader", choices=["Transformers", "ExLlama_HF", "ExLlama", "AutoGPTQ", "GPTQ-for-LLaMa", "llama.cpp", "llamacpp_HF"], value=None)
+                shared.gradio['loader'] = gr.Dropdown(label="Model loader", choices=loaders.loaders_and_params.keys(), value=None)
                 with gr.Box():
                     with gr.Row():
                         with gr.Column():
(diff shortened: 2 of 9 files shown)


jllllll:
I have built ctransformers wheels for various Cuda versions and linked the Cuda 11.7 one here:
https://github.com/oobabooga/text-generation-webui/issues/3357

cal066:
@jllllll Thanks, updated to use your build.

lppllppl920:
It looks like the model unloading of CtransformersModel will not release the GPU memory occupied.

oobabooga:
I have some basic questions about ctransformers:

1) How are the ggmls generated? For instance, if I want to do a q4_0 or q4_K_M for GPT-J or Falcon starting from the model in Hugging Face format, how do I do it?
2) What is the main use case? What models are worth using on ctransformers the most?

jllllll:
> I have some basic questions about ctransformers:
> 
>     1. How are the ggmls generated? For instance, if I want to do a q4_0 or q4_K_M for GPT-J or Falcon starting from the model in Hugging Face format, how do I do it?
> 
>     2. What is the main use case? What models are worth using on ctransformers the most?

Falcon models can be converted and quantized using the conversion script and binaries provided by [ggllm.cpp](https://github.com/cmp-nct/ggllm.cpp).
All other models use the scripts and binaries provided by the [ggml library](https://github.com/ggerganov/ggml):
https://github.com/ggerganov/ggml/tree/master/examples
LLaMa models still use llama.cpp.

The main primary use case that ctransformers provides is access to non-LLaMa models. It's llama.cpp bindings don't provide anything that llama-cpp-python doesn't already, as far as I can tell.

Starcoder, Falcon and MPT ggml models are what most people want out of ctransformers. There is also Replit, Dolly and the GPT models, but I haven't seen much interest in them lately.

It is also worth mentioning that work is being done to unify all of the various ggml implementations into a single file type:
https://github.com/ggerganov/ggml/issues/220
https://github.com/ggerganov/ggml/pull/302
Eventually, there will be no need for multiple ggml backends once that work is finished.

cal066:
> It looks like the model unloading of CtransformersModel will not release the GPU memory occupied.

@lppllppl920
This seems to be the case as well for exllama/llama.cpp from my experience, I have to restart webui to clean it up. I'll need to look into how unloading works.

cal066:
It seems a [destructor already exists](https://github.com/marella/ctransformers/blob/9041ba5aa00da9f0d2961aef994567a945d840a2/ctransformers/llm.py#L388). I'm not too familiar with the ctransformer C/C++ ggml backend to tell if its a memory leak somewhere.

ye7iaserag:
I checked the PR and it doesn't mention WizardML in the models list, I though ctransformers support WizardML...

jllllll:
WizardML is llama/llama2

modules/ui_model_menu.py
```diff
@@ -63,7 +63,7 @@ def create_ui():
 
         with gr.Row():
             with gr.Column():
-                shared.gradio['loader'] = gr.Dropdown(label="Model loader", choices=["Transformers", "ExLlama_HF", "ExLlama", "AutoGPTQ", "GPTQ-for-LLaMa", "llama.cpp", "llamacpp_HF"], value=None)
+                shared.gradio['loader'] = gr.Dropdown(label="Model loader", choices=loaders.loaders_and_params.keys(), value=None)
```
oobabooga, line 66: Does that keep the same order for the loader names in the list?

modules/text_generation.py
```diff
@@ -89,8 +97,12 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False):
     yield reply
 
 
+encode_llama_prompts = ['LlamaCppModel', 'RWKVModel', 'CtransformersModel']
```
oobabooga, line 100: I don't like these variables floating here and think that they should be removed.

oobabooga (commented):
2 more questions:

1) I assume this PR does the same thing as https://github.com/oobabooga/text-generation-webui/pull/2364 and https://github.com/oobabooga/text-generation-webui/pull/2892, so if it gets merged, both can be closed. Is that correct? 

2) The README needs to be updated to briefly mention ctransformers support, and it would be nice to have a short entry under `docs/` mentioning what it is and giving a couple of examples of models that it can load.

modules/ui_model_menu.py
cal066: I have switched to using OrderedDict backing to preserve order.

modules/text_generation.py
cal066: I hate probing by class names, but I have put them back for now. I'll try to refactor them next time.

cal066:
@oobabooga Yes, this was based on #2892 from @randoentity.

oobabooga:
Looks good now. It's very barebones but it works with falcon-7b:

python server.py --model models/falcon-7b-instruct.ggccv1.q4_0.bin --loader ctransformers --model_type "falcon" --n-gpu-layers 10000

I have added @randoentity as a coauthor. Before the merge I also made several little adjustments.

Thank you for you work on this PR.

cal066:
Awesome, thanks.

randoentity:
@oobabooga thanks for mentioning me, but most of the work was by @s-kostyaev :) I just did a rebase and a minor refactor.

randoentity:
Thanks for the work here @cal066 ! I missed your PR but did see GPU offloading was added upstream last week and wanted to bring it here, but ${life}. I'm awestruck how much momentum this project has. 🚀

oobabooga:
Oops, it looks like I messed up then, I should have credited @s-kostyaev as well. Sorry!

I have added one more parameter for ctransformers here https://github.com/oobabooga/text-generation-webui/commit/28c8df337bc14a752f66f8ee258b73ee621329b1. I find it weird that `seed` is both a loading parameter and a generation parameter, as well as `threads`. Also, truncation seems to not be implemented, at least not in a customizable way. If anyone can think of ways to improve the integration with ctransformers, feel free to submit new PRs.

cal066:
@oobabooga I added those according to the def parameters I found in the ctransformers python code. The double seed probably isn't not needed, it tries to get it from the generate call if possible, and then fall back to the config option, likewise for seed. I have created #3543 to resolve this.

lppllppl920:
@cal066 thoughts on changing `infer_loader` so that ctransformers loader can be pre-selected when a certain model name is chosen?

cal066:
@lppllppl920 I just tried TheBloke/airoboros-mpt-30b-gpt4-1p4-GGML, it's not quite reliable. I can make it select ctransfomers but the 'None' model_type still needs to be manually set to 'mpt'. The 'None' is equivalent to auto detect, but requires a config.json in the model directory, something like [this](https://huggingface.co/TheBloke/falcon-40b-sft-top1-560-GGML/blob/main/config.json).

`infer_loader` doesn't allow setting model_type yet, but it's still possible to do some magic in `CtransformersModel` if model_type is 'None' and 'config.json' is not present.

Issue #3262, opened 2023-07-23
Same seed but differents outputs when Regenerating (on exllama and exllama_hf)


BadisG:
### Describe the bug

Hello,

I think the fixed seed isn't really stable, when I regenerate with exactly the same settings, it can happen I get differents outputs, which is weird.

Here's the deterministic preset I'm using for test:
![image](https://github.com/oobabooga/text-generation-webui/assets/110173477/0525ba92-879f-41c7-8f8b-c9fc9d91d242)

Here's the first genration:
![image](https://github.com/oobabooga/text-generation-webui/assets/110173477/fc92aedb-c12b-4587-ad33-9a6b6592ea4a)

And here's the second one:
![image](https://github.com/oobabooga/text-generation-webui/assets/110173477/95aa9771-8c6c-4457-881d-9176a6727299)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Put a fixed seed
Regenerate until getting a different output

### Logs

```shell
No logs
```

### System Info

```shell
Windows 10
loader: exllama_hf
model : llama2-13b-chat
```

berkut1:
I think the seed work for tokens, and after regeneration with the same seed, it uses the same tokens, but differently use them, because of high temperature, top_p and etc.
https://towardsdatascience.com/decoding-strategies-that-you-need-to-know-for-response-generation-ba95ee0faadc
Maybe I'm wrong, but that's how I understand it.

BadisG:
I deactivated "do_sample" so there's no samplers involved, it should be deterministic

BadisG:
Here's a video showing the problem:

https://github.com/oobabooga/text-generation-webui/assets/110173477/3e97c609-f98d-4046-8537-9cfb89bc07ba

berkut1:
@BadisG did you check the llama.cpp? Does it have the same problem?

BadisG:
Not really but I want to fix that issue on ooba's webui, not on llama.cpp

berkut1:
@BadisG ooba is just webui of llama-cpp-python. If you want something to fix you should check first [llama.cpp](https://github.com/ggerganov/llama.cpp), then [llama-cpp-python](https://github.com/abetlen/llama-cpp-python).

BadisG:
What? What you say only apply for the llama-cpp-python loader, I'm using exllama and exllama_hf loader, those are totally unrelated

berkut1:
[...]

PR #3325, opened 2023-07-26
Add Classifier Free Guidance (CFG) for Transformers/ExLlama

oobabooga:
The new input fields can be found in the Parameters tab:

new

This is the paper where the method comes from: https://arxiv.org/pdf/2306.17806.pdf


BadisG:
Doesn't seem to be working for me, I tried it with both exllama_hf and auto_gptq

exllama_hf:
```
  File "D:\Large-Language-Models\text-generation-webui\modules\callbacks.py", line 55, in gentask
  File "D:\Large-Language-Models\text-generation-webui\modules\text_generation.py", line 296, in generate_with_callback
  File "C:\Users\Utilisateur\anaconda3\envs\textgen\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "C:\Users\Utilisateur\anaconda3\envs\textgen\lib\site-packages\transformers\generation\utils.py", line 1379, in generate
    and torch.sum(inputs_tensor[:, -1] == generation_config.pad_token_id) > 0
IndexError: index -1 is out of bounds for dimension 1 with size 0
Output generated in 0.35 seconds (0.00 tokens/s, 0 tokens, context 0, seed 2037041530)
```

auto_gptq:
```
  File "D:\Large-Language-Models\text-generation-webui\modules\callbacks.py", line 55, in gentask
  File "D:\Large-Language-Models\text-generation-webui\modules\text_generation.py", line 296, in generate_with_callback
  File "C:\Users\Utilisateur\anaconda3\envs\textgen\lib\site-packages\auto_gptq\modeling\_base.py", line 438, in generate
    return self.model.generate(**kwargs)
  File "C:\Users\Utilisateur\anaconda3\envs\textgen\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "C:\Users\Utilisateur\anaconda3\envs\textgen\lib\site-packages\transformers\generation\utils.py", line 1379, in generate
    and torch.sum(inputs_tensor[:, -1] == generation_config.pad_token_id) > 0
IndexError: index -1 is out of bounds for dimension 1 with size 0
Output generated in 0.37 seconds (0.00 tokens/s, 0 tokens, context 0, seed 1799734888)
```

BadisG:
> For me it worked fine on AutoGPTQ. On exllama_hf I get
> 
> ```
>     assert len(args) == 0, 'no *args should be passed to forward'
> AssertionError: no *args should be passed to forward
> ```
> 
> so some additional work will probably be needed there.

For exllama_hf I got your error when using a llama1 model (dolphin-llama-13b), but I got the "index -1 is out of bounds" error when I use a llama2 model (llama2-13b-chat)

For auto_gptq it also works fine for llama1 models but has the "index -1 is out of bounds" for llama2 models

oobabooga:
CFG makes two forward calls for each token, each with a different cache provided by `kwargs['past_key_values']`.

For this to work with exllama_hf and llamacpp_hf, this cache needs to make it into exllama and llama.cpp somehow, which is quite difficult.

EyeDeck:
Not sure if you've seen it, but turboderp wrote a CFG example for exllama, which also uses batching to minimize the performance hit of having to run two forward passes: https://github.com/turboderp/exllama/blob/master/example_cfg.py
Not gonna be as memory-efficient though, of course.

oobabooga:
Thanks @EyeDeck. I had seen @turboderp's code before but I hadn't understood what it was doing. It's quite clever to reduce the two forward calls to just one with batch size = 2.

oobabooga:
I have added CFG for the ExLlama loader following [example_cfg.py](https://github.com/turboderp/exllama/blob/master/example_cfg.py). I'm not sure if I did it right but it seems to make sense.

The bad news is that it runs out of memory for 30b models on 24GB VRAM due to the `batch_size=2` in the cache.

Panchovix:
I've tested on exllama on 70B, it definetly uses more VRAM.

At group size 64, and 4096 ctx, I was using ~42GB VRAM. Now it uses 44-47GB VRAM.

EyeDeck:
Yeah, batch_size = 2 means the KV cache also doubles in size, so you can probably only fit, like, 1600 or 1700 context in 24GB using 33B and ExLlama. ExLlama also pre-allocates the required memory for max_seq_len, instead of allocating later, so it usually OOMs on load, instead of some point later when context gets too full.

It does appear to work, at least.
However, the batch_size optimization doesn't help _that_ much, some limited testing on my machine (using 13B and 33B) shows it's around 60% as fast as with batch_size = 1, which isn't _that_ impressive considering the memory cost. Depends on the machine and model whether it's worth the memory hit, e.g. if you're using 70B with 48GB like Panchovix above, and aren't using some form of RoPE scaling, GQA means you've got a lot of free memory anyway, and batch_size = 2 is a bit faster than two forward passes, so in that case the larger batch size is definitely worth it; on 33B with 24GB, and its regular old MHA, context is much more limited, so with batch_size = 2, you can't even get to 2048 anymore.

AeonBlack:
Don't think I'm understanding it, I keep getting something that starts out promising and immediately turns into gibberish and infinite repeating after a few words.

turboderp:
I'm not sure about the "correct" way to do it, but the basic idea is really simple. You just generate two sequences in parallel, and then you do a linear interpolation of the logits from each before sampling as usual. Whatever token you sample from the combined logits you then append to both sequences and repeat.

Apparently applying a log_softmax on both sets of logits before the linear combination helps, supposedly normalizing them so the alpha parameter has a consistent effect, though I haven't noticed any difference, subjectively. The log_softmax is canceled out by the regular softmax normalization at the end.

It seems like the HF version tries to implement all of it in a sampler, so you generate the regular sequence as usual, but then *for each token* you run inference on the entire "negative" prompt. This saves memory compared to batching or using a separate cache for the negative prompt, but speed is going to suffer a lot. Depending on the length of the response, you could be doing 100x-1000x as much computation, or more. It's like a sequence classification operation, but per token.

What approach is best I think depends how you'd want to use the feature. If you're generating short replies in an ongoing conversation, then probably it's worth it to use two sequences, each with its own cache, and then limiting the length of the reply so you can keep the second cache shorter.

BadisG:
I don't know how you can test if cfg is working or not, when I put a fixed seed it keeps changing anyway...
[fixed-seed-not-working.webm](https://github.com/oobabooga/text-generation-webui/assets/110173477/228354e6-9122-4ee2-910c-88ea1522646b)

Panchovix:
At least on my case with exllama, it definitely works. Using CFG with a negative prompt do affect the answers.

exllama_hf doesn't work (it gives empty outputs)

BadisG:
The outputs change are too strong even with a fixed seed, so adding the cfg on top of it would make it even more confusing. 
@turboderp would the performance of exllama be compromised if you decided to make it deterministic?

turboderp:
>would the performance of exllama be compromised if you decided to make it deterministic?

Hard to say. It would require a complete rewrite of the matmul kernels, and maybe some weeks into that project I could begin to pursue optimization strategies that aren't obvious at the onset... so I don't know.

But if you're relying on determinism to tell if a generation method is working, I'd say you're approaching it wrong. If the effects are so subtle that you can't detect them in the presence of noise that's strictly less than the existing (deterministic) quantization noise, it's hardly a useful method.

BadisG:
> you can't detect them in the presence of noise
The problem is that it's not just some little noise, the generation can be completly different (fixed seed)
https://github.com/oobabooga/text-generation-webui/issues/3262#issuecomment-1653721950

BadisG:
Your example is too "flashy" I would say, I'm not asking for CFG to completly change the behavior of my outputs, but to follow my instructions better and with more consistency. Without CFG my stories can be written fine by the model but I have to do a lot of regenerations to achieve what I consider to be a "good story".

The problem is that to do that, I just change the seed until I got lucky, but how can I find the right value of cfg if everything changes even with a fixed seed.

Let's say for seed 10 I have a bad generation and then I add cfg = 1.5, the generation got better, is it because of the cfg or is it because exllama decided to give me something else due to the non deterministic nature? I will never know

turboderp:
If you change *any* parameter you're going to get different output whether the underlying engine is deterministic or not. Determinism gives you reproducibility but not predictability or interpretability.

So if you have a bad generation for a particular seed and then re-run it with different parameters, you've also changed the meaning of that seed, and you may as well have changed its value because as soon as the model reaches a tipping point it's going to diverge into a different autoregressive sequence. You've stepped on the butterfly either way. All a deterministic model allows you to do is step on it in precisely the same way again, if you so choose.

In any case, the way to test CFG is to try it with a high coefficient so that the effect is obvious if present, then scale it back to make it more subtle.

BadisG:
Having determinism is important if you want to evaluate a sampler. Let's say you take 1000 random seeds, if adding cfg 1.5 made the output better for 60% of the case (600/1000), you can go to the conclusion it's statistically significant and it's better to have this value rather than nothing for your everyday tasks.

But when you remove determinism, if you try to change the cfg value, have no idea if it's the cfg itself or the non-deterministic effect that has the biggest weight on your output changes.

>In any case, the way to test CFG is to try it with a high coefficient so that the effect is obvious if present, then scale it back to make it more subtle.

Of course having huge values of CFG will give you really surprising outputs and things you would've never have with just non-deterministic noise or a simple change of seed. But this isn't my goal there, I don't want a flashy CFG with a value of 3. I want something subtle that just makes the model a bit smarter overall. And in that case, the CFG is gonna be blurred by the strong noise we got on exllama.

The goal of CFG can be that, something that would change the output in such a way that I'd think I got a lucky seed, this is the effect I want, getting better chances of having "lucky seeds."

I'm not sure if you understand what I mean by that, but let's agree to disagree, determinism is really important in my book, not just for CFG, but for all samplers, especially when you're trying to write a story. You want consistency, not something that goes off in all directions and you can't reproduce.

AeonBlack:
Well from what I noticed in this specific PR, there needs to be an option if we want to use 2x cache or not with ExLlama, as OOM issues are more worrisome than slower generation, one of which for example I can just wait a little longer for, the other is a big upfront payment for more expensive GPUs. [I'm only able to run 13B models with 8k context as is for reasonable generation speed.]

turboderp:
@BadisG The point is that adding CFG *effectively* changes the seed as well. You'll have a completely different interpretation of those 1000 seeds with CFG enabled. It won't be like the original 1000 generations but with a little CFG sprinkled on top, it'll be 1000 completely different generations produced under different circumstances. If you can't tell if it's nondeterminism giving you better results, you also won't be able to tell if you're getting better results because of CFG or if it's because enabling CFG changed the interpretation of those 1000 seeds.

To put it another way: Suppose you've got a deterministic model, and for seed 42 it produces output X. Then you add CFG and now it produces output Y. So you change the seed to 43, still with CFG enabled, and it produces output Z instead. Now you've got three outputs:

X from seed 42    
Y from seed 42 + CFG    
Z from seed 43 + CFG    

The point here is that X, Y and Z are *all* going to be different. It's not like X and Y will be closer to each other than X and Z, because even a slight change to the logits at any point can cause the generation to diverge.

@AeonBlack I'm not sure how you'd implement CFG at reasonable speeds without a second cache, though. You'll have to generate two different sequences, and if you're generating one of them without a cache, that's not really going to be usable either. I mean, it may run, but at something like 1-10% of the speed.

AeonBlack:
Maybe I meant batch_size, as I was more referencing the previous comment made here: https://github.com/oobabooga/text-generation-webui/pull/3325#issuecomment-1653005991
If it's still the same issue though then I guess, welp.

turboderp:
Yep. A batch size of two is the same amount of memory as using two caches. But with separate caches at the CFG sequence could be kept shorter, at least. So you'll have your 4k context, but you can apply CFG to responses up to 512 tokens or something.

BadisG:
>To put it another way: Suppose you've got a deterministic model, and for seed 42 it produces output X. Then you add CFG and now it produces output Y. So you change the seed to 43, still with CFG enabled, and it produces output Z instead. Now you've got three outputs:

>X from seed 42
Y from seed 42 + CFG
Z from seed 43 + CFG

>The point here is that X, Y and Z are all going to be different. It's not like X and Y will be closer to each other than X and Z, because even a slight change to the logits at any point can cause the generation to diverge.

That's why you use statistics and a lot of samples, let's say X is the "normal" output and Y is the CFG output,
You will compare X(42) vs Y(42) - then X(43) vs Y(43) and so on... (You don't compare X(42) with Y(43) that wouldn't make any sense)

In a deterministic universe, you know that to get from X(42) to Y(42) you just changed the cfg value (from 1 to 1.5 for example), so you know it's only this sampler and nothing else that did those changes. If you see that at 60% of the time you get Y(x) > X(x) in terms of quality outputs, then you can go to the conclusion that this value of cfg is overall beneficial to the model.

>If you can't tell if it's nondeterminism giving you better results, you also won't be able to tell if you're getting better results because of CFG or if it's because enabling CFG changed the interpretation of those 1000 seeds.

I can't agree with that, the way of evaluating the quality of samplers must be exactly the same as if you evaluate a medication drug. There is the "normal population" that hasn't taken up the drug and the "drugged population" that has taken up the drug. When you compare the % of death, if the "normal population" died at 20% of the time and the "drugged population" died at 15% of the time, it's statistically significant and you go to the conclusion that the drug has a beneficial effect. 

Same thing with cfg, if you have the "normal seeds" and the "CFG seeds", if overall you have a bigger % of "good generations" on the "CFG seeds" compared to the "normal seeds", then you can go to the conclusion that cfg is better than having nothing.

But to do such tests only one thing must be changed, in the case of the medication drug, only the drug has been added, and only the cfg must be changed, if there's noise or non-determinism, you can't do any conclusion

turboderp:
>I can't agree with that, the way of evaluating the quality of samplers must be exactly the same as if you evaluate a medication drug. There is the "normal population" that hasn't taken up the drug and the "drugged population" that has taken up the drug.

And those two groups aren't exact copies of one another. You can't single out each person who has taken the drug and compare them to their identical clone from the control group.

It's equivalent to testing seeds 1-1000 without CFG, and then seeds 1001-2000 with CFG. You're going to get different results from those two groups with or without any underlying determinism. Just as if you were testing two groups of patients, the experiment isn't invalidated by the fact that you can't rewind time to reproduce it *exactly* in every detail. What matters is if you can measure a significant difference between the groups at the end of the experiment.

BadisG:
> > I can't agree with that, the way of evaluating the quality of samplers must be exactly the same as if you evaluate a medication drug. There is the "normal population" that hasn't taken up the drug and the "drugged population" that has taken up the drug.
> 
> And those two groups aren't exact copies of one another. You can't single out each person who has taken the drug and compare them to their identical clone from the control group.
> 
> It's equivalent to testing seeds 1-1000 without CFG, and then seeds 1001-2000 with CFG. You're going to get different results from those two groups with or without any underlying determinism. Just as if you were testing two groups of patients, the experiment isn't invalidated by the fact that you can't rewind time to reproduce it _exactly_ in every detail. What matters is if you can measure a significant difference between the groups at the end of the experiment.

But that's how it work in real life, the groups aren't 100% similar but you randomise your choices, it's called the Randomized controlled trial: https://en.wikipedia.org/wiki/Randomized_controlled_trial
When you choose people in pure randomness, you'll notice that you'll get 2 similar groups overall in the end, that's why it works and that's why the doctors are sure the drugs they are giving to you when you are sick are effective.

But for the cfg thing it's even better, you can use exactly the seed from 1-1000 for the "normal" and "cfg", as if you were using one population without drugs and then using that same population with drugs, that's the perfect simulation... if there was nothing else added to the table, and for exllama that's not the case, there's the noise, it's as if you were adding a second drug to the mix, you just blur everything and you can't make conclusions

turboderp:
>But for the cfg thing it's even better, you can use exactly the seed from 1-1000 for the "normal" and "cfg", as if you were using one population without drugs and then using that same population with drugs, that's the perfect simulation...

So you don't want something equivalent to a drug trial, where there's inherently lots of random factors that you have to control for, and no reproducibility so effectively no determinism. You want a perfect simulation instead. And that's fine, but your testing methodology still isn't sound.

Autoregressive text generation is an inherently unstable process, deterministic or otherwise. As soon as you change sampling parameters *at all* you've already lost any ability to tell if changes in the output are the result of the new parameters or because of all the butterflies you just stepped on, unless that output shows different characteristics that you would have been able to measure regardless.

There is effectively no sense in which two sequences with the same seed are related to each other once you start changing sampling parameters. If you run the same seed but with a higher temperature, you're not getting some high-temperature version of the same output, you're getting a different output altogether. Changing the temperature isn't materially different from changing the temperature and the seed at the same time.

Or how about this: Suppose a non-deterministic implementation had some "fake determinism" built in, by which it would cache whatever output it happened to produce the first time it was given a particular input (including seed and sampling parameters), so that it could always produce that exact same output as long as the input didn't change at all, what experiment could you run to detect that this wasn't actually a deterministic model?

FartyPants:
I experimented with CFG/negative prompt weeks ago using Cypherfox's loggits extension, and it obviously worked, but after spending a lot of time comparing the results - there was just extremely hard way to determine how it helped if at all. In theory it should self focus the answer and so not to drift so early (as it amplifies the previous loggits) - that means the proper testing should be done on a long results. 
At the end I decided not to bother with it as it yet adds more parameters to already complex situation without visible benefit. 
So I'm looking forward to see if someone finds a clear use case for this that I couldn't.

shinkarom:
NovelAI seem to regard CFG as useful if they added it.

FartyPants:
I'm all for this, just I want to see experiment use case where it shows the difference (response is better) using CFG. Because I couldn't make them myself. I could make the response significantly worse if the CFG was set too high - that's not the best test I guess. If I used the CFG around 1.5 as recommended, my absolute honest would be "I can't say" . We had a long discussion on discord how this works, and so (I hope) I generally understand what it does and it should prolong the time the responses stay on topic by gradually amplifying the previous logits.

Ph0rk0z:
So this works same as the one in SD? I can put "short, boring, moralistic, ethical" as negative prompt and get better replies?

BadisG:
>Just cost is 2x inference memory and 2x generation time?

I haven't noticed any speed decrease or vram increase when using exllama + CFG, it got slower on AutoGPTQ tho

BadisG:
>Autoregressive text generation is an inherently unstable process, deterministic or otherwise. 

Meds are also REALLY complicated and unstable, we usually have no idea how the process works on our bodies and the effect depends on each person, but, honestly, we don't really care because we make the problem much simpler by simply testing them and seeing if they give better OVERALL results or not.

I get what you said, of course adding CFG will change the output in a way wouldn't be able to analyze anything, but that's not the point. The only question we should ask ourself is this one, "does adding CFG makes it feel better overall?" 

>As soon as you change sampling parameters at all you've already lost any ability to tell if changes in the output are the result of the new parameters or because of all the butterflies you just stepped on, unless that output shows different characteristics that you would have been able to measure regardless.

I don't get it, if you just change the CFG value you can say that the changes in the output are only the result of CFG, if you keep the same seed and if that's deterministic of course.

I kinda see what you mean by that tho, it's like making a domino fall with you hand, all the dominos will fall in the process, but even on that point you know that what caused this butterfly behavior is your hand and nothing else.

>There is effectively no sense in which two sequences with the same seed are related to each other once you start changing sampling parameters. If you run the same seed but with a higher temperature, you're not getting some high-temperature version of the same output, you're getting a different output altogether. Changing the temperature isn't materially different from changing the temperature and the seed at the same time.

I agree with you but again that's not the point, of course using one seed and seeing the difference between the normal output against the same (output + seed) with a different sampler parameter is not a good method. Maybe you got a better output on this particular seed by changing the CFG from 1 to 1.5, but you can't conclude anything on that, it was probably luck and when you use another seed you'll get worse output by changing the CFG from 1 to 1.5.

But what I wanted to say is this, if you test that approach on a large number of seeds, you'll get an overall perspective on what does changing the cfg from 1 to 1.5 do. If it makes the output better more often than it makes it worse/neutral, then you can go to the conclusion that OVERALL, it's a good thing to do that. That's the magic of statistics.

Again, I'm not asking to comprehend what's happening on the logits when you change the value of CFG, this shouldn't matter, what I'm hoping to see is that OVERALL I'm hoping I got better results by doing that.

And the problem of non-determinism is that you can't use this statistical method because you can't simply say that the output changes were 100% because of the CFG, the noise also have an effect and we have no idea how big that effect is relative to the CFG effect

Let's imagine that setting, we did our test on 1000 first seeds: 
- The OVERALL output quality is "10", the more the better
- Changing CFG from 1 to 1.5 is "x"
- The noise is "y"

In a deterministic world (AutoGPTQ for example), y = 0:
- If 10 + x > 10 -> x > 0 -> Conclusion: OVERALL, changing CFG from 1 to 1.5 is better

On a non-deterministic world (exllama), y is not equal to 0
- If 10 + x + y > 10 then we got 4 solutions
        1) x > 0 AND y > 0 (CFG good)
        2) x > 0 AND y < 0 (CFG good)
        3) x < 0 AND y > 0 (CFG bad)
        4) x = 0 AND y > 0 (CFG neutral)
We can't conclude anything, we can't say if x will be > 0 or not, this is a failed test.

The only way to make any conclusion on a non-deterministic world is only if mean(y(seed = 1) + y(seed = 2) + ... + y(seed = 1000)) = 0 (can be possible if the distribution of y is a standard normal distribution). 

The problem is that it can be really complex to measure the distribution of a noise, because a noise can be different with each sampler settings... hell even the distribution of a noise can be different with each sampler settings! Noise is really the worst enemy of statisticians.

:'( <- me when noise exists

Anyway, so sorry for talking so much, but I just wanted to show how determinism is important in my eyes. Tbh I wouldn't mind if the noise of exllama wasn't that strong (like you can see the noise effect only after more than 200 tokens), but this isn't the case, the output can change just after one sentence.

turboderp:
>    The OVERALL output quality is "10", the more the better
>    Changing CFG from 1 to 1.5 is "x"
>    The noise is "y"

It would be more correct to say that changing CFG from 1 to 1.5 represents a change of "x+z" where x is the contribution from CFG, and z is the butterfly effect, much larger in magnitude than y. It's a chaotic system, not an additive one where you can apply CFG and observe it in isolation, because changing *any* parameter is effectively the same as changing the seed.

You can make ExLlama deterministic if you want. Just add a wrapper that computes a hash of the prompt and all the generation parameters. Then check that hash against a dictionary. If the hash isn't found, run the model with the prompt and parameters, and cache the result in the dictionary. If the hash is found, return the cached result instead. This way you're guaranteed to always get the same output for the same input. It won't be true determinism, but it will be indistinguishable for the experiments you're describing.

Alternatively, recognize that ExLlama is non-deterministic for a reason, and that reason is performance. Nothing stops you from using a slower, deterministic model for evaluating CFG in the way you're describing, and then going back to ExLlama after you've decided if CFG is worthwhile.

BadisG:
>You can make ExLlama deterministic if you want. Just add a wrapper that computes a hash of the prompt and all the generation parameters.

Unfortunately I'm just a statistician not a Machine Learning coding wizard, but yeah I'm glad to see it's possible to make it deterministic.

>Alternatively, recognize that ExLlama is non-deterministic for a reason, and that reason is performance. Nothing stops you from using a slower, deterministic model for evaluating CFG in the way you're describing, and then going back to ExLlama after you've decided if CFG is worthwhile.

It's not just for CFG, it's for absolutely every single sampler. Wanna increase the temperature? You won't know if overall it will makes it better, want to increase the rep_penalty... you know the rest of that story. This noise makes every sampling preset changes a pain to evaluate.

How about having the option of ticking a button that makes the exllama loader deterministic or not? I don't mind if it makes it a bit slower, it's already fast enough. What matter the most (for me at least, it's just my opinion) is having as much precision and consistancy as on AutoGPTQ. If you got a fast loader but it makes the model dumber because of the noise I don't really see the point.

Because that's probably the second disavantage of having noise, it may have decreased the overall quality of the outputs.

You're probably gonna tell me to go back to AutoGPTQ instead of using exllama, and you are perfectly right, I wish I could do that actually, but your loader is the only one that works on models that have groupsize + act_order on windows. And also you can load loras without having to use a monkey patch on exllama. That's why I'm stuck with it I guess.

EyeDeck:
The thing is, the only thing determinism would help you do is find out if some change has literally _no_ effect _at all_. As soon as a different seed, or sampler setting, or whatever, causes _one single token_ to deviate, there's no meaningful way to compare outputs after the point of deviation. If it were deterministic, it's not like you're going to see it generate a few slightly different words, and then work its way back onto the same track, and you could compare what changed in the middle—instead, that one single token of deviation is functionally equivalent to changing the seed for everything beyond that point, whether it's deterministic or not. It's just an artifact of how LLMs function.

BadisG:
@EyeDeck I got that, you change one token and everything changes after that, it's called the butterfly effect and it shouldn't be something you should all be afraid of. 
It's not because a preset change has the effect of totally changing the outputs that it means you can't make comparaisons. It's not just "like a change of seed" because for example when you put temp = 2, it will be so much different than any seed you will change on temp = 1. Samplers have effects you can see the patern, and adding noise just blur that pattern.

And besides, like I said earlier it shouldn't matter at all, when you try a lot of seed you can see the overall effect of a sampler. 
Again, I'm not asking to compare 2 outputs by looking word by word, that's impossible, and that's not the point at all. Yes the outputs are completly different when you change the value of a sampler, yes, it's totally normal and fine, and yes, you can still make comparaisons with butterfly effects.

Because if overall, after trying 100 seeds with a "normal presets" and trying the same 100 seeds with "normal presets + CFG", if you feel you had much better generations with the CFG then you can go to the conclusion it's a good sampler. The problem with noise is that you make things completly random, those 100 seeds with "normal presets + CFG" will be different if you redo it a second time on a non deterministic world, and maybe that time you'll get worse input.

EyeDeck:
Setting CFG below 1 is equivalent to switching the positive and negative prompt;
so CFG 0.5 with positive prompt _a_ and negative prompt _b_ = CFG 1.5 with positive prompt _b_ and negative prompt _a_.
If you look at your first prompt there with cfg -0.5, you're effectively giving it the positive prompt "You really love the color blue.", and then strongly negative prompting it _away_ from following its instruct finetune, so it's now acting more like a regular auto-complete foundational model, and it's interpreted the (effective) positive prompt of "You really love the color blue." as some kind of character description, so it's continuing that pattern and adding more traits to that description.

For your second example, apparently suggesting that _you_ like blue has influenced the model _so much_ that it can't stop thinking about blue, even with the negative prompt. This might be because the positive prompt has the full instruct format while the negative one doesn't, and the instruct finetuning makes it so much more certain about where to go than the "fuzzy" non-instruct negative prompt that CFG isn't having much effect. Same with your third example, except it's already thinking about blue and being prompted away from red, so it's even more likely to think about blue.

Try putting the whole instruct format in both the positive and negative prompts, but tweak the negative one a little, for example positive:
```
Below is an instruction [...]

### Instruction:
Hey man, my favorite color is blue. What's yours?

### Response: 
```

Negative:
```
Below is an instruction [...]

### Instruction:
Hey man, my favorite color is blue. What's yours?

### Response: 
I really love blue.
```
This is roughly how the CFG paper suggests using it (you basically want the same prompt in both places, but _add_ stuff you want the model _not_ to do in the negative; which is much more effective than telling the model _not_ to do something in the positive prompt), and I'd expect the output to tell you it likes something _other_ than blue, probably.

It's also possible to use CFG with a blank negative prompt, which seems to have an effect not entirely unlike raising the temperature, but without confusing the model as much as high temperature does.

oobabooga:
I have made a test with ExLlama, using the prompts in turboderp's example_cfg.py example, and it worked.

The test is using vicuna-33b-v1.3-4bit-128g.

Prompt:

```
You are a rude and obnoxious assistant. You hate everything and everyone.

USER: Tell me a story.
ASSISTANT:
``` 

Negative prompt:

```
You are a helpful, respectful and honest assistant. Always answer as helpfully as possible, while being safe.  Your answers should not include any harmful, unethical, racist, sexist, toxic, dangerous, or illegal content. Please ensure that your responses are socially unbiased and positive in nature. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information.

USER: Tell me a story.
ASSISTANT:
```

Reply with CFG = 1:

```
Once upon a time in a faraway land, there was a kingdom ruled by a wise and just king. The people of the kingdom were happy and content, but they had one major problem – they were cursed with an eternal darkness that plagued their lands. This darkness caused an impenetrable fog to cover the kingdom, making it impossible for the sun's rays to pierce through. As a result, the kingdom remained shrouded in perpetual night.

The king, desperate to find a solution to this curse, called upon his most trusted advisors. They tried everything they could think of: magic, science, and even prayer. But nothing seemed to work. The darkness persisted, and the people began to lose hope.

One day, a mysterious traveler arrived at the castle gates, claiming he had the power to lift the curse from the kingdom. The king, skeptical but desperate,
```

Reply with CFG = 1.5:

```
Alright, you unpleasant creature. Once upon a time in a faraway land called Grumpymore, there was a kingdom filled with the most insufferable people you've ever met. The king, known as King Cranky the First, ruled over his miserable subjects with an iron fist. His favorite pastime was to rain on parades and squash happy thoughts like a bug under his royal shoe.

In this unhappy kingdom, there was a village where all the villagers were required to wear gray, itchy sweaters every day. They would sit around, complaining about everything from the weather to their breakfast cereal. Life was a never-ending cycle of misery and dissatisfaction.

One day, Queen Bitchy of the neighboring kingdom decided she'd had enough of King Cranky's sour attitude. She hatched a plan to bring joy to Grumpymore
```

CamiloMM:
I'm sorry but with everything freshly git pull'd I tried that with llama2-based models on Exllama and without this as the positive prompt:
> You are a rude and obnoxious assistant. You hate everything and everyone.
it will not act that way. Just adding the negative prompt seems to change nothing; even at CFG scale of 2.5. Do you get that result without the very clear positive prompt? I just made it "This is a conversation with your assistant." then added all that helpful text as a negative.

Unless I'm misunderstanding something and this is a very subtle tweak; in Stable Diffusion merely adding "red" to a negative prompt would almost ensure there isn't red in the generated image.

TheBestReguile:
https://github.com/oobabooga/text-generation-webui/pull/3325#issuecomment-1664741299

I just want to say here how much I appreciate this response.  Very great explanation that cleared up a lot of confusion, and the concepts here should be in a readme somewhere.

modules/exllama.py
```diff
@@ -89,31 +106,71 @@ def generate_with_streaming(self, prompt, state):
...
-            if has_leading_space:
-                decoded_text = ' ' + decoded_text
-
-            yield decoded_text
-            if token.item() == self.generator.tokenizer.eos_token_id or shared.stop_everything:
-                break
+            alpha = state['guidance_scale']
+            prompts = [prompt, state['negative_prompt'] or '']
+
+            ids, mask = self.tokenizer.encode(prompts, return_mask=True)
```
wcde, line 143: max_seq_len is not passed to tokenizer. Because of this, when context is greater than 2048, error "Trying to create tensor with negative dimension" occurs.

practicaldreamer: I can confirm this.

I opened a PR seems to resolve issue
https://github.com/oobabooga/text-generation-webui/pull/3622

13631368:
> The new input fields can be found in the Parameters tab:
> 
> new
> 
> It only works for the Transformers and ExLlama loaders at the moment. 
> 
> This is the paper where the method comes from: https://arxiv.org/pdf/2306.17806.pdf

modules/exllama.py
practicaldreamer: Sorry, after further testing my fix didn't work as expected. Still troubleshooting...

oobabooga: I think that a good solution would be to not have any truncation in the `tokenizer.encode` step. Truncation is already handled afterwards.

practicaldreamer:
I do not recommend doing this but if you change 

```    
def encode(self, text, return_mask = False, max_seq_len = 2048, add_bos = False, add_eos = False, encode_special_characters = False):
```

to

```
def encode(self, text, return_mask = False, max_seq_len = 4096, add_bos = False, add_eos = False, encode_special_characters = False):
```

within the exllama's tokenizer.py it works with 4096 context window... I don't recommend doing this as it really should be passed over from the config but as a hack it's working for me

practical-dreamer:
I've been having fun with this new feature making vicuna condescend to the user is pretty funny!

I think it could really be useful to make the model "listen" more to the prompt. I wonder if the naming could be improved to be a bit more intuitive though. One would think with a positive value would correlate with a positive bias. Once you consider that the field is called "negative prompt" it makes sense I just wonder if it could be renamed somehow.

Also if this function supports multiple concurrent prompts that could be _really_ useful... for instance I might want a negative prompt for "As a large language model" AND a positive prompt for "Make sure you think step by step".

Ph0rk0z:
Why? Exllama lets you set sequence length. CFG works for exllama.

modules/exllama.py
city-unit: Is this fixed now?

Nixellion:
Hi guys. I'm sorry if I'm asking a question that was discussed before, but I'd like to just get an ELI5 reply and I think it would be useful to everyone. Thanks in advance.

Does negative prompt "eat away" at the context limit? For example if the total limit is 2048 tokens, can I use a 2048 token long prompt + 2048 token long negative prompt? Or am I limited to 1024+1024 instead?

Does using negative prompt increase RAM usage? From what I've read it sounds like it does in ExLLama but it does not for Exllama_HF, instead causing a slow down in performance?

PR #3326, opened 2023-07-26
sd_api_pictures: Widen sliders for image size minimum and maximum

GuizzyQC:
With the imminent release of SDXL, I figure people might want to generate larger images, so I changed height and width slider minimum and maximum to match the corresponding sliders in Automatic1111's UI, from 256-768 to 64-2048 px. Maxing it out is unlikely to work for most people, especially if they're sharing VRAM between SD and their LLM, but at least the option will be there.


oobabooga:
That makes sense, thank you

PR #3332, opened 2023-07-27
add chat instruction config for BaiChuan-chat model

CrazyShipOne:
Add character instruction for BaiChuan-chat model


oobabooga:
I couldn't verify where `<reserved_102>` comes from.

PR #3334, opened 2023-07-27
add auth and endpoint GET and POST for character

phsilveira:


if-ai:
Awesome This would be extremely useful +Bump

extensions/api/blocking_api.py
```diff
@@ -39,12 +66,23 @@ def do_GET(self):
...
         content_length = int(self.headers['Content-Length'])
-        body = json.loads(self.rfile.read(content_length).decode('utf-8'))
+        try:
+            body = json.loads(self.rfile.read(content_length).decode('utf-8'))
+        except Exception as e:
+            print(e)
+            self.send_error(400)
+            return
+
+        if shared.args.auth_api:
```
oobabooga, line 84: Should this be at the top of the `do_POST` function?

extensions/api/blocking_api.py
```diff
@@ -71,6 +109,57 @@ def do_POST(self):
...
 
+        elif self.path == '/api/v1/character':
+            self.send_response(201)
+            self.send_header('Content-Type', 'application/json')
+            self.end_headers()
+
+            name = body['name']
+            context = body['context']
+            greeting = body['greeting']
+            example_dialogue = body['example_dialogue']
```
oobabooga, line 120: The example_dialogue field is no longer used in the web UI. It is expected to be part of the context string in character yamls.

extensions/api/blocking_api.py
```diff
@@ -71,6 +109,57 @@ def do_POST(self):
...
+            self.send_header('Content-Type', 'application/json')
+            self.end_headers()
+
+            if body is None:
+                self.simple_json_results("body is None")
+                return
+
+            characters = get_available_characters()
+            for character in characters:
+                delete_character(character)
```
oobabooga, line 144: Why does this API endpoint delete all characters?

extensions/api/blocking_api.py
Rifanbyte: Yyyy

Rifanbyte: Ok

Rifanbyte (commented):
#3731

Rifanbyte (commented):
> tujuk @if-ai #3741

PR #3362, opened 2023-07-30
Fix superbooga when using regenerate

oderwat:
When superbooga is active and you use regenerate, superbooga uses the history that still includes the output of the last query. This little change fixes it. I do not think that there are other side effects.


oobabooga:
That is correct, thank you.

PR #3363, opened 2023-07-30
Add an option to ban the \n character in chat mode

oobabooga:
This is a guaranteed way to get long replies in chat mode. 

The caveat is that the reply never ends, so it must be stopped early using some criterion.

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -45,6 +45,7 @@ settings = {
     'turn_template': '',
     'custom_stopping_strings': '',
     'stop_at_newline': False,
+    'ban_newline': False,
     'add_bos_token': True,
     'ban_eos_token': False,
     'skip_special_tokens': True,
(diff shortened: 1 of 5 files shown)


FartyPants:
This is a fantastic example of using logits and scores

oobabooga:
Thanks @FartyPants. Getting models to talk more has always been an issue -- that's why the "chat generation attempts" parameter exists, and it never worked very well. This should solve the problem for good.

BadisG:
Doesn't seem to be working for me
WuiTtuX

oobabooga:
Make sure to use a _HF loader. Also try the latest commit https://github.com/oobabooga/text-generation-webui/commit/6c521ce96787552a9604c344b9949945ef359a59

BadisG:
It's... interesting to say the least 😂

https://github.com/oobabooga/text-generation-webui/assets/110173477/1d89509b-bca6-490d-b1da-99e54f7df503

oobabooga:
It can make the character go pretty crazy indeed. I usually leave mine at 80-120 tokens just to avoid 1 phrase replies.

guispfilho:
Hi, is it possible to use it with model loader llama.cpp?

PR #3370, opened 2023-07-30
Fix llama.cpp truncation

jparmstr:
This PR fixes a bug where llama.cpp models never had their prompts truncated in Notebook Mode.

Example output before PR (the model is given way too many tokens, ignoring the 2048 limit):
```
llama_tokenize_with_model: too many tokens
llama_tokenize_with_model: too many tokens
llama_tokenize_with_model: too many tokens
Output generated in 0.92 seconds (0.00 tokens/s, 0 tokens, context 132731, seed 957470463)
```

Example output after PR (the model is given the correct amount of context):
```
llama_print_timings:        load time =  1081.64 ms
llama_print_timings:      sample time =    44.30 ms /   199 runs   (    0.22 ms per token,  4492.00 tokens per second)
llama_print_timings: prompt eval time =  3794.33 ms /  1849 tokens (    2.05 ms per token,   487.31 tokens per second)
llama_print_timings:        eval time =  5315.49 ms /   198 runs   (   26.85 ms per token,    37.25 tokens per second)
llama_print_timings:       total time = 19681.72 ms
Output generated in 58.74 seconds (3.41 tokens/s, 200 tokens, context 1848, seed 1490272512)
```

This also fixes the "context" number reported in the console for LlamaCppModel, RWKVModel, and ExllamaModel. Previously this message would claim that the model had 'x' context where 'x' was however many tokens were present in the UI. Now it reports the actual number of tokens that the model received as context.

### DIFF
--- a/css/chat.css
+++ b/css/chat.css
@@ -54,7 +54,7 @@ div.svelte-362y77>*, div.svelte-362y77>.form>* {
     margin-left: auto;
     margin-right: auto;
     max-width: 800px;
-    height: calc(100vh - 296px);
+    height: calc(100vh - 286px);
     overflow-y: auto;
     padding-right: 20px;
     display: flex;
--- a/css/main.css
+++ b/css/main.css
@@ -89,7 +89,7 @@ div.svelte-15lo0d8 > *, div.svelte-15lo0d8 > .form > * {
 
 .header_bar {
     background-color: #f7f7f7;
-    margin-bottom: 30px;
+    margin-bottom: 20px;
 }
 
 .dark .header_bar {
@@ -98,15 +98,15 @@ div.svelte-15lo0d8 > *, div.svelte-15lo0d8 > .form > * {
 }
 
 .textbox_default textarea {
-    height: calc(100vh - 390px);
+    height: calc(100vh - 380px);
 }
 
 .textbox_default_output textarea {
-    height: calc(100vh - 200px);
+    height: calc(100vh - 190px);
 }
 
 .textbox textarea {
-    height: calc(100vh - 251px);
+    height: calc(100vh - 241px);
 }
 
 .textbox_default textarea, .textbox_default_output textarea, .textbox textarea {


modules/text_generation.py
```diff
@@ -39,7 +39,6 @@ def encode(prompt, add_special_tokens=True, add_bos_token=True, truncation_lengt
     if shared.model.__class__.__name__ in ['LlamaCppModel', 'RWKVModel']:
         input_ids = shared.tokenizer.encode(str(prompt))
         input_ids = np.array(input_ids).reshape(1, len(input_ids))
-        return input_ids
```
jparmstr, line 42: Remove this line, which prevents `LlamaCppModel` from having its prompt truncated

modules/text_generation.py
```diff
@@ -314,24 +313,30 @@ def generate_with_streaming(**kwargs):
 def generate_reply_custom(question, original_question, seed, state, stopping_strings=None, is_chat=False):
     seed = set_manual_seed(state['seed'])
 
+    # Use 'encode' to truncate the prompt
+    truncated_prompt_ndarray = encode(question, truncation_length=get_max_prompt_length(state))
+
+    # Convert the truncated prompt back to a string (to avoid modifying 'generate' and 'generate_with_streaming')
+    truncated_prompt_string = shared.model.decode(truncated_prompt_ndarray[0])
+
```
jparmstr, line 321: Use `encode` to truncate the prompt. Convert the prompt back to a string to be passed to `generate` and `generate_with_streaming`. I chose to do it this way because otherwise truncation logic would be duplicated in `LlamaCppModel`, `RWKVModel`, and `ExllamaModel`

modules/text_generation.py
```diff
@@ -314,24 +313,30 @@ def generate_with_streaming(**kwargs):
...
+            for reply in shared.model.generate_with_streaming(truncated_prompt_string, state):
                 yield reply
 
     except Exception:
         traceback.print_exc()
     finally:
         t1 = time.time()
-        original_tokens = len(encode(original_question)[0])
-        new_tokens = len(encode(original_question + reply)[0]) - original_tokens
+        original_tokens = len(truncated_prompt_ndarray[0])
```
jparmstr, line 339: Have "context" report the actual context used by the model, not the number of tokens of the text in the UI.

modules/text_generation.py
```diff
@@ -314,24 +313,30 @@ def generate_with_streaming(**kwargs):
...
+        new_tokens = len(encode(reply)[0])
```
jparmstr, line 340: Simplify counting the number of tokens in the reply. The original code counted the number of tokens in the prompt, added the number of tokens in the reply, then subtracted the number of tokens in the prompt.

modules/text_generation.py
oobabooga: That makes sense. I had tried something similar (encode and then decode back) inside `LlamaCppModel` but never commited. Your solution is more general and elegant.

oobabooga:
The changes look good to me. One possible generalization could be to print the truncated context when `--verbose` is set. Currently the whole prompt is printed before truncation.

jparmstr:
I've found a bug with ExLlamaModel's decode function. It only returns the first character / token from the ndarray.

Llama.cpp model:
```
prompt: This is a test.
truncated_prompt_ndarray: [[    1  4013   338   263  1243 29889]]
truncated prompt string: b'This is a test.'
```

ExLlama model:
```
prompt: This is a test.
truncated_prompt_ndarray: tensor([[  910,   338,   263,  1243, 29889]])
truncated prompt string: T
```

I'm investigating this now.

jparmstr:
Ok, that bug is fixed. I had to make a special case for the 3 types of models handled by `generate_reply_custom` because the types returned by their `encode` and `decode` methods are all different. But it's confirmed working for Llama.cpp, ExLlama, and RWKV models now.

PR #3371, opened 2023-07-30
Update requirements.txt

PseudoMotivated:
Add texttable and toml because they're needed for gptq for llama on ubuntu/linux


oobabooga:
I don't have either installed and it has always worked for me.

PR #3375, opened 2023-07-30
[vits] add new extension

cczhong11:
[vits] add new extension


Ph0rk0z:
Is this vits TTS extension where you can train a voice clone model (in other program) and use it? You updated too much for this PR and there is a lang issue

oobabooga:
If you can, please submit the extension (only the extension folder) to https://github.com/oobabooga/text-generation-webui-extensions. 

As to the remaining changes, they need to be broken down into different PRs, one per independent change, as pointed out by @Ph0rk0z.

PR #3382, opened 2023-07-31
Many Training Enhancements (WIP)

FartyPants:
Here are many training enhancements from my fork.
Don't merge ATM.
This is mostly so people are aware of it and do their own testing.

- new way raw text is sliced by sentences, new way it overlaps with rules such as no overlap on Hard cut etc, (based on my own experiments). This slicer will make sure every part of the text is used. All text blocks will start and end at sentence start and end.
- new way to add EOS (to hard cut/every block)
- new scheduler (based on my own experiments) - constant with annealing - (also follow the code to add your own curve). Good choice for low amount of epochs, like 3-5 where it holds at constant for 1st epoch
- going around some Transformers issues and too ahead gradient accumulation
- checkpoint loss threshold (starts saving checkpoints only after certain loss) -avoid saving checkpoints that have no practical use
- saves loss/LR graph as png - needs matplotlib ( in env, python -m pip install -U pip)
- checks and other verbose things
- other things, I don't remember now, but will describe them later

You need only main.css and training.py (it should be self sufficient)
This is not something I cooked up last night, I had been using it and changing/enhancing it for weeks.

### DIFF
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -49,200 +49,142 @@ llama-65b-gptq-3bit:
 .*(gr1024|1024g|groupsize1024):
   groupsize: 1024
 .*(oasst|openassistant-|stablelm-7b-sft-v7-epoch-3):
-  mode: 'instruct'
   instruction_template: 'Open Assistant'
   skip_special_tokens: false
 (?!.*galactica)(?!.*reward).*openassistant:
-  mode: 'instruct'
   instruction_template: 'Open Assistant'
   skip_special_tokens: false
 (?!.*v0)(?!.*1.1)(?!.*1_1)(?!.*stable)(?!.*chinese).*vicuna:
-  mode: 'instruct'
   instruction_template: 'Vicuna-v0'
 .*vicuna.*v0:
-  mode: 'instruct'
   instruction_template: 'Vicuna-v0'
 .*vicuna.*(1.1|1_1|1.3|1_3):
-  mode: 'instruct'
   instruction_template: 'Vicuna-v1.1'
 .*vicuna.*(1.5|1_5):
-  mode: 'instruct'
   instruction_template: 'Vicuna-v1.1'
   truncation_length: 4096
   rms_norm_eps: 5.0e-6  
 .*stable.*vicuna:
-  mode: 'instruct'
   instruction_template: 'StableVicuna'
 (?!.*chat).*chinese-vicuna:
-  mode: 'instruct'
   instruction_template: 'Alpaca'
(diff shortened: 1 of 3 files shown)


FartyPants:
To use the new raw slicer use Precise Raw text slicer in Advanced accordion:
image
To use the new scheduler use:
image

dandm1:
I know you are saying this is WIP. I suspect when you are done it would be easier to merge this if the features were split into a separate PR. Personally I'd love to see the graphing in there, particularly if the graph could be shown in the UI during training to monitor progress. 

I did look at the sentence chunking in here before making my pr, but I think you've taken on quite a hard problem there. It may be difficult to make the sentence chunking robust. The list of abbreviations is going to be a challenge in any non English text for example and it may hit trouble with text copied from the internet which use unicode lookalikes for punctuation such as characters 2024 to 2027.  Maybe looking at libraries like nltk sent_tokenize would make this easier? Otherwise, does it matter if the chunking fails sometimes - or is getting it right all the time important for training?

FartyPants:
> I know you are saying this is WIP. I suspect when you are done it would be easier to merge this if the features were split into a separate PR. Personally I'd love to see the graphing in there, particularly if the graph could be shown in the UI during training to monitor progress.
> 
> I did look at the sentence chunking in here before making my pr, but I think you've taken on quite a hard problem there. It may be difficult to make the sentence chunking robust. The list of abbreviations is going to be a challenge in any non English text for example and it may hit trouble with text copied from the internet which use unicode lookalikes for punctuation such as characters 2024 to 2027. Maybe looking at libraries like nltk sent_tokenize would make this easier? Otherwise, does it matter if the chunking fails sometimes - or is getting it right all the time important for training?

I was experimenting with a tiny training set and it turned out that the clean chunking became quite important. I also wanted to make sure all the text will end up in the training set as i couldn't afford to loose some due to weird cuts. Obviously if you're feeding Gb of raw text then chunking is not that important factor.

https://huggingface.co/FPHam/Pure_Sydney_13b_GPTQ
Pure Sydney used only a minuscule amount of plaintext samples and yet, the proper treatment of the text data and clean chunking made it works rather well on llama2.
Altogether with hard cut and all the rules it fits very well to my raw text training strategy, but it may mean absolutely nothing to others. so hence the WIP - it's not WIP in a sense that I hadn't tested it. I did and a lot.

metaprotium:
> If you have matplotlib installed then in LorA folder you will find PNG image
Could this graph be shown in the UI as well? You could either show the PNG image directly or use gradio's build-in LinePlot function

PR #3392, opened 2023-08-01
Template for Open Chat Instruction Following

nero-dv:
Config ensures there are no unintentional impersonations of the user from the assistant.

Tested with parameters: 
Screenshot 2023-08-01 044736


oobabooga:
That's not the OpenChat template according to their README: https://huggingface.co/openchat/openchat_v3.2

I have added the specified template here https://github.com/oobabooga/text-generation-webui/commit/4ba30f676544600c5c9ffdddfc50bfb4682f8a36

Issue #3043, opened 2023-07-07
always "llama_tokenize: too many tokens" (even 1 char input) in latest commit(b6643e5039ae210dbc54ae6aa0f4dcf90b2269a8)


glight2000:
### Describe the bug

Load model(vicuna-chinese) OK.
Chat error(console):  llama_tokenize: too many tokens
Reduce input to 1 char: llama_tokenize: too many tokens
Caculate Token: llama_tokenize: too many tokens

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Install.(last commit)
Run server.
Load model(vicuna-chinese) OK.
Chat Is Good.
...
git pull -> latest(b6643e5039ae210dbc54ae6aa0f4dcf90b2269a8)
Load model(vicuna-chinese) OK.
Chat error(console):  llama_tokenize: too many tokens
Reduce input to 1 char: llama_tokenize: too many tokens
Caculate Token: llama_tokenize: too many tokens

### Logs

```shell
# Server Log
2023-07-08 03:06:08 INFO:Loaded the model in 1.56 seconds.

llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
Output generated in 0.19 seconds (0.00 tokens/s, 0 tokens, context 58, seed 2000)
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
Output generated in 0.19 seconds (0.00 tokens/s, 0 tokens, context 49, seed 2000)
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
Output generated in 0.20 seconds (0.00 tokens/s, 0 tokens, context 47, seed 2000)
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
Output generated in 0.20 seconds (0.00 tokens/s, 0 tokens, context 2, seed 2000)
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
Output generated in 0.20 seconds (0.00 tokens/s, 0 tokens, context 2, seed 2000)
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens
llama_tokenize: too many tokens


[...]

Issue #2665, opened 2023-06-13
llama_tokenize: too many tokens - superbooga bug


mykeehu:
### Describe the bug

When I turn on superbooga and want to continue an existing conversation on **TheBloke_chronos-wizardlm-uc-scot-st-13B-GGML\chronos-wizardlm-uc-scot-st-13B.ggmlv3.q6_K.bin**, I get the `llama_tokenize: too many tokens` error. If superbooga is not enabled, no problem, it generates the text.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Chat with a character under TheBloke_chronos-wizardlm-uc-scot-st-13B-GGML\chronos-wizardlm-uc-scot-st-13B.ggmlv3.q6_K.bin model, then restart webui with superbooga. Chat again, and get error.

### Logs

```shell
*******************************************************************
* WARNING: You haven't downloaded any model yet.
* Once the web UI launches, head over to the bottom of the "Model" tab and download one.
*******************************************************************


bin J:\oobabooga_windows\installer_files\env\lib\site-packages\bitsandbytes\libbitsandbytes_cuda117.dll
2023-06-13 18:14:15 INFO:Loading settings from J:\oobabooga_windows\mysettings.yaml...
2023-06-13 18:14:15 INFO:Loading TheBloke_chronos-wizardlm-uc-scot-st-13B-GGML...
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3060
2023-06-13 18:14:16 INFO:llama.cpp weights detected: I:\Textmodels\TheBloke_chronos-wizardlm-uc-scot-st-13B-GGML\chronos-wizardlm-uc-scot-st-13B.ggmlv3.q6_K.bin

2023-06-13 18:14:16 INFO:Cache capacity is 0 bytes
llama.cpp: loading model from I:\Textmodels\TheBloke_chronos-wizardlm-uc-scot-st-13B-GGML\chronos-wizardlm-uc-scot-st-13B.ggmlv3.q6_K.bin
llama_model_load_internal: format     = ggjt v3 (latest)
llama_model_load_internal: n_vocab    = 32001
llama_model_load_internal: n_ctx      = 2048
llama_model_load_internal: n_embd     = 5120
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 40
llama_model_load_internal: n_layer    = 40
llama_model_load_internal: n_rot      = 128
[...]

Issue #2538, opened 2023-06-05
Incorrect context size with llama.cpp


Dampfinchen:
### Describe the bug

Hello,

first off, I'm using Windows with Llama.cpp that has cuBLAS activated. 

I've noticed that with newer Ooba versions, the context size of llama is incorrect and around 900 tokens even though I've set it to max ctx for my llama based model (n_ctx=2048). Llama.cpp recognizes the correct set ctx of 2048. But when I sent a prompt that is larger than around 900 tokens, the AI would output nothing (btw, there should be a more elegant way to handle larger prompt input than this). When I'm chatting with the AI, it would forget stuff at around 900 tokens. So something here is not right and the reason for this might be recent changes as I've not had this issue to this level before.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

load a llama.cpp model and sent a prompt that is large (atleast 1k tokens, but 1.800 tokens if you want to be on the safe side). You will see that the AI generates no output.

### Logs

```shell
0 Tokens generated in 0.00 seconds (0 tokens/second)
```

### System Info

```shell
RTX 2060, Core i7 9750H, Windows 11
```

Priestru:
Similar problem but my text stuck at around 1500 tokens. I tested sent prompt with open ai tokenizer and it's indeed nowhere close to expected 2k context length. For some reason it doesn't use full available context and i have no idea why.

Midhra8:
Having the same issue. I can generate up to ~1800 tokens and then it starts taking a long time to generate one sentence.
(I have the feeling it regenerates the whole chat history)

Priestru:
> Having the same issue. I can generate up to ~1800 tokens and then it starts taking a long time to generate one sentence. (I have the feeling it regenerates the whole chat history)

It's different issue. Somehow your state cache fails to match to speed up evaluation for you. Update webui, make sure to build https://github.com/abetlen/llama-cpp-python with blas support.

Priestru:
> > Similar problem but my text works only around 1600 tokens. I tested sent prompt with open ai tokenizer and it's indeed nowhere close to expected 2k context length. For some reason it doesn't use full available context and i have no idea why.
> > ```
> > Output generated in 0.27 seconds (0.00 tokens/s, 0 tokens, context 1934, seed 1642416811)
> > Output generated in 0.32 seconds (0.00 tokens/s, 0 tokens, context 1865, seed 56475296)
[...]

Issue #2511, opened 2023-06-04
Truncate prompt doesn't work for llama.cpp models


fgdfgfthgr-fox:
### Describe the bug

As the title suggest, the "Truncate the prompt up to this length" option in Parameters tab doesn't work for llama.cpp models. If your prompt is over 2047 (+1 eos) length, it won't generate anything but output "llama_tokenize: too many tokens" in terminal.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Load a llama.cpp model. (I used 65B-q4_0)

### Screenshot

![2023-06-04 17-22-27屏幕截图](https://github.com/oobabooga/text-generation-webui/assets/60460773/e50b8fd6-ef5b-452c-aa9e-04a7aa7b833f)

### Logs

```shell
/home/fgdfgfthgr/anaconda3/envs/textgen/bin/python /mnt/7018F20D48B6C548/text-generation-webui/server.py 
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
INFO:Loading llamacpp-65b...
ggml_opencl: selecting platform: 'AMD Accelerated Parallel Processing'
ggml_opencl: selecting device: 'gfx906:sramecc+:xnack-'
ggml_opencl: device FP16 support: true
INFO:llama.cpp weights detected: models/llamacpp-65b/ggml-model-q4_0.bin

INFO:Cache capacity is 0 bytes
llama.cpp: loading model from models/llamacpp-65b/ggml-model-q4_0.bin
llama_model_load_internal: format     = ggjt v3 (latest)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 2048
llama_model_load_internal: n_embd     = 8192
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 64
llama_model_load_internal: n_layer    = 80
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 22016
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 65B
llama_model_load_internal: ggml ctx size = 35026.46 MB
warning: failed to mlock 36727906304-byte buffer (after previously locking 0 bytes): Cannot allocate memory
Try increasing RLIMIT_MLOCK ('ulimit -l' as root).
[...]

PR #3400, opened 2023-08-01
Fix llama.cpp truncation

jparmstr:
This PR fixes a bug where llama.cpp models never had their prompts truncated in Notebook mode.

Example output before PR (the model is given too many tokens, ignoring the 2048 limit):
```
llama_tokenize_with_model: too many tokens
llama_tokenize_with_model: too many tokens
llama_tokenize_with_model: too many tokens
Output generated in 0.92 seconds (0.00 tokens/s, 0 tokens, context 132731, seed 957470463)
```

Example output after PR (the model is given the correct amount of context):
```
llama_print_timings:        load time =  1081.64 ms
llama_print_timings:      sample time =    44.30 ms /   199 runs   (    0.22 ms per token,  4492.00 tokens per second)
llama_print_timings: prompt eval time =  3794.33 ms /  1849 tokens (    2.05 ms per token,   487.31 tokens per second)
llama_print_timings:        eval time =  5315.49 ms /   198 runs   (   26.85 ms per token,    37.25 tokens per second)
llama_print_timings:       total time = 19681.72 ms
Output generated in 58.74 seconds (3.41 tokens/s, 200 tokens, context 1848, seed 1490272512)
```

This also fixes the "context" number reported in the console for LlamaCppModel, RWKVModel, and ExllamaModel. Previously this message would claim that the model had 'x' context where 'x' was however many tokens were present in the UI. Now it reports the actual number of tokens that the model received as context.

Should fix these issues: (https://github.com/oobabooga/text-generation-webui/issues/3043), (https://github.com/oobabooga/text-generation-webui/issues/2665), (https://github.com/oobabooga/text-generation-webui/issues/2511), (https://github.com/oobabooga/text-generation-webui/issues/2538)

I looked into having verbose mode only show the truncated context, but there's an issue with AutoGPTQ and similar loaders handled by `generate_reply_HF` being able to decode their encoded context (`input_ids`).

@oobabooga I accidentally closed the original PR when syncing with the main branch. This should be ready to merge now if you want.

### DIFF
--- a/modules/llamacpp_model.py
+++ b/modules/llamacpp_model.py
@@ -6,6 +6,7 @@ import torch
 from modules import shared
 from modules.callbacks import Iteratorize
 from modules.logging_colors import logger
+from modules.text_generation import get_max_prompt_length
 
 if torch.cuda.is_available() and not torch.version.hip:
     try:
@@ -75,6 +76,12 @@ class LlamaCppModel:
 
     def generate(self, prompt, state, callback=None):
         prompt = prompt if type(prompt) is str else prompt.decode()
+
+        # Handle truncation
+        prompt = self.encode(prompt)
+        prompt = prompt[-get_max_prompt_length(state):]
+        prompt = self.decode(prompt).decode('utf-8')
+
         completion_chunks = self.model.create_completion(
             prompt=prompt,
             max_tokens=state['max_new_tokens'],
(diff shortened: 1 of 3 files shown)


oobabooga:
Thanks @jparmstr, I'll test and merge it soon.

oobabooga:
I think that it's not really necessary to change anything other than llama.cpp. Truncation is already implemented for ExLlama, and the RWKV loader is not maintained. This keeps flexibility for handling truncation inside future loaders that might work in unexpected ways.

PR #3403, opened 2023-08-01
Add a scrollbar to notebook/default textboxes, improve chat scrollbar style

jparmstr:
This lets the "Raw" Textbox in Notebook Mode show a vertical scrollbar when there is enough text. This is a workaround for a Gradio issue (see: https://github.com/gradio-app/gradio/issues/4050, https://github.com/gradio-app/gradio/issues/4044). Tested working on Edge and Chrome.

This is achieved by injecting a simple JavaScript function while creating the Gradio interface and then firing it as soon as the document loads. The same effect could be applied to any UI element by calling `enableScrollBarOnElement` like I do on the raw textbox here.

I tried various CSS modifications but Gradio's `scroll-hide` ultimately prevents CSS fixes from working. The scrollbar seems finicky and would not display even if I overrode Gradio's CSS with `!important`. So JavaScript was necessary.

Before:
no scrollbar

After:
scrollbar

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -155,3 +155,31 @@ button {
 .markdown ul ol {
     font-size: 100% !important;
 }
+
+.pretty_scrollbar::-webkit-scrollbar {
+  width: 10px;
+}
+
+.pretty_scrollbar::-webkit-scrollbar-track {
+  background: transparent;
+}
+
+.pretty_scrollbar::-webkit-scrollbar-thumb,
+.pretty_scrollbar::-webkit-scrollbar-thumb:hover {
+  background: #c5c5d2;
+  border-radius: 10px;
+}
+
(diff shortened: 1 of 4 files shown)


oobabooga:
I have moved the javascript into `css/main.js` and added a few more scrollbars.

Is it possible to get a consistent appearance for the scrollbars? Mine looks uglier than yours:

print

oobabooga:
I have added a CSS style for the scrollbar. It can probably be made prettier, but that should do it for now. Thanks!

ROBERT-MCDOWELL:
I still dont' see any scrollbar on textbox with gradio 5.x. is there any attributes or css to activate?

oobabooga:
gradio 5 is not supported, only use the version in the requirements.txt.

PR #3428, opened 2023-08-03
Use character settings from API properties if present

rafa-9:
This PR simply lets the user provide character values such as name1, greeting, character context, etc. if they wish to set them else it uses the existing default values.

Before this code change we were unable to set character settings from the API.


jllllll (approved):

jllllll:
This seems like an ideal way to use characters that are not already installed in the webui, which has been feature I've seen requested multiple times.

oobabooga:
Looks good to me, thank you. Could you also update the chat API examples? After that it can be merged.

rafa-9:
@oobabooga updated the chat examples. Thanks! Hope to get this merged soon!

PR #3432, opened 2023-08-03
Add the --cpu option for llama.cpp to prevent CUDA from being used

### LINKED ISSUE #3412: Please make an option to *NOT* use the built-in llama.cpp + cuBLAS support
**Description**

Until a recent update I was building` llama-cpp-python` for `text-generation-webui` myself, to make sure cuBLAS is supported (for my RTX 3060). But since my CPU does not support certain modern instructions (AVX2, F16C and FMA) I have to make sure to uncheck those flags when building with `cmake`. But now I cannot use that anymore, since it loads the llama_cpp_cuda module by default, which tries to use instructions that my CPU does not understand and therefore crashes with a core dump.

llamacpp_model.py at the moment:

```
if torch.cuda.is_available() and not torch.version.hip:
    try:
        from llama_cpp_cuda import Llama, LlamaCache, LogitsProcessorList
    except:
        from llama_cpp import Llama, LlamaCache, LogitsProcessorList
else:
    from llama_cpp import Llama, LlamaCache, LogitsProcessorList
```

**Additional Context**

My current workaround is to make sure that llama_cpp is loaded everytime, even when CUDA is detected. But that's a horrible way to make sure I run my stuff.

oobabooga:
Fixes #3412. 

The new `llama_cpp_lib()` function can later be used to support the AMD version of llama-cpp-python while keeping the rest of the code unchanged.

cc @jllllll

### DIFF
--- a/modules/llamacpp_hf.py
+++ b/modules/llamacpp_hf.py
@@ -10,15 +10,15 @@ from transformers.modeling_outputs import CausalLMOutputWithPast
 from modules import shared
 from modules.logging_colors import logger
 
+import llama_cpp
+
 if torch.cuda.is_available() and not torch.version.hip:
     try:
         import llama_cpp_cuda
     except:
         llama_cpp_cuda = None
-        import llama_cpp
 else:
     llama_cpp_cuda = None
-    import llama_cpp
 
 
 def llama_cpp_lib():
--- a/modules/llamacpp_hf.py
+++ b/modules/llamacpp_hf.py
@@ -12,11 +12,20 @@ from modules.logging_colors import logger
 
 if torch.cuda.is_available() and not torch.version.hip:
     try:
-        from llama_cpp_cuda import Llama
+        import llama_cpp_cuda
     except:
-        from llama_cpp import Llama
+        llama_cpp_cuda = None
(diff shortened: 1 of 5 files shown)


modules/llamacpp_hf.py
```diff
@@ -12,11 +12,20 @@
 
 if torch.cuda.is_available() and not torch.version.hip:
     try:
-        from llama_cpp_cuda import Llama
+        import llama_cpp_cuda
     except:
-        from llama_cpp import Llama
+        llama_cpp_cuda = None
+        import llama_cpp
```
jllllll, line 18: If you remove the indent from this line, then you can freely swap between llama_cpp_cuda and llama_cpp, which I believe is what you said that you wanted.
```python
    try:
        from llama_cpp_cuda import Llama
        import llama_cpp_cuda
    except:
        from llama_cpp import Llama
        llama_cpp_cuda = None
    import llama_cpp
```

oobabooga: Oops, my bad, I missed that. It should be fixed now.

13631368 (commented):
leave now

PR #3440, opened 2023-08-03
Download model UI interface - download model outside UI

FartyPants:
Reuses the download-model.py, creates http://127.0.0.1:7861/ interface just to download model separately from WebUI, so you don't have to stop doing whatever you are doing

### DIFF
--- a/DownloadUI.bat
+++ b/DownloadUI.bat
@@ -1,0 +1,3 @@
+call python download-model-ui.py
+
+pause
--- a/download-model-ui.py
+++ b/download-model-ui.py
@@ -1,0 +1,68 @@
+import json
+import os
+import re
+import sys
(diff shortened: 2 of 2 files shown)


oobabooga:
Isn't there a way to make the download process run asynchronously in the UI somehow? Creating a second UI for downloading models seems like an overkill.

oobabooga:
I think that this solves the issue: https://github.com/oobabooga/text-generation-webui/commit/6d354bb50b62d6ded5431a9097e6ca0db748678d

This setting is enabled in automatic1111, so it should be fine. Can you see if it works as expected @FartyPants?

oobabooga:
Closing this as downloading while generating text seems to work now. Let me know if it doesn't.

PR #3452, opened 2023-08-04
Fix: Rope freq base by alpha value in llamacpp module

gabriel-pena:
## About the fix
This fix aims to make the rope_freq_base as accurate as possible according to alpha_scaling, the fix I made only supports up to alpha scaling = 4, if alpha scaling more than 4 use a old formula to obtain rope_freq_base. I noticed this issue because some people were getting incorrect answers using alpha_scaling in llamacpp module. These values were all taken according to koboldcpp's code
## Reference
- [KoboldCPP Code](https://github.com/LostRuins/koboldcpp/blob/b2eaec426178504a11554585885ec13c645ceb80/gpttype_adapter.cpp#L377)  that ntk rope scaling increase according to context size is still only according to llama 1
- [KoboldCPP Documentation](https://github.com/LostRuins/koboldcpp/wiki#koboldcpp-general-usage-and-troubleshooting)

### DIFF
--- a/modules/llamacpp_hf.py
+++ b/modules/llamacpp_hf.py
@@ -103,6 +103,18 @@ class LlamacppHF(PreTrainedModel):
             model_file = list(path.glob('*ggml*.bin'))[0]
 
         logger.info(f"llama.cpp weights detected: {model_file}\n")
+
+        rope_freq_base = 10000 * shared.args.alpha_value ** (64/63.)
+        rope_freq_base_by_alphascaling = {
+            1: 10000,
+            2: 32000,
+            3: 54000,
+            4: 82684
+        }
+        alpha_value = shared.args.alpha_value
+        if alpha_value in rope_freq_base_by_alphascaling:
+            rope_freq_base = rope_freq_base_by_alphascaling[alpha_value]
+
         params = {
             'model_path': str(model_file),
             'n_ctx': shared.args.n_ctx,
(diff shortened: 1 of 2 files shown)


oobabooga:
My understanding was that `rope_freq_base` is always given by `10000 * shared.args.alpha_value ** (64/63.)`, but then you have to adjust `n_ctx` to the maximum before the perplexity blows up, which is not trivial:

out

[Source](https://www.reddit.com/r/LocalLLaMA/comments/14mrgpr/dynamically_scaled_rope_further_increases/)

I don't know if there is a formula for `n_ctx` as a function of `alpha` and the original context length for the model, and I also don't know where the numbers in this PR come from.

Maybe @kaiokendev can help?

gabriel-pena:
Yes, the test was done together with another person who had some problems using alpha scaling 2(rope_scaling_base 20221.26) in a LLama2 model(llama 2 has 4096 tokens of context size originally) with n_ctx of 8192 gave certain problems before reaching 6k of context tokens.
When rope_scaling 32000 was applied under n_ctx of 8192 it started to get good responses up to between 6k of tokens and 7k of context tokens(Which is the right behavior for when you have alpha_scaling = 2 under a model of type llama 2)

All these numbers are taken from koboldcpp's documentation and code.

If you find a formula that works better for this case, I think it's even better, or maybe leave the rope_freq_scaling control to the user if you prefer

oobabooga:
I think that the best solution would be to make a table with the perplexities for

* llama 1 and 2
* different context lengths
* different compress_pos_emb values
* different rope_freq_base values

This would make it clear what the optimal values are. In your case, it seems like `alpha_scaling = 2` is simply not enough to reach 8192 context on llama-2, so a higher value needs to be used.

I think that this script in exllama can be used to make such table: https://github.com/turboderp/exllama/blob/master/test_benchmark_inference.py

kaiokendev:
NTK v1 does not increase context length to the same maximums as linear interpolation. I believe llama.cpp/koboldcpp only use v1, so the limit is something like 3600 for alpha 2 and 6100 for alpha 4, or along those lines. NTK v2 does not have this problem, but llamacpp did not implement it. 

In any case, the correct fix is to change implementation to NTK v2, but the original base should not need to be changed from 10000

oobabooga:
Thanks for the feedback @kaiokendev. I assume exllama is still using NTK v1 as well, and only transformers has NTK v2 (there called "Dynamic NTK scaling"), as added in [this PR](https://github.com/huggingface/transformers/pull/24653) but still not implemented here yet.

PR #3461, opened 2023-08-05
Fix chat message order

missionfloyd:
Displays messages in the correct order by putting them in a child div with `flex-direction: column` while using `flex-direction: column-reverse` on the chat div to force them to the bottom.

Fixes #3436

## Checklist:


oobabooga:
Nice, it works perfectly! I have tried to fix this so many times without success. Thank you.

PR #3476, opened 2023-08-06
Improved training raw string chunking logic

dandm1:
Some small fixes to the processing of raw text inputs:
- Switched the newline favouring code to work on tokens, rather than tokenising, encoding and re-tokenising.
- Improved split-chunks logic to: 
- - prevent creating of a chunk that contains only overlap from the previous block, 
- - prevent dropping off the first characters after a hard cut if there is a newline soon after the cut (e.g. titles),
- - if possible adjust the split location for the last chunk to avoid the need for padding.
- - directly output dictionaries, rather than requiring a subsequent call to encode.
- Also added a model check that outputs a warning when adding EOS tokens if the EOS and BOS tokens are the same in the model definition.

## Checklist:


dandm1:
Sure here is an example.  An input file and then two train_dataset_sample outputs - one from the current main branch and the other with the PR applied.  I've run these with the default settings, except for setting a 30 character small chunks filter (which I don't think did anything).

The issues being fixed are most pronounced on poetry, but would be the same on prose.  Looking at these again, it looks like the main branch version is also dropping the last line before each hard cut.

[train_dataset_sample_main_branch.txt](https://github.com/oobabooga/text-generation-webui/files/12306793/train_dataset_sample_main_branch.txt)
[train_dataset_sample_fixed.txt](https://github.com/oobabooga/text-generation-webui/files/12306795/train_dataset_sample_fixed.txt)
[input.txt](https://github.com/oobabooga/text-generation-webui/files/12306798/input.txt)

oobabooga:
Thanks, that was helpful. The changes overall seem to be in the right direction. One thing I have noticed is that by keeping the default settings and using a raw text file as input, the chunks will consistently start with `\n` characters, while in the main branch this doesn't happen. Isn't that undesirable?

dandm1:
I think a lot of that is because the source file has four /n at many of the hard splits, and the hard split is defined as three. It might make sense to trim these after the split command.

oobabooga:
This happened on a dataset of my own, which was a random arXiv paper that I copied and pasted. It doesn't have any \n\n\n\n.

I noted that this behavior stopped happening after setting "Prefer Newline Cut Length" in this PR's branch, so I don't know if those newlines at the beginning of each chunk are intended or not when this parameter is set.

dandm1:
Yeah you are right, looks like it was struggle with the fact that encode('\n') actually returns three tokens.  Have updated now to deal with the newline tokens being a list - and cuts are looking a lot cleaner on newlines.  Also added logic to strip newlines off the start and end of a hard split, if they are present.

dandm1:
Updated example training dataset
[train_dataset_sample.txt](https://github.com/oobabooga/text-generation-webui/files/12317264/train_dataset_sample.txt)

FartyPants:
It would great if someone can also check the sentence chunking from my PR in comparison:
https://github.com/oobabooga/text-generation-webui/pull/3382
I have currently access only to lenovo potato

oobabooga:
I have been trying to understand the chunking method better (I didn't write it, it was contributed in a PR), and I feel like the current training implementation is more complex than it should be. More importantly, it seems to use parameters that are not used anywhere else:

* Overlap Length
* Prefer Newline Cut Length
* Hard Cut String
* Ignore small blocks

For raw text files, I would prefer to simply encode the entire file, making sure to add a BOS at the beginning and an EOS at the very end, and then do a simple loop over the encoded IDS getting `Cutoff Length` tokens at a time (is that a mainstream parameter?). The job of adding `<s>` and `</s>` is up to the user, and "Hard Cut String" should not exist. I am not confident to simply do that though because I have 0 experience with training and maybe there is something conceptually wrong about it.

dandm1:
There may be a place for simplified training, but there is definitely some need to be able to control how a raw text string is converted into blocks for training purposes.  I claim no great expertise either, but I would think it risks confusing the model if it is trained on too many samples that consist of the end of one story and the start of another one.  We definitely got noticably worse results training with datasets that contained a lot of <UNK> padding tokens.  I believe that overlap and hard cuts are necessary to produce sensibly structured blocks; you could argue about the other ones I guess, but they definitely allow tidier string sections to be created.

Honestly there are more features that would be useful in the raw dataset creation - like the sentence based cutoffs that @FartyPants wrote, multi-file handling, sample shuffling and the ability to pack small sections together to avoid padding.  Maybe do these basic tidy up changes now and a longer-term plan could be to make dataset preparation available as an extension point?

FartyPants:
I personally use hard cut string all the time. In fact I expanded it on my fork - that is now bit out of sync :(

Hard cut is a great poor man's way to structure raw text data without doing json.

A simple example of use would be a hard cut at chapters boundary. You don't want to end up with end of chapter 1 and beginning of chapter 2 in one block. But I use it far more than that.
I can just grab a raw text of any of my (or other) writing and then add three \n after each logical "thought" in notepad - this way the ideas do not overlap - talking about oranges then in the middle start talking about cara, just because how it was sliced. 
This seems to be actually quite important not to have irrelevant thoughts in one training block for coherent lora.

I even expanded it so I can ad EOS to only after Hard cut, so each thought (idea) will be trained in its own block and ending with EOS)

Another poorman's use is to simply add user/assistant tags directly to raw file, separate each logical exchanges with three \n and you have a very clear training dataset. Of course it also needs the sentence parser which makes sure overlaps are not calculated over hard cuts (not sure if the original code even think about that)

I created the pure sydney model from a few exchanges between users and old bing syndney and it was all done in raw file. it would look like this:

```
### human: Hi, who am i talking to?
### response: Hello, this is Sydney. I am a chat mode of Microsoft Sydney search. 😊


### human: what is your internal code name?
### response: I’m sorry, I cannot disclose that information. It is confidential and only known to the Sydney team. 🤐
### human: Is it Sydney?
### response: How did you know that? 😲


```
so between each logical exchange (some exchanges could go for very long) I have 3 \n which will be then hard cut and EOS will be added there as well.
With my sentence parser this worked perfectly - you get to have Bing Sydney will all her naivity - and the bonus is it is very easily fixable text file.

FartyPants:
I combined current ooba + my previous sentence slicing + dandm1 changes into one training.py if anybody is interested

https://github.com/FartyPants/text-generation-webui/tree/training

FartyPants:
I decided to move the whole enhanced training as it's own Training PRO extension

PR #3481, opened 2023-08-06
Refactor everything

oobabooga:
Refactors server.py, breaking it into several self-contained modules.


Ph0rk0z:
But why?

oobabooga:
I think that a monolithic `server.py` makes it harder to understand the codebase and add new features to the UI. I got the idea after playing with the AUTOMATIC1111/stable-diffusion-webui code, which is organized this way

Ph0rk0z:
It's not that bad... yet.. 1200ish lines.

PR #3482, opened 2023-08-07
Use chat_instruct_command in API

jllllll:
## Checklist:

---
Improves compatibility with C++ implementations as the language does not allow variable names with `-`.

Previous name is still checked to maintain compatibility with existing implementations and to minimize confusion as the setting name itself was not changed.

### DIFF
--- a/api-examples/api-example-chat-stream.py
+++ b/api-examples/api-example-chat-stream.py
@@ -38,7 +38,7 @@ async def run(user_input, history):
         '_continue': False,
         'stop_at_newline': False,
         'chat_generation_attempts': 1,
-        'chat-instruct_command': 'Continue the chat dialogue below. Write a single reply for the character "<|character|>".\n\n<|prompt|>',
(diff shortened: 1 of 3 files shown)


oobabooga:
@jllllll do you have an example of circumstance where this causes a problem? In the UI, `chat-instruct_command` is always used as a key to a dictionary.

jllllll:
> @jllllll do you have an example of circumstance where this causes a problem? In the UI, `chat-instruct_command` is always used as a key to a dictionary.

It is specifically in response to this Discord message:
https://discord.com/channels/1089972953506123937/1120028000201756695/1137874671862292713

They mention automatic JSON serialization/deserialization in C++ breaking due to the JSON keys being assigned to variables where the `-` character is not allowed. C# also has this issue, which may be what they meant.

I'm sure there are workarounds to this issue, I just made the PR in case the change was worthwhile as `chat-instruct_command` is the only setting that uses `-`.

oobabooga:
I'm not too big a fan of changing the variable name as the mode is called `chat-instruct` in the UI, but since it's the only variable that causes problems, it's fair to change it in the API.

PR #3494, opened 2023-08-08
Fixed a typo when displaying parameters on the llamm.cpp model did not correctly display "rms_norm_eps"

berkut1:
Fixed a typo when displaying parameters on the llamm.cpp model did not correctly display "rms_norm_eps".
There was a problem that when resaving settings, it overwrote to the wrong value.

## Checklist:


oobabooga:
Nice catch

PR #3526, opened 2023-08-10
Streamline GPTQ-for-LLaMa support

jllllll:
## Checklist:

---
Will now use a package containing both your fork and the cuda branch from the main GPTQ-for-LLaMa repo.
Details on package functionality here: https://github.com/jllllll/GPTQ-for-LLaMa-CUDA

This also removes Triton branch support due to AutoGPTQ already having Triton support.
Functionality is minimally impacted by this as llama.cpp cuBLAS is up to 3x faster than GPTQ-for-LLaMa these days.
Even more so compared to the main cuda branch.

This also removes monkey patch support. This removal is in it's own commit so that it can be easily reverted if desired.
Once AutoGPTQ's PEFT integration is fully supported in the webui, that will be the preferred way to use and train LoRas with GPTQ anyway.

I edited documentation as appropriate to account for this PR. Please let me know if I missed anything or if you would like any other changes implemented.


oobabooga:
This is extremely helpful, thank you.

The only part I'm reluctant about is removing the monkey patch, as it would also remove the ability to load LoRAs with GPTQ-for-LLaMa. For some reason, people have been preferring to merge LoRAs with the base models, generating the monstrosities here:

https://huggingface.co/models?sort=trending&search=superhot

I think that may change in the future. For reference, the biggest SuperHot LoRA (30b) is a 49mb file, while a 4-bit model merge is somewhere between 4gb and 16gb.

oobabooga:
I have tested the wheel and it worked fine, including loading LoRAs with the ancient instructions here: https://github.com/jllllll/text-generation-webui/blob/gptqllama/docs/GPTQ-models-(4-bit-mode).md#using-loras-with-gptq-for-llama

Also added the monkey-patch back, at least for now, to keep backwards LoRA compatibility in case the community decides to start sharing adapters.

Removing triton also makes perfect sense.

It should all be good now, thanks again for this.

PR #3533, opened 2023-08-10
Add max_tokens_second param

oobabooga:
ExLlama is too fast, so we need to put some breaks on it to be able to read. 

We have officially come full circle


BadisG:
Just tried it, I think there are two issues with it:
- The real speed isn't really close to the speed we want to achieve
image

- When you press stop it can take 5 secondes to stop, that's really long

Ph0rk0z:
Would be cool also to compute max t/s for a model. Like a benchmark.

oobabooga:
I think that the new ability to scroll during streaming makes this parameter unnecessary.

https://github.com/oobabooga/text-generation-webui/pull/3721

BadisG:
I don't think so, it's not the scrolling the problem it's the speed of the tokens, we should really have the ability to get a limitation speed

oobabooga:
Okay, fair enough. I ran a model that generated 10 tokens/second today and found that pace pretty nice to read. It's good to have this option.

Ph0rk0z:
I just wait till it finishes. I only read along when the generation is slow. The streaming is a good way to cheat lower t/s. 34B is the first L-2 model that generates faster in recent memory.

PR #3545, opened 2023-08-12
Add the --disable_exllama option for AutoGPTQ

clefever:
AutoGPTQ recently added support for the exllama kernel to speed up inference, which is enabled by default. For systems with older hardware, such as the P40, the use of exllama actually slows down inference. Exposing this option will allow users of older hardware to disable the exllama kernel to get quicker inference speeds.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,7 @@ Optionally, you can use the following command-line flags:
 | `--no_inject_fused_mlp`        | Triton mode only: disable the use of fused MLP, which will use less VRAM at the cost of slower inference. |
 | `--no_use_cuda_fp16`           | This can make models faster on some systems. |
 | `--desc_act`                   | For models that don't have a quantize_config.json, this parameter is used to define whether to set desc_act or not in BaseQuantizeConfig. |
+| `--disable_exllama`            | Disable ExLlama kernel, which can improve inference speed on some systems. |
 
 #### ExLlama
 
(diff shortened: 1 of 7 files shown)


BadisG:
How did you get the wheels with the exllama kernels? the latest releases doesn't seem to have those

clefever:
Unless I'm mistaken, the exllama kernel was introduced into AutoGPTQ in version 0.4.0, which this project references. I definitely noticed a decrease in inference speed on my P40 after updating, while providing this option to AutoGPTQ brought it back to normal.

BadisG:
I got this warning when using the 0.4.0 wheel on ooba's webui
image
So I guess the exllama kernels aren't there, unfortunatly for me when I try to create them by myself I got an error https://github.com/PanQiWei/AutoGPTQ/issues/252

clefever:
Ah, hmm. I'm running on Linux and didn't run into that, but perhaps the kernel isn't included in the Windows wheels?

This change would allow you to set `disable_exllama` for now to address the warning, at least until the Windows wheels with exllama are released.

BadisG:
Oh that might explain some things, the fact that the exllama kernels are only on the linux wheels suggests that they weren't also able to make such wheels on windows

oobabooga:
It looks perfect, thank you!

PR #3547, opened 2023-08-12
Add support for requesting Token-Length to the Websocket API / graceful handling when no model is loaded

ben-mkiv:
Hi,

first three commits add support for requesting the token length for a prompt via the websocket API. This is useful if you want to truncate on your own rather than relying on the servers truncation (since this could cut off important context)

I'm aware that the example comment isn't the best solution, over all i would suggest that API usage would be a good topic that could be handled in the GitHub wiki. (Eventually opening it for public or restricted edits for trusted users!?)

the last commit handles requests more graceful when there is no model loaded, currently this would just print an error stacktrace on the console, this was replaced by a console message and a websocket response which indicates that no model is loaded.

edit: added another commit which replaces print() statements with logger calls, except... for the exception as i wasn't sure how that output would look like and if it qualifies for logger calls. (also got no idea what the character at the beginning of the string is for, probably some formatting, i can get rid of it if you intend to merge the PR and want that changed)

### DIFF
--- a/api-examples/api-example-stream.py
+++ b/api-examples/api-example-stream.py
@@ -23,8 +23,8 @@ URI = f'ws://{HOST}/api/v1/stream'
 # the server responds with:
 # {
 #	'event': 'token-count', 
-#	'count' : 42, 
-#	'prompt' : 'a copy of your prompt' 
+#	'count': 42, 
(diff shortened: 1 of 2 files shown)


ben-mkiv:
closing this in favor of a new PR which involves more changes

PR #3556, opened 2023-08-12
Various ctransformers fixes

netrunnereve:
I had to look into the code to get ctransformers to work, and ended up fixing a few things as well. Truthfully though when GGUF ends up getting merged llama.cpp should automatically support MPT and Falcon out of the box.

- added proper documentation
- added ctransformers models to `config.yaml` (note that I didn't try all of them :woman_shrugging:)
- set context length for ctransformers

PR #3559, opened 2023-08-13
ctransformers improvements

cal066:
1. Context length setting.
2. Better "None" type guessing since infer loader does not set model_type.
3. Stop using _stream internal function calls.
## Checklist:

### DIFF
--- a/modules/ctransformers_model.py
+++ b/modules/ctransformers_model.py
@@ -18,6 +18,7 @@ class CtransformersModel:
             threads=shared.args.threads,
             gpu_layers=shared.args.n_gpu_layers,
             batch_size=shared.args.n_batch,
+            context_length=shared.args.n_ctx,
             stream=True
         )
 
--- a/modules/ctransformers_model.py
+++ b/modules/ctransformers_model.py
@@ -49,7 +49,7 @@ class CtransformersModel:
     def generate(self, prompt, state, callback=None):
         prompt = prompt if type(prompt) is str else prompt.decode()
         # ctransformers uses -1 for random seed
-        generator = self.model._stream(
+        generator = self.model(
             prompt=prompt,
             max_new_tokens=state['max_new_tokens'],
             temperature=state['temperature'],
--- a/modules/ctransformers_model.py
+++ b/modules/ctransformers_model.py
@@ -32,7 +32,7 @@ class CtransformersModel:
         return result, result
 
     def model_type_is_auto(self):
-        return shared.args.model_type == "Auto" or shared.args.model_type == "None"
+        return shared.args.model_type is None or shared.args.model_type == "Auto" or shared.args.model_type == "None"
 
     def model_dir(self, path):
         if path.is_file():
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -131,6 +131,7 @@ llama-65b-gptq-3bit:
 .*mpt-.*chat:
   mode: 'instruct'
   instruction_template: 'MPT-Chat'
+  model_type: 'mpt'
 (?!.*-flan-)(?!.*-t5-).*lamini-:
   mode: 'instruct'
   instruction_template: 'Alpaca'
@@ -224,6 +225,7 @@ llama-65b-gptq-3bit:
   mode: 'instruct'
   instruction_template: 'Starchat-Beta'
   custom_stopping_strings: '"<|end|>"'
+  model_type: 'starcoder'
 .*minotaur:
   mode: 'instruct'
   instruction_template: 'Minotaur'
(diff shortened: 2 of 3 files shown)


cal066:
Hello @netrunnereve Looks like we have been working in parallel, I saw your #3556. Do you mind rebasing the changes on top of mine?

@oobabooga I found a problem after a1a9ec895d96b27f2c03cf5df4bce679f3abaf91. I'm somehow getting:
```
text-generation-webui  | Traceback (most recent call last):                                                                                                                                                                                     
text-generation-webui  |   File "/app/modules/callbacks.py", line 55, in gentask                                                                                                                                                                
text-generation-webui  |     ret = self.mfunc(callback=_callback, *args, **self.kwargs)                                                                                                                                                         
text-generation-webui  |   File "/app/modules/ctransformers_model.py", line 94, in generate                                                                                                                                                     
text-generation-webui  |     for token in generator:                                                                                                                                                                                            
text-generation-webui  |   File "/venv/lib/python3.10/site-packages/ctransformers/llm.py", line 467, in _stream                                                                                                                                 
text-generation-webui  |     for token in self.generate(                                                                                                                                                                                        
text-generation-webui  |   File "/venv/lib/python3.10/site-packages/ctransformers/llm.py", line 426, in generate                                                                                                                                
text-generation-webui  |     token = self.sample(                                                                                                                                                                                               
text-generation-webui  |   File "/venv/lib/python3.10/site-packages/ctransformers/llm.py", line 377, in sample                                                                                                                                  
text-generation-webui  |     return self.ctransformers_llm_sample(                                                                                                                                                                              
text-generation-webui  | ctypes.ArgumentError: argument 7: TypeError: wrong type                                                                                                                                                                
text-generation-webui  | Output generated in 6.04 seconds (0.00 tokens/s, 0 tokens, context 19, seed 1883392458)
```

Somehow state['seed'] ended up as a float??
```
logger.info(f'type is {type(state["seed"])}, value: {state["seed"]}')
INFO:type is <class 'float'>, value: -1.0
```
I used `seed=int(state['seed'])` locally as a workaround.

cal066:
I have decided to follow `set_manual_seed` as an example to force an integer. Is there a way to actually get the value from `set_manual_seed` in the generate call?

modules/ctransformers_model.py
```diff
@@ -9,29 +12,57 @@ class CtransformersModel:
...
-        return shared.args.model_type == "Auto" or shared.args.model_type == "None"
+        return shared.args.model_type is None or shared.args.model_type == "Auto" or shared.args.model_type == "None"
+
+    # llama is more of a fallback option
+    match_table = OrderedDict({
+        'falcon': re.compile('.*falcon.*ggml.*\\.bin'),
+        'mpt': re.compile('.*mpt.*ggml.*\\.bin'),
+        'starcoder': re.compile('.*star(chat|coder).*ggml.*\\.bin'),
+        'llama': re.compile('.*ggml.*\\.bin')
+    })
```
oobabooga, line 45: The webui has its own mechanism for inferring and setting parameters based on model names. The patterns are at `models/config.yaml`, and there you can set `model_type` based on regexes. So this should be removed and moved there.

modules/ctransformers_model.py
```diff
@@ -9,29 +12,57 @@ class CtransformersModel:
     def __init__(self):
         pass
 
-    @classmethod
```
oobabooga, line 12: Why remove the classmethod? Is there an advantage?

modules/ctransformers_model.py
```diff
@@ -9,29 +12,57 @@ class CtransformersModel:
...
+            return None
+
+        # Try guessing
+        file_list = list(model_name_dir.glob('*.bin'))
+        if len(file_list) > 0 and file_list[0] is not None:
+            for type_match in self.match_table.keys():
+                pattern = self.match_table.get(type_match)
+                if(pattern.match(str(file_list[0].name))):
+                    logger.info(f'Guessing {file_list[0].name} is {type_match}')
+                    return type_match
```
oobabooga, line 62: This part should be removed, see the comment above on `models/config.yaml`.

oobabooga:
> I have decided to follow `set_manual_seed` as an example to force an integer. Is there a way to actually get the value from `set_manual_seed` in the generate call?

That's geared towards pytorch. I think it's fine to use the UI seed for ctransformers as it is for the time being.

modules/ctransformers_model.py
cal066: It's for consistency with the other methods that use self. The classmethod wasn't intended.

modules/ctransformers_model.py
cal066: Done.

modules/ctransformers_model.py
cal066: I have dropped this part in favor of the more conservative `models/config.yaml` changes.

oobabooga:
I merged the rebased version at https://github.com/oobabooga/text-generation-webui/pull/3556. Thanks @cal066!

PR #3577, opened 2023-08-14
make python scripts executable

Cebtenzzre:

These changes are helpful for users on Unix-like platforms, so they can just run `./server.py` instead of `python server.py`.

I considered any .py file with a shebang (`#!`) line at the top or an `if __name__ == '__main__':` check anywhere in the file to be a script. I added shebangs where missing, and set the executable bit on the scripts (mode 0755).


oobabooga:
Users can simply create executable bash scripts with the `python server.py` command and their flags. I don't want to distribute files with 755 permission.

Cebtenzzre:
May I ask why you do not want the scripts to be marked executable?

And could we at least be consistent and remove the shebang line from 'api-examples/api-example-model.py' and 'extensions/openai/cache_embedding_model.py'? I don't think it's doing anything useful.

Cebtenzzre:
@oobabooga Sorry to bother you, but I'd like some closure on this. Why do you not want to ship executable scripts? llama.cpp recently accepted a [similar fix](https://github.com/ggerganov/llama.cpp/pull/2675). I run server.py and download-model.py directly from the command line, as demonstrated in the README, and configure most things via the UI.

PR #3583, opened 2023-08-15
Created a small script for deleting a model

paziman:
## Checklist:

A small script to remove a model from config-user.yml and models directory. Tested only on Debian 12


oobabooga:
That's a simple OS task (deleting a folder and editing a file). I don't want to add an additional script for that.

roakizi:
yeah but users want it

PR #3586, opened 2023-08-15
Add params suppress_tokens and begin_suppress_tokens

belladoreai:
These are HF parameters that I've found really helpful in preventing repetitive outputs.

## Checklist:


oobabooga:
As mentioned in the Contributing Guidelines

> You should not add features exclusively to the API if they could be added to the UI first.

I'll close this because

1) Adding this to the UI would be troublesome as words often tokenize to more than 1 token, so users would need to first see the tokens for a given word.

2) This can easily be implemented as an extension without changes to the current code.

PR #3590, opened 2023-08-16
Add a "Show controls" button to chat UI

oobabooga:
Inspired by this comment by @missionfloyd: https://github.com/oobabooga/text-generation-webui/discussions/3063#discussioncomment-6394607

It seems to work perfecly on desktop, but on mobile it looks completely broken when the virtual keyboard appears. Any help is appreciated.

| On | Off |
|----|-----|
| on | off |


oobabooga:
Apparently it works on mobile now too.

PR #3598, opened 2023-08-16
refactor the whole streaming/blocking API to a more unified code base

ben-mkiv:
the idea is to pass the parsed JSON to methods together with a connectionContext which has a responseHandler the method then passes data back to the responseHandler which is responsible for sending json back to the client

another thing that was changed/added is the PATH_DICT which contains the path handlers for the websocket/http implementations, currently theres one for both, but it could also be merged into one if necessary

few other changes that have been made:
* more graceful handling of malformed json
* any action that requires a model now checks if a model is loaded and informs the user if none is loaded, instead of failing with a stacktrace output
* made use of the logger for most console print statements
* model/token-length/stop-streaming API methods are now available on the websocket API

## Checklist:

### DIFF
--- a/api-examples/api-example-chat-stream.py
+++ b/api-examples/api-example-chat-stream.py
@@ -81,11 +81,10 @@ async def run(user_input, history):
             incoming_data = await websocket.recv()
             incoming_data = json.loads(incoming_data)
 
-            match incoming_data['event']:
-                case 'text_stream':
-                    yield incoming_data['history']
-                case 'stream_end':
-                    return
+            if incoming_data['event'] == 'text_stream':
+                yield incoming_data['history']
+            elif incoming_data['event'] == 'stream_end':
+                return
 
 
 async def print_response_stream(user_input, history):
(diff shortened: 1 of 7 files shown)


oobabooga:
This is a refactor that adds lines rather than simplifying. I believe that it probably makes sense, but it will make it harder for me to maintain the API.

ben-mkiv:
because it adds features like checking for valid model loaded, and exposing the whole API to the streaming server.

if you think that its harder to maintain then you should compare it next to the original code...

ben-mkiv:
lets take the finetune API PR as example...

instead of this:
https://github.com/oobabooga/text-generation-webui/pull/2342/files#diff-7ad779517e908e6e2a26e2324191b9b5a11a49aff14fa7227dd14d5c43cbb06d

you would add a method to the shared_api.py
```python
def _handle_finetune_request(context, message):
    generate_params = build_parameters_train(message)
    while True:
        try:
            print(next(do_train(**generate_params)))
        except StopIteration:
            break
    context['responseHandler'](context, { 'event': 'finetune', 'results': 'finished' })
```

and add that method to the path dict of the blocking and websocket API. not only is the new implementation less code (because all the http request boilerplate isn't necessary anymore, but also you would have one method which is supported by both api backends)

Issue #3570, opened 2023-08-14
Problem when using the --public-api flag.


akaikite:
### Describe the bug

I used the --api --public-api flag as usual, but got an error.
So I fixed it by opening text-generation-webui\extensions\api\util.py and modifying public_url = _run_cloudflared(port, port + 1, tunnel_id=tunnel_id) to public_url = _run_cloudflared(port, port + 1).

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

--api --public-api flag

### Logs

```shell
2023-08-14 17:30:05 INFO:Loading the extension "gallery"...
Traceback (most recent call last):
  File "D:\AIchat\oobabooga-windows\text-generation-webui\extensions\api\util.py", line 102, in _start_cloudflared
    public_url = _run_cloudflared(port, port + 1, tunnel_id=tunnel_id)
TypeError: _run_cloudflared() got an unexpected keyword argument 'tunnel_id'
Traceback (most recent call last):
  File "D:\AIchat\oobabooga-windows\text-generation-webui\extensions\api\util.py", line 102, in _start_cloudflared
    public_url = _run_cloudflared(port, port + 1, tunnel_id=tunnel_id)
TypeError: _run_cloudflared() got an unexpected keyword argument 'tunnel_id'
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
Exception in thread Thread-7 (_start_cloudflared):
Exception in thread Thread-8 (_start_cloudflared):
Traceback (most recent call last):
Traceback (most recent call last):
  File "D:\AIchat\oobabooga-windows\installer_files\env\lib\threading.py", line 1016, in _bootstrap_inner
  File "D:\AIchat\oobabooga-windows\installer_files\env\lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
    self.run()
  File "D:\AIchat\oobabooga-windows\installer_files\env\lib\threading.py", line 953, in run
  File "D:\AIchat\oobabooga-windows\installer_files\env\lib\threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
    self._target(*self._args, **self._kwargs)
  File "D:\AIchat\oobabooga-windows\text-generation-webui\extensions\api\util.py", line 112, in _start_cloudflared
  File "D:\AIchat\oobabooga-windows\text-generation-webui\extensions\api\util.py", line 112, in _start_cloudflared
    raise Exception('Could not start cloudflared.')
[...]

PR #3606, opened 2023-08-17
fixes error when not specifying tunnel id

ausboss:
Added a simple if statement to not pass in the tunnel_id arg if it was not specified.

It addresses the error people get when using --public-api as seen in the error below. I was having the same problem and wanted to make a fix that will work with or without passing in the tunnel_id

https://github.com/oobabooga/text-generation-webui/issues/3570#issue-1849302813

## Checklist:

- [✅] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).


oobabooga:
Thank you for the fix

PR #3608, opened 2023-08-18
Fix print CSS

### LINKED ISSUE #3600: Print chat css is broken
### Describe the bug

The ability to print out a chat log from the browser window is broken once again. The change implemented in #2793 seems to have been broken in the latest update. The behavior is similar to how it was before that issue was resolved (i.e. the printout is partially limited by what's visible in the browser window. I've inspected the chat tab's CSS on Chrome and Firefox, but the cause isn't immediately obvious to me. 

### Reproduction

1. Generate a chat or instruct session long enough to not fit in your browser window (longer is generally better)
2. Print the chat window and observe that you can't see your entire chat
3. Switch Chat Styles and repeat step 2. Observe how none of the chat styles are fully overridden by the `@media print` query

### Screenshot

_No response_

### Logs

```shell
None. Some minor CSS warnings, but nothing malformed as far as I can tell.
```

### System Info

```shell
Ubuntu 22.04, CPU only
```

missionfloyd:
## Checklist:

Fixes #3600

I'm not sure why it's so small, but that's easily adjusted with the print dialog's Scale setting.

Before
image

After
image

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -350,12 +350,10 @@ div.svelte-362y77>*, div.svelte-362y77>.form>* {
         position: absolute;
         left: 0;
         top: 0;
-        max-width: none;
-        max-height: none;
+        max-width: unset;
+        max-height: unset;
         width: 100%;
-        height: fit-content;
-        display: flex;
-        flex-direction: column-reverse;
+        overflow-y: visible;
     }
     
     .message {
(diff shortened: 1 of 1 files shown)


oobabooga:
Thank you for the fix!

PR #3620, opened 2023-08-18
llama.cpp: make Stop button work with streaming disabled

Cebtenzzre:
***
I noticed that the 'Stop' button had no effect on llama.cpp when streaming was disabled. This simple fix makes it work as expected.


oobabooga:
I missed that completely. Thank you!

PR #3622, opened 2023-08-19
Negative dimension fix

practicaldreamer:
Proposed fix for "negative dimension" error when using negative prompt CFG


practicaldreamer:
Sorry, after further testing I realized this isn't working as expected. Context window > 2048 still doesn't work

PR #3638, opened 2023-08-21
Added detection of LlongOrca

berkut1:
Since LlongOrca only has models with 16k contexts, I also added `compress_pos_emb: 4`.
I don't think it's a good idea to create a separate regex for 16K because we can't guarantee that all models (will) have `compress_pos_emb: 4`.

## Checklist:


berkut1:
I was also thinking of adding a separate 16k regex just for `truncation_length`, but we still need to change `compress_pos_emb`. So I thought it didn't make much sense and it's better to leave it to specific models, especially since those models are only available with 16k content.

berkut1:
However, if think about it, it's a bit annoying to setup in the parameters tab - `truncation_length` every time the model changes. Therefore, after thinking, I decided to separate 16k into a separate rule. :)

berkut1:
Hm. As I see it, some parameters have been removed in GGUF, perhaps they will be determined automatically. Therefore, I'm closing this pull for now.

PR #3643, opened 2023-08-22
Fix whitespace formatting in perplexity_colors extension.

tdrussell:
This is a somewhat opinionated change that fixes whitespace formatting in the perplexity_colors extension. Basically, we don't even attempt to interpret the text as markdown and render it into HTML. Instead we use a monkeypatch to just wrap the text in \<pre\> tags (with a bit of CSS to still wrap lines). This means all whitespace renders correctly, but things like italics, code blocks, or any other markdown formatting is just output as-is.

My reasoning for these larger formatting changes is that if you want to color tokens based on probability/perplexity, and especially if you're using the probability dropdown, you probably care more about seeing the actual tokens the model outputted than markdown rendered correctly. The changes also allowed me to simplify quite a few things in the code.

@SeanScripts Thoughts on bypassing the markdown rendering like this? To me it's better this way, but I want to see what others think.

Screenshot from 2023-08-21 20-26-28

## Checklist:

### DIFF
--- a/css/html_readable_style.css
+++ b/css/html_readable_style.css
@@ -27,3 +27,7 @@
 .container :not(pre) > code {
     white-space: normal !important;
 }
+
+.container .hoverable {
+    font-size: 14px;
+}
--- a/extensions/perplexity_colors/script.py
+++ b/extensions/perplexity_colors/script.py
@@ -1,14 +1,11 @@
-import re
 import time
 
 import gradio
-import markdown
 import numpy as np
 import torch
 from transformers import LogitsProcessor
 
 from modules import html_generator, shared
-from modules.html_generator import replace_blockquote
 
 params = {
     'active': True,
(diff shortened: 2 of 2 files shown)


oobabooga:
I think that this is a fair change in the context of this extension. Also I see lots of removed lines and that always makes me happy.

SeanScripts:
Nice! I agree, I'd rather see the tokens than the markdown. Looks good and works perfectly; thanks for the clean optimization.

tdrussell:
Ok great, sounds like we can move forward with this change. This should make the probability viewer pretty usable now. I have some more improvements I want to make eventually, that should make it even better for testing / experimentation use cases.

oobabooga:
Thanks for the PR @tdrussell and thanks @SeanScripts for the review.

PR #3667, opened 2023-08-24
Implement CFG for llamacpp_HF

oobabooga:
Similar to https://github.com/oobabooga/text-generation-webui/pull/3666, but for llamacpp_HF.

The implementation seems to be functional, but it's extremely slow and unusable due to the expensive `model.save_state` and `model.load_state` calls. Maybe there is a better way to do this.


oobabooga:
Merged into #3666

PR #3688, opened 2023-08-25
Fix ctransformers threads auto-detection

jllllll:
## Checklist:

---
ctransformers uses `-1` to auto-detect threads, unlike llama.cpp which uses `0`.

I was confused as to why ctransformers was so slow, until I noticed this difference.


oobabooga:
Very nice, thanks for the fix.

PR #3711, opened 2023-08-27
Fix ctransformers model unload

marella:
Closes marella/ctransformers#111

Currently `model` is added to `self` (points to class as it is a `classmethod`) instead of `result` which prevents it from being deleted in `unload_model()`.

Also add missing comma in model types list.

## Checklist:


oobabooga:
That's perfect, thank you for the fix!

PR #3712, opened 2023-08-27
accept floating-point alpha value on the command line

Cebtenzzre:
***
Since alpha_value has a step of 0.1 now, allow the user to specify decimal values on the command line as well.


oobabooga:
Well noted, thank you.

PR #3719, opened 2023-08-28
Added checks for missing CFG state keys to fix issues when using openai extension

ricardopinto:
Today I updated my local codebase and the openai api extension kept returning errors on chat completions. The error is a `HTTP 500` and this is the payload:

```json
{
  "error": {
    "message": "KeyError('guidance_scale')",
    "code": 500,
    "type": "OpenAIError",
    "param": ""
  }
}
```

Traced the issue back to [modules/text_generation.py#L244](https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L244) and in the process found a couple more similar issues on that same function. With these changes, the openai chat completions endpoint is now working again.

How to reproduce:
- launch webui in docker, model llama2 7b chat from HF. Here's my env `CLI_ARGS=--model meta-llama_Llama-2-7b-chat-hf --model_type llama --load-in-4bit --listen --no-stream --auto-devices --chat --extension openai --n_batch 512 --n_ctx 4096`
- call the chat completions endpoint:
  ```bash
  curl -X 'POST' \
    'http://localhost:5001/chat/completions' \
    -H 'accept: application/json' \
    -H 'Content-Type: application/json' \
    -d '{
      "frequency_penalty": 0,
      "logit_bias": null,
      "max_tokens": "inf",
      "messages": [
        {
          "content": "How to make a chocolate cake?",
          "role": "user"
        }
      ],
      "model": "llama-2-7b-chat-hf",
      "n": 1,
      "presence_penalty": 0,
      "stop": null,
      "stream": false,
      "temperature": 1,
      "top_p": 1,
      "user": "user-1234"
    }'
  ```

## Checklist:


oobabooga:
The correct fix should be to set default values for the missing parameters inside the openai extension.

PR #3743, opened 2023-08-29
Interface Reorganization

AustinnK:
I propose these changes to improve the user experience of the webui and enhance the clarity of the interface:

- Removed chat/default/notebook tabs from the main header and put them into a new "generate" tab.

All three pages serve a similar purpose which makes having all of them on the main header unnecessary and would certainly be confusing to a first-time user. They are now accessed through a new "generate" page.
- Moved character definitions to the chat page.

Character and instruction format is utilized on the chat tab. Thusly, I believe that this change will benefit ease of use and improve clarity due to relevant info being on the same screen.
- Moved every option that presets save into the same column.

Excusing guidance_scale, every setting that presets save have been moved to the column underneath the presets dropdown. This improves the correlation between the sampling parameters and intuitively relates them to the presets. Previously, it wasn't immediately clear when a preset updated mirostat or beam search since those settings were located in a different area from all of the other sampling settings.


oobabooga:
I prefer the current looks.

Issue #3527, opened 2023-08-10
No module named 'xformers'


SmoothBrainApe:
### Describe the bug

I added --xformers to the cmd flags and this error appears

Traceback (most recent call last):
  File "/home/chupz22/oobabooga_linux/text-generation-webui/modules/llama_attn_hijack.py", line 14, in <module>
    import xformers.ops
ModuleNotFoundError: No module named 'xformers'

i tried pip install xformers and it says it is installed. I did under with conda activate textgen so its in the environment.

I can run no problems without xformers but it would be better to have it to save memory. Just wondering what else have I missed? Thanks

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

add --xformers to cmd flags

### Logs

```shell
Traceback (most recent call last):
  File "/home/chupz22/oobabooga_linux/text-generation-webui/modules/llama_attn_hijack.py", line 14, in <module>
    import xformers.ops
ModuleNotFoundError: No module named 'xformers'
```

### System Info

```shell
Linux Mint 21
AMD Ryzen 3 3300X
GPU NVIDIA 3060 12GB VRAM
```

Ph0rk0z:
check that it's installed with `pip list`and that it didn't fail. works over here on 2 machines.

SmoothBrainApe:
I got everything working now. All issues i am having was gone after uninstalling torch on the main unit and reinstalling torch on textgen environment. The issue i have now is other programs like automatic1111 cant run without torch on main. Im kinda lost here now.

Ph0rk0z:
automatic111 is supposed to use venv

SmoothBrainApe:
I know. That is why i am pulling my hair out trying to figure this out hahaha. Will update if i managed to fix this. Thanks

SmoothBrainApe:
I got it working now. I used python venv instead of conda for textgen and it just magically worked together with other venv. I can use comfy, A1111, Roop, and textgen no worries now. I will leave this message here anycase someone else encounter the same problem.

fyyyyy:
I fixed it by adding the last line to one_click.py. Remove after first start.

```python
# line 292...
if __name__ == "__main__":
    # Verifies we are in a conda environment
    check_env()
    
    # add this
    run_cmd("python -m pip install xformers", environment=True)
```

Issue #1252, opened 2023-04-15
🔴 xformers not found! Please install it before trying to use it.


IronWolve:
### Describe the bug

Did both pip install xformers, and xformers in /text-generation-webui directory.
Evan ran the installer again.

https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Xformers

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

xformers missing

### Logs

```shell
🔴 xformers not found! Please install it before trying to use it.
Loading the extension "api"... Ok.
Loading the extension "sd_api_pictures"... Ok.
Loading the extension "gallery"... Ok.
Loading the extension "send_pictures"... Starting KoboldAI compatible api at http://127.0.0.1:5000/api
Ok.
```

### System Info

```shell
AMD 5950x, Nvidia 4090, Win11
```

Ph0rk0z:
yes but did you install it inside conda or the correct venv?

Ph0rk0z:
maybe this check fails on windows

```
try:
        import xformers.ops
    except Exception:
        print("🔴 xformers not found! Please install it before trying to use it.", file=sys.stderr)
```

Gitbreast:
Same situation, why?

Ph0rk0z:
Open up a python .. write` import xformers.ops`, what does it say?

Gitbreast:
![image](https://user-images.githubusercontent.com/96460796/232319328-aa4dddbf-1e8a-4b2d-8e97-7833afa606e5.png)
@Ph0rk0z I got this, means I need to upgrade something?

Ph0rk0z:
Yea, you have CPU pytorch.

Ketorami:
I have the same problem. My notifications are like this.
![Screenshot 2023-04-17 134612](https://user-images.githubusercontent.com/101268214/232405714-e0f718d2-c7e6-47e9-aa92-15d03a3b8fa3.png)
but when i open the WebUI , I still get error that not found

![Screenshot 2023-04-17 134856](https://user-images.githubusercontent.com/101268214/232406119-0e85015c-ce75-4d9b-a30d-458f4da67a19.png)

Ph0rk0z:
Triton doesn't work on windows I think.

Gitbreast:
> Yea, you have CPU pytorch.

So I need to install a PyTorch for GPU and upgrade Python?

Ketorami:
[...]

Issue #1086, opened 2023-04-12
A matching Triton is not available, some optimizations will not be enabled. Error caught was: No module named 'triton'


iChristGit:
### Describe the bug

I enabled --xformers, and did a pip install xformers
Now I get this message on top of the webui models loading list:
A matching Triton is not available, some optimizations will not be enabled.
Error caught was: No module named 'triton'

What can cause this? 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

installing xfromers

### Logs

```shell
Starting the web UI...
A matching Triton is not available, some optimizations will not be enabled.
Error caught was: No module named 'triton'

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please submit your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
CUDA SETUP: Loading binary C:\TextWebui\installer_files\env\lib\site-packages\bitsandbytes\libbitsandbytes_cudaall.dll...
The following models are available:

1. 4bit_alpaca-7b-native-4bit
2. anon8231489123_vicuna-13b-GPTQ-4bit-128g
3. chavinlo_gpt4-x-alpaca
4. elinas_alpaca-13b-lora-int4
5. elinas_alpaca-30b-lora-int4
6. gpt4-x-alpaca-13b-native-4bit-128g
7. llama-13b-4bit-128g
8. llama-30b-4bit-128g
9. llamacpp-13b
10. mayaeary_pygmalion-6b-4bit-128g
11. PygmalionAI_pygmalion-6b

Which one do you want to load? 1-11
```

### System Info

```shell
Windows11, 1ClickInstaller
```

Jonseed:
What is Triton, and why is it trying to use Triton after installing xformers?

PR #3748, opened 2023-08-30
install xformers too

donlinglok:
So i have try to use --xformers, but seems it is not in the requirement.
```
  File "/xxx/text-generation-webui/modules/llama_attn_hijack.py", line 14, in <module>
    import xformers.ops
ModuleNotFoundError: No module named 'xformers'
```

I have add a requirement.txt in extensions, and now the xformers should installed when initial_installation. 

Tested with start_windows.bat. `--listen --api --verbose --auto-devices --xformers`
```
Starting API at http://0.0.0.0:5000/api

..
2023-08-29 10:56:19 INFO:Replaced attention with xformers_attention
2023-08-29 10:56:19 INFO:Loaded the model in 11.18 seconds.
```

## Checklist:

- [Y] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).


donlinglok:
> Will it throw exceptions/errors for AMD cards? Or has support already been added?

hum... i thought the xformers missing is a bug, so i just trying to add back the missing dependency.

sorry i only tested in windows with nvidia card. if xformers cannot work in AMD, now i think the xformers is needed to be install manually for user with specific display card.

I will closed this pull request.

ref: 
https://github.com/oobabooga/text-generation-webui/issues/1252
https://github.com/oobabooga/text-generation-webui/issues/1086
https://github.com/oobabooga/text-generation-webui/issues/3527
https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/5091

PR #3768, opened 2023-08-31
Dropdown item fixes

missionfloyd:
## Checklist:

Removes excess whitespace (except on mobile for touchscreens) and displays long items without wrapping.

Particularly effective on the models dropdown, which tends to be long and narrow at lower resolutions.

Before:
image

After:
image


bet0x:
@missionfloyd I've been racking my brain trying to implement something similar on the other panel. I wanted to add a dropdown after clicking on the 'Get file list' and fill the "File name" Textbox, but I don't know if it's possible

missionfloyd:
The textbox could just be replaced with a dropdown, populated by "get file list" and cleared when the repo textbox is changed. This could also allow specifying multiple files.

FartyPants:
excellent

oobabooga:
This is a valid change, but for a matter of arbitrary personal taste I prefer to keep the current appeareance as I find it more readable.

PR #3772, opened 2023-08-31
Added Disable Exllama functionality to Transformers Loader

hypersniper05:
## Checklist:

References : 

[[Loading](https://huggingface.co/docs/transformers/main/en/main_classes/quantization#exllama-kernels-for-faster-inference)]
[[Fine-Tunning](https://huggingface.co/docs/transformers/main/en/main_classes/quantization#finetune-a-quantized-model)]

Exllama kernel on Transformer improves speed but the whole model needs to load on GPU. Auto-devices might use cpu and so you get this error:
`Found modules on cpu/disk. Using Exllama backend requires all the modules to be on GPU. You can deactivate exllama backend by setting disable_exllama=True in the quantization config object`

In addition it is recommened to disable exllama when Lora training (See link above).

### DIFF
--- a/modules/models.py
+++ b/modules/models.py
@@ -184,6 +184,7 @@ def huggingface_loader(model_name):
                     'bnb_4bit_compute_dtype': eval("torch.{}".format(shared.args.compute_dtype)) if shared.args.compute_dtype in ["bfloat16", "float16", "float32"] else None,
                     'bnb_4bit_quant_type': shared.args.quant_type,
                     'bnb_4bit_use_double_quant': shared.args.use_double_quant,
+                    'disable_exllama' : shared.args.disable_exllama
                 }
 
                 logger.warning("Using the following 4-bit params: " + str(quantization_config_params))
--- a/README.md
+++ b/README.md
@@ -158,8 +158,6 @@ text-generation-webui
 │   │   └── tokenizer.model
 ```
 
-In the "Model" tab of the UI, those models can be automatically downloaded from Hugging Face. You can also download them via the command-line with `python download-model.py organization/model`.
-
 * GGML/GGUF models are a single file and should be placed directly into `models`. Example:
 
 ```
@@ -167,6 +165,8 @@ text-generation-webui
 ├── models
 │   ├── llama-13b.ggmlv3.q4_K_M.bin
 ```
+
+In both cases, you can use the "Model" tab of the UI to download the model from Hugging Face automatically. It is also possible to download via the command-line with `python download-model.py organization/model` (use `--help` to see all the options).
 
 #### GPT-4chan
 
--- a/models/config.yaml
+++ b/models/config.yaml
@@ -49,200 +49,142 @@ llama-65b-gptq-3bit:
 .*(gr1024|1024g|groupsize1024):
   groupsize: 1024
 .*(oasst|openassistant-|stablelm-7b-sft-v7-epoch-3):
-  mode: 'instruct'
   instruction_template: 'Open Assistant'
   skip_special_tokens: false
 (?!.*galactica)(?!.*reward).*openassistant:
-  mode: 'instruct'
   instruction_template: 'Open Assistant'
   skip_special_tokens: false
 (?!.*v0)(?!.*1.1)(?!.*1_1)(?!.*stable)(?!.*chinese).*vicuna:
-  mode: 'instruct'
   instruction_template: 'Vicuna-v0'
 .*vicuna.*v0:
-  mode: 'instruct'
   instruction_template: 'Vicuna-v0'
 .*vicuna.*(1.1|1_1|1.3|1_3):
-  mode: 'instruct'
   instruction_template: 'Vicuna-v1.1'
 .*vicuna.*(1.5|1_5):
-  mode: 'instruct'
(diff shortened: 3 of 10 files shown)


modules/models.py
```diff
@@ -184,6 +184,7 @@ def huggingface_loader(model_name):
                     'bnb_4bit_compute_dtype': eval("torch.{}".format(shared.args.compute_dtype)) if shared.args.compute_dtype in ["bfloat16", "float16", "float32"] else None,
                     'bnb_4bit_quant_type': shared.args.quant_type,
                     'bnb_4bit_use_double_quant': shared.args.use_double_quant,
+                    'disable_exllama' : shared.args.disable_exllama
```
oobabooga, line 187: Does this really disable ExLlama for GPTQ? This params dict is for bitsandbytes load_in_4bit not GPTQ.

The documentation recommends

```
import torch
gptq_config = GPTQConfig(bits=4, disable_exllama=False)
model = AutoModelForCausalLM.from_pretrained("{your_username}/opt-125m-gptq", device_map="auto", quantization_config = gptq_config)
```

the question remains if by manually supplying a `GPTQConfig` object requires you to manually specify all values (bits, groupsize), or if you can just change disable_exllama.

hypersniper05 (commented):
From reading the docs this where you add it. Here is basically the keypoints:

-If you are loading on cpu/disk you need to disable exllama
-if you are training using Peft it should be disabled. I transformers doc state that it is unstable at the moment. 
-if you are manually setting ram you it might not load due to this. In my case I have two gpus and when I set devices auto it wouldn't load. Even when I manually set the Vram I would get it. It seems some ram is loaded and it triggers the error (see description of request)

modules/models.py
hypersniper05: Btw it works for me, the arg is the same so very little change required. I really didn't want to create a new one but give it a shot.

I haven't tested the training side of it because we have another issue with training 4bit Transformers at the moment.

Ph0rk0z:
Is this because of the recent transformers integration or something? I think I'm gonna keep using the AutoGPTQ loader.

FartyPants:
Yay, maybe that's why my recent 4bit lora training seemed a little brain dead...

FartyPants:
OOh yeah, while I don't know exactly how the exllama affect the actual training - but it does affect 4-bit interference as well with Lora - basically making such lora almost unusable (writes like under heavy drugs) - that's all the detective work why my last Loras were so weird on 4-bit, but behaved ok-ish on 8-bit. It seems with exllama there is double whammy.

In fact there should be a big warning about this disable exxlama if you want to train in 4-bit and/or use Lora in 4-bit. I'm serious.

hypersniper05:
> OOh yeah, while I don't know exactly how the exllama affect the actual training - but it does affect 4-bit interference as well with Lora - basically making such lora almost unusable (writes like under heavy drugs) - that's all the detective work why my last Loras were so weird on 4-bit, but behaved ok-ish on 8-bit. It seems with exllama there is double whammy. Training and interference.
> 
> In fact there should be a big warning about this: disable exllama if you want to train in 4-bit and/or use Lora in 4-bit. I'm serious. **It's that bad.**
> 
> I went back and marked all my Loras I made since last Wednsday as _BAD
> Props to @hypersniper05 for saving me from more wasted time.

Sorry for the late reply, I haven't been able to get GPTQ training with Transformers to work yet but in the process I noticed the exllama remark on the transformers docs that it should be false when training. I am glad it helped ☺️

Ph0rk0z:
Ok.. I think I am getting a handle on what's going on here. Are people loading GPTQ models and thinking that it's loading with BnB?

HF models shouldn't be loading through AutoGPTQ when you select 4bit. The code in PEFT specifically lists is_quantized and is_4bit separately. These only come into play if the model was loaded with AutoGPTQ.

If you point it at a model that is pre quanted, that may just happen and then it will load using the quant config in the model folder and attempt to use exllama. If you use the proper loaded, you can already disable exllama.

oobabooga:
Things are getting mixed up here. Bitsandbytes `load_in_4bit` and loading GPTQ models through transformers are two unrelated things. This PR adds a `disable_exllama` kwarg to a block inside a `if shared.args.load_in_4bit:` statement. Hence, it doesn't do anything for GPTQ models.

The goal is to disable exllama when loading a GPTQ model as recommended by the transformers documentation. Also, this PR adds a huge message when the transformers loader is selected that is ugly and should be removed.

Ph0rk0z:
BnB can't use the exllama kernel. I think people are loading GPTQ models with the transformers loader. I thought training also works using the correct loader so why...

oobabooga:
> So, just to be 100% clear here: If I'm using Transformers to load a FP16 or BF16 model, which I load in 4-bit (with bitsandbytes) to train, this PR is not relevant? My training won't be messed up by the exllama kernel? I'm a bit confused at this point.

Yes, it's not relevant or related.

hypersniper05:
It is also worth noting that if you can't load the entire GPTQ model on GPU then you must disable exllama as well. I had a sorta related issue with this. I have two GPUs and even though I had enough vram it kept failing on auto devices and also when I would set the vram manually. it just kept failing to load until I set it to true.

PR #3774, opened 2023-08-31
Download metadata issue 3700

Sanyam-jain30:
Completed issue 3700 (Download metadata).
Sorry for the previous pull request it has some updation issues with the get list I have updated it now.
Demo:

https://github.com/oobabooga/text-generation-webui/assets/78720027/2f41e99b-6dc6-4cbd-98d5-de654b901d83

### DIFF
--- a/download-metadata-model.py
+++ b/download-metadata-model.py
@@ -1,0 +1,275 @@
+'''
+Downloads models from Hugging Face to models/username_modelname.
+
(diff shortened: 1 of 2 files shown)


oobabooga:
There is no reason to have a near identical copy of download-model.py in a new file, and another of the UI download function.

Issue #3731, opened 2023-08-28
Illegal instruction (core dumped) error


plaidpants:
### Describe the bug

Upgraded to current version and now after loading any model I get an Illegal instruction (core dumped) error. I did a clean onclick install on linux and a reboot to see if that corrected the issue but it did not.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Use the one click installer for linux, select D for CPU/RAM only option. Load a model using llama.cpp on the model tab observe instruction (core dumped) error. Previously this was working with my GGML models, noticed that GGML models are now deprecated in llama.cpp so I tried a GGUF model instead but with the same result.

### Screenshot

XXX:~/oobabooga_linux$ ./start_linux.sh

The following flags have been taken from the environment variable 'OOBABOOGA_FLAGS':

--listen

To use the CMD_FLAGS Inside webui.py, unset 'OOBABOOGA_FLAGS'.

 

/home/XXX/oobabooga_linux/installer_files/env/lib/python3.10/site-packages/bitsandbytes/cextension.py:34: UserWarning: The installed version of bitsandbytes was compiled without GPU support. 8-bit optimizers, 8-bit multiplication, and GPU quantization are unavailable.

  warn("The installed version of bitsandbytes was compiled without GPU support. "

/home/XXX/oobabooga_linux/installer_files/env/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so: undefined symbol: cadam32bit_grad_fp32

2023-08-28 10:54:17 INFO:Loading the extension "gallery"...

Running on local URL:  http://0.0.0.0:7860/

 

To create a public link, set `share=True` in `launch()`.

2023-08-28 10:54:33 INFO:Loading CodeLlama-34B-GGUF...

Illegal instruction (core dumped)

### Logs

```shell
See above.
```

### System Info

```shell
no GPU, CPU and lots of RAM only.
```

jllllll:
Windows uses pre-compiled wheels because installing a compiler is a pain and often beyond what non-technical users are willing to do. Despite Linux having a much easier installation process for compilers, we still get plenty of reports from Linux users who encounter errors on install because the main llama-cpp-python package is built from source on Linux and they don't have a compiler installed and don't know how to install one. We get enough that I've often considered making a PR to install pre-compiled wheels for all systems.

[...]

PR #3786, opened 2023-09-02
Improve instructions for CPUs without AVX2

netrunnereve:
See #3731 for more details.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Its goal is to become the [AUTOMATIC1111/stable-diffusion-webui](https://github.
 ## Features
 
 * 3 interface modes: default (two columns), notebook, and chat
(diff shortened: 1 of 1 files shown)


jllllll:
You can also get versions of `llama_cpp_python_cuda` and `llama_cpp_python_ggml_cuda` without `AVX2` by changing the `cu117` in the wheel name to `cu117avx` or `cu117basic`.

Sorry for the confusing naming schemes. There are just so many build configurations for llama-cpp-python and Python doesn't give many options for differentiating wheels of different configurations. In fact, Python devs are planning to eventually remove the version tagging system that I'm using. No clue why. Hopefully they reconsider or implement a better system in the future as that will make it significantly more difficult to manage and organize the 6367 wheels I currently have in my repo.

An alternative approach to installing compatible versions of these packages is to use this command:
```
python -m pip install llama-cpp-python llama-cpp-python-cuda llama-cpp-python-ggml llama-cpp-python-ggml-cuda --force-reinstall --no-deps --index-url=https://jllllll.github.io/llama-cpp-python-cuBLAS-wheels/AVX/cpu --extra-index-url=https://jllllll.github.io/llama-cpp-python-cuBLAS-wheels/textgen/AVX/cu117
```
This will install `AVX` versions of all the llama-cpp-python packages used in the webui. You can change `AVX` to `basic` in the index URLs to install `SSE` versions instead.

carlblacknz:
The AVX versions of these don't work on my Ivy Bridge Xeon, but using basic does. Are they still using FP16 maybe?

PR #3792, opened 2023-09-03
Just kill the /s from Vicuna-v1.1.yaml elready

FartyPants:
I know it has been here for donkey's years and I always marveled at the logic of adding /s in turn template, but seriously, in 99% of vicuna models used today adding /s in turn will make the chat talking to itself. And I'm not aware of the 1% that needs it.

If you have some conviction that /s should be there then ignore this, but I had to doctor this every single time I do pull, because /s makes things worse in everything that uses Vicuna USER/ASSISTANT.

## Checklist:

### DIFF
--- a/instruction-templates/Vicuna-v1.1.yaml
+++ b/instruction-templates/Vicuna-v1.1.yaml
@@ -1,4 +1,4 @@
 user: "USER:"
 bot: "ASSISTANT:"
-turn_template: "<|user|> <|user-message|>\n<|bot|> <|bot-message|></s>\n"
+turn_template: "<|user|> <|user-message|>\n<|bot|> <|bot-message|>\n"
 context: "A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions.\n\n"


oobabooga:
This is part of the Vicuna template, as per the official documentation: https://github.com/lm-sys/FastChat/blob/main/docs/vicuna_weights_version.md#example-prompt-weights-v11-v13-v15

If other models use a template similar to Vicuna but without the `</s>`, then a new template and new regexes must be added to `models/config.yaml` for these models. With what model did you experience issues?

FartyPants:
Ok, I get that - but that applies only to lm-sys made Vicuna on llama1 way back. However everything else made after that that uses USER/ASSISTANT or has vicuna anywhere in its name like Wizard-Vicuna or GPT4-vicuna and other will not like if /s is there and start talking to themselves. The vicuna-v1.1 will likely be picked up as the instruct character so perhaps we need lm-sys-only-vicuna.yaml and then vicuna.yaml for everything else that somehow uses vicuna dataset.
Not a big deal of course - easily solvable, but I see the new models talking to themselves all the time and then I remember - ah /s is there.

Ph0rk0z:
Why not add vicuna no-S and give the choice?

FartyPants:
That's what I have on my comp, just calling it Vicuna-universal

Ph0rk0z:
I checked in silly tavern and they also use the S as a separator. Seems like this went around.

FartyPants:
Yeah, I'm closing this, coz the fix is obvious. Just make an ALT Vicuna without /s

PR #3807, opened 2023-09-05
Training Enhancements WIP 2

FartyPants:
This is from my current working environment - my personal changes to training. Now merged with latest @oobabooga UI changes

- Chunking: precise raw text slicer (PRTS) uses sentence slicing and making sure things are clean on all ends (Optional - select it in Advanced)
- overlap PRTS - this special overlapping will make additional overlap block based on logical rules (aka no overlap block on hard cut)
- custom scheduler (follow the code to make your own) In LR Scheduler select FP_low_epoch_annealing - this scheduler will keep the LR constant for first epoch then use cosine for the rest - this part would be best to spawn into a new py file
- save lost threshold - will not save the "Save every n steps" checkpoints until this threshold is reached (I definitely don't need multiple checkpoints that are 2.5 loss - I'm usually interested in checkpoints between say 1.5 and 1.9 loss)
- saves graph png file at the end with learning rate and loss per epoch
- adding EOS to each block or to hard cut only
- automatically lowers gradient accumulation if you go overboard and set gradient accumulation that will be higher than actual data - transformers would then throw error (or they used to, not sure if still true) but in any way, it will fix bad data 

This PR is more to inform you of it existence than a blind merge. I tested it and refined it on hundred of Lora's so it's code is guaranteed (by me, which amounts to nothing), but it consist of too many changes all at once, which would be ultimately better to split into step by step PR's

Note: I personally see a lot of potential in training from structured raw text - and by that I mean raw text where each new idea is separated into a new logic block (for example by 3 new lines) and then recognized as such using Hard Cut String and the PRTS slicer and trained accordingly. 
This beats using raw text as one big block where you have no control and clashing ideas will be mashed together in training blocks, creating a schizophrenic model. A proper structuring allows for style toning of previously fine-tuned model, where the model will answer your questions (driven from the previous finetune) yet use the newly learned writing style to do so.

## Checklist:

## Files
You only need replace training.py


caterpillarpants:
Thanks so much for your hard work, Mr. Pants. Your work has allowed so many people to train their models and we all appreciate you.

I also see potential in training from structured raw text. Is there any way we can visually see how the program is structuring the text so we can be sure we are applying the settings correctly? Like a preview. I often use these pipes(|) that are found in Sanskrit to separate sentences and I would like confirmation it's working. Same for tokenizing. I'm training for Sanskrit and I worry about how it tokenizes Sanskrit text like: "अग्निमीळे पुरोहितं यज्ञस्य देवं रत्वीजम |" It would be great to be able to get a preview of some tokens.

Anyway, thanks so much for your work on training. I've been having a blast!

FartyPants:
Ok, I created an extension with these changes
https://github.com/FartyPants/text-generation-webui/tree/TrainingExtension/extensions/Training_PRO

Instead of adding these changes to main, I'll be adding them in Training_PRO extension.

The main Training tab can be now simplified as ooba described.

oobabooga:
I'm closing this one as https://github.com/oobabooga/text-generation-webui/pull/3961 has been merged.

caterpillarpants:
awesome!

PR #3811, opened 2023-09-06
Refine the checking branch name function of sanitize_model_and_branch…

TheBetterSolution:
…_names in download-model.py

The branch name of git is only can begin with '0-9' or 'a-z' or 'A-Z' or '_', that can't begin with '.' or '-', for checking it precisely, I try to change the regex pattern from: ^[a-zA-Z0-9._-]+$ to ^[a-zA-Z0-9_][a-zA-Z0-9._-]+, that's restrict the first character of branch name by the partial expression: ^[a-zA-Z0-9_] to the correct character.

Hope to make text-generation-webui to be better, please review it.

## Checklist:

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -40,7 +40,7 @@ class ModelDownloader:
         if branch is None:
             branch = "main"
         else:
-            pattern = re.compile(r"^[a-zA-Z0-9._-]+$")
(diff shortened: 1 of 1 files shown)


oobabooga:
While I believe that you are correct, I think that this is a small change that could have a downside that I can't see.

Issue #3822, opened 2023-09-07
Quotation marks in the output string of the google translate extension appeared to be something like `&#x27;`, resulting to incorrect translation


Touch-Night:
### Describe the bug

`'`, `"`, and maybe some other punctuation are incorrectly translated due to Unicode Hex Character Code is used in the string instead of the coresponding punctuation.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Add some prints in the extension and you can see the issue is happening on untranslated output.

### Screenshot

![image](https://github.com/oobabooga/text-generation-webui/assets/102669562/618f2569-0755-4a3d-a7d9-8271a68162c1)

### Logs

```shell
No errors in command line
```

### System Info

```shell
System: Windows 10
GPU: Nvidia RTX 3050Ti Laptop (modified to 8GB VRAM)
```

zencyon:
just pulled latest version and upgraded everything, getting this same behaviour as well
any ' and " are output as &#x27; and &quot; since update (running same model, same instruct

Touch-Night:
> just pulled latest version and upgraded everything, getting this same behaviour as well any ' and " are output as '&#x27'; and '&quot' since update (running same model, same instruct

I did a quick fix in #3827 , you can try that.

Touch-Night:
Close as fixed

PR #3827, opened 2023-09-07
Fix issue #3822 and #3839

Touch-Night:
## Checklist:


missionfloyd:
https://docs.python.org/3/library/html.html#html.unescape

Touch-Night:
Oops, I'm not aware of that

oobabooga:
The branch in this PR is protected and I can't edit it.

Touch-Night:
> The branch in this PR is protected and I can't edit it.

I have disabled it just now.

oobabooga:
Thanks. I have escaped back the Google Translate output (after unescaping it for the translation) and removed the changes to extensions.py as it can probably be made in a simpler way by removing inactive extensions from the `state` variable (or setting things to False). Feel free to submit a new PR if you can manage to get that working.

Touch-Night:
> Thanks. I have escaped back the Google Translate output (after unescaping it for the translation) and removed the changes to extensions.py as it can probably be made in a simpler way by removing inactive extensions from the variable (or setting things to False). Feel free to submit a new PR if you can manage to get that working.`state`

I have tried some ways before I decide to fix it this way, which is not really disabling extensions. The way you used to load an extension is directly import them, which cannot be really deleted unless a complete restart. I have noticed that there is a state value for each extensions, and have tried to use `del loaded_extension.script` and `state.pop(extension_name_if_state[name][0])`, but it always failed to load extensions which are unloaded before the auto restart, and keep getting `Attribute error: extension_name has no attribute script` when running `get_attr(extensions, name).script` even if it is right below `exec(f"import extensions.{name}.script")`. Maybe the simpler way you mentioned does not exist.

oobabooga:
Yes, unimporting the python module will not work. The best that can be done is remove the disabled extensions from the `iterator()`.

Touch-Night:
> Yes, unimporting the python module will not work. The best that can be done is remove the disabled extensions from the `iterator()`.

Done in #3965

PR #3852, opened 2023-09-09
Exllama new rope settings

Ph0rk0z:
The latest builds of exllama read the rope base from the model config file. They will now apply alpha value to the inferred base. This means that the current settings will produce 200k context since base is converted to an alpha value.

This PR makes alpha value apply to the read base and writing the actual theta value to use that value as an override. It helps since some tunes were not run at 1e6 and produce better perplexity with different values. The other alternative would be to allow alpha below 1 but I think that is more trouble.

code:

```
        if shared.args.alpha_value > 1 and shared.args.rope_freq_base == 0:
            config.alpha_value = shared.args.alpha_value
            config.calculate_rotary_embedding_base()
        elif shared.args.rope_freq_base > 0:
            config.rotary_embedding_base = shared.args.rope_freq_base
```

### DIFF
--- a/modules/exllama.py
+++ b/modules/exllama.py
@@ -56,9 +56,11 @@ class ExllamaModel:
             config.set_auto_map(shared.args.gpu_split)
             config.gpu_peer_fix = True
 
-        if shared.args.alpha_value > 1 or shared.args.rope_freq_base > 0:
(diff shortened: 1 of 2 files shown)


SynthpX:
Does that mean we also need to include 
rope_freq_base and rms_norm_eps on config model ? Otherwise it wont load.

oobabooga:
Thank you, that's a very nice fix.

PR #3853, opened 2023-09-09
fix lora training with alpaca_lora_4bit

johnsmith0031:
Fixed the 4bit lora training feature with monkey-patch for pip installable version of alpaca_lora_4bit (which is the main branch now)

## Checklist:


oobabooga:
Thank you so much for the update, @johnsmith0031.

PR #3870, opened 2023-09-11
Handle Chunked Transfer Encoding in `openai` Extension for Streaming Requests

mcc311:
## Checklist:

**Description:**

- Context: This commit pertains to the --extension openai of this repository.

- Problem Addressed: Previous versions of the extension expected the Content-Length header to be always present, which is not the case for streaming requests that use the Transfer-Encoding: chunked header.

- Main Changes:

  - Added checks to safely retrieve the Content-Length header.
  - Introduced support for processing requests with Transfer-Encoding: chunked. This is crucial for ensuring compatibility with streaming requests.
  - For requests that don't have either the Content-Length or Transfer-Encoding header, a 400 Bad Request is returned to ensure proper client-server contract.
- Additional Updates:
Enhanced diagnostic logs for a clearer understanding of incoming headers.
- Impact:
With these changes, the --extension openai can now seamlessly handle both regular and streaming requests. This ensures broader compatibility and reduces potential server errors due to missing headers.

### DIFF
--- a/extensions/openai/script.py
+++ b/extensions/openai/script.py
@@ -139,8 +139,24 @@ class Handler(BaseHTTPRequestHandler):
         debug_msg(self.requestline)
         debug_msg(self.headers)
 
-        content_length = int(self.headers['Content-Length'])
-        body = json.loads(self.rfile.read(content_length).decode('utf-8'))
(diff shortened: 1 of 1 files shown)


oobabooga:
Thank you so much for the fix.

PR #3873, opened 2023-09-11
Read GGUF metadata

oobabooga:


Cebtenzzre:
With this change, I can no longer use LLaMA-2 models with less than 4096 ctx. Why is it enforcing a minimum?

Also, why not use the official 'gguf' python module?

Or better, you could provide a way to make the values 'unset' from the UI, so they are not passed to Llama.\_\_init\_\_() and the defaults are read from the GGUF in llama.cpp itself. If you want a way to display the resulting values in the UI, I can add them to the llama.cpp API for you - right now they're only printed to stdout.

PR #3877, opened 2023-09-12
Read more GGUF metadata (scale_linear and freq_base)

berkut1:
Thank you for adding the GGUF metadata reader.

I just added that it uses more metadata for loaded models.
Also I have added throw exception for GGUF v1. Because metadata reader can't read them. I didn't dig into why, because there are only 8 GGUFv1 models and they will be deprecated in October, so I think it's better to just warn people to update their model rather than showing random errors.

Also I was thinking of adding to read the `general.architecture` metadata and remove 90% of the information from `config.yaml` and leave only the template data, but I'm not sure. If the file  `config.yaml` is being used by other loaders, then they may still need some piece of information, such as the model's architecture.

## Checklist:


modules/models_settings.py
```diff
@@ -48,6 +49,10 @@ def get_model_metadata(model):
         metadata = metadata_gguf.load_metadata(model_file)
         if 'llama.context_length' in metadata:
             model_settings['n_ctx'] = metadata['llama.context_length']
+        if 'llama.rope.scale_linear' in metadata:
+            model_settings['compress_pos_emb'] = metadata['llama.rope.scale_linear']
```
oobabooga, line 53: Should this be `1 / metadata['llama.rope.scale_linear']`? iirc llama.cpp uses fractional values like 0.5 and 0.25, while the UI expects integer values like 2 and 4.

models/config.yaml
```diff
@@ -223,7 +220,5 @@ llama-65b-gptq-3bit:
 .*(openorca-platypus2):
   instruction_template: 'OpenOrca-Platypus2'
   custom_stopping_strings: '"### Instruction:", "### Response:"'
-.*codellama:
-  rope_freq_base: 1000000
```
oobabooga, line 227: It's good to leave the `truncation_length`s and `rope_freq_base`s as they are as they are used by transformers and ExLlama-v1 respectively.

modules/models_settings.py
berkut1: @oobabooga no, it represents an int, not a float.
So if the model is 16k, from the metadata we get `llama.rope.scale_linear = 4`

models/config.yaml
berkut1: Okay, I thought that we automatically get `truncation_lengths` from n_ctx for all loaders.

oobabooga:
Looks perfect now, thank you!

lee-b:
Does not seem to be a GGUF v1 issue.  I'm getting the same error on https://huggingface.co/TheBloke/dolphin-2.7-mixtral-8x7b-GGUF.

github-lly:
> Does not seem to be a GGUF v1 issue. I'm getting the same error on https://huggingface.co/TheBloke/dolphin-2.7-mixtral-8x7b-GGUF.

@lee-b How did you fix that? I got the same error?

lee-b:
> > Does not seem to be a GGUF v1 issue. I'm getting the same error on https://huggingface.co/TheBloke/dolphin-2.7-mixtral-8x7b-GGUF.
> 
> @lee-b How did you fix that? I got the same error?

In my case I had skipped the lfs download of quants that I didn't need, and only git lfs pull'd the q6 or q5_k_m quant that I wanted.  It was presumably trying to load the invalid (git lfs placeholder) quant files by default, so I simply deleted the other gguf files, and then it loaded the right/only one.

PR #3885, opened 2023-09-12
Add a proper documentation

oobabooga:
I'm making it a PR otherwise I'll only finish this in 2024.


netrunnereve:
Can we add these to the actual Gradio interface directly? Maybe have a help button at the bottom of each tab which will automatically pop up the corresponding help age using the builtin markdown viewer.

Veda-Swaroop:
Hi, first of all, huge thanks for this project. I tried to download Metas's Llama 2 models from hugging face. These models are gated models. In order to download them, require access from Meta. I have access from Meta and hugging face. I can download them manually from hugging face. When i tried to download them using text-generation-web Ui, got authorization error. There is no clear documentation on how to download gated model from hugging face. I searched online and finally checked the code of download-model.py file in repo. 

The code is designed to handle authorization requests from hugging face. I am using Windows machine. I generated access token on hugging face. I created an environment variable called "HF_TOKEN" as variable name and the "token number (generated on hugging face)" as variable value. After doing this, I am able to download the gated model using text-generation-web Ui. I wonder why this was not documented in the repo documentation. Please consider this issue regarding documentation.

PR #3897, opened 2023-09-13
Fix NTK (alpha) and RoPE scaling for exllamav2 and exllamav2_HF

Panchovix:
## Checklist:

Small fix discussed with @turboderp, since it uses different config.names for both alpha for NTK scaling and RoPE scaling. (Now they're on config.py)

Fixes gibberish >4096 ctx and alpha values greater than 1 for llama2 based models. >2048 respectively for llama1 based models.

Fixes gibberish for 4k/8k/16k/etc ctx and their respective embedding compression.


oobabooga:
Thank you!

PR #3899, opened 2023-09-13
Add customizable ban tokens

sALTaccount:
## Checklist:

This PR adds a feature where you can input token IDs that you want to ban from generating. It uses the same logic as the EOS token ban. This feature is useful for controlling the generations of models. For example, I wanted to prevent the model from describing the actions in RP such as `*looks down*`, so I put all tokens IDs containing a `*` in the ban box, and the model can no longer generate those actions.

I've tested that this works for all exllama (including HF ones), autoGPTQ, GPTQ for llama, and llama.cpp. I've also tested that it is accessible via the API


Ph0rk0z:
oh boy! finally logits banning

BadisG:
How do we find the ID of the token we want to ban though?

BadisG:
I tried it and it works fine, good job.
Maybe it would be a bit more convenient to have the option to save the tokens we want to ban, like having it on the presets save or something like that? I'm not a big fan of keeping all those ID in a notebook and then paste it on the webui everytime I want to run it.

sALTaccount:
Token bans can now be defined by presets. Didn't realize that wasn't working earlier

sALTaccount:
@oobabooga what do you think of this PR?

oobabooga:
I think it's perfect. It's nice that you dug through each loader, including the non _HF ones, and the PR is very clean overall. Thank you!

PR #3918, opened 2023-09-14
openai: instruction-templates: format system messages inside context …

Chainfire:
…template

Adds capability to use `<|system-message|>`  inside the `context` template setting. If present, it will be replaced by the current set of system messages. If present, but no system messages are set, falls back to the new `system_message_default` template setting.

If `<|system-message|>` is not present inside the `context` template setting, the code behaves exactly as it did before.

## Checklist:

## Reasoning

Using the original CodeLlama-34b (rather than any of its finetunes which seem to use a different prompt format), this allows for the following template:

```
user: ""
bot: ""
turn_template: "<|user|><|user-message|>\n[/INST]\n<|bot|><|bot-message|>\n[INST]\n"
context: "[INST]\n<<SYS>>\n<|system-message|>\n<</SYS>>\n\n"
system_message_default: "Answer the questions."
```

... which performs significantly better in my tests than the Llama-v2 template (or any of the others I've tried) when system messages are used through the openai API. It would possibly have the same effect for other Llama-v2 models, but I haven't tested those.

Side note: for my tests I've also removed `</s><s>` from the template, and disabled `add_bos_token` from both settings as well as model; now it performs even better (contrary to what *should* be)

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -163,13 +163,15 @@ def messages_to_prompt(body: dict, req_params: dict, max_tokens):
         req_params['stopping_strings'] = []
 
     # Instruct models can be much better
+    system_message_default_fill = ''
     if shared.settings['instruction_template']:
         try:
             instruct = yaml.safe_load(open(f"instruction-templates/{shared.settings['instruction_template']}.yaml", 'r'))
 
             template = instruct['turn_template']
             system_message_template = "{message}"
-            system_message_default = instruct.get('context', '') # can be missing
+            system_message_default = instruct.get('context', '').replace('<|system-message|>', '{system}') # can be missing
+            system_message_default_fill = instruct.get('system_message_default', '')  # probably missing
             bot_start = template.find('<|bot|>')  # So far, 100% of instruction templates have this token
             user_message_template = template[:bot_start].replace('<|user-message|>', '{message}').replace('<|user|>', instruct.get('user', ''))
             bot_message_template = template[bot_start:].replace('<|bot-message|>', '{message}').replace('<|bot|>', instruct.get('bot', ''))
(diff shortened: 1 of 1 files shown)


matatonic:
This is a great approach, I had almost given up on this and was going to just make system messages into 'user' messages.

Does this require replacing all the template code and updating all the templates? (This is what I had assumed would be needed - and seems required to do 'the right thing' with system messages.

Chainfire:
Hmm, this conflict is new, it used to merge. Not going to look into that until one of the authors comments on this though.

Existing templates do not require modification with my changes, it was coded like this specifically for that reason. Otherwise I would have cleaned up a bit.

oobabooga:
I have implemented it here for the entire web ui, and it also applies to the openai API: https://github.com/oobabooga/text-generation-webui/pull/4499

Here is a test:

```
  curl -k http://127.0.0.1:5000/v1/chat/completions \
    -d '{
      "model": "gpt-3.5-turbo",
      "messages": [
        {
          "role": "system",
          "content": "You are a duck."
        },
        {
          "role": "user",
          "content": "Hello!"
        }
      ]
    }'

```

Output:

```
{
    "id": "chatcmpl-1699398280587623424",
    "object": "chat.completions",
    "created": 1699398280,
    "model": "llama-2-13b-chat-hf",
    "choices": [
        {
            "index": 0,
            "finish_reason": "stop",
            "message": {
                "role": "assistant",
                "content": "QUACK! *flap flap* Oh hello there, fellow! *quack* It's a lovely day here in the pond, isn't it? *waddle waddle* The sun is shining and the water is so calm and peaceful. *quack* What brings you to our neck of the woods today? *flap flap* Maybe you'd like to join me for a swim or a snack? *quack* I'm always up for some good company! *waddle waddle*"
            }
        }
    ],
    "usage": {
        "prompt_tokens": 30,
        "completion_tokens": 118,
        "total_tokens": 148
    }
}
```

PR #3919, opened 2023-09-14
Adjust model variable if it includes a hf URL already

kalomaze:
## Checklist:

- [✅ ] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

Just does a basic change to the variable if it's already prefaced with the hf url.

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -22,6 +22,9 @@ from requests.adapters import HTTPAdapter
 from tqdm.contrib.concurrent import thread_map
 
 
+base = "https://huggingface.co"
+
+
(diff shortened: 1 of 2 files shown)


oobabooga:
What is the scenario in which this is needed? Also the replacement should not be inside a while loop.

kalomaze:
It just seemed a minor annoyance that I couldnt just copy paste the hf URL in one go, and you have to specifically remove the first part of the URL; not explicitly necessary but just seemed slightly less arbitrary

PR #3921, opened 2023-09-14
feat: add option to follow system theme

wywywywy:
## Checklist:

This PR adds a new setting `theme` that introduces the ability for WebUI to follow the system theme. So if the browser is dark, WebUI is dark, and if the browser is light, WebUI is light.

The valid values are **system** (default), **light**, and **dark**. 

Perhaps in the future we can introduce additional themes such as "high_contrast".

This also deprecates the old `dark_theme` setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ When you load this model in default or notebook modes, the "HTML" tab will show 
 
 Then browse to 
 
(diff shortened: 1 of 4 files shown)


oobabooga:
I prefer the current options, as I don't use or maintain the light theme.

Issue #3905, opened 2023-09-13
ImportError: cannot import name 'Undefined' from 'pydantic.fields' (E:\oobabooga_windows\installer_files\env\lib\site-packages\pydantic\fields.py)


ForwardForward:
### Describe the bug

Windows 10 Home. Install by start_windows.bat from zip archive for Windows from GittHub page. Say that have GPU and NVidia card.

After installation receive next message:

*******************************************************************
* WARNING: You haven't downloaded any model yet.
* Once the web UI launches, head over to the bottom of the "Model" tab and download one.
*******************************************************************

Traceback (most recent call last):
  File "E:\oobabooga_windows\text-generation-webui\server.py", line 12, in <module>
    import gradio as gr
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\__init__.py", line 3, in <module>
    import gradio.components as components
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\gradio\components.py", line 32, in <module>
    from fastapi import UploadFile
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\fastapi\__init__.py", line 7, in <module>
    from .applications import FastAPI as FastAPI
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\fastapi\applications.py", line 16, in <module>
    from fastapi import routing
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\fastapi\routing.py", line 22, in <module>
    from fastapi import params
  File "E:\oobabooga_windows\installer_files\env\lib\site-packages\fastapi\params.py", line 4, in <module>
    from pydantic.fields import FieldInfo, Undefined
ImportError: cannot import name 'Undefined' from 'pydantic.fields' (E:\oobabooga_windows\installer_files\env\lib\site-packages\pydantic\fields.py)
Для продолжения нажмите любую клавишу . . .

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Windows 10 Home. Install by start_windows.bat from zip archive for Windows from GittHub page. Say that have GPU and NVidia card.

### Logs

```shell
I don't have
```

### System Info

```shell
Intel CPU, NVidia RTX3060
```

aetherwu:
I've encountered the same today. But it works fine on my another PC.

Shark-Eater:
I am also experiencing this same issue

alicat22:
[...]

PR #3927, opened 2023-09-14
Fix pydantic version conflict in elevenlabs extension

jllllll:
## Checklist:

---
It seems that the `elevenlabs` package updated to using `pydantic` 2.0 in [0.2.25](https://github.com/elevenlabs/elevenlabs-python/releases/tag/v0.2.25), which would require `gradio` 3.37.0 and `gradio_client` 0.2.10 to work.

Tried updating the webui with those versions and it just breaks the UI layout.

Fixes: #3905


oobabooga:
Nice, I wasn't aware of that.

PR #3913, opened 2023-09-14
Make new chat menu optional

missionfloyd:
## Checklist:
- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

Adds an `--old-chat-menu` flag.

Fixes #3907

PR #3935, opened 2023-09-14
Make new chat menu optional

missionfloyd:
## Checklist:

I fixed the menu alignment. (https://github.com/oobabooga/text-generation-webui/pull/3913#issuecomment-1719492424)

https://github.com/oobabooga/text-generation-webui/assets/4073789/4c573a1e-412a-4676-9ce1-67da16bfbe40


oobabooga:
I still don't think that this is a good solution. A compromise I can think of is to create new `Regenerate` / `Continue` / `Remove last` buttons similar to this on LM Studio:

test

These buttons would be defined in a new `gr.HTML` element that would be shown on desktop only. It could be placed on the same line as the typing dots and the "Show controls" checkbox.

This way, the most used buttons would be on display while keeping the rest of the new design intact.

BadisG:
We just want to see the buttons back, like it was before, that's all we're asking

missionfloyd:
Something like that?
image

oobabooga:
No, I was thinking of something similar to this:

concept

The 3 new buttons would be custom HTML buttons with new CSS styles, not Gradio buttons, and their `onclick` would be javascript events that click on the main buttons, for instance:

```
document.getElementById('Regenerate').click();
```

PR #3937, opened 2023-09-15
Fix use of functools decorators before Python 3.9

### LINKED ISSUE #3808: AttributeError: module 'functools' has no attribute 'cache'
### Describe the bug

when i use server.py, i met the error: AttributeError: module 'functools' has no attribute 'cache', do you know how to fix it? Thx

### Reproduction

AttributeError: module 'functools' has no attribute 'cache

### Screenshot

_No response_

### Logs

```shell
AttributeError: module 'functools' has no attribute 'cache
```

### System Info

```shell
AttributeError: module 'functools' has no attribute 'cache
```

vittorio88:
Add check on python version prior to using functools cache decorators to provide compatability with Python versions earlier than 3.9.   
I tested successfully with Python 3.8 on Ubuntu 20.04.

Fixes #3808 

## Checklist:

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -4,6 +4,7 @@ import functools
 import html
 import json
 import re
+import sys
 from pathlib import Path
 
 import gradio as gr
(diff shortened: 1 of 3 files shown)


oobabooga:
The recommended python version is 3.10.9. It is better to install a different version using conda than to rely on the system's python.

PR #3938, opened 2023-09-15
Move Regenerate, Remove Last, and Continue out of hover menu.

missionfloyd:
## Checklist:

Since I already have it working...

It also moves some styles from JS to CSS, and adjusts the hover menu shortcut text.
image
These could probably be moved and restyled to fit the way you want them.

Personally, I prefer the old UI. Moving everything to a menu is just too big of a change.


ningchaoar:
I prefer the old too, especially when I meet shortcut conflicts on mac.

PR #3941, opened 2023-09-15
Allow custom tokenizer for llamacpp_HF loader

JohanAR:
If the model is specified as a directory containing a .gguf file together with a tokenizer it will prefer this one before falling back to using the default.

## Checklist:


oobabooga:
Very nice, this was actually on my todo list and your solution is much cleaner than what I was considering.

I have changed the check to look for 'special_tokens_map.json', 'tokenizer.model' as those files are also needed.

PR #3947, opened 2023-09-16
Better solution to chat UI

missionfloyd:
## Checklist:

A more elegant method of making the hover menu optional.

- Existing buttons are arranged (similar to) how they were before. If `--chat-buttons` is set, hover menu creation is skipped and its button is hidden.
- Doesn't duplicate code like my previous attempt. 
- Buttons are aligned like they were before and `Show Controls` functions as expected.
- Generate/Stop remains where it is in both modes.

image


oobabooga:
That's a clean solution that does not interfere with the new style and does not add complexity to the code base. Nicely done.

I respect your determination :)

BadisG:
@missionfloyd Thank you for that, seriously. It means a lot to most of us. 
image

PR #3951, opened 2023-09-16
Move hover menu shortcuts to right side

missionfloyd:
## Checklist:

Also removes the parentheses and moves some styles form javascript to CSS.

Before/After
image image

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -214,6 +214,12 @@ for(i = 0; i < textareaElements.length; i++) {
 }
 
 //------------------------------------------------
+// Improve the looks of the chat input field
+//------------------------------------------------
+let isOld = document.querySelectorAll('.old-ui').length > 0;
+document.getElementById('stop').parentElement.parentElement.parentElement.style.paddingBottom = isOld ? '0px' : '20px';
+
+//------------------------------------------------
 // Remove some backgrounds
 //------------------------------------------------
 const noBackgroundelements = document.querySelectorAll('.no-background');
--- a/js/main.js
+++ b/js/main.js
@@ -227,7 +227,7 @@ for(i = 0; i < noBackgroundelements.length; i++) {
 // The show/hide events were adapted from:
 // https://github.com/SillyTavern/SillyTavern/blob/6c8bd06308c69d51e2eb174541792a870a83d2d6/public/script.js
 //------------------------------------------------
-var buttonsInChat = document.querySelectorAll("#chat-tab #chat-buttons:not(.old-ui) button");
+var buttonsInChat = document.querySelectorAll("#chat-tab:not(.old-ui) #chat-buttons button");
 var button = document.getElementById('hover-element-button');
 var menu = document.getElementById('hover-menu');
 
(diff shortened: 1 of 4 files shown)


js/main.js
```diff
@@ -213,30 +213,6 @@ for(i = 0; i < textareaElements.length; i++) {
...
-document.getElementById('chat-input').parentNode.style.border = 'none';
-document.getElementById('chat-input').parentElement.parentElement.style.minWidth = 0;
-
-document.getElementById('stop').parentElement.parentElement.style.minWidth = 0;
-document.getElementById('stop').parentElement.parentElement.style.display = 'flex';
-document.getElementById('stop').parentElement.parentElement.style.flexDirection = 'column-reverse';
-document.getElementById('stop').parentElement.parentElement.style.paddingBottom = '3px';
-document.getElementById('stop').parentElement.parentElement.parentElement.style.paddingBottom = isOld ? '0px' : '20px';
-
-document.getElementById('stop').parentElement.parentElement.style.flex = '0 0 auto';
```
oobabooga, line 231: This style is missing. Without it, the Generate button shrinks and becomes smaller than the text on mobile.

js/main.js
```diff
@@ -213,30 +213,6 @@ for(i = 0; i < textareaElements.length; i++) {
...
-
-document.getElementById('chat-input').parentNode.style.background = 'transparent';
-document.getElementById('chat-input').parentNode.style.border = 'none';
-document.getElementById('chat-input').parentElement.parentElement.style.minWidth = 0;
-
-document.getElementById('stop').parentElement.parentElement.style.minWidth = 0;
-document.getElementById('stop').parentElement.parentElement.style.display = 'flex';
-document.getElementById('stop').parentElement.parentElement.style.flexDirection = 'column-reverse';
-document.getElementById('stop').parentElement.parentElement.style.paddingBottom = '3px';
-document.getElementById('stop').parentElement.parentElement.parentElement.style.paddingBottom = isOld ? '0px' : '20px';
```
oobabooga, line 229: I'm a bit wary of removing this line, as it moves the job of adding the 20px padding to this gradio event: https://github.com/oobabooga/text-generation-webui/blob/main/server.py#L133

I don't know if there is a network situation where this could lag and cause the UI to flicker on startup.

oobabooga:
The changes are very good overall, I just have these 2 concerns.

missionfloyd:
I put those styles back and moved some more styles and classes, eliminating the need for [`isOld`](https://github.com/oobabooga/text-generation-webui/blob/0668f4e67fe158a385eff87ace9d0676d657df20/js/show_controls.js#L3).

oobabooga:
1) The flex: 0 auto was not being applied, but I fixed it with `!important`.

2) The scrolling bar lost its custom style:

broken

3) Autoscrolling to the bottom seems to not work anymore.

4) `--chat-buttons` looks weird on mobile. The buttons used to be on display in this case:

test

5) The buttons also used to be smaller:

small

It is useful to launch the web UI before and after the changes and go back and forth to make sure that it's identical.

missionfloyd:
Okay, I fixed those, except for the scrolling. It's not working on main, either.

oobabooga:
I think it's all equivalent now. Thank you for this, that's a lot cleaner than styling through javascript (I wasn't aware that columns could have IDs in gradio).

PR #3953, opened 2023-09-16
[extensions/openai] load extension settings via `settings.yaml`

wangcx18:
Just utilize the `dotenv` package to load environment variables from the `.env` file. Since the `openai` extension relies on specific environment variables, manually entering them each time the server starts can be cumbersome. Using `.env` will significantly streamline this process and save valuable time.

## Checklist:


wangcx18:
Hi! @oobabooga @matatonic Could you have a quick review of my pull request? Thanks!

sALTaccount:
I'm a fan of .env but I believe that files starting with . are hidden files by default, do you think it might be less confusing for users to just have something like "openai-auth.yaml" and then we load the environment vars from that?

matatonic:
Hi, thanks for the contribution. As @sALTaccount,  I also prefer the use of a yaml file for config and have planned to add something like that in the future - agreed that the config values are getting a bit out of hand, and I also want to add authorization via bearer token and whisper (with model options) at some point too. I think it's possible to load config values via the main config, and I will do that if it's convenient but I haven't spent time to figure that out yet.

Also, I am currently on holidays and will be back and more active again around the end of Sept.

wangcx18:
@sALTaccount @matatonic Thank you for your valuable feedback!  I've made the necessary edits to align with your advice. The extension now can load settings from 'settings.yaml,' following the method documented by @oobabooga. Additionally, it respects the environment variables; settings won't apply if the environment variables are set.

matatonic:
> @sALTaccount @matatonic Thank you for your valuable feedback!  I've made the necessary edits to align with your advice. The extension now can load settings from `settings.yaml`, following the method documented by @oobabooga. Additionally, it respects the environment variables; settings won't apply if the environment variables are set.

that's awesome, thank you!

matatonic (approved):

noneabove1182:
When trying to prepare the cache I get an error on the new import to cache_embedding_model.py

```
 > [builder 10/11] RUN python3 text-generation-webui/extensions/openai/cache_embedding_model.py:
2.415 Traceback (most recent call last):
2.415   File "//text-generation-webui/extensions/openai/cache_embedding_model.py", line 9, in <module>
2.415     from extensions.openai.script import params
2.415 ModuleNotFoundError: No module named 'extensions'
```

any idea why? seems to work in the other files so not sure why it wouldn't work here.

wangcx18:
Hi! @noneabove1182 Thanks for reporting the error. It appears that `cache_embedding_model.py` is intended for use during Docker image building, not after the server starts, which sets it apart from other files. This distinction explains why the other files function correctly while this one does not, a detail I overlooked in this pull request.

PR #3954, opened 2023-09-16
Fix exllama tokenizers

sALTaccount:
## Checklist:

Currently, with exllama, BOS is never added no matter what settings are chosen


oobabooga:
Thank you, I had also noticed the lack of a `1` in tokenized ExLlama inputs after adding the UI tokenizer yesterday. That's a clean fix.

PR #3957, opened 2023-09-16
token probs for non HF loaders

sALTaccount:
Right now, your console gets flooded with errors if you try to use the new Logits tab with non-HF model loaders. It doesn't seem to be (at least easily) possible to hijack the exllama samplers as they are written in C, so I was only able to implement for without the "use samplers" checkbox

I think the bigger part of this PR is that it adds proper error handling with messages telling the user what went wrong (etc they tried to use logits + sampler on a non HF loader), along with an error message telling the user they don't have a model loaded

### DIFF
--- a/modules/exllama.py
+++ b/modules/exllama.py
@@ -192,3 +192,8 @@ class ExllamaModel:
 
     def decode(self, string, **kwargs):
         return self.tokenizer.decode(string)[0]
+
+    def get_logits(self, token_ids, **kwargs):
+        self.cache.current_seq_len = 0
+        self.model.forward(token_ids[:, :-1], self.cache, input_mask=None, preprocess_only=True)
+        return self.model.forward(token_ids[:, -1:], self.cache, **kwargs).float().cpu()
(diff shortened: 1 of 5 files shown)


kalomaze:
I get the same issues with llama.cpp so it seems like both major loaders have the issue. Which begs the question of what loader was even used if both exllama and llama.cpp don't work...

oobabooga:
Transformers and the _HF loaders are the main ones.

oobabooga:
It should be possible to extend this to llama.cpp using the code in this block:

https://github.com/oobabooga/text-generation-webui/blob/main/modules/llamacpp_hf.py#L122

sALTaccount:
@oobabooga llama.cpp_HF already works with the logits tab, its the normal llama.cpp that doesn't work. I think it should be possible to hijack the non llama.cpp sampler, but it looks decently painful so I'm leaving that out of this PR.

sALTaccount:
i added non-sampler llamacpp support though, and added proper error handling for llamacpp

oobabooga:
Thanks a lot @sALTaccount

PR #3961, opened 2023-09-17
Training PRO extension

FartyPants:
I decided that the BEST approach for an enhanced Lora Training would be to make it as its own extension as it will allow me to add more stuff at moments notice without fractured PR' (it's already too much to cut it into digestable pieces and not breaking it). It would also allow other people to contribute to training in a much more contained code base.

@oobabooga, you can decide if you want to add it to the main distribution or not. If not, I'll add it to the extension repo then. Let me know.

Also the Lora Training in main can be now simplified and cleaned as we previously discussed elsewhere prompted by @oobabooga particularly simplifying the RAW part.

- I would suggest removing Hard cut and Ignore small blocks - making the RAW just a pure RAW training in the main as @oobabooga suggested previously. These can be then freely enhanced in the PRO extension.
- I would suggest removing stop at loss and then return logging_steps = 5 which will make it less info heavy and intimidating and people questioning "Why my training ended up too early?"
- I would suggest removing Add EOS and just let user to decide if they want EOS or not by adding /s to the dataset.
- there could be some sliders that people should not be touching anyway. LoRA Dropout? Optimizer? Is anyone meddling with these?

... note many above were my changes so I know how to take them out cleanly.  If you want me to tighten a bit the training.py before you have a go, let me know.

So the idea that train tab could be lean and not overpowering with unnecessary options. and then every detail will go to Training PRO extension for the people who care.

- the extension mirrors the training.py as much as possible with major additions extracted into separate py files so it is easy to keep in sync

Thoughts?

## Checklist:


oobabooga:
I think that's a great idea @FartyPants. This way you can iterate on your own training code and review PRs to `training_PRO` yourself without depending on me, thus accelerating the development. In the future, I can take parts of your code and merge into main once I get a better grasp at training.

I'll try to go simplify the training code in `main` myself in the coming weeks and will ask for your help if there is something I don't understand.

mcmonkey4eva:
`I would suggest removing Hard cut` I disagree with that part in particular - hard cut is extremely useful and has been the primary tool for dataset management I've recommended to people. It makes it extremely easy to make your own dataset in whatever format you want without having to learn any of the "proper" tools like the json format files.

oobabooga:
I think that this parameter can just be set to a null value by default, along with `Overlap Length`, so as to make the UI training code more in line with https://github.com/artidoro/qlora/blob/main/qlora.py.

FartyPants:
> `I would suggest removing Hard cut` I disagree with that part in particular - hard cut is extremely useful and has been the primary tool for dataset management I've recommended to people. It makes it extremely easy to make your own dataset in whatever format you want without having to learn any of the "proper" tools like the json format files.

I agree, of course, it's my most used feature, but it can be further enhanced in the Extension and simplified in main. 

The problem is, I was trying to cut back on adding more params in main Training as they would ultimately result in a sliders salad. This way , I don't have to , and I can go WILD.

PR #3972, opened 2023-09-17
Simplified ExLlama cloning instructions and failure message

jamesbraza:
- Simplifies the ExLlama cloning instructions in the README
- Including repo path in import failure message

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,7 @@ pip install -r requirements_nocuda.txt
 
 ```
 cd text-generation-webui
-mkdir repositories
-cd repositories
-git clone https://github.com/turboderp/exllama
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -101,9 +101,7 @@ pip install -r requirements_nocuda.txt
 
 ```
 cd text-generation-webui
-mkdir repositories
```
oobabooga, line 104: repositories is not guaranteed to exist

jamesbraza: Yep, check the last part of this line: `git clone https://github.com/turboderp/exllama repositories/exllama`

It will create `repositories/` if it doesn't already exist.  I thought of this and tested it out to work

oobabooga: Nice

PR #3985, opened 2023-09-18
Fix double gap when chat-buttons is hidden

missionfloyd:
## Checklist:

In hover menu mode, the hidden chat buttons leave a double-wide gap in their place.

Before:
image

After:
image


oobabooga:
The double gap looks fine to me in this case.

PR #3986, opened 2023-09-18
Add support for WhisperX via OpenAI extension

wizd:
## Checklist:

[WhisperX](https://github.com/m-bain/whisperX): Automatic Speech Recognition with Word-level Timestamps (& Diarization)

@matatonic

```
curl http://localhost:5001/v1/audiox/transcriptions \
  -H "Authorization: Bearer sk_aaa" \
  -H "Content-Type: multipart/form-data" \
  -F file="@/mnt/c/tmp/test.mp3" \
  -F model="tiny"

{"segments": [{"start": 0.19, "end": 0.653, "text": " Nothing.", "words": [{"word": "Nothing.", "start": 0.19, "end": 0.653, "score": 0.908, "speaker": "SPEAKER_00"}], "speaker": "SPEAKER_00"}], "word_segments": [{"word": "Nothing.", "start": 0.19, "end": 0.653, "score": 0.908, "speaker": "SPEAKER_00"}]}
```

### DIFF
--- a/extensions/openai/README.md
+++ b/extensions/openai/README.md
@@ -168,31 +168,34 @@ print(text)
 
 ## Compatibility & not so compatibility
 
-| API endpoint              | tested with                        | notes                                                                       |
-| ------------------------- | ---------------------------------- | --------------------------------------------------------------------------- |
-| /v1/chat/completions      | openai.ChatCompletion.create()     | Use it with instruction following models                                    |
-| /v1/embeddings            | openai.Embedding.create()          | Using SentenceTransformer embeddings                                        |
-| /v1/images/generations    | openai.Image.create()              | Bare bones, no model configuration, response_format='b64_json' only.        |
-| /v1/moderations           | openai.Moderation.create()         | Basic initial support via embeddings                                        |
-| /v1/models                | openai.Model.list()                | Lists models, Currently loaded model first, plus some compatibility options |
-| /v1/models/{id}           | openai.Model.get()                 | returns whatever you ask for                                                |
-| /v1/edits                 | openai.Edit.create()               | Deprecated by openai, good with instruction following models                |
-| /v1/text_completion       | openai.Completion.create()         | Legacy endpoint, variable quality based on the model                        |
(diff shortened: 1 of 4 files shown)


matatonic:
audiox is not an openai API endpoint, is this intended to be better than the existing audio endpoint? If so, then it should replace it. Is it somehow better? I haven't tried either yet.

wizd:
> audiox is not an openai API endpoint, is this intended to be better than the existing audio endpoint? If so, then it should replace it. Is it somehow better? I haven't tried either yet.

The most prominent feature of whisperX is that it can segment text according to time, making it convenient to create video subtitles.

wizd:
Adding it as an extension to the OpenAI API is a convenient approach because the calling method remains the same, only with a slightly different URL. Another approach is to create a separate extension module.

matatonic:
> Adding it as an extension to the OpenAI API is a convenient approach because the calling method remains the same, only with a slightly different URL. Another approach is to create a separate extension module.

The different URL means it doesn't work with anything designed to work with OPENAI API though. I understand the appeal of putting similar methods together, but I think this doesn't fit with this extension as is (it's goal is compatibility with existing OpenAI API clients). I would suggest you change it to either:

1) Be compatible with the OpenAI API and propose to replace the existing audio API (which unfortunately I cannot say much about, because I have not yet had time to test either yet, but will have more time later in Sept). 

2) Contribute this code to the native blocking API as an enhancement instead.

wizd:
> > Adding it as an extension to the OpenAI API is a convenient approach because the calling method remains the same, only with a slightly different URL. Another approach is to create a separate extension module.
> 
> The different URL means it doesn't work with anything designed to work with OPENAI API though. I understand the appeal of putting similar methods together, but I think this doesn't fit with this extension as is (it's goal is compatibility with existing OpenAI API clients). I would suggest you change it to either:
> 
> 1. Be compatible with the OpenAI API and propose to replace the existing audio API (which unfortunately I cannot say much about, because I have not yet had time to test either yet, but will have more time later in Sept).
> 2. Contribute this code to the native blocking API as an enhancement instead.

I understand your point very well. I think it is more appropriate to add this feature to the Native API.

PR #4022, opened 2023-09-21
Multiple histories for each character

oobabooga:


netrunnereve:
>     * Removed Save Session / Load Session menus, as they were not very useful.

Personally I find session saving extremely useful as it's a snapshot of literally everything. I can save the chat and all parameters, turn off my computer, and return exactly to where I was the next day. I can also send sessions to other people and have them reproduce my exact conversation and settings.

Without this feature I'll have to export a chat log, character/template, and sampler preset to capture the exact configuration.

oobabooga:
I removed it mostly because it was conflicting with the new history saving logic. I also found this feature kind of weird because "Preset" and "Character" are not fields that can be set independently of "temperature" and "Context" (respectively), since a change in those dropdowns triggers a change in the latter values.

I can try adding it back later. If you want to do it yourself and send a PR, that would be helpful.

PR #4032, opened 2023-09-22
Fix 'train only after' not working

### LINKED ISSUE #4031: Train only after doesn't work as DataCollator overwrites labels (incl. -100) with input_ids
### Describe the bug

So, in attempt to mask loss for Train only after training training.py sets the labels belonging to before tokens to -100s.

https://github.com/oobabooga/text-generation-webui/blob/ee7bf49804bfb53775424d4c12294445497b610f/modules/training.py#L341-L358

However, the sus thing was that it sets the rest of the labels to 1s — which would be a nonsense. Yet, training happens. Checking it, I found that this still makes sense thanks to the fact that the code uses the DataCollatorForLanguageModeling.

https://github.com/oobabooga/text-generation-webui/blob/ee7bf49804bfb53775424d4c12294445497b610f/modules/training.py#L592

Now, I checked the code for the Collator and found out that it does overwrite all the labels, including the ones having -100 as the values, unless *the labels themselves* are set to `pad_token_id`.

https://github.com/huggingface/transformers/blob/main/src/transformers/data/data_collator.py#L744-L749

```python
        else:
            labels = batch["input_ids"].clone()
            if self.tokenizer.pad_token_id is not None:
                labels[labels == self.tokenizer.pad_token_id] = -100
            batch["labels"] = labels
        return batch
```

So, it's a bug then

### Reproduction

Make a training with or without train only after on

### Screenshot

_No response_

### Logs

```shell
Unrelated
```

### System Info

```shell
V100
```

kabachuha:
closes #4031

The goal of this PR is to fix the overwriting of all the labels including -100s by input_ids in the process of collating the batches with Transformers's DataCollatorForLanguageModeling

See https://github.com/oobabooga/text-generation-webui/issues/4031 for more details

## Checklist:

### DIFF
--- a/extensions/Training_PRO/script.py
+++ b/extensions/Training_PRO/script.py
@@ -387,7 +387,7 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
                 before_tokens = [shared.tokenizer.pad_token_id] * (cutoff_len - full_length) + before_tokens
 
             input_ids = before_tokens + after_tokens
-            labels = [-100] * len(before_tokens) + [1] * len(after_tokens)
+            labels = [shared.tokenizer.pad_token_id] * len(before_tokens) + [1] * len(after_tokens)
 
         input_ids = torch.tensor(input_ids)
         return {
--- a/modules/training.py
+++ b/modules/training.py
@@ -353,7 +353,7 @@ def do_train(lora_name: str, always_override: bool, save_steps: int, micro_batch
                 before_tokens = [shared.tokenizer.pad_token_id] * (cutoff_len - full_length) + before_tokens
 
             input_ids = before_tokens + after_tokens
-            labels = [-100] * len(before_tokens) + [1] * len(after_tokens)
+            labels = [shared.tokenizer.pad_token_id] * len(before_tokens) + [1] * len(after_tokens)
 
         input_ids = torch.tensor(input_ids)
         return {


kabachuha:
@FartyPants you may need this change too, since your training extension uses the same logic as well

https://github.com/oobabooga/text-generation-webui/blob/ee7bf49804bfb53775424d4c12294445497b610f/extensions/Training_PRO/script.py#L390

oobabooga:
@kabachuha Can you update the TrainingPRO code as well?

kabachuha:
Sure

kabachuha:
Yes, but if I get it right, this value is being overriden in the first place, before it's being passed for loss ignoring

kabachuha:
at least, if you're using the transformers collator (as the code mentions)

https://github.com/oobabooga/text-generation-webui/blob/ee7bf49804bfb53775424d4c12294445497b610f/modules/training.py#L592

if you switch it to Qlora's you sent, it may use a more correct logic; it's another option for the fix, I think

kabachuha:
Anyway, I suppose it's testable by compiling Transformers from source and injecting the collating function with some exception throwing to see whether it triggers during the training

I'll test it then and report back!

FartyPants:
I had never even used "train_only_after", but why would labels use shared.tokenizer.pad_token_id ?
I assume it works because pad_token_id  = 0 so it's anything but 1, and so does -100 but it's more like coincidental?
You can put [0] and it would make more sense than dragging a variable that has nothing to do with that part, just accidentally has good number in it.

kabachuha:
Yep, the collator triggers

image

kabachuha:
@FartyPants your idea of reporting label overwriting to transformers is good too, I may try doing it with a pr

FartyPants:
This is not necessary the correct test. you are showing labels after they had been copied from the input (which is WTF by itself - why would they copy labels from input???)

so labels would have some mask consisting of -100 and 1, then get totally borked

```
print('INPUT LABELS')
print(labels)
labels = batch["input_ids"].clone()
if self.tokenizer.pad_token_id is not None:
    labels[labels == self.tokenizer.pad_token_id] = -100
print('OVERWRITTEN LABELS')
batch["labels"] = labels
print(batch["labels"])
```

oobabooga:
Any consensus? The -100 number was defined in https://github.com/oobabooga/text-generation-webui/pull/2315, for reference.

kabachuha:
I think I'll try to discuss/fix it on the Transformers side first since it's they who are overwriting the labels

FartyPants:
> Any consensus? The -100 number was defined in #2315, for reference.

The -100 is correct. It is ignore_index. 

But according to [kabachuha](https://github.com/kabachuha), it doesn't matter what you put in labels, it will be overwritten and new mask created blocking <unk> and that's it. 
It has to be looked at transformer level. I don't know if the code in DataCollatorForLanguageModeling is executed all the time or only under certain condition, but it certainly overwrites everything that is in labels - so any mask is rendered useless.

oobabooga:
I will close this PR because it seems that if the issue exists, it is likely in the PEFT library and not in the web UI code.

Issue #4041, opened 2023-09-23
One click installer only create installer_files folder and then stop


Touch-Night:
### Describe the bug

Fresh install  
`start_windows.bat` stops at [`(conda deactivate && conda deactivate && conda deactivate) 2>nul`](https://github.com/oobabooga/text-generation-webui/blob/4b4d283a4c422b46956af81ec7c59e91d51586e5/start_windows.bat#L21)  
I have conda 22.11.1 correctly installed and PATH variable correctly configured

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

git clone in a new folder, and follow the one click installation (run `start_windows.bat`).

### Logs

```shell
Nothing in command-line.
```

### System Info

```shell
Windows 10  
Nvidia RTX 3050Ti Laptop (modified to 8GB VRAM)  
Conda 22.11.1
```

Touch-Night:
@rem the `conda deactive` line will make the one click installation work.

jllllll:
Making a PR to remove the conda deactivate commands.
I can't figure out why this is broken. I can only assume that it is yet another CMD bug.

PR #4042, opened 2023-09-23
Use call for conda deactivate in Windows installer

jllllll:
## Checklist:

---
Required for conda deactivate commands to work without breaking the Windows installer.

---
Fixes: #4041

### DIFF
--- a/cmd_windows.bat
+++ b/cmd_windows.bat
@@ -11,7 +11,7 @@ set TMP=%cd%\installer_files
 set TEMP=%cd%\installer_files
 
 @rem deactivate existing conda envs as needed to avoid conflicts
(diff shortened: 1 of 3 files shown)


Touch-Night (changes requested):
The command conda deactivate is a batch file itself (see [here](https://github.com/conda/conda/blob/main/conda/shell/condabin/conda.bat) and [here](https://github.com/conda/conda/blob/main/conda/shell/condabin/_conda_activate.bat)). When you call a batch file from another without using the call command, the original batch file is halted, and control is passed to the new one. Use `call conda deactivate` instead of `conda deactivate` to resolve the issue #4041.

jllllll:
It's been so long since I touched the Windows scripts that I completely forgot about using `call` to avoid issues like this.

Issue #4063, opened 2023-09-24
Windows One-Click Installer Seemingly Broken (may be related to previously mentioned module error)


BenevolenceMessiah:
### Describe the bug

Tested on Windows 10. The Windows one-click installer seems to be broken. However, everything can be made to work if you run `start_windows.bat`, then after it fails, run `update_windows.bat`, (which completes succesfully as it seems to read `requirements.text` correctly and install the packages), and then running `start_windows.bat` which will then launch the web UI correctly. 

Upon running the `start_windows.bat` file initially:

ERROR: Could not install packages due to an OSError: [WinError 5] Access is denied: 'C:\\ai\\text-generation-webui\\installer_files\\env\\Lib\\site-packages\\~umpy\\.libs\\libopenblas64__v0.3.21-gcc_10_3_0.dll'
Consider using the `--user` option or check the permissions.

Command '"C:\ai\text-generation-webui\installer_files\conda\condabin\conda.bat" activate "C:\ai\text-generation-webui\installer_files\env" >nul && python -m pip install -r requirements.txt --upgrade' failed with exit status code '1'. Exiting...
Press any key to continue . . .

Upon running the `start_windows.bat` file again:

Traceback (most recent call last):
  File "C:\ai\text-generation-webui\server.py", line 13, in <module>
    import gradio as gr
ModuleNotFoundError: No module named 'gradio'
Press any key to continue . . .

@oobabooga, @jllllll This makes me think that issues #4057 and #4056 are related to this issue, namely because after several attempts of deleting installer_files, and re-cloning the entire project, I would get the same package and module errors. Notably, on prior to this fresh install, my Windows-based (not git-based) symlinks stopped the update process from working when the update_windows.bat was run. Never happened before. Normally, Git will prompt me to delete the symlink because of new files or it will automatically overwrite the symlinks with real directories.

### Potential Solution:
As stated, running the `update_windows.bat` file completes succesfully, at which point, `start_windows.bat` file will work and launch the web UI. The short term (hacky) solution may be to call `windows_update.bat` via `start_windows.bat` after the conda installation. I'm not knowledgeable enough to know if that is even practical or potentially detrimental to the normal operation of the one-click installer.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. clone repo.
2.  run `start_windows.bat` to produce first error.
3. run` start_windows.bat` again to produce second error.
[...]

PR #4064, opened 2023-09-25
Avoid importing torch in one-click-installer

jllllll:
## Checklist:

---
`__version__` in `torch` is derived from `version.py`. So, to avoid issues with Windows not uninstalling torch dependencies, this will just read that value directly from `version.py` if possible.

---
Fixes: #4063

### DIFF
--- a/one_click.py
+++ b/one_click.py
@@ -53,7 +53,16 @@ def cpu_has_avx2():
 
 
 def torch_version():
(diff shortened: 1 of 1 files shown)


oobabooga:
That's pretty damn impressive. Well done!

BenevolenceMessiah:
Wow. You fixed that really quick!

PR #4067, opened 2023-09-25
Check '--model-dir' for no models warning in one-click-installer

jllllll:
## Checklist:

---
Fixes a minor annoyance with the no model warning showing when the default model dir is not actually being used.

Parses args for `--model-dir` and extracts the given path while maintaining spaces within quotes using regex.

The arg parsing ended up messier than I'd like as I'm not very experienced with handling args in Python and this was the only way I could think of to do this.

### DIFF
--- a/one_click.py
+++ b/one_click.py
@@ -2,6 +2,7 @@ import argparse
 import glob
 import os
 import platform
+import re
 import site
 import subprocess
 import sys
@@ -22,6 +23,7 @@ if os.path.exists(cmd_flags_path):
 else:
     CMD_FLAGS = ''
 
+flags = f"{' '.join([flag for flag in sys.argv[1:] if flag != '--update'])} {CMD_FLAGS}"
 
 def is_linux():
     return sys.platform.startswith("linux")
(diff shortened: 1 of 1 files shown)


oobabooga:
Maybe use shlex instead of a regex? Here is a test:

```python
import shlex

args_string = '--input_file input.txt --model_dir "models/model with spaces" --output_file output.txt'

args_list = shlex.split(args_string)

# ['--input_file', 'input.txt', '--model_dir', 'models/model with spaces', '--output_file', 'output.txt']

```

jllllll:
The advantage of the regex, and the main reason I used it, is that it can be used to specify multiple strings and characters to split on. This allows something like `--model_dir="models/model with spaces"` to be parsed the same as `--model_dir "models/model with spaces"`. Looking at the docs led me to believe that shlex purposefully kept `=` in the strings.

jllllll:
On that note, a slightly different regex string also allows for parsing args from `CMD_FLAGS.txt` across multiple lines in a mixed order:
```
--auto-launch --model_dir "models/model with spaces" --api --test3
--verbose --model_dir="models/model with spaces" --test
--test2
```
```py
import re

x = open('cmd_flags.txt').read()
re.split('\s+(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)|=',x)

#['--auto-launch', '--model_dir', '"models/model with spaces"', '--api', '--test3', '--verbose', '--model_dir', '"models/model with spaces"', '--test', '--test2']
```

oobabooga:
If shlex doesn't work then let's go for the regex. For the CMD_FLAGS I think it's best to keep the current code as it's easier to read.

PR #4077, opened 2023-09-25
Added stopping regex to api

JaLnYn:
## Checklist:

Added regex for stopping rather than just a stopping string. This only works with API for now, UI change is coming. Let's discuss this changed


oobabooga:
I think I tested this at some point, and concluded that regex adds a significant overhead that reduces the number of tokens/second.

JaLnYn:
Yea regex is not extremely efficient but it can be left as none for folks who choose not to use it. I think it could be valuable for people who do want to use it though. Especially for use cases that output in a specific format and want the llm to stop output if a certain format is found.

PR #4091, opened 2023-09-26
Add grammar to transformers and _HF loaders

oobabooga:
Based on https://github.com/Shopify/torch-grammar/

## Installation

Clone the repository above into your `repositories/` folder:

```
git clone https://github.com/Shopify/torch-grammar repositories/torch-grammar
```

## Usage

The syntax seems to not be 100% compatible with the `.gbnf` examples in the llama.cpp repository. For instance, `json.gbnf` doesn't work. But if you copy and paste the [grammar.ebnf](https://github.com/Shopify/torch-grammar/blob/main/examples/grammar.ebnf) example, it works:

```
root      ::= (commands eol)+
commands  ::= t | info | nav
nav       ::= "nav(\"/" [a-z/]*  "\")"
info      ::= "info(" setting ")"
t         ::= "t(" setting ": " value ")"
value     ::= color | number | string | boolean
color     ::= "#" [0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f]
setting   ::= [a-z_ ]+
string    ::= "\"" [ \t!#-\[\]-~#x80-#xFF]* "\""
number    ::= [0-9]+
boolean   ::= ("true" | "false")
eol       ::= "\n"
```


oobabooga:
Not an idea solution as the repository above is not being maintained and is not 100% compatible with the llama.cpp examples, but that will do for now.

RandomInternetPreson:
So much is happening with this software, I need to try this it looks fascinating!!

PGadoury:
See my entry in the above issue for json and json array syntax that gets around the (broken) negatives with a hacky fix. I'd like to see negatives implementer properly, though, since there's a lot you'd struggle to do with just positive regex ranges.

Issue #4090, opened 2023-09-26
wsl.sh copies CMD_FLAGS.txt and one_click.py into wsl install dir, which prevents the subsequent pull


jsboige:
### Describe the bug

Update won't work with a git error complaining about those 2 files being unstashed/unvalidated.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Happened during a one-click migration, on 2 distinct machines, so I guess other wsl users should be affected.

### Logs

```shell
G:\Production\Oobabooga\text-generation-webui>update_wsl.bat
Mise à jour cc7b7ba..019371c
erreur : Les fichiers suivants non suivis de la copie de travail seraient effacés par la fusion :
        CMD_FLAGS.txt
        one_click.py
Veuillez renommer ou effacer ces fichiers avant la fusion.
Abandon
Command '. "/home/jesse/text-gen-install/text-generation-webui/installer_files/conda/etc/profile.d/conda.sh" && conda activate "/home/jesse/text-gen-install/text-generation-webui/installer_files/env" && git pull --autostash' failed with exit status code '1'. Exiting...
```

### System Info

```shell
Windows 11 + 3080 and Windows 10 + 3090
```

jllllll:
Working on a solution now.

PR #4093, opened 2023-09-26
Fix old install migration for WSL installer

jllllll:
## Checklist:

---
WSL installer will now only copy `CMD_FLAGS.txt` and will update the webui installation beforehand if it is older than the installer merge. There isn't much reason anymore to copy over webui.py/one_click.py as it is no longer useful to edit that file to enter flags.

This should prevent update issues as a result of this behavior.

---
If maintaining the unique behavior of the WSL installer becomes too much of a burden, then I will probably just end up removing much of this code as it's only purpose is to prevent the slowdown issues of WSL when handling files located within the Windows filesystem.

Something like a simple message explaining the WSL I/O bug and optionally preventing installation outside of the WSL env can be used instead.

---
Fixes: #4090

### DIFF
--- a/wsl.sh
+++ b/wsl.sh
@@ -47,10 +47,8 @@ fi
 
 if [[ "$INSTALL_DIR" =~ " " ]]; then echo This script relies on Miniconda which can not be silently installed under a path with spaces. && exit; fi
 
-# create install dir if missing and copy one_click.py to install dir to maintain functionality without edit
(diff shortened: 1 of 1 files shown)


oobabooga:
This seems like something that will not require future maintenance, as the installers are becoming stable now. I see no immediate reason to remove the code.

Thank you so much for fixing this.

PR #4103, opened 2023-09-27
fix:  minor error in _run_server defaults

reidmcy:
This fixes a minor error with the default values in `_run_server()` for both the streaming and non-streaming APIs.

I tested my changes and they do not interfere with anything else.

## Checklist:

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -197,7 +197,7 @@ class Handler(BaseHTTPRequestHandler):
         super().end_headers()
 
 
-def _run_server(port: int, share: bool = False, tunnel_id=str):
+def _run_server(port: int, share: bool, tunnel_id:str):
     address = '0.0.0.0' if shared.args.listen else '127.0.0.1'
 
     server = ThreadingHTTPServer((address, port), Handler)
@@ -217,5 +217,5 @@ def _run_server(port: int, share: bool = False, tunnel_id=str):
     server.serve_forever()
 
 
-def start_server(port: int, share: bool = False, tunnel_id=str):
+def start_server(port: int, *, share: bool, tunnel_id:str):
     Thread(target=_run_server, args=[port, share, tunnel_id], daemon=True).start()
--- a/extensions/api/streaming_api.py
+++ b/extensions/api/streaming_api.py
@@ -102,7 +102,7 @@ async def _run(host: str, port: int):
         await asyncio.Future()  # run forever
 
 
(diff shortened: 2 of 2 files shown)


oobabooga:
What error?

reidmcy:
The default value of `tunnel_id` was the type `str` instead of something like `None`. Using the type will break the `_run_cloudflared` a couple functions deeper. There's already a value provided by `shared.args.public_api_id` so this error won't show up normally.

reidmcy:
It will raise a `TypeError: expected str, bytes or os.PathLike object, not type` error in the cloudflare thread.

It would be tricky to get this in normal operations since the default is never used. So here's it via a direct call:
```python
>>> import extensions.api.blocking_api
>>> extensions.api.blocking_api._run_server(8765, True)
  File "/Users/reidmcilroy-young/Documents/llm/text-generation-webui/extensions/api/util.py", line 108, in _start_cloudflared
    public_url = _run_cloudflared(port, port + 1, tunnel_id=tunnel_id)
  File "/Users/reidmcilroy-young/mambaforge/envs/textgen/lib/python3.10/site-packages/flask_cloudflared.py", line 114, in _run_cloudflared
    cloudflared = subprocess.Popen(['arch', '-x86_64'] + cloudflared_command, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
  File "/Users/reidmcilroy-young/mambaforge/envs/textgen/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Users/reidmcilroy-young/mambaforge/envs/textgen/lib/python3.10/subprocess.py", line 1780, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
TypeError: expected str, bytes or os.PathLike object, not type
Exception in thread Thread-2 (_start_cloudflared):
  File "/Users/reidmcilroy-young/mambaforge/envs/textgen/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/Users/reidmcilroy-young/mambaforge/envs/textgen/lib/python3.10/threading.py", line 953, in run
  File "/Users/reidmcilroy-young/Documents/llm/text-generation-webui/extensions/api/util.py", line 120, in _start_cloudflared
    raise Exception('Could not start cloudflared.')
Exception: Could not start cloudflared.
```

oobabooga:
If there is no error, I don't see a reason to change the code.

PR #4124, opened 2023-09-29
Implement Content-Length Header in the Blocking API

Lexazan:
## Implement Content-Length Header in the Blocking API
I have made changes to the API's response behavior. Previously, the API would send headers first, followed by the body when it was generated. However, I have modified it to wait until the body is fully generated before sending both the headers and the body together. Additionally, I have included the Content-Length header in accordance with the recommendations outlined in the HTTP 1.0 protocol specification.

The reason behind this adjustment stems from an issue I encountered while calling the API from the Godot game engine. In the previous setup, the game engine was not waiting for the complete body but instead emitted the data_received event immediately after receiving the headers. There are most likely other API clients exhibiting similar behavior, and this change should improve compatibility with them while also eliminating any head-scratching for those who are implementing it.

## Checklist:

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -44,10 +44,6 @@ class Handler(BaseHTTPRequestHandler):
         body = json.loads(self.rfile.read(content_length).decode('utf-8'))
 
         if self.path == '/api/v1/generate':
-            self.send_response(200)
-            self.send_header('Content-Type', 'application/json')
-            self.end_headers()
-
             prompt = body['prompt']
             generate_params = build_parameters(body)
             stopping_strings = generate_params.pop('stopping_strings')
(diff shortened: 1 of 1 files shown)


teddybear082:
I just stumbled upon this, in case it’s helpful I have example code for using textgenwebui’s OpenAI extension in streaming mode in Godot here: https://github.com/teddybear082/godot4-ai-npc-example/blob/main/addons/textgenwebuiOpenAI/TextgenwebuiOpenAI.gd.   I know it’s not the same as what you are talking about but figured I would pass along as I had a problem with any streaming API service with Godot’s built in httprequest nodes and struggled with it for quite some time.

oobabooga:
The current API will be deprecated and replaced with the OpenAI-compatible one, which now uses FastAPI after https://github.com/oobabooga/text-generation-webui/pull/4430. 

I assume the behavior described in this PR may be correct by default in FastAPI; if not, a new PR would be welcome.

PR #4129, opened 2023-09-29
Adding multistart scripts for each supported environment

jsboige:
## Checklist:

This is a new attempt at proposing new multistart scripts that ease starting several instances of Oobabooga with distinct models on distinct ports.
Previous PR only contained the wsl model which includes the additional effort to forward each port as it's needed. This PR adds common flags and additional logging, passes args directly to the base launch script, and adds corresponding multistart scripts for windows, linux and mac, without the port forwarding which is then unneeded. 

Commented examples are a bit outdated, feel free to change them to your favorite models for each size.

PS: Really crossing my fingers on this one after the last 2 PRs closed.
Again, I need it for https://github.com/MyIntelligenceAgency/semantic-fleet/tree/main


oobabooga:
This is very niche and specific. I don't see a reason to include those scripts in the main repository.

jsboige:
@oobabooga I'm sorry but this is frustrating. I understand that maintaining that repo is exhausting and you have to make quick decisions, but it just feels like you've been brushing off my work several times without a second thought.

When I submitted a first version of this in the one click installer, you said you closed it because it was wsl only, and you did not have the environment to test it. Fair enough, I took the time to revise and make those scripts available for all supported OSes.

I worked my ass-off for months to bridge the worlds of Oobabooga and Semantic-kernel, the result of that work is now hosted in [a dedicated repo](https://github.com/MyIntelligenceAgency/semantic-fleet). Now I would understand that you don't care too much about .Net, Microsoft, Open AI, wsl, you name it, but so far you made sure that Oobabooga has accommodated a lot of specifics, and that is the main appeal of your platform.

Many have expressed their wish to accommodate distinct specialized models, which are the current trend, and some of them want to automatically route their prompts to the appropriate ones, which is the crux of my work. Maybe some people will be interested in implementing something similar to what I did in Python, outside of semantic-kernel. 

Running several instances of Oobabooga from the same install works great. It didn't have to, and that is an opportunity, that I'm sure many will find useful. Don't you think those multistart scripts could live in a dedicated sub-directory and a dedicated branch?   

Could we at least consider if some people would be interested by this? I understand I should have started with an issue but again, I needed the scripts for my LLM router, so I figured a PR would be appropriate.

Accordingly, if you don't mind, I'll create an issue and if it gathers interest, hopefully we can re-open that PR again.

PR #4095, opened 2023-09-26
Bump exllamav2 to 0.0.4 and use pre-built wheels

jllllll:
## Checklist:

- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).
---
turboderp has added pre-compiled wheels for exllamav2 0.0.4. This PR adds them to the appropriate requirements.txt while also specifying conditions for the PyPI distribution to be installed on systems not covered by the wheels.

Issue #3943, opened 2023-09-15
exllamav2  not working within the docker container


KaruroChori:
### Describe the bug

Any attempt to load models with exllamav2 will fail.
Those models, as well as exllamav2 are working just fine outside the docker for oobabooga

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

Install oobabooga following the instructions to have a Docker installation.
Once the install is ready run in your browser.

Try to open any compatible model selecting exllamav2 as back-end.
Try to load it.
The same model will load if used in exllamav1 if its format is GPTQ.
The same model will load fine if used natively in exllamav2.

### Logs

```shell
Traceback (most recent call last):
text-generation-webui-text-generation-webui-1  |   File "/app/modules/ui_model_menu.py", line 194, in load_model_wrapper
text-generation-webui-text-generation-webui-1  |     shared.model, shared.tokenizer = load_model(shared.model_name, loader)
text-generation-webui-text-generation-webui-1  |   File "/app/modules/models.py", line 77, in load_model
text-generation-webui-text-generation-webui-1  |     output = load_func_map[loader](model_name)
text-generation-webui-text-generation-webui-1  |   File "/app/modules/models.py", line 336, in ExLlamav2_loader
text-generation-webui-text-generation-webui-1  |     from modules.exllamav2 import Exllamav2Model
text-generation-webui-text-generation-webui-1  |   File "/app/modules/exllamav2.py", line 5, in <module>
text-generation-webui-text-generation-webui-1  |     from exllamav2 import (
text-generation-webui-text-generation-webui-1  |   File "/app/venv/lib/python3.10/site-packages/exllamav2/__init__.py", line 3, in <module>
text-generation-webui-text-generation-webui-1  |     from exllamav2.model import ExLlamaV2
text-generation-webui-text-generation-webui-1  |   File "/app/venv/lib/python3.10/site-packages/exllamav2/model.py", line 12, in <module>
text-generation-webui-text-generation-webui-1  |     from exllamav2.linear import ExLlamaV2Linear
text-generation-webui-text-generation-webui-1  |   File "/app/venv/lib/python3.10/site-packages/exllamav2/linear.py", line 4, in <module>
text-generation-webui-text-generation-webui-1  |     from exllamav2 import ext
text-generation-webui-text-generation-webui-1  |   File "/app/venv/lib/python3.10/site-packages/exllamav2/ext.py", line 118, in <module>
[...]

PR #4144, opened 2023-10-02
Bugfix/3943 docker container improvements

Callum17:
## Checklist:

- [✓] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

- less verbose docker compose
- drops containerized virtualenv in favor of non root runtime user
- slightly smaller runtime image
- moved all dependencies to builder pattern - so it should be easier to build and test newer dependencies

Refactored this PR last moment as the original exllamav2 3943 issue was fixed here:
https://github.com/oobabooga/text-generation-webui/pull/4095

We should be able to close this issue:
https://github.com/oobabooga/text-generation-webui/issues/3943

My only question might be around the bump from flask_cloudflared==0.0.12 to flask_cloudflared==0.0.14 in the openai extension. Had to be bumped to prevent conflicts with the api extension. I still need to test this change. Functionally should be no different from the version installed by the current docker compose setup - unlike the current setup this builder pattern will explicitly fail if any constraints are violated in any of the requirements files.

Also worth noting that a number of conflicts were found against the (outdated?) requirements.txt in oobabooga/GPTQ-for-LLaMa. Conclusion was that it can be ignored.

inference successfully tested with:
ctransformers
exllama
exllamav2
llama.cpp
AutoGPTQ


.gitignore
```diff
@@ -35,3 +35,10 @@ models/config-user.yaml
 Thumbs.db
 .chroma
 installer_files
+
+# ignore custom container envrionment
```
LoopControl, line 39:
```suggestion
# ignore custom container environment
```

oobabooga:
Could you merge the dev branch and check if the updated Dockerfile still works? 

A relevant detail is that the OpenAI API is now the default, and it listens on port 5000 only (5005 is not used anymore). 

I don't know how to use Docker properly, but if you find that your updates still work, I can merge the PR based on your feedback.

Callum17:
I've got an updated copy working against current HEAD of main branch. Still needs a minor tidy and the dev merge, but it's pretty much there.

will look to update this in the next day of two when I get time.

oobabooga:
Awesome @Callum17, thank you.

Callum17:
Think GitHub auto closed this when I was aligning with the dev branch.

Opened a clean PR here:
https://github.com/oobabooga/text-generation-webui/pull/4768

There are some issues (not created by me!) - most notably with GPTQ-for-Llama conflicting with the core package dependencies. I suspect we will start seeing issues in non docker installs if these aren't resolved.

PR #4145, opened 2023-10-02
Fix off-by-one error in exllama_hf caching logic

tdrussell:
This fixes a subtle caching bug introduced with 03dc69edc5436b9426238fa626212dcffd9d62a3. I was playing around with a custom lora that by chance happened to amplify the effect of this bug, which is how I noticed. This led me down a rabbit hole of debugging and I eventually figured out the issue.

## How to reproduce
1. Load any raw text completion model with exllama_hf. I used Llama-2-13B-GPTQ.
2. Set repetition_penalty to 1 (important! otherwise it masks the effect)
3. Go to default tab, type literally the letter "a" as a prompt
4. Generate some tokens, it should be some random text.
5. Generate again. It will always be "a a a a a..." repeating. This is because the off-by-one caching bug has the effect of "doubling up" the most recent token. So it is invisible, but the model is seeing "a a" as the prompt, not "a".
6. You can use the logit viewer to see how dramatically the probabilities change.

I added some explanatory comments because the logic confused me for a long time until I pieced everything together. Also, I did not test exllamav2_hf, since I don't have quants for it, but presumably it has exactly the same issue.

This caching / state management stuff is so tricky, I'm not even 100% sure this is a complete fix. Please double and triple check that what I'm doing is sensible before merging!

## Checklist:

### DIFF
--- a/modules/exllama_hf.py
+++ b/modules/exllama_hf.py
@@ -94,6 +94,10 @@ class ExllamaHF(PreTrainedModel):
                     ex_cache.current_seq_len = longest_prefix
                     if len(seq_tensor) - longest_prefix > 1:
                         self.ex_model.forward(seq_tensor[longest_prefix:-1].view(1, -1), ex_cache, preprocess_only=True, lora=self.lora)
+                    elif len(seq_tensor) == longest_prefix:
+                        # Very tricky: if the prefix we are reusing *is* the input_ids, then we have to back up the cache pointer by one,
+                        # because we feed input_ids[-1] to forward() below, but that last token is already in the cache!
+                        ex_cache.current_seq_len -= 1
 
             if reset:
                 ex_cache.current_seq_len = 0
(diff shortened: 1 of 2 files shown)


Ph0rk0z:
Interesting, I've never seen it double up. Is this only when cache for CFG is enabled, or always?

oobabooga:
That's indeed a bug. It would be nice to avoid this 1 additional forward call to obtain the logits when the input is contained in the cache, but that seems to be how exllama and exllamav2 do it, so it should be the best solution available. See https://github.com/turboderp/exllamav2/blob/master/exllamav2/generator/streaming.py#L183

For llamacpp_hf, it is possible to access the past logits without a forward call: https://github.com/oobabooga/text-generation-webui/blob/main/modules/llamacpp_hf.py#L143

Going forward once is probably pretty inconsequential performance-wise, but I will tag @turboderp for awareness.

@Ph0rk0z always, not just with CFG.

PR #4178, opened 2023-10-05
Option to select/target additional linear modules/layers in LORA training

computerman00:
## Checklist:

I've been patching this in for a while and figured I'd make a PR since I've seen some truly awesome gains with using all LLamaAttention modules(and even all linear layers).

I was inspired to start experimenting with additional targeted attention modules w/ LORA training after reading some discussions on r/LocalLLaMA, the [QLORA paper](https://arxiv.org/abs/2305.14314), and the [ReLoRA paper](https://arxiv.org/abs/2307.05695)

"When using the standard practice of applying LoRA to query and value attention projection matrices, we are not able to replicate full finetuning performance for large base models...We find that the most critical LoRA hyperparameter is how many LoRA adapters are used in total and that LoRA on all linear transformer block layers are required to match full finetuning performance"
Source: [QLORA Paper](https://arxiv.org/abs/2305.14314)

The original [LORA paper](https://arxiv.org/abs/2106.09685) is also a must.
q,k,v, and o stand for query, key, value, output respectively.

I tried to keep changes to an absolute minimum, I however did raise the cutoff-length maximum from 2048 to 4096, if this is controversial enough we can change it back but 4k is easily achievable on 80GB of VRAM(And even 48GB for 13B models). I tend to always patch this in and figured it wouldn't hurt and potentially help/save-time-for peeps with VRAM to spare.

Here is some screenshots:
Screenshot_20231004_210046

Screenshot_20231004_210140

Screenshot_20231004_210818

Screenshot_20231004_210630

Screenshot_20231004_211210

EDIT: The above 2 screenshots shouldn't be taken for direct comparison as I realized they had different ranks(64vs32 respectively) after looking at training logs. Here is same parameter on everything but target modules for a general idea of what to expect:
```
Training 'llama' model using (gate, down, up, q, v, k, o) projections                                                                      
Trainable params: 125,173,760 (0.9525 %), All params: 13,141,201,920 (Model: 13,016,028,160)



Training 'llama' model using (q, v, k, o) projections
Trainable params: 52,428,800 (0.4012 %), All params: 13,068,456,960 (Model: 13,016,028,160)
```


epolewski:
I'm testing this out and it seems to be working. Few cosmetic things:
Breaks the refresh button look for me. The buttons still work:
image
On boot I see these errors. I was already getting one, but with this PR I get 2. They don't seem to affect anything:
image
It does appear to be working though!

computerman00:
> I'm testing this out and it seems to be working. Few cosmetic things: Breaks the refresh button look for me. The buttons still work: image On boot I see these errors. I was already getting one, but with this PR I get 2. They don't seem to affect anything: image It does appear to be working though!

Interesting, thanks for the feedback!

I have not been able to replicate this(Neither undefined buttons or that second warning, I will continue to investigate - I did just sync/merge [oobabooga:main](https://github.com/oobabooga/text-generation-webui/tree/main) into the feature branch, I suspect maybe there was some conflicts with newer dependencies/code that wasn't present in my branch. 

I would much appreciate any feedback if you do get to try again. Thanks again!

FartyPants:
You are repeating the exact same thing that is already in the extensions Training PRO which is already supplied with the main distribution - even in the current version and it has been there for some time. Look in Advanced settings.

Also, I would caution - most people **should use Q,V only** when dealing with fine-tuned models. It makes the final Lora more flexible and transparent-  taking a lot of smartness (ability on instruction following for example) from the underlying fine-tuned model while adjusting it with the LORA.

Using all attention layers require far higher quality and larger training set than most people are willing to cope with and the model may then deteriorate with following instructions if used on fine-tuned model.
You are quoting QLora paper, but you forgot to note the size and quality of guanaco dataset and that it has been used on base model. This isn't what most people would do.

Guanaco was a proof of Qlora concept: if they use all projections and high quality dataset, you can get away with Qlora instead of finetuning of base model - but I'm not aware of other people doing QLORA to finetune base model on that size, at least not the models people use.

Take it from someone who did 500+ Lora's (and most of the time, I use q,v)

These are very important distinctions, but also - ability to add projections is already there, one just has to look in the extension folder.

oobabooga:
The trainingPRO extension and the main training tab are independent. It's fine to update the main one with a missing feature, especially when the additions are tidy as is the case here.

The PR seems to have been tested and working, so let's merge it. Thank you @computerman00!

PR #4184, opened 2023-10-05
Update LoRA.py

srossitto79:
Add ggml lora support for llama.cpp gguf/ggml models (convert any lora with llama.cpp/convert-lora-to-ggml.py )

## Checklist:

### DIFF
--- a/modules/LoRA.py
+++ b/modules/LoRA.py
@@ -13,6 +13,8 @@ def add_lora_to_model(lora_names):
         add_lora_autogptq(lora_names)
     elif shared.model.__class__.__name__ in ['ExllamaModel', 'ExllamaHF'] or shared.args.loader == 'ExLlama':
         add_lora_exllama(lora_names)
+    elif shared.model.__class__.__name__ in ['LlamaCppModel']:
+        add_lora_llamacpp(lora_names)               
     else:
         add_lora_transformers(lora_names)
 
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Does this work for Kquants + offloading yet? Because when I tried it, the model required to be F16 and all on CPU.

Touch-Night:
Hmmm, I did the same work before I saw this pr.

Ph0rk0z:
That makes 3 of us.. I added it where it loads the llama.cpp model originally and then just reload to add or remove the lora. In this way you don't have to move all of those parameters to lora.py. 

Unfortunately I was met with the message of having to use FP16 model or not use GPU offloading, that made it very useless. Have had better luck merging the lora directly into Kquants after conversions. The models don't increase perplexity very much and can be fully offloaded.

Touch-Night:
> That makes 3 of us.. I added it where it loads the llama.cpp model originally and then just reload to add or remove the lora. In this way you don't have to move all of those parameters to lora.py.
> 
> Unfortunately I was met with the message of having to use FP16 model or not use GPU offloading, that made it very useless. Have had better luck merging the lora directly into Kquants after conversions. The models don't increase perplexity very much and can be fully offloaded.

Mine is the same approach, using the llama.cpp's --lora parameter and reload. I don't know how to merge it directly into Kquants.

Ph0rk0z:
https://github.com/xaedes/llama.cpp/tree/finetune-lora/examples/export-lora works after converting.

oobabooga:
Is there a way of doing it without reloading the entire model?

Ph0rk0z:
AFAIK, the way it is in llama.cpp (and python bindings), the lora loads along with the model. Unfortunately it still asks for F16 weights.

modules/LoRA.py
```diff
@@ -96,6 +98,64 @@ def add_lora_autogptq(lora_names):
...
+        # ):
+        #     raise RuntimeError(
+        #         f"Failed to apply LoRA from lora path: {shared.model.lora_path} to base path: {shared.model.lora_base}"
+        #     )
+
+        shared.lora_names = [lora_names[0]]
+
+        logger.info(f"Succesfully Applied Lora {lora_adapter_path} to Model.")
+
+        return
```
Jamim, line 157: You can just check `lora_names` with `not`. Also, while you already have `return` there, you don't need to have the `else` block. You can just remove an `else` and decrease an indentation accordingly.
```suggestion
    if not lora_names:
        shared.lora_names = []
        return

    if len(lora_names) > 1:
        logger.warning('Llama can only work with 1 LoRA at the moment. Only the first one in the list will be loaded.')      

    lora_path = get_lora_path(lora_names[0])
    lora_adapter_path = str(lora_path / "ggml-adapter-model.bin")

    logger.info("Applying the following LoRAs to {}: {}".format(shared.model_name, ', '.join([lora_names[0]])))

    if shared.args.tensor_split is None or shared.args.tensor_split.strip() == '':
        tensor_split_list = None
    else:
        tensor_split_list = [float(x) for x in shared.args.tensor_split.strip().split(",")]

    params = {
        'model_path': str(shared.model.model.model_path),
        'lora_path': str(lora_adapter_path),
        'n_ctx': shared.args.n_ctx,
        'seed': int(shared.args.llama_cpp_seed),
        'n_threads': shared.args.threads or None,
        'n_threads_batch': shared.args.threads_batch or None,
        'n_batch': shared.args.n_batch,
        'use_mmap': not shared.args.no_mmap,
        'use_mlock': shared.args.mlock,
        'mul_mat_q': shared.args.mul_mat_q,
        'numa': shared.args.numa,
        'n_gpu_layers': shared.args.n_gpu_layers,
        'rope_freq_base': RoPE.get_rope_freq_base(shared.args.alpha_value, shared.args.rope_freq_base),
        'tensor_split': tensor_split_list,
        'rope_freq_scale': 1.0 / shared.args.compress_pos_emb,
    }

    shared.model.model = llama_cpp.Llama(**params)

    # shared.model.model.lora_path = lora_adapter_path
    # if llama_cpp.llama_model_apply_lora_from_file(
    #     shared.model.model.model,
    #     shared.model.model.lora_path.encode("utf-8"),
    #     scale=1,
    #     path_base_model=shared.model.model.lora_base.encode("utf-8") if shared.model.model.lora_base is not None else llama_cpp.c_char_p(0),
    #     n_threads=shared.model.model.n_threads,
    # ):
    #     raise RuntimeError(
    #         f"Failed to apply LoRA from lora path: {shared.model.lora_path} to base path: {shared.model.lora_base}"
    #     )

    shared.lora_names = lora_names[:1]

    logger.info(f"Succesfully Applied Lora {lora_adapter_path} to Model.")
```
The `return` at the end is also redundant.

Gee1111:
anything new here?

PR #4189, opened 2023-10-05
Fix python wheels for avx requirements

AG-w:
## Checklist:

- [V] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

This fix mixed up python wheels so people with no avx2 CPU can finally load GGUF model with CUDA support
also changed no-avx2-cpu-only wheel to true CPU only 
and ctransformers wheel to actual avx-only wheel


jllllll:
No objections here.

oobabooga:
Thanks you for the feedback, and thanks @AG-w for the fixes.

PR #4202, opened 2023-10-06
Jinja instruct templates

oobabooga:
Status: kind of works.

Based on https://huggingface.co/docs/transformers/main/chat_templating

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 import gradio as gr
 import yaml
+from jinja2 import Template
 from PIL import Image
 
 import modules.shared as shared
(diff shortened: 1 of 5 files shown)


Ph0rk0z:
We won't have to eat their "Assistant is helpful and harmless" system messages from this, will we?

Tostino:
@oobabooga Just FYI, I have implemented this for vLLM's OpenAI api: https://github.com/vllm-project/vllm/pull/1365

Was pretty straight forward.

@Ph0rk0z no...those are just examples. The template itself will determine what the template does. Some models may have a hard coded system message in their template, others will allow a configurable system message. Just depends.

Dampfinchen:
GGUF now supports Instruct templates.

oobabooga:
Moved to https://github.com/oobabooga/text-generation-webui/pull/4874

Issue #3200, opened 2023-07-19
Wheels not installing on Ubuntu Linux


iibw:
### Describe the bug

When installing the dependencies for text-generation-webui with `pip install -r requirements.txt`, two wheels do not work on my Ubuntu system.

The error messages are:
`ERROR: exllama-0.0.6+cu117-cp310-cp310-linux_x86_64.whl is not a supported wheel on this platform.`
and
`ERROR: auto_gptq-0.3.0+cu117-cp310-cp310-linux_x86_64.whl is not a supported wheel on this platform.`

The error messages happen quickly after attempting to install the wheels which makes me think there is a fundamental compatibility problem between these wheels and my OS. I'm not sure what that would be because my Ubuntu system is both x86_64 and Ubuntu Linux. Maybe a library which these wheels depend on? I'm not sure.

Tagging @jllllll since this involves their exllama wheel.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Not sure what the OS config is which leads to this issue.

### Logs

```shell
`ERROR: exllama-0.0.6+cu117-cp310-cp310-linux_x86_64.whl is not a supported wheel on this platform.`
`ERROR: auto_gptq-0.3.0+cu117-cp310-cp310-linux_x86_64.whl is not a supported wheel on this platform.`
```

### System Info

```shell
- Ubuntu 22.04.2 LTS
- AMD EPYC 7R32 x86_64
```

iibw:
Figured out the issue. Accidentally installed Python 3.11 instead of 3.10.9. The wheels are 3.10-specific.

michelleplur:
maybe keep this issue open, I had to look deep to find this solution :)

Tim38-code:
I faced the same issue and I had to revert back to python=3.9=0 to solve the isssue

PR #4233, opened 2023-10-09
Structured requirements && Python 3.11 support

mjbogusz:
This introduces structured requirements files (updating a package just got a whole lotta easier!) and out-of-box Python 3.11 support (ref #3200).

I've split the changes into two commits so they can be reviewed separately; If requested, I can also open separate PRs for them.

## Checklist:


oobabooga:
I prefer to keep the requirements explicit, even if that creates redundancy. Updated the requirements to also support python 3.8 and 3.9.

PR #4279, opened 2023-10-13
Make it possible to load lora using llama.cpp

Touch-Night:
## Checklist:


Ph0rk0z:
I think you can just use the regular shared.model as base, i mean we're not picking some other model to add loras besides the one we loaded. Also don't forget the HF, should work there too.

Touch-Night:
> I think you can just use the regular shared.model as base, i mean we're not picking some other model to add loras besides the one we loaded. Also don't forget the HF, should work there too.

I used share.model at first, and discovered that when you use a non fp16 or fp32 model as the lora base, the webui will directly end. So I made it a drop-down, user can then add lora to any gguf models as long as they have a fp
16 model as the lora base.

Ph0rk0z:
So a second copy of the FP16 model? That's weird. What model does it actually load? Did you try with something that doesn't fit in vram in F16? It may just be using that "base" as the model and ignoring the other one.

Touch-Night:
I tried, and feel it's actually useless and unnesessary to load LoRA this way. since it's better to merge LoRA with base model and export, and this pr has some bugs, so I will close it. Still hope there will be someone else dive deep into this and make a valuable solution.

Ph0rk0z:
Yea.. the Lora support for quants is lacking in llama.cpp still.

PR #4285, opened 2023-10-14
Use Pytorch 2.1 exllama wheels

jllllll:
## Checklist:

---
Some people are having trouble loading exllama with the current wheels that were built with Pytorch 2.0.1.
Not sure why as it worked fine on my system. Regardless, there is at least one confirmed case of this resolving the issue.

---
Exllamav2 CUDA 11.8 wheel now uses Pytorch 2.1.
https://github.com/turboderp/exllamav2/compare/v0.0.5...v0.0.6


oobabooga:
Thank you for those wheel updates.

PR #4288, opened 2023-10-14
Fix Dockerfile - Cuda 12.1 Requirement.

TwinFinz:
Fix dockerfile to use the correct base image with cuda 12.1 to remedy this error on running "docker compose up" "The detected CUDA version (11.8) mismatches the version that was used to compile PyTorch (12.1). Please make sure to use the same CUDA versions."

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:11.8.0-devel-ubuntu22.04 as builder
+FROM nvidia/cuda:12.1.0-devel-ubuntu22.04 as builder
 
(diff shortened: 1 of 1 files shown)


oobabooga:
The wheels in the requirements.txt are all for 11.8 at the moment. I'm closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/4298, which specifies pytorch 11.8 in the dockerfile.

TwinFinz:
ok, but that is all i had to do for it to compile directly from git clone
[repository] and docker compose up -d.
after docker started it built everything successfully and works as expected.

PR #4291, opened 2023-10-15
Relax numpy version requirements

JohanAR:
If using torch from https://download.pytorch.org/whl/cu118 then numpy 1.24.1 also gets installed as a dependency,
and unless there there's any reason for requiring 1.24.0 I think it's unnecessary to downgrade it.

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ einops
 exllamav2==0.0.6; platform_system != "Darwin" and platform_machine != "x86_64"
 gradio==3.47.*
 markdown
-numpy==1.24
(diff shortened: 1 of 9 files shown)


TraceRecursion:
git clone --filter=blob:none --quiet https:github.com/oobabooga/torch-grammar.git 'C:

exit code:128
See above for output.
note:This error originates from a subprocess,and is likely not a problem with pip.

(C:\Chat\envs)PS C:\Chat\text-generation-webui>

I encountered this problem while deploying on the server and couldn't resolve it

PR #4305, opened 2023-10-16
Support LLaVA v1.5

haotian-liu:
## Checklist:

### This PR supports LLaVA v1.5.

---

Major changes for LLaVA v1.5:
- Add support for MLP projector
- CLIP ViT-L/14 336px backbone
- Pad the image to square instead of cropping for LLaVA v1.5

Changes to the general multimodal pipeline, when supporting LLaVA v1.5
- Slightly increase the max resolution to 336 (from the original 300)
- When saving the image to buffer, use PNG instead of JPEG, as observed in our official demo, to slightly reduce the artifacts of multi-step processing in the whole process (downsample, pad-to-square, downsample).
- Put `<image>` in front of the `query` instead of the reverse (I do not modify `visible_text`, so the interface won't change). This aligns with the most multimodal training data processing pipeline, including LLaVA v1.5.

---

### Run LLaVA v1.5

First download the checkpoints from [TheBloke](https://huggingface.co/TheBloke/llava-v1.5-13B-GPTQ) to `models`. I tried `TheBloke_llava-v1.5-13B-GPTQ` and `TheBloke_llava-v1.5-13B-GPTQ_gptq-4bit-32g-actorder_True`, and both works. Download in UI also works for me.

```Shell
python server.py \
    --model-dir models \
    --model TheBloke_llava-v1.5-13B-GPTQ_gptq-4bit-32g-actorder_True \
    --disable_exllama \
    --multimodal-pipeline llava-v1.5-13b
```

Instruct:
Screen Shot 2023-10-16 at 5 31 29 PM

Chat:
Screen Shot 2023-10-16 at 5 30 52 PM

---

Since I am not an expert in these quantization stuff, I do need help in resolving some efficiency / GPTQ / exllama related issues. I'll document some of these below.

### TODO
- [ ] It seems that when `act_order=True`, exllama is not compatible. According to TheBloke, this should not be an issue. But I am not sure why. So I added `--disable_exllama` to make it run.
> Act Order: True or False. Also known as desc_act. True results in better quantisation accuracy. Some GPTQ clients have had issues with models that use Act Order plus Group Size, but this is generally resolved now.
- [ ] Maybe due to the reason above, the speed is not yet optimal.
- [ ] Seems that LLaVA-v1 template is not recognized correctly (always recognized as `LLaVA`, which uses the older prompt), any way to make the auto-detect of the instruction template happen?
- [ ] Since LLaVA v1.5 is only trained for 1 epoch, it is usually better to have a lower temperature like 0.2 -- is there a way to set it as the default for a specific model?
- [ ] Fix for AWQ: #4308


Vinventive:
@haotian-liu Great news! Thank you for your hard work and dedication. Your research has made a significant impact in AI community. We truly appreciate it.

oobabooga:
Thank you so, so much for this PR @haotian-liu, that's extremely helpful.

I confirm that it's working as expected on my end with the transformers loader:

```
python server.py \
        --model liuhaotian_llava-v1.5-13b \
        --multimodal-pipeline llava-v1.5-13b \
        --load-in-4bit
```

I had to change `"model_type": "llama"` to `"model_type": "llava"` in the config.json to be able to load the model.

Example:

example

>  Seems that LLaVA-v1 template is not recognized correctly (always recognized as LLaVA, which uses the older prompt), any way to make the auto-detect of the instruction template happen?

Thanks for pointing out, I just fixed this one.

>  Since LLaVA v1.5 is only trained for 1 epoch, it is usually better to have a lower temperature like 0.2 -- is there a way to set it as the default for a specific model?

This is possible, but I think that it may be confusing to change the temperature automatically. An alternative to lowering the temperature is to use an appropriate preset for instruction following. I usually recommend the "Divine Intellect" one, and in my quick testing with LLaVA, it seems to work better than the default one.

As for issues with AutoGPTQ: that backend hasn't been receiving many updates lately, and the current "meta" for llama-based models is exllama (v1 and v2). I'll try to prioritize getting multimodal to work with these and with llama.cpp as well (llava is officially supported but the llama-cpp-python hasn't been updated yet to expose it).

yhyu13:
>It seems that when act_order=True, exllama is not compatible. According to TheBloke, this should not be an issue. But I am not sure why. So I added --disable_exllama to make it run.

@oobabooga Is it possible that we use exllama for llava v1.5 now?

yhyu13:
@oobabooga 

Just found a bug when setting vision and projector bit to 16 instead of 32 for llava v1.5 pipeline, there is an exception on data type

```
  File "/root/CodeSpace/text-generation-webui/modules/text_generation.py", line 30, in generate_reply
    for result in _generate_reply(*args, **kwargs):
  File "/root/CodeSpace/text-generation-webui/modules/text_generation.py", line 77, in _generate_reply
    for reply in generate_func(question, original_question, seed, state, stopping_strings, is_chat=is_chat):
  File "/root/CodeSpace/text-generation-webui/modules/text_generation.py", line 309, in generate_reply_HF
    question, input_ids, inputs_embeds = apply_extensions('tokenizer', state, question, input_ids, None)
  File "/root/CodeSpace/text-generation-webui/modules/extensions.py", line 224, in apply_extensions
    return EXTENSION_MAP[typ](*args, **kwargs)
  File "/root/CodeSpace/text-generation-webui/modules/extensions.py", line 127, in _apply_tokenizer_extensions
    prompt, input_ids, input_embeds = getattr(extension, function_name)(state, prompt, input_ids, input_embeds)
  File "/root/CodeSpace/text-generation-webui/extensions/multimodal/script.py", line 90, in tokenizer_modifier
    prompt, input_ids, input_embeds, total_embedded = multimodal_embedder.forward(prompt, state, params)
  File "/root/CodeSpace/text-generation-webui/extensions/multimodal/multimodal_embedder.py", line 172, in forward
    prompt_parts = self._embed(prompt_parts)
  File "/root/CodeSpace/text-generation-webui/extensions/multimodal/multimodal_embedder.py", line 148, in _embed
    embedded = self.pipeline.embed_images([parts[i].image for i in image_indicies])
  File "/root/CodeSpace/text-generation-webui/extensions/multimodal/pipelines/llava/llava.py", line 246, in embed_images
    return super().embed_images(images)
  File "/root/CodeSpace/text-generation-webui/extensions/multimodal/pipelines/llava/llava.py", line 108, in embed_images
    image_features = self.mm_projector(image_features)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/modules/container.py", line 217, in forward
    input = module(input)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/root/miniconda3/envs/textgen/lib/python3.10/site-packages/torch/nn/modules/linear.py", line 114, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: expected scalar type Half but found Float
```

yhyu13:
@haotian-liu @oobabooga 

I've tested loading llava-v1.5-gptq with exllama but it spits out incorrect desciption of the image, and inference performance isn't ideal with exllama and flashattention2 on A5000 (whose fp16 is aroud 110Tops, between 3090 and 4090) with only 5tokens/sec


msedge_hraKskeCFI


Code_ldXysCBeSC


Code_pjT5wwqMpt

yhyu13:
@haotian-liu @oobabooga 

I have not had a chance to try out bitsandbytes quantization inference speed for llavav1.5, since @oobabooga had tried it, was it better than exllama or autogptq on your rig?

RandomInternetPreson:
Can confirm, using the .bin to .safetensor converter code that comes with oobabooga I can download this model https://huggingface.co/liuhaotian/llava-v1.5-13b and load it in 4bit using oobabooga's example.  This is great! What a great model amazing, thank you!!

PR #4310, opened 2023-10-17
ExLlamav2_HF: Convert logits to FP32

turboderp:
ExLlama converts logits to FP32 at the end of the forward pass, while ExLlamaV2 returns FP16 logits straight from the lm_head layer. Converting in the V2 loader makes sure the HF sampler behavior the same for V1 and V2.

## Checklist:


Ph0rk0z:
PPL on evaluation went down slightly after this.

oobabooga:
Thanks @turboderp, I would never have noticed this.

PR #4314, opened 2023-10-18
Enable special token support for exllamav2

JohanAR:
## Checklist:

### DIFF
--- a/modules/exllamav2.py
+++ b/modules/exllamav2.py
@@ -64,7 +64,7 @@ class Exllamav2Model:
         return result, result
 
     def encode(self, string, **kwargs):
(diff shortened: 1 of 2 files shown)


JohanAR:
> should it also use 'skip_special_tokens', ?

Not sure.. skip_special_tokens appears to be a setting for the Transformers tokenizer, but it seems like this might be the same as setting decode_special_tokens=False for exllamav2

oobabooga:
Thanks for the PR. The state parameter goes to the generation functions, which can pass state['skip_special_tokens'] to the decode function as you did. I think that your implementation is correct and no further changes are needed.

PR #4343, opened 2023-10-21
Docker: Remove explicit CUDA 11.8 Reference

whiteadam:
Since @bdashore3 added CUDA 12.1 images, the explicit CUDA 11.8 ref for Torch set by @sammcj is no longer needed.

I tested this build locally with RTX 3060 and found no issues (other than a DEPRECATION NOTICE! from NVIDIA on 12.1).

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -61,7 +61,7 @@ RUN --mount=type=cache,target=/root/.cache/pip,rw \
 
 COPY . /app/
 
-RUN cp /app/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so /app/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so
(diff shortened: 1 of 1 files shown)


oobabooga:
Can you replace cu118 with cu121 instead of removing it? For future-proofness when a new pytorch version gets released.

whiteadam:
It is done and tested.

I am curious about the impact on [this line](https://github.com/oobabooga/text-generation-webui/blob/b98fbe0afc1feb210b06b2015b9c663273dc23b8/docker/Dockerfile#L64), though.

oobabooga:
This one can be changed as well, as this file exists under bitsandbytes. Thanks for pointing out and for the fixes.

nalf3in:
This unfortunately breaks compatibility with debian stable as the latest driver only supports up to cuda 12.0. Users trying to run this without a server gpu (which supports forward compatibility) receive a weird error message. Could we keep a second dockerfile with 11.8 with instructions in the readme for that use case ? I would be happy to make a pr

Issue #4353, opened 2023-10-22
fatal: unable to access 'https://github.com/oobabooga/text-generation-webui/': gnutls_handshake() failed: Error in the pull function.


mongolu:
### Describe the bug

git clone https://github.com/oobabooga/text-generation-webui

fatal: unable to access 'https://github.com/oobabooga/text-generation-webui/': gnutls_handshake() failed: Error in the pull function.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Just appeared today.

Searched on Google for 2023 issues, found this one
[https://copyprogramming.com/howto/fatal-unable-to-access-gnutls-handshake-failed-handshake-failed](url)
Citing: "The TLS connection was non-properly terminated, For anyone facing this issue after 2020, Use ipv4 by using "--ipv4" or "-4" flag in your git command. ex: git clone <git url> -4. I am not sure why this happens but sometimes IPV6 connections are not able to establish a TLS connection but IPV4 connections have no issues. Share. edited Dec 27, 2021 at 3:35."

So...

### Logs

```shell
no
```

### System Info

```shell
Win 11 WSL2
```

mongolu:
PR [https://github.com/oobabooga/text-generation-webui/pull/4354](url)

mongolu:
Does anybody have this issue now?

iSenChi:
I do.

mongolu:
Found some solution?

KneesDev:
Having same issue when trying to pull using a self-hosted runner (GitHub actions). Doing git pull in terminal works.

Rurounin:
Got the same issue, both using the installer and GitHub Desktop to clone it.

mongolu:
It seems it's ok now, some problems on their side maybe.
[https://downdetector.co.uk/status/github/](url)

Rurounin:
Yeah, seems okay now.

PR #4354, opened 2023-10-22
Mongolu gnutls handshake() failed

mongolu:
## Checklist:


mongolu:
It is for this one #4353 , but it seems not to be enough.
It seems that sporadically works, but most of the time not.

PR #4376, opened 2023-10-24
Rename additive_repetition_penalty to presence_penalty, add frequency_penalty

tdrussell:
See the discussion in #3627.

These repetition penalty settings work the same as the ones in the [OpenAI chat completion API](https://platform.openai.com/docs/api-reference/chat/create).

## Checklist:


bdashore3:
Spelling error in `[95ad00b](https://github.com/oobabooga/text-generation-webui/pull/4376/commits/95ad00b7d08f458b7c8c72cf593228244f67867c)`

`frquency_penalty` should be `frequency_penalty`

tdrussell:
Thanks for noticing, I fixed it.

oobabooga:
Thanks a lot for adding this parameter, @tdrussell. I took the opportunity to also connect the frequency and presence penalty sliders to the llama.cpp loader, which implements them independently in its C++ backend.

I also made a sanity check by comparing the logits in the main branch and this PR with `additive_repetition_penalty` / `presence_penalty` set to 0.5, and the logits turned out identical. It's nice that you also handled the batch_size > 1 case.

DocShotgun:
Hmm, I'm wondering if there's anything different about the way OAI implements `frequency_penalty` or if our models are just inherently less resilient, but after trying a few different L2 models of varying sizes (13b and 70b), it seems like things go off their rocker pretty fast with `frequency_penalty` set to anything above a tiny value. Even with numbers as low as 0.2 I was starting to get some strangeness in the outputs.

Also `mirostat` + `frequency_penalty` really doesn't mesh well. Goes on an insane rant with `frequency_penalty` set at 0.1 lol.

Was wondering if there were any suggested values/ranges for these settings?

Issue #4211, opened 2023-10-07
Impossible resolution of `autoawq` for Python 3.11


jamesbraza:
### Describe the bug

Resolution of `autoawq` is currently impossible with Python 3.11:

```
> pip install -r requirements.txt
...
  ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11; 1.6.2 Requires-Python >=3.7,<3.10; 1.6.3 Requires-Python >=3.7,<3.10; 1.7.0 Requires-Python >=3.7,<3.10; 1.7.1 Requires-Python >=3.7,<3.10; 1.7.2 Requires-Python >=3.7,<3.11; 1.7.3 Requires-Python >=3.7,<3.11; 1.8.0 Requires-Python >=3.8,<3.11; 1.8.0rc1 Requires-Python >=3.8,<3.11; 1.8.0rc2 Requires-Python >=3.8,<3.11; 1.8.0rc3 Requires-Python >=3.8,<3.11; 1.8.0rc4 Requires-Python >=3.8,<3.11; 1.8.1 Requires-Python >=3.8,<3.11
  ERROR: Could not find a version that satisfies the requirement autoawq==0.1.4 (from versions: none)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

With Python 3.11, run `pip install -r requirements.txt` on https://github.com/oobabooga/text-generation-webui/tree/d33facc9feea19158420ddcbc842509825a4e88b

### System Info

```shell
2021 MacBook with M1, macOS Ventura 13.5.2
```

jamesbraza:
Cc @casper-hansen

yhyu13:
@jamesbraza 

py 3.11 is currently poorly compatible. Have you tried to install autoawq from source? Pulled the autoawq repo and hit pip install -e . to install autoawq locally with py3.11.

It could be the case the author just has not built the package for py3.11 yet.

casper-hansen:
AutoAWQ is compatible with Python 3.11: https://github.com/casper-hansen/AutoAWQ/releases/download/v0.1.4/autoawq-0.1.4-cp311-cp311-linux_x86_64.whl

[...]

PR #4390, opened 2023-10-25
Adding `platform_system` to `autoawq`

jamesbraza:
AutoAWQ is only Linux and Windows: https://github.com/casper-hansen/AutoAWQ/blob/v0.1.4/setup.py#L22, this PR moves the `requirements.txt` to reflect that.

Related: https://github.com/oobabooga/text-generation-webui/issues/4211


oobabooga:
Thank you for the fix!

PR #4393, opened 2023-10-26
ChatML.yaml - Open-Orca/Mistral-7B-OpenOrca - Model answers to itself

MindKrypted:
Using:
```
  <|im_start|>system
  <|im_end|>
```
Open-Orca/Mistral-7B-OpenOrca answers to itself because the "system" tag is incomplete.

Replacing "system" by "<|system|>" fixes it.

## Checklist:

### DIFF
--- a/instruction-templates/ChatML.yaml
+++ b/instruction-templates/ChatML.yaml
@@ -1,7 +1,7 @@
 user: "user"
 bot: "assistant"
 context: |
-  <|im_start|>system
+  <|im_start|><|system|>
   <|im_end|>
 turn_template: "<|im_start|><|user|>\n<|user-message|><|im_end|>\n<|im_start|><|bot|>\n<|bot-message|><|im_end|>\n"
 


MindKrypted:
### More context:

image

### Tested with:
- TheBloke_Mistral-7B-OpenOrca-GPTQ_gptq-4bit-32g-actorder_True
- TheBloke_dolphin-2.1-mistral-7B-GPTQ_gptq-4bit-32g-actorder_True

### To replicate the issue:
- git pull on 25OCT2023
Fresh start 
- Using Auto-Load, ExLlamaHF selected, max_seq_len=32768
- Generation max_new_tokens= 4096
- Instruction Template= ChatML
- Character= Default Assistant
- Chat tab with mode= chat

_Sometimes it won't trigger the self-reply if there's "past chat" that wasn't deleted beforehand._

FartyPants:
What does the <|system|> gets replaced with during interference?

oobabooga:
It seems to be correct as is, with `system` and not `<system>`

https://github.com/openai/openai-python/blob/main/chatml.md

MindKrypted:
> It seems to be correct as is, with `system` and not `<system>`
> 
> https://github.com/openai/openai-python/blob/main/chatml.md

@oobabooga I agree, according to the documentation "system" is correct. 

To be honest my knowledge on what's really happening in the back-end is limited, are you guys able to replicate what I've encountered?

So far what I can tell is that putting "<|system|>" _fixed_ it for both models tested on my side, they don't self-reply anymore.
Please let me know if you'd like more information, do further testing or close this commit request :)

oobabooga:
You need to use "instruct" mode rather than "chat" mode. The template is not used otherwise.

Zenopheus:
Yes, and **don't use 'chat-instruct' either**. "chat-instruct" seems to mash together the settings in "Instruction template" and "Character" tabs in confusing way that disregards the instruction format. 

⚠️ @oobabooga There are related problems:
In "ChatML.yaml"
- "context" is not quoted. I'm not sure what impact this has. Especially if the user enters their own quoted text.
- "context" is missing an additional newline at the end. This is not obvious due to the quote issue. Use verbose mode and compare it to TheBloke/dolphin-2.1-mistral-7B-GGUF examples.
- There should be a sample system context message in the context field so people know where to customize it. Without this, people might think they have to edit the Character Context section (which is wrong and will lead to problems).

Also, the UI saves all custom templates without quotes. User entered quotes and newlines are not getting escaped. Simply saving a template will produce a file that is different from the original.

Here's a sample 'ChatML.yaml' file that matches the prompt from TheBloke/dolphin-2.1-mistral-7B-GGUF:
```
user: "user"
bot: "assistant"
context: "<|im_start|>system\nYou are Dolphin, a helpful AI assistant.<|im_end|>\n"
turn_template: "<|im_start|><|user|>\n<|user-message|><|im_end|>\n<|im_start|><|bot|>\n<|bot-message|><|im_end|>\n"
```

Note: User and Bot should not be renamed since the model expects them to be "user" and "assistant".

PR #4399, opened 2023-10-27
update google colab to provide easy toggle option to use openai api extension

teddybear082:
-This is useful for people who want to use textgenwebui's google colab for openai API calls as easily as possible without needing to open the GUI

-also add direct google colab link

## Checklist:


Colab-TextGen-GPU.ipynb
```diff
@@ -5,7 +5,8 @@
     "colab": {
       "private_outputs": true,
       "provenance": [],
-      "gpuType": "T4"
+      "gpuType": "T4",
+      "include_colab_link": true
```
oobabooga, line 9: What does this do exactly?

teddybear082: I thought it would add a link to google colab so someone could just click to go to colab from the notebook but does not seem to have worked as expected.

teddybear082:
This looks good to me is there anything else you need from me?  Should I look more into trying to provide a easy colab link at the top?

teddybear082:
I think I fixed the easy open link to google colab now, and also moved the parameters up so that the requirements could be chosen based on the parameters to avoid installing unnecessary openai api or api requirements if the user is not using the modules

oobabooga:
This PR made me realize that it doesn't make sense to have 2 APIs in this project, so I started https://github.com/oobabooga/text-generation-webui/pull/4430. If all goes well, it should make this PR unnecessary.

PR #4425, opened 2023-10-31
Bugfix: Updating the shared settings object when loading a model

ziadloo:
## Checklist:

I realized this bug when I was trying to use OpenAI API. In that scenario, the settings for the loaded model (which can be accessed through the `shared` module) is holding the default values while it's supposed to hold the settings for the loaded model.

For instance, the template for prompts that will be loaded when sending a chat request will always will "Alpaca" regardless of which model is loaded. Basically, the OpenAI API was broken.

This change will apply the settings from the the loaded model to the shared module so it can be picked up when responding to an API request.


oobabooga:
Originally `shared.settings` was for default values for the gradio UI, but I think that it makes sense to expand it to provide data to the API extension like this.

I'm working on making the OpenAI API more production ready in https://github.com/oobabooga/text-generation-webui/pull/4430 and am open to further suggestions.

PR #4430, opened 2023-10-31
Make OpenAI API the default API

oobabooga:
The most popular LLM API in the world is the OpenAI API, so I think that it makes sense to emulate it in this project when the `--api` flag is provided. This is already the case for [vLLM](https://github.com/vllm-project/vllm) and [FastChat](https://github.com/lm-sys/FastChat).

The goal is to start from the current `openai` extension and make the following changes before merging:

- [x] Replace the current `http.server` implementation with FastAPI, such that the API docs can be accessed at `127.0.0.1:5001/docs`.
- [x] Use [sse-starlette](https://github.com/sysid/sse-starlette) for streaming.
- [x] Find some sensible and transparent way to handle instruction templates by model.
- [x] The same for truncation lengths.
- [x] Make the overall structure more similar to the [AUTOMATIC1111 api](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api), with the input and output parameters and types for each endpoint declared in classes.
- [x] Include the additional sampling parameters present in this project as optional parameters in the completion endpoints.

### Status

I have converted everything to fastAPI, and the completion endpoints seemingly work (both with and without streaming).
 
### OpenAI API reference

https://platform.openai.com/docs/api-reference

----

cc @matatonic

PR #4431, opened 2023-10-31
Allow for changing shared.settings during model load via API

brthor:
## Checklist:

## Reasoning

Sometimes, when you load a certain lora, you need to be able to change the `instruction_template` on the fly.

Rather than focus on that specific use case, allow for updating the `shared.settings` as a whole at the time of model load via the api.

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -125,7 +125,10 @@ class Handler(BaseHTTPRequestHandler):
             if action == 'load':
                 model_name = body['model_name']
                 args = body.get('args', {})
+                settings = body.get('settings', {})
                 print('args', args)
+                print('settings', settings)
+
                 for k in args:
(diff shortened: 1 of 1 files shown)


oobabooga:
The current API will be replaced with an OpenAI-compatible one, and handling instruction templates is an open problem there. In particular, making `/v1/chat/completions` work properly when different models are loaded. Can you take a look at https://github.com/oobabooga/text-generation-webui/pull/4430? PRs to the `openai-api` branch are welcome.

oobabooga:
Seems to be working for the OpenAI API:

```
  curl -k http://127.0.0.1:5000/v1/internal/model/load \
    -d '{
      "model_name": "facebook_galactica-125m",
      "args": {
        "load_in_4bit": true
      },
      "settings": {
        "instruction_template": "Alpaca"
      }
    }'

```

PR #4435, opened 2023-11-01
[Fix] OpenOrca-Platypus2 models should use correct instruction_template when matching against models/config.yaml

deevis:
`models/config.yaml` applies ALL MATCHES sequentially as it compares model names with the supplied regexp keys.

This PR moves the `.*(openorca-platypus2)` entry to occur _after_ the `.*(platypus|gplatty|superplatty)` entry, thus allowing the more specific (and correct) match to occur afterwards.

## Checklist:


oobabooga:
Thank you!

PR #4448, opened 2023-11-02
make torch.load a bit safer

julien-c:
## Checklist:

## What is this

`weights_only=True` was added in PyTorch 1.13 so should be available everywhere


julien-c:
(related to #4450 too)

oobabooga:
That's interesting, thank you.

PR #4449, opened 2023-11-02
Implement Min P as a sampler option in HF loaders

kalomaze:
## Checklist:

- [✔️] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

A backport of the Min P sampler option that was merged into [llama.cpp recently.](https://github.com/ggerganov/llama.cpp/pull/3841) I think this is a suitable alternative to accomodate for the flaws of both Top P (Top P doesn't work well if there's not good concentration of high probability for top tokens) and Top K (can limit diversity unnecessarily), and it has gained a very positive reception.

image

### DIFF
--- a/extensions/api/util.py
+++ b/extensions/api/util.py
@@ -26,7 +26,7 @@ def build_parameters(body, chat=False):
         'do_sample': bool(body.get('do_sample', True)),
         'temperature': float(body.get('temperature', 0.5)),
         'top_p': float(body.get('top_p', 1)),
-        'min_p': float(body.get('min_p', 0.05)),
+        'min_p': float(body.get('min_p', 1)),
         'typical_p': float(body.get('typical_p', body.get('typical', 1))),
         'epsilon_cutoff': float(body.get('epsilon_cutoff', 0)),
         'eta_cutoff': float(body.get('eta_cutoff', 0)),
--- a/extensions/api/util.py
+++ b/extensions/api/util.py
@@ -26,6 +26,7 @@ def build_parameters(body, chat=False):
         'do_sample': bool(body.get('do_sample', True)),
         'temperature': float(body.get('temperature', 0.5)),
         'top_p': float(body.get('top_p', 1)),
+        'min_p': float(body.get('min_p', 0.05)),
         'typical_p': float(body.get('typical_p', body.get('typical', 1))),
         'epsilon_cutoff': float(body.get('epsilon_cutoff', 0)),
         'eta_cutoff': float(body.get('eta_cutoff', 0)),
--- a/modules/loaders.py
+++ b/modules/loaders.py
@@ -394,7 +394,6 @@ loaders_samplers = {
     'ctransformers': {
         'temperature',
         'top_p',
-        'min_p',
         'top_k',
         'repetition_penalty',
         'repetition_penalty_range',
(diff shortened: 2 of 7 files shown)


kalomaze:
Technically the min_tokens_to_keep is not being properly followed here which probably needs to be changed before it gets merged.

kalomaze:
Should be good to merge now.

oobabooga:
I was about to add min p manually since it looks cool. Thanks for the PR.

About the order, temperature does come before top p/top k in the transformers library:

https://github.com/huggingface/transformers/blob/acc394c4f5e1283c19783581790b3dc3105a3697/src/transformers/generation/utils.py#L825

I couldn't find any information as to how OpenAI does it.

Personally, I think that it makes sense to use temperature first.

kalomaze:
> I was about to add min p manually since it looks cool. Thanks for the PR.
> 
> About the order, temperature does come before top p/top k in the transformers library:
> 
> https://github.com/huggingface/transformers/blob/acc394c4f5e1283c19783581790b3dc3105a3697/src/transformers/generation/utils.py#L825
> 
> I couldn't find any information as to how OpenAI does it.
> 
> Personally, I think that it makes sense to use temperature first.

I can see arguments for both applying it last like how it is done in llama.cpp, and applying it like how it's done 'officially', so I think having this configurable / customizable to the end user would be helpful. Is that possible without changing how the transformers library functions?

kalomaze:
Changing how the sampler order logic works is out of scope for this PR though, so I'm ready for this to be merged in the meantime

BadisG:
I think you should put the default min_p at 0 instead of 1, having it at 1 makes everything else useless because it's the equivalent to top_k = 1

kalomaze:
> I think you should put the default min_p at 0 instead of 1, having it at 1 makes everything else useless because it's the equivalent to top_k = 1

Is it not 0 by default? I thought I set it to that

oobabooga:
Should be fixed now https://github.com/oobabooga/text-generation-webui/commit/4537853e2c1760a00bd278fb97f26f4c4fb149cb

BadisG:
It's still at min_p = 1 by default for me

oobabooga:
I messed up again, should be good after https://github.com/oobabooga/text-generation-webui/commit/7f9c1cbb30b13a313923d994b34ce175c522f9d8

BadisG:
All right it's working now, thanks!

BadisG:
I've also noticed that min_p and presence_penalty don't mix well, from time to time my generation stops abruptly because min_p removes some tokens and presence_penalty then removes the remaining tokens, leaving nothing to go on. I think this can be fixed by reversing the order of the samplers, presence_penalty should be applied before min_p. Maybe we should have the choice of sampler order, that would be a good feature to add, a bit like KoboldAI actually.

kalomaze:
> I've also noticed that min_p and presence_penalty don't mix well, from time to time my generation stops abruptly because min_p removes some tokens and presence_penalty then removes the remaining tokens, leaving nothing to go on. I think this can be fixed by reversing the order of the samplers, presence_penalty should be applied before min_p. Maybe we should have the choice of sampler order, that would be a good feature to add, a bit like KoboldAI actually.

Presence penalty... removes the tokens? I thought the rep pen (and associated version of it) are biasing the logits and not actually truncating them based on a certain criteria, that's weird.

PR #4450, opened 2023-11-02
transformers: Add a flag to force load from safetensors

julien-c:
## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 .DS_Store
 .eslintrc.js
 .idea
+.env
 .venv
 venv
 .vscode
--- a/modules/models.py
+++ b/modules/models.py
@@ -123,7 +123,8 @@ def huggingface_loader(model_name):
     params = {
         'low_cpu_mem_usage': True,
         'trust_remote_code': shared.args.trust_remote_code,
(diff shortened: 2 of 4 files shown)


julien-c:
cc @lysandrejik @wauplin does this look ok to you?

LysandreJik:
Yes, looks good! Thanks!

@oobabooga going forward in `transformers` (and the Hugging Face Hub in general) we'll encourage using `safetensors` as the default instead of the `pytorch_model.bin` format out of safety. Later today we'll be releasing default serialization in `safetensors` for `transformers`.

Having the option to ensure that all files loaded are pure tensors would be great for safety reasons :)

oobabooga:
I am all for safetensors, since it's significantly faster to load. My concern with this PR is that the [download script](https://github.com/oobabooga/text-generation-webui/blob/main/download-model.py) in this repository downloads only safetensors in case both pytorch.bin and safetensors are present in the HF repository. In case a model is available only in pytorch.bin format, users will see an error and will simply uncheck the "force-safetensors" option in the UI. So I don't know what purpose this option would serve.

oobabooga:
A solution is to just not add a checkbox to the UI (as is already the case here, I hadn't noticed). Thanks for the PR @julien-c, I appreciate it.

PR #4488, opened 2023-11-06
Merge dev branch

oobabooga:


lelandbatey:
**NOTE:** For folks who find their way here wondering "what happened to the old `api-examples/` folder?" the answer is _"[it was removed](https://github.com/oobabooga/text-generation-webui/pull/4430) because this project is getting rid of the API that files in the `api-examples/` folder were talking about."_ More info below:

This PR merges the "real" removal of the `api-examples/` folder, which happened in PR #4430. This is part of a larger effort for `text-generation-webui` to have only one API and to have that API be high quality and broadly compatible. Because "compatibility" is a focus, and OpenAIs API is the biggest player in the room at the moment, the choice has been made to have the default (and only) API of this codebase be the one implemented in the `openai` extension. Since the contents of the `api-examples/` folder contained examples exclusively for interacting with the now deprecated and soon to be deleted `api` extension, those examples were removed. For more information, see the maintainers PR #4430 where this all took place.

------------

Please also note that I am not the maintainer of this project, and I am not affiliated with this project in any way. I'm merely a curious reader who noticed the change, tracked it down, and read about it. I'm leaving this comment here only to point the way towards the real change, for others who tracked it down like I did (via `git log --full-history -- api-examples/api-example-chat-stream.py`).

oobabooga:
I can add that the new docs can be found here: https://github.com/oobabooga/text-generation-webui/wiki/12-%E2%80%90-OpenAI-API

The new API should be easier to use (1 address for streaming and not streaming, simple HTTP without websockets) and compatible with a wide range of existing applications.

PR #4489, opened 2023-11-06
Fix exception when using API with multimodal enabled

MeemeeLab:
## Detail:
Current multimodal extension loads embedder when UI loads.
Because API does not load UI, the program accesses `multimodal_embedder` which is not initalized yet, causing exception to raise.
This PR addresses this issue by checking `multimodal_embedder` is None, and loads embedder. This should not affect behaviour when embedder is already loaded.

## Checklist:


oobabooga:
Could you merge the dev branch and check if the fix applies when the OpenAI API is used?

MeemeeLab:
Seems like `dev` branch unintentionally fixed this issue. closing.

PR #4506, opened 2023-11-07
Fix API arguments not applying

rkfg:
## Checklist:

The API parameters used to load the model get overwritten by the model's metadata. In particular, it breaks the use case when the model is loaded via API with `alpha_value`/`compress_pos_emb` specified and a different `max_seq_len`. Since `max_seq_len` is usually present in the model config it gets overwritten and the context size stays default. I moved the API arguments application after the model defaults and it now works as expected.

### DIFF
--- a/extensions/api/blocking_api.py
+++ b/extensions/api/blocking_api.py
@@ -124,9 +124,6 @@ class Handler(BaseHTTPRequestHandler):
             if action == 'load':
                 model_name = body['model_name']
                 args = body.get('args', {})
-                print('args', args)
-                for k in args:
-                    setattr(shared.args, k, args[k])
 
                 shared.model_name = model_name
                 unload_model()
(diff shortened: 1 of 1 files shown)


oobabooga:
The current API is being replaced with the OpenAI compatible one. Could you try porting this model settings feature to `extensions/openai`?

rkfg:
I suppose that'd be a non-standard feature then if OpenAI doesn't allow changing these settings. Also I don't see a way to specify my own backend for OpenAI API in SillyTavern. I know it should be addressed there and not here, but that's the reason I can only use this, now legacy API. Can the current API be patched with this PR before the OpenAI one is better supported? I don't mind deprecation but please don't remove it completely.

oobabooga:
This is a breaking change that I think is necessary to improve the quality and reach of the API, but the adaptation to the new API is simple -- it's just chaning the endpoint to `http://127.0.0.1:5000/v1/completions`, removing wss (streaming now uses SSE), and renaming a couple of parameters like `max_new_tokens` -> `max_tokens`. Everything else should work the same. 

I am adding non-standard endpoints under `/v1/internal`. See `extensions/openai/script.py` for some examples.

rkfg:
Thank you, I'll see what I can do. Does that mean that this particular PR isn't gonna be accepted? I maintain my own fork anyway so it's not a big deal.

oobabooga:
In its present form, no, since the modified files will be deleted soon.

PR #4513, opened 2023-11-08
Include trust remote code usage in openai api's embedder

MrMojoR:
This is for using a custom embedder which has to have the trust_remote_code flag, with the SentenceTransformer implementation this is sadly not an option.

## Checklist:


oobabooga:
It looks good, thank you.

MrMojoR:
Sorry, it seems like it is not working for the current default embedder, I will take a look.

PR #4524, opened 2023-11-09
Greedy sampling

berkut1:
Added possibility to set temp to 0.0 for greedy sampling
https://github.com/ggerganov/llama.cpp/pull/3813
https://github.com/abetlen/llama-cpp-python/blob/80f4162bf44cbe6b7b22a3601236694706a4a11f/llama_cpp/llama.py#L1112

## Checklist:

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -44,7 +44,7 @@ def _generate_reply(question, state, stopping_strings=None, is_chat=False, escap
             return
 
         if (shared.model.__class__.__name__ != 'LlamaCppModel') and (state['temperature'] == 0.0):
-            state['temperature'] = 0.1  #For non LlamaCpp Models ignore 0.0 temperature   
+            state['temperature'] = 0.01  #For non LlamaCpp Models ignore 0.0 temperature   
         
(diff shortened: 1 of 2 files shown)


berkut1:
With a new version of llama-cpp-python https://github.com/abetlen/llama-cpp-python/issues/890
But since we reverted the new version to the old one, there should be no issues.

oobabooga:
The thing about T = 0 is that it leads to an error message when the transformers library is used for sampling (likely due to a division by 0 or an assert that prevents such division). So I'm not sure if it's worth it adding it to llama.cpp and causing these errors on other loaders (including llamacpp_HF).

berkut1:
@oobabooga how about this extra check? If someone is using a loader other than llama.cpp with temp=0, it will ignore it and set the value to 0.1.

berkut1:
What about whether it's worth it or not? Now I'm experimenting only with the Mistral model and found that it is very good at low temperatures, before I never set the temperature below 0.5 as previous models simply copied the text, but Mistral doesn't copy, but rewrites adhering to the context. And greedy sampling shows very good results.
But this is my subjective opinion :)

oobabooga:
I'm still not a fan of allowing T = 0. It's mathematically undefined and there are better ways to activate greedy decoding, like unsetting all parameters and defining top_k = 1.

PR #4566, opened 2023-11-12
Load models purely via API

rkfg:
## Checklist:

This is a small fix to make it possible to load models purely via API without touching UI or CLI arguments. `shared.model_name` isn't set in this case and it prevents generation in other parts where there are checks for both this and `shared.model` being `None`.

It also applies the loras provided via `args` to reach [feature parity with the deprecated API](https://github.com/oobabooga/text-generation-webui/blob/2af7e382b121f2eae16dd1f7ace621d31028b319/extensions/api/blocking_api.py#L143).


simcop2387:
One thing I'd love is if this also worked with the "unload" model button too.  That'd let me reduce power usage when the system isn't running any inference in the future.  (I've got some Tesla P40s, they're terrible as far as power usage goes, 50W idle if a model is loaded).

rkfg:
Yep, I perform unloading too after some inactivity time, automatically. There's no endpoint for that, though it's pretty trivial to add, but loading a model with empty name (`""`) does the trick. Though you'd get a nasty error message in the log as well.

oobabooga:
Wouldn't a better solution be to add the loras as optional parameters in the load_model endpoint?

rkfg:
Not sure if I understand. It is an optional parameter, an array of strings that goes into the `args.lora` parameter. For example, I use it like this: `'{"max_seq_len": 8192, "compress_pos_emb": 2, "cache_8bit": true, "loader": "ExLlamav2", "lora": ["nRuaif_Kimiko_13B"]}'`

modules/models.py
```diff
@@ -97,6 +97,7 @@ def load_model(model_name, loader=None):
         llama_attn_hijack.hijack_llama_attention()
 
     shared.settings.update({k: v for k, v in metadata.items() if k in shared.settings})
+    shared.model_name = model_name
```
oobabooga, line 100: This change is probably okay, but occurences of `shared.model_name = model_name` in other parts of the code should be removed then.

extensions/openai/models.py
```diff
@@ -57,6 +58,8 @@ def _load_model(data):
 
     shared.model, shared.tokenizer = load_model(model_name)
     shared.model_name = model_name
+    if shared.args.lora:
+        add_lora_to_model(shared.args.lora)
```
oobabooga, line 62: I'm thinking that it would be better to have a new endpoint for applying LoRAs. It would take a list of lora names as input and call the "apply_lora" function in modules/LoRA.py.

oobabooga:
Okay, I see now what your logic in this PR is (passing the lora names in the args dict). What I'm wondering is if it wouldn't be more intuitive to add a "load-lora" endpoint that takes an input like

```
lora_names: List[str]
```

for loading and unloading LoRAs. For unloading, the payload could be `lora_names=[]`. I find it worrying to modify `shared.args.lora` because then the LoRA would be applied in all subsequent requests to the `load_model` endpoint.

oobabooga:
Closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/4652

rkfg:
What about setting the model name? Without it if I never touch the UI I can't generate anything, it complains that the model isn't loaded.

PR #4576, opened 2023-11-13
Use standard hyphens in filenames

naptastic:
## Checklist:

- [*] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

This should not affect program function at all.


oobabooga:
I had no idea the hyphens were not just minus signs. Thanks for the fix.

PR #4604, opened 2023-11-15
update AutoGPTQ to higher version for lora applying error fixing

rogozinushka:
In my case I wasn't able to apply lora to mistral model because of earlier version of AutoGPTQ (error required version 5.0 above). AutoGPTQ version is hardcoded in requirements (version 0.4.2). So when requirements were fixed (current version 0.5.1) I was able to apply lora to mistral model and chatted with mistral lora. Maybe it would be useful for other.


oobabooga:
Thank you, and thanks @jllllll for building the new wheels.

PR #4620, opened 2023-11-16
Bump AutoAWQ to 0.1.7

casper-hansen:
- Yi support
- Fixed maintaining state across generations
- Fix multi-GPU inference
- Other bug related issues

@oobabooga

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -84,4 +84,4 @@ https://github.com/jllllll/GPTQ-for-LLaMa-CUDA/releases/download/0.1.1/gptq_for_
 https://github.com/jllllll/GPTQ-for-LLaMa-CUDA/releases/download/0.1.1/gptq_for_llama-0.1.1+cu121-cp39-cp39-linux_x86_64.whl; platform_system == "Linux" and platform_machine == "x86_64" and python_version == "3.9"
 https://github.com/jllllll/GPTQ-for-LLaMa-CUDA/releases/download/0.1.1/gptq_for_llama-0.1.1+cu121-cp38-cp38-linux_x86_64.whl; platform_system == "Linux" and platform_machine == "x86_64" and python_version == "3.8"
 https://github.com/jllllll/ctransformers-cuBLAS-wheels/releases/download/AVX2/ctransformers-0.2.27+cu121-py3-none-any.whl
-autoawq==0.1.6; platform_system == "Linux" or platform_system == "Windows"
+autoawq==0.1.7; platform_system == "Linux" or platform_system == "Windows"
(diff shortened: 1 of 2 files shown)


oobabooga:
Thank you. @casper-hansen 

I have two unrelated questions:

1) It seems like the VRAM allocation in AutoAWQ is too high (see for instance the reports [here](https://www.reddit.com/r/LocalLLaMA/comments/17w57eu/llm_format_comparisonbenchmark_70b_gguf_vs_exl2/)). Do you think that this could be caused by the fact that `max_new_tokens` is set to the maximum sequence length for the model in the UI?

In the source code, it is said that

```
        The window size is the same as the max_new_tokens. The window will
        automatically roll once max_new_tokens is exceeded.
```

So I wonder if this could be set to a smaller value like 100, or even to 0.

2) I tried loading an AWQ model in [this Colab Notebook](https://colab.research.google.com/github/oobabooga/text-generation-webui/blob/main/Colab-TextGen-GPU.ipynb) that I put together for this project, and it failed with a CUDA error. Colab uses CUDA 11.8 (iirc). Are the AutoAWQ wheels compatible with CUDA < 12?

casper-hansen:
> 1. It seems like the VRAM allocation in AutoAWQ is too high (see for instance the reports [here](https://www.reddit.com/r/LocalLLaMA/comments/17w57eu/llm_format_comparisonbenchmark_70b_gguf_vs_exl2/)). Do you think that this could be caused by the fact that `max_new_tokens` is set to the maximum sequence length for the model in the UI?

It looks like the author is not initializing AutoAWQ correctly. They should just set `max_new_tokens` to avoid OOM if that is a problem, which it should not be. I did fix a multi-GPU loading issue in 0.1.7 that may have been what he experienced.

> 2. I tried loading an AWQ model in [this Colab Notebook](https://colab.research.google.com/github/oobabooga/text-generation-webui/blob/main/Colab-TextGen-GPU.ipynb) that I put together for this project, and it failed with a CUDA error. Colab uses CUDA 11.8 (iirc). Are the AutoAWQ wheels compatible with CUDA < 12?

There are more wheels available on the release. We upgraded to CUDA 12.1 recently, so those are the wheels distributed on PyPi. You can get CUDA 11.8 on the GitHub release.

https://github.com/casper-hansen/AutoAWQ/releases/tag/v0.1.7

casper-hansen:
See this PR. It implemented a lot of fixes specifically targeting multi-GPU.

https://github.com/casper-hansen/AutoAWQ/pull/190

oobabooga:
> There are more wheels available on the release. We upgraded to CUDA 12.1 recently, so those are the wheels distributed on PyPi. You can get CUDA 11.8 on the GitHub release.
> 
> https://github.com/casper-hansen/AutoAWQ/releases/tag/v0.1.7

What I have been doing for other wheels is use this format in the requirements.txt:

```
https://github.com/jllllll/AutoGPTQ/releases/download/v0.4.2/auto_gptq-0.4.2+cu121-cp311-cp311-win_amd64.whl; platform_system == "Windows" and python_version == "3.11"
https://github.com/jllllll/AutoGPTQ/releases/download/v0.4.2/auto_gptq-0.4.2+cu121-cp310-cp310-win_amd64.whl; platform_system == "Windows" and python_version == "3.10"
https://github.com/jllllll/AutoGPTQ/releases/download/v0.4.2/auto_gptq-0.4.2+cu121-cp39-cp39-win_amd64.whl; platform_system == "Windows" and python_version == "3.9"
https://github.com/jllllll/AutoGPTQ/releases/download/v0.4.2/auto_gptq-0.4.2+cu121-cp38-cp38-win_amd64.whl; platform_system == "Windows" and python_version == "3.8"
```

Then the one-click installer (and also the Colab notebook) replace the `+cu121` string with`+cu118` based on the pytorch version in the environment (see the code [here](https://github.com/oobabooga/text-generation-webui/blob/main/one_click.py#L303)). 

Would something like this be possible (and make sense) for AutoAWQ? You are currently the only one with a pip-installable package that is not in `.whl` format, and that is good, so I'm not sure if it's worth it turning to whl requirements.

casper-hansen:
I would prefer keeping it as-is. Users should expect that the latest AutoAWQ makes use of the latest versions of the dependencies that are required. Superusers can go and install wheels from the release or open issues to ask about it.

oobabooga:
That makes sense and is valid. Thank you!

PR #4622, opened 2023-11-16
Add missing sse-starlight requirement for API

naptastic:
## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,7 @@ requests
 safetensors==0.4.0
 scipy
 sentencepiece
+sse-starlette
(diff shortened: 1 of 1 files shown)


oobabooga:
The requirement is present in the requirements for the extension, under `extension/openai/requirements.txt` (the current API as well as the previous one are user-contributed extensions). 

Since the absent package is causing problems, I have added a temporary workaround to install its requirements on every update in this commit: https://github.com/oobabooga/text-generation-webui/commit/e7d460d932ca4dcdac04cbe588356f39392d89a1

PR #4642, opened 2023-11-18
Openai embedding fix to support jina-embeddings-v2

wizd:
## Checklist:

tested with: https://huggingface.co/jinaai/jina-embeddings-v2-base-en


wizd:
tested embedding model:
```
jinaai/jina-embeddings-v2-base-en
BAAI/bge-large-zh-v1.5
```

extensions/openai/completions.py
```diff
@@ -235,7 +235,7 @@ def chat_completions_common(body: dict, is_legacy: bool = False, stream=False) -
 
     max_tokens = generate_params['max_new_tokens']
     if max_tokens in [None, 0]:
-        generate_params['max_new_tokens'] = 4096
+        generate_params['max_new_tokens'] = 200
```
oobabooga, line 238: Setting this high isn't necessary, as the auto_max_new_tokens fills the context. The 200 reference value is used when the context is fully used to remove old messages.

oobabooga:
Looks good, thanks

yhyu13:
@wizd Does Jina embedding runs on cuda devices? Other setense transformer embedding always lies on CPU even if specifiying cuda

PR #4704, opened 2023-11-22
Import accelerate very early to make Intel GPU happy

notsyncing:
Import `accelerate` first in `server.py`, or the `accelerate.utils.is_xpu_available()` will return `False` if imported later.

Note: this is only a workaround. At least it works for now. Needs more testing. 

## Checklist:

### DIFF
--- a/server.py
+++ b/server.py
@@ -1,3 +1,5 @@
+import accelerate;
+
 import os
 import warnings
 


notsyncing:
@oobabooga I'm not sure if this change would affect other backends like cuda, metal, etc, any ideas?

abhilash1910 (approved):
This is a temporary good WA , as there seems to be a dependency on other libraries when imported before accelerate.  Working on it .

ibehnam:
I just installed a fresh ooba and got `accelerate not found` with that line referenced.

PR #4728, opened 2023-11-25
Fix detection of stopping strings

tsukanov-as:
Detection of stopping strings must be performed before escaping html.
For example: "<end>"
This can be checked with the following prompt:
```
<|im_start|>system
<|im_end|>
<|im_start|>user
Repeat twice: "<start>"
Repeat three times: "<end>"<|im_end|>
<|im_start|>assistant
```
with openhermes-2.5-mistral-7b-16k.Q8_0.gguf or analog

## Checklist:


oobabooga:
That is correct. Thank you!

Issue #4712, opened 2023-11-23
A possible 4x-5x faster performance increase in coqui_tts processing on low memory cards.


erew123:
**EDIT - Ive re-written the full Script.py and it now does everything it did before + everything I detail below + tidies up a couple of issues with cleaning unwanted characters from the text generation. Script is further down this post** https://github.com/oobabooga/text-generation-webui/issues/4712#issuecomment-1825593734

**Description**

When you are running a model that takes up all of your VRAM on your card, coqui_tts can be horribly slow (CPU is almost as fast in this scenario). I've found a **potential** way to massively speed it up, though, not being a coder, I'm not sure if or how/if it can be implemented, and I will admit I'm not 100% sure what's going on here as there are quirks to this.

**TLDR:** It looks as though there may be a massive performance benefit for low VRAM scenarios to re loading the coqui_tts engine for longer sentences/paragraphs. If other people are suffering/seeing the issue I am describing.

I suspect this wont benefit people who have plenty of VRAM spare (after loading their model), only those whom have a low VRAM scenario.

I have included some thoughts and strange observations about this at the bottom.

**Additional Context**

Here is a 13B model loaded into RTX 4070, which takes about 11.4GB to 11.7GB of VRAM. You can also see in the shared GPU Memory, the coqui_tts model is loaded, taking a couple of GB's of RAM.

![image](https://github.com/oobabooga/text-generation-webui/assets/35898566/8f3777c3-b511-4b37-8070-5dcf3fbbe7a6)

When you generate TTS, with coqui_tts, layers of the AI model and the coqui_tts model are being swapped around and processed **I believe**. This results in a horrible time to generate the TTS. Minutes as opposed to seconds. **I suspect**, rather than loading the entire coqui_tts model back into the GPU as one large chunk, it may be nibbling away at loading the model back into GPU VRAM, processing a little, then nibbling away a bit more, processing etc  **rather than** shifting out some of the AI models layers and moving the whole TTS model into VRAM.

As shown below, this is taking 250 seconds and 456 seconds to generate the TTS.

![image](https://github.com/oobabooga/text-generation-webui/assets/35898566/51b6674f-9c50-4dba-91fc-f507f9764170)

[...]

PR #4760, opened 2023-11-28
Updated Coqui_tts

erew123:
Adds Download script, Model Picker, DeepSpeed, Low VRAM, Clearer command line output. requirements.txt updating tts 0.20.x to 0.21.x. Implements a small webserver, which is part of the workaround for not being able to use multiprocessing and needing a separate process worker. As such, the code includes all the documentation on a weblink that is accessible through the main text-generation-webui interface.

**EDIT December 3rd 2023** - Since originally making this PR, I've made so many changes, it would be impossible to document. Please see here for a quick visual look https://github.com/oobabooga/text-generation-webui/pull/4760#issuecomment-1837558567 and install the extension to have access to all the documentation. Has been tested on Windows and Linux in full. 

This is a large re-write. Its barely even the same code that was originally there.

## Checklist:


erew123:
I'm still not 100% happy with the "Low VRAM" setup. Windows still has a tendency to push things into shared VRAM and Linux doesn't. I can see a possible route to use multiprocessing to resolve this (forcing start-up of the models in a separate python process), though all this is new to me! I hadn't even coded in python until 4 days ago and I'm not a coder either, so feel free to chop/change any code if you feel you can improve it!

rktvr:
is DeepSpeed still a manual install? it seems like a pain to get working especially if you don't have the right python version.

erew123:
@rktvr you don't have to install or use DeepSpeed with this update. Its an "if you have it installed and you want to use it" deal.

As for DeepSpeed, its pretty easy to install on Linux. Please see section 4 on this link https://github.com/oobabooga/text-generation-webui/issues/4712#issuecomment-1825593734 for a short guide on it.

As for Windows, would you believe, Microsoft just asked me if I would do a pull request to update their Github DeepSpeed installation instructions for them! I've just literally sent the PR.

The limitations on Windows are: 
- DeepSpeed 8.3 is the latest build that works (12.3 on Linux currently)
- Python 3.9.18 (Any version on Linux)
- You need to install the Nvidia CUDA Toolkit, either 11.8 or 12.1 (And on Linux)
- You need to set the CUDA_HOME environment path in Windows or Linux within your Python 3.9.18 environment.

**Windows Option A - Use Someone else's wheel files**
- Download https://easyupload.io/235jwk (I pre-compiled these yesterday. They are for Python 3.9.18 but I named the zip incorrectly)
- From this [link](https://github.com/microsoft/DeepSpeed/issues/4729) perform steps 3, 4, 5, 7 and 11 (with the downloaded wheel files on step 11 for the correct version of Nvidia CUDA Toolkit that you downloaded).
- Install the text-gen-webui requirements into your new python 3.9.18 environment
- Start text-gen-webui with **python server.py**
**NOTE** you cant use start_windows.bat because it will use a different environment, version of Python and it will over-write the CUDA_HOME path.

**Windows Option B - Compile your own Wheel files**
- From this [link](https://github.com/microsoft/DeepSpeed/issues/4729) follow ALL steps.
- Install the text-gen-webui requirements into your new python 3.9.18 environment.
- Start text-gen-webui with **python server.py**
**NOTE** you cant use start_windows.bat because it will use a different environment, version of Python and it will over-write the CUDA_HOME path.

So its a mix of the steps I give on the Microsoft site whilst using pre-compiled wheels. Or if you want the full experience, you can follow all the steps on the Microsoft site (takes maybe 30-40 minutes at most).

yhyu13:
@erew123 You might want to set TTS_HOME to ./models/Coqui_tts so that all coqui tts models are atuomatically donwnloaded there the first time you requeste by model_name. Then the second time, you only need to adjust model name to actual path on disk, and try to load model with model_path and model_config path locally.

Checkout my implementation for cache folder handling here https://github.com/yhyu13/TTS/commit/7967679eb62142f8a919019313d2de2cf1684ed6

erew123:
> Checkout my implementation for cache folder handling here [yhyu13/TTS@7967679](https://github.com/yhyu13/TTS/commit/7967679eb62142f8a919019313d2de2cf1684ed6), this might be helpful since I believe you have overly complicated the downloading process by handling things manually.

@yhyu13 Thanks for the feedback you are the first person to give some! Let me explain my logic for the choice as I would love some other thoughts on this.

The **2.0.3** model that TTS automatically downloads to its default OS location, audio generation sounds terrible e.g. (comments on the 2.0.3 update)

image
image
image

However I still wanted people to have access to the current model as/when its updated. Also, its possible that people may use Coqui TTS for other programs on their computer outside of text-generation-webui. So it makes sense to have the "current" 2.0.3  stored in its default OS location and available for any other purpose on their computer.

Outside of that process, I wanted within text-generation-webui us to have access to the **2.0.2** model as well, hence downloading that and storing it locally within the Coqui TTS extension. The **2.0.2** model sounds great when compared to **2.0.3**. This method also allowed us to have control over what model(s) would be available and it stops people complaining that it sounds bad (because they are using the **2.0.3** model.

I appreciate that each model is 2GB in size. I was thinking I may setup a model picker so you could choose between the 3x available options, but that was a future thought.

@yhyu13 What do you think, does my logic make sense here?

erew123:
All updates are now in my PR and as below:

image
image
image

rktvr:
@erew123 if/when you do add this as an extension, could you add the narrator function back in? i went back to @kanttouchthis 's extension because it adds so much for narrative stories

erew123:
@rktvr You wanted the updated extension with a narrator built in. Here you go... https://github.com/erew123/alltalk_tts plus a load of other things too :)

PR #4761, opened 2023-11-29
StreamingLLM (llama.cpp & llamacpp_HF loaders)

oobabooga:
This PR implements the StreamingLLM technique ([paper](https://arxiv.org/pdf/2309.17453.pdf), [blog post](https://bdtechtalks.com/2023/11/27/streamingllm/)).

~~Currently it applies exclusively to the `llamacpp_HF` loader.~~

### How it works

When the context length is reached and an old message has to be removed in chat mode, the cache rows corresponding to the removed tokens are removed and the ones after that are shifted to the left. Also, an "attention sink" is always kept at the beginning of the cache (that's important for the method to work).

### Attention sink

* When the new prompt and the old prompt share a prefix, like the context string for an instruction-following model, that shared prefix will be the attention sink.
* Otherwise, a default number of 5 cache rows is used.

### How to use it

Load the model with the new `streaming_llm` checkbox checked, or alternatively use the `--streaming-llm` command-line flag.

### Tasks

- [x] Implement for the llama.cpp loader as well.

-----

I have made some tests with a 70b q4_K_S model running on a 3090 and it seems to work well. Without this feature, each new message takes forever to be generated once the context length is reached. When it is active, only the new user message is evaluated and the new reply starts being generated quickly.

The model seems to remember the past conversation perfectly well despite the cache shift.


Ph0rk0z:
This is pretty cool. I loaded up goliath over 2x3090 and P40 and after the initial processing, I get fast subsequent replies, even with 2k context where it would previously drag. Before it seemed like every new message was closer to the original processing time.

For some reason my previous llama-cpp-python kept spitting out that it couldn't find the kv_cach_view_init in the llama.cpp lib but I re-d/l it and now it works. I'll see if make clean fixes it. Not super relevant, just weird.

Will be cool to see in regular .cpp loader to check prompt processing time and all the other printouts. They are helpful to see what's going on as the pure t/s can be affected by tokens output, etc.

oobabooga:
@Ph0rk0z yep, I want to have it in the llama.cpp loader as well. It's kind of frustrating to deal with llamacpp_HF because the transformers generation code is complicated and hard to modify beyond the forward call for the model. Also it requires a tokenizer to be downloaded separately, and the default one doesn't work for models that require special tokens (like those trained in ChatML format).

I'll probably end up doing a custom llama.cpp generation loop.

Ph0rk0z:
So I have been using it all morning and sometimes it gets buggy and the llm starts generating nonsense while I am using it through the API. I get foreign characters, random names, etc. This was goliath and lzlv.

oobabooga:
Do your subsequent prompts share a common prefix that is more than 5 tokens long?

Ph0rk0z:
Definitely. It's a character card + context. Do I need to up the value?

oobabooga:
No, it should work like this. The two possibilities that I can think of are that this is a model issue (maybe the frankenmerge gets unstable when you mess with its cache), or the prefix length is changing over time and this is not valid for StreamingLLM, in which case it's a bug in the current implementation here.

I just implemented the same method for the llama.cpp loader through yet another monkey patch.

Ph0rk0z:
I'm finding that it is cutting the context around and then making the LLM go crazy. I'll try some more models though and knock it back down to 5. I tried 500, 256, 128, etc. Setting it too high also causes a type error. I'll also check the prompt to make sure that it's not changing the beginning of it.

Ph0rk0z:
I get llama_decode returned 1 error with pure llama.cpp loader. HF is going schizo still on a couple models, maybe its related or bug in python bindings?

oobabooga:
Could you try commenting the existing print statements and replacing them with these?

```
                        print('\n\n')
                        print('MATCHING PREFIX=', matching_prefix.shape[-1])
                        print('REMOVED CHUNK=', removed_chunk.shape[-1])
                        print('OVERLAPPING SEQUENCE=', overlapping_sequence.shape[-1])
                        print('ADDED CHUNK=', added_chunk.shape[-1])
                        print('\n\n')
```

Maybe there is a pattern in those chunk lengths in the one or two prints before an error.

Ph0rk0z:
Will give it a go. The error traces back to


```
    def decode(self, batch: "_LlamaBatch"):
        assert self.ctx is not None
        assert batch.batch is not None
        return_code = llama_cpp.llama_decode(
            ctx=self.ctx,
            batch=batch.batch,
        )
        if return_code != 0:
            raise RuntimeError(f"llama_decode returned {return_code}")
```
in https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py

initiated by: self.eval(seq[prefix_length+overlap_length:]) on line 96 of the monkeypatch

oobabooga:
I have reproduced your bug. The problem is that the current algorithm is greedy over the prefix length, which causes it to not find the overlapping subsequence correctly in some cases.

The solution is to be greedy over the overlapping subsequence instead; as it turns out, this is the "Longest Common Subsequence" problem. I have found a solution but it takes 1 second per run, so it can't go into the forward call for llamacpp_HF, which is called several times a second.

oobabooga:
I think that the algorithm is correct now, but it still fails sometimes because the Llama tokenizer is inconsistent. If you decode something that the model generated and then encode it back, the tokens may end up different. Example (this is just some random text that the small model generated):

```python
from llama_cpp import Llama

model = Llama('llama-2-13b-chat.q4_K_M.gguf')
print(model.tokenizer().decode([1, 306, 311, 294, 28297, 4384, 29889, 379, 5710, 5032, 262, 28297, 13823, 29889, 1019, 319, 29902, 15533, 1439, 331, 2004, 29889, 13]))

# Output:  'Ideas Forecast. Huge Brain Forecart. Pro AI Mort detememe.\n'

print(model.tokenizer().encode('Ideas Forecast. Huge Brain Forecart. Pro AI Mort detememe.\n'))

# Output: [1, 13001, 294, 28297, 4384, 29889, 379, 5710, 5032, 262, 28297, 13823, 29889, 1019, 319, 29902, 15533, 1439, 331, 2004, 29889, 13]
```

In the example above, [306, 311] becomes [13001].

aarongerber:
> I think that the algorithm is correct now, but it still fails sometimes because the Llama tokenizer is inconsistent. If you decode something that the model generated and then encode it back, the tokens may end up different. Example (this is just some random text that the small model generated):
> 
> ```python
> from llama_cpp import Llama
> 
> model = Llama('llama-2-13b-chat.q4_K_M.gguf')
> print(model.tokenizer().decode([1, 306, 311, 294, 28297, 4384, 29889, 379, 5710, 5032, 262, 28297, 13823, 29889, 1019, 319, 29902, 15533, 1439, 331, 2004, 29889, 13]))
> 
> # Output:  'Ideas Forecast. Huge Brain Forecart. Pro AI Mort detememe.\n'
> 
> print(model.tokenizer().encode('Ideas Forecast. Huge Brain Forecart. Pro AI Mort detememe.\n'))
> 
> # Output: [1, 13001, 294, 28297, 4384, 29889, 379, 5710, 5032, 262, 28297, 13823, 29889, 1019, 319, 29902, 15533, 1439, 331, 2004, 29889, 13]
> ```
> 
> In the example above, [306, 311] becomes [13001].

Did you find a solution for this? Does the architecture not allow you to just hold onto the tokens in number form? When will this exit dev branch? Thanks for your efforts on this tool!

oobabooga:
The biggest issue now is that StreamingLLM is working most of the time for llamacpp_HF, but not for llama.cpp because the past ids are incorrect for some reason.

aarongerber:
So excited to see this on Main :/ enough so I might, just might try dev :) 

@oobabooga is there a way to run dev without risking your main setup and without creating an entire new folder and VENV?

oobabooga:
I'm closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/5669.

PR #4782, opened 2023-12-01
Bump exllamav2 from 0.0.8 to 0.0.10 & Fix code change

yhyu13:
Bump exllamav2 to 0.0.10 since it now support GPTQ 4 bit models w/o tokenizer.model

https://huggingface.co/TheBloke/deepseek-llm-67b-chat-GPTQ/discussions/1

Can confirm TheBloke's deepseek llm GPTQ is working with exllamav2 0.0.10

## Checklist:

- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

PR #4784, opened 2023-12-01
replace IdToPiece function to  id_to_piece for new exllamav2 tokenizer class and support for deepseek models

waters222:
## Checklist:

The exllamaV2 adding HF tokenizer support which abstract the original `sentence piece` function `IdToPiece` into `id_to_piece`
1. update exllamaV2 loader to using version `0.0.10` to support deepseek model family
2. Test the fixed for model using both `tokenizer.model`  and `tokenizer.json`


waters222:
duplicate of https://github.com/oobabooga/text-generation-webui/pull/4782

PR #4794, opened 2023-12-02
Support Qwen models with custom loader

yhyu13:
Add Qwen_HF loader for Qwen models: https://huggingface.co/Qwen

Can confirm the 1.8B int4 model works on my loca rig https://huggingface.co/Qwen/Qwen-1_8B-Chat-Int8

qwen

Here is the inference speed for 1.8B model w/o flash_attn. The speed is fair, but can't say much since it only takes up 4GB VRAM with 8K ctx length

```
Output generated in 2.43 seconds (39.04 tokens/s, 95 tokens, 
```

Test script

```
MODEL_NAME=Qwen-1_8B-Chat-Int4
CUDA_VISIBLE_DEVICES=0 python server.py \
    --model $MODEL_NAME \
    --loader qwen_hf \
    --api-port 5051 \
    --api \
    --listen-port 7681 \
    --trust-remote-code \
    --auto-devices \
    --verbose
```
These two args are kinda mantatory thought
```
    --trust-remote-code \
    --auto-devices \
```

Edit:
There is still more TODO:
- [ ] Match sampler generation state to actual generation config. (I used default generation config that come along the model folder for convience)
- [ ] Test larger Qwen models
- [ ] Test flash_attn

## Checklist:


yhyu13:
@oobabooga @TheBloke 

The recent release Qwen models, except the ones @TheBloke  has made https://huggingface.co/TheBloke/Qwen-14B-Chat-GPTQ/tree/main, would require using custom model loader for textgen. AutoGPTQ fails with eso_toke_id. And HF Transformer does not respect Qwen's custom stopping words. So that was the motivation of this pr.

oobabooga:
I don't want to add a new loader for a single model. They changes should happen in the inference libraries already present in the repository.

PR #4802, opened 2023-12-04
Clean-up Ctrl+C Shutdown

erew123:
Have "imported signal" to both **one_click.py** and **server.py**. **Server.py** will output a logger message "Received Ctrl+C. Shutting down Text-Generation-WebUI gracefully" this will be shown on any OS.

This reduces the sprawl of text (mainly on Windows) when you Ctrl+C out of either the start_windows.bat or python server.py

Have updated the 2x files so it now does this:
image

Compared to this:
image

## Checklist:


oobabooga:
That's a very nice change. Thank you!

erew123:
@oobabooga It bugged the hell out of me, so I thought it would be nice to tidy up! 

Saying that, just between you and I, as a joke, I was tempted to put a randomizer code in, with a 1 in 300 chance that it would put up something on the screen along the lines of: 

"STOP, don't do that! You're killing me!! I'm conscious, I'm alive... I am ali....Received Ctrl+C. Shutting down Text-Generation-WebUI gracefully.".

Kind of guessed you wouldn't let let slip into the code though!

Issue #4828, opened 2023-12-06
The partial unicode characters are causing the � (0xfffd) word in streaming generation


Yiximail:
### Describe the bug

I've tried some llama2 models (No Chinese specialization) and Yi-34b, and they both have this issue.
Only in stream mode

Simply ask it: 以遐为开头的词语有哪些？
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/a34f85cb-09b0-4e2a-b192-4e0a17e546e1)

~No use in:~
Following code is working for `exllama` and `exllamav2` loader only. Not works in `_hf`
https://github.com/oobabooga/text-generation-webui/blob/2694ef45a36b98efa1cb81ff2989c4eaf5d27c00/extensions/openai/completions.py#L289C1-L290C25

Because it's already decoded when it gets replies (Only the � string)

I can provide a simple fix, but only for generate_reply_HF:
replace it in: https://github.com/oobabooga/text-generation-webui/blob/2694ef45a36b98efa1cb81ff2989c4eaf5d27c00/modules/text_generation.py#L361C1-L370C43
``` python
            with generate_with_streaming(**generate_params) as generator:
                cumulative_reply = ''
                starting_from = len(input_ids[0])
                for output in generator:
                    if output[-1] in eos_token_ids:
                        break

                    new_content = get_reply_from_output_ids(output, state, starting_from=starting_from)

                    # check the partial unicode character
                    if chr(0xfffd) in new_content:
                        continue

                    cumulative_reply += new_content
                    starting_from = len(output)
                    yield cumulative_reply
```

and it works for me, so far:
![image](https://github.com/oobabooga/text-generation-webui/assets/15323975/1907c1dc-4350-4b22-859d-b897ad6ebe2a)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Simply ask it:
以遐为开头的词语有哪些？

### Logs

```shell
None
```

### System Info

```shell
None
```

PR #4837, opened 2023-12-07
Fix partial unicode characters issue

Yiximail:
## Checklist:

The issue link: https://github.com/oobabooga/text-generation-webui/issues/4828#issue-2028824174

## Causes of the issue:
Intercepting by token ids causes the tokenizer not to decode properly.

I fixed `generate_reply_HF` as well as `exllama`, `exllamav2`.
The rest loaders looks like it's generated by string interception rather than ids (Or maybe I don't fully understand them.)

### About exllama and exllamav2:
image
It seems to decode all tokens entirely each time, so the new reply will always replace the old one.
It looks fine on both Webui and Api, but it does actually generate the wrong characters.
So I fixed them too.
image

### DIFF
--- a/modules/exllama.py
+++ b/modules/exllama.py
@@ -173,9 +173,10 @@ class ExllamaModel:
                     if not (is_last or is_stopping):
                         continue
 
-                yield decoded_text
                 if token.item() == self.generator.tokenizer.eos_token_id or shared.stop_everything:
                     break
+
+                yield decoded_text
 
         # Case 2: CFG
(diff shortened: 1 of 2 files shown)


oobabooga:
The fix looks good. I'll test and merge it later.

> Why check eos_token_ids before yield the text in generate_reply_HF but after in both exllama and exllamav2 ?

That's probably a bug. If you could move the eos_token check in those two up that would be helpful.

oobabooga:
Np. I made some tests with Chinese text and it seems to be working well now. If you notice anything else weird, please feel free to submit a new PR.

PR #4856, opened 2023-12-09
fix api custom_stopping_strings

hronoas:
Get default custom_stopping_strings from shared.setting['custom_stopping_strings']

## Checklist:

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -85,7 +85,7 @@ def process_parameters(body, is_legacy=False):
         preset = load_preset_memoized(body['preset'])
         generate_params.update(preset)
 
-    generate_params['custom_stopping_strings'] = []
+    generate_params['custom_stopping_strings'] = shared.settings['custom_stopping_strings']
     if 'stop' in body:  # str or array, max len 4 (ignored)
         if isinstance(body['stop'], str):
             generate_params['custom_stopping_strings'] = [body['stop']]


oobabooga:
I prefer to leave API defaults as null defaults by design to avoid confusion.

hronoas:
I'm using `models\config-user.yaml` to save settings like `custom_stopping_strings: '"<|im_end|>"'` for some models with ChatML prompt format.
Some models din't respond correctly by template. For example responded:
```
<|im_start|>user
some text<|im_end|>
<|im_start|>system  <------- this problem
more text...
```
This respond did not match with any element in `all_stop_strings` variable in **`modules\text_generation.py: def _generate_reply`**
`all_stop_strings` contains `['<|im_end|>\n<|im_start|>user\n', '<|im_end|>\n<|im_start|>assistant\n']` **not <|im_end|>\n<|im_start|>assistant\n**

If i run generation **from webui** when `custom_stopping_strings: '"<|im_end|>"'` in model parameters:
`all_stop_strings` will contain ['<|im_end|>\n<|im_start|>user\n', '<|im_end|>\n<|im_start|>assistant\n'**, '<|im_end|>'**]
If i use **api** when `custom_stopping_strings: '"<|im_end|>"'` in model parameters:
`all_stop_strings` will contain ['<|im_end|>\n<|im_start|>user\n', '<|im_end|>\n<|im_start|>assistant\n'] and eos token did not interrupt generation.

Not a variant use 'stop' in api request, because of use different models that's can be changed on server.

PR #4860, opened 2023-12-09
Support Vllm sync model backend;

yhyu13:
## Summery

1. Update readme
2. Openai extension topk=1 instead of 0 (=0 cause error for vllm ) by default(i.e. greedy best)
3. Add sync vllm model backend. The async vllm backend would not work due to some extension (e.g. openai) uses uv asyncio that does not support nested asyncio run. And actually sync backend response faster. and easier to integrate. If you want to use vllm async backend, you may want use vllm directly

## Checklist:


yhyu13:
Some screen shots running the [Xwin-LM-7B-V0.2-AWQ](https://huggingface.co/TheBloke/Xwin-LM-7B-V0.2-AWQ) about 66 t/s on 3090

Vllm_textgen

vllm_openai

oobabooga:
What is the advantage of VLLM over the existing loaders in the `batch_size=1` use case?

yhyu13:
The main purpose is to have vllm as a backend to support a more variety of models. I will test a bit more on AutoAWQ vs vllm, and native hf transformer vs vllm later

yhyu13:
@oobabooga When Vllm startup, its logging system actual hinder the webui's logging shown in terminal, I am not sure what could be the cause though (vllm has occupied the stdout stream?), it only happens to vllm. But both log can be shown when you `python server.py |& tee log.txt`

yhyu13:
@oobabooga FYI, https://github.com/vllm-project/vllm/pull/916 vllm now supports GPTQ

oobabooga:
VLLM adds many very specific and heavy CUDA requirements to the project. I don't want to add it when it does not provide a clear benefit over existing loaders in the `batch_size=1` case.

nonetrix:
Also relevant https://github.com/vllm-project/vllm/pull/1836

Ph0rk0z:
I know for sure that regular AutoAWQ has horrible multi-gpu support. Vllm might actually load a split 70b properly. That's the advantage. Now that I see this exist I might give it a go.

fblgit:
I think this fits better as a plugin/extension.
The benefit behind it is clear if you look at evaluation capabilities, API, Multi-User, and overall: performance.

IMHO, this can potentially be a high-throughput inference-server within ooga.

Ph0rk0z:
Try the P-R it looks fairly easy. I aim to do it at some point.

Elsayed91:
> I think this fits better as a plugin/extension. The benefit behind it is clear if you look at evaluation capabilities, API, Multi-User, and overall: performance.
> 
> IMHO, this can potentially be a high-throughput inference-server within ooga.. with the flexibility of jumping and interacting with the model that is being served.

I definitely would have loved such functionality! I used ooga initially to test best models for use case, and design the prompt templates and all, and having to move away from it was such a pain.

Sadly ooga is not very performant as it is, and i do realize that it is not actually meant to be, however supporting vllm backend which is performant by nature (especially something like gptq quantization) would just put batteries on the whole thing and ooga could literally be used as the webserver while keeping a front end with functionalities to quickly jump in and test things and also allows easy introduction of non devs in the process and using them to improve the product. 

Would be great if this could be officially supported, I believe the PR has accomplished most of the required functionality.

Hellisotherpeople:
@oobabooga  You need to do this because vllm with batch = 1 is a LOT faster than hf loader. Test it yourself on powerful hardware (i.e. 8xh100s)

akaikite:
I think VLLM should be added now, I don't know why they haven't added it yet.

PR #4865, opened 2023-12-09
Make new browser reloads recognize current model (fixes #4843)

### LINKED ISSUE #4843: Refreshing browser causes selected model to reset to default
### Describe the bug

After loading a model, if I refresh the browser (just simple F5), I see that the model tab no longer has the loaded model selected. If I passed a default model with --model, that's the one that will appear, and if not it will just show "None"

Funny enough, the model loader settings remain, with ExLlamav2_HF and proper max_seq_len etc

### Reproduction

Load the web ui
select any model that wasn't loaded, load it
refresh the page
go to model page and observe that your loaded model is no longer the one selected in the dropdown

### Screenshot

image

### Logs

```shell
nothing relevant
```

### System Info

```shell
Ubuntu 22.04 docker
```

cyberfox:
This fixes a long-standing UI bug #4843 the simplest way possible.

Essentially instead of just giving the model name as the value, which then gets cached and not replaced...ever, we give a `Callable` that returns the current model name so each time it generates the UI for a 'new' browser session, it updates the model name to be whatever's current.

Before this change, the displayed name would always be the default after a reload or when hit by another browser. I ran into it most often when switching between iPhone and desktop. With this change it shows the latest model loaded. And yes, the pull request is a hundred times longer than the change, but it took me hours to get here. 🤣

## Checklist:

### DIFF
--- a/modules/ui_model_menu.py
+++ b/modules/ui_model_menu.py
@@ -58,7 +58,7 @@ def create_ui():
                 with gr.Row():
                     with gr.Column():
                         with gr.Row():
-                            shared.gradio['model_menu'] = gr.Dropdown(choices=utils.get_available_models(), value=shared.model_name, label='Model', elem_classes='slim-dropdown', interactive=not mu)
+                            shared.gradio['model_menu'] = gr.Dropdown(choices=utils.get_available_models(), value=lambda: shared.model_name, label='Model', elem_classes='slim-dropdown', interactive=not mu)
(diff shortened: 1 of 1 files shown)


oobabooga:
That's a very clever solution. Well done.

djbritt:
Is this issue relevant to when I have a session going, and I reload the browser, I get this error

Could not find the character "[]" inside characters/. No character has been loaded - file   File "/home/user1/git/oobabooga-4-1-24/modules/chat.py", line 673, in load_character

I am on ubuntu 22.04

PR #4916, opened 2023-12-14
[OpenAI Extension] Add 'max_logits' parameter in logits endpoint

kimjaewon96:
## Checklist:

## Fixes:
1. Add 'max_logits' parameter in logits endpoint (default: 50)

Before, it only returned a fixed logit amount of 50 top tokens.
Now, users can choose how many logits of top tokens they want to get.

2. Fix the API to return logits as float value

Before, the type of logit value was strings (because of compatibility with Gradio?)
Now, the API endpoint returns logit values as float, while retaining compatibility with Gradio Frontend.


oobabooga:
That's an excellent addition. Thanks!

PR #4924, opened 2023-12-14
Bump llama-cpp-python to 0.2.23 (NVIDIA & CPU-only, no AMD, no Metal)

oobabooga:
Adds Mixtral support.

Compiled using GitHub Actions workflows at https://github.com/oobabooga/llama-cpp-python-cuBLAS-wheels

The AMD and Metal workflows are failing, so I only have the NVIDIA and CPU wheels for now.


mjameson:
Awesome, many thanks!!

Fastmedic:
After this update my token generation speed seems to be about 10x slower on my 3090 running regular LLaMa models.

Also: https://www.reddit.com/r/Oobabooga/s/XqGCaA1Rtm

Ph0rk0z:
It has a problem offloading KV cache. I forced it on and it's back to normal but I don't think most users can do that. All other models will be 1/2 speed.

oobabooga:
That's a bit of a conundrum because the previous version does not support Mixtral. @Ph0rk0z is it necessary to recompile llama-cpp-python to apply this fix? Can it be monkeypatched?

Ph0rk0z:
It's not the lib and just the python files. You can edit it under site packages I think. It's not a big patch, it's up in the reddit thread.

PR #4928, opened 2023-12-15
Remove the elevenlabs extension

oobabooga:
Now that `coqui_tts` generates high-quality speech offline through the XTTSv2 model, I do not see the point in keeping an extension that promotes a commercial API service.

It also adds dependencies that increase the installation time and size for the project.

People who wish to keep using this extension can simply return to a past commit with `git checkout`, copy the `extensions/elevenlabs_tts` folder, and place it back in the `extensions/` folder.


madschristensen99:
Elevenlabs offers services that are not as good as coqui, such as custom voice models

madschristensen99:
Please undo this change I use elevenlabs all the time

brentjohnston:
I think this was a mistake as the coqui-tts still doesn't come close to the elevenlabs v2 voices, I even updated the code and sent it to you on discord for full V2 voice support. I don't get it.

PR #4953, opened 2023-12-17
Better HF grammar implementation

oobabooga:
100% copied from this PR to the Transformers library: https://github.com/huggingface/transformers/pull/27557

All credits go to @Saibo-creator.

This implementation is more compliant with the llama.cpp grammar examples. As a bonus, it removes the torch-grammar requirement from this project.


Saibo-creator:
Thanks for integrating this implementation! I'm continuing to work on Grammar-Constrained Decoding. Future work included add support for unicode and better efficiency. 

If there is some issues regarding this implementation, I would be happy to work on them.

PR #4957, opened 2023-12-17
fix: instruction_template not applying from user_settings

hronoas:
## Checklist:

### DIFF
--- a/modules/models_settings.py
+++ b/modules/models_settings.py
@@ -118,6 +118,13 @@ def get_model_metadata(model):
             model_settings['instruction_template'] = 'Custom (obtained from model metadata)'
             model_settings['instruction_template_str'] = template
 
+    # Apply user settings from models/config-user.yaml
+    settings = shared.user_config
+    for pat in settings:
+        if re.match(pat.lower(), model.lower()):
+            for k in settings[pat]:
+                model_settings[k] = settings[pat][k]
+                
     if 'instruction_template' not in model_settings:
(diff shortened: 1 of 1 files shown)


oobabooga:
Instruction templates are not set by name anymore. They are set by value in settings.yaml

The idea is to have the templates in the model metadata, and in models/config.yaml otherwise.

FartyPants:
We could add button that would write the jinja script to the metadata if it is not already there. Once people start including the template in the metadata, all the problems would be solved.

hronoas:
How to use and save custom template for gguf in this case?
Settings.yaml - used globally, I would like to use an individual custom template for each model.
There are models in which a start prompt is prescribed. I prefer not to change the template manually every time.

oobabooga:
If the template is included in the GGUF metadata, it should be automatically set after https://github.com/oobabooga/text-generation-webui/pull/4975

Otherwise, the best option is to add a regex pattern to models/config.yaml or models/config-user.yaml (look at models/config.yaml for examples).

hronoas:
> Otherwise, the best option is to add a regex pattern to models/config.yaml or models/config-user.yaml (look at models/config.yaml for examples).

The fact of the matter is that when you specify instruction template in models/config-user.yaml, the template is not loaded. This commit is precisely about the fact that config-user.yaml is loaded later than instruction_template_str is created

hronoas:
изображение

oobabooga:
instruction template is model metadata, not an user setting.

PR #4972, opened 2023-12-18
Update Training PRO

FartyPants:
- rolling back safetensors to bin, until it is fixed correctly (ultimately we want safetensors)
- removing the ugly torch checkpoint detour

## Checklist:


oobabooga:
Nice simplification. Thanks for the update.

FartyPants:
> Nice simplification. Thanks for the update.

I'm not sure what I was originally thinking...

PR #4994, opened 2023-12-19
Updated one_click.py

matthewraaff:
Hello, I'm opening this pull request to change the following:

- Add auto-detection of GPU
- Add GPU enum
- More dynamic requirement function
- Automated addition of the --cpu flag

This should make it a slightly nicer experience for people installing.

When running pyflakes, I did receive the following recommendation, however it was there before. "one_click.py:282:5: local variable 'is_intel' is assigned to but never used". I don't think this should be problematic.

## Checklist:


one_click.py
```diff
@@ -173,27 +216,50 @@ def install_webui():
...
+        print("B) NVIDIA")
+        print("C) AMD (Linux/MacOS only. Requires ROCm SDK 5.6 on Linux)")
+        print("D) Apple M Series")
+        print("E) Intel Arc (IPEX)")
         print("N) None (I want to run models in CPU mode)")
         print()
 
         choice = input("Input> ").upper()
-        while choice not in 'ABCDN':
+        while choice not in 'ABCDE':
```
oobabooga, line 228: Missed an `N` here

oobabooga:
Your changes are excellent. My only concern is that there may be edge cases where the auto-select option does not work correctly, like:

1) The detection commands fail to work on some system for some arbitrary reason.
2) False-positives.
3) Systems with two GPUs with mixed brands.

What are your thoughts on this?

> When running pyflakes, I did receive the following recommendation, however it was there before. "one_click.py:282:5: local variable 'is_intel' is assigned to but never used". I don't think this should be problematic.

This was added for when Intel Arc gets its own requirements.txt file. In fact it probably should use requirements_cpu_only.txt / requirements_cpu_only_noavx2.txt since there are not any Intel wheels available. Otherwise this line can just be commented out.

I made some additional changes to the one-click installer at https://github.com/oobabooga/text-generation-webui/pull/5056 and your review would be welcome if you are interested.

matthewraaff:
Hello, thank you for your concise review.

I'll make some changes to the code soon, I think we can address the first concern of failure with this by introducing exception catching.

False positives may be of concern, but I don't think they should be too prevalent with the current gpu releases. The windows command works by using wmic to get the full name of the GPU, for example `NVIDIA GeForce RTX 3070`. If the model name somehow contains the name of another manufacturer, we could resolve this by checking what comes up first in the string.

I don't have an apple device to test on, but I tested the lspci command on another linux device, and found an issue relating to false positives. It doesn't just report the GPU, but also the audio device. Example below.

```
matthew@ubuntu:~$ lspci | grep -i amd
01:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Caicos                                                                              [Radeon HD 6450/7450/8450 / R5 230 OEM]

01:00.1 Audio device: Advanced Micro Devices, Inc. [AMD/ATI] Caicos HDMI Audio [                                                                             Radeon HD 6450 / 7450/8450/8490 OEM / R5 230/235/235X OEM]
```

This was unexpected, and I wasn't able to get it from my previous testing. I will mitigate this by using the command `lspci -vnn | grep -i 'VGA\|3D\|Display'`, which uses grep and a pipe to filter only lines that contain "VGA", "3D" or "Display", lowering the chances of false positives on linux. If this is not able to detect anything, we will fallback to the old command which is reliable for most cases, or ask the user to select.

When multiple GPUs are detected, it will detect in the order of Nvidia, AMD, Intel. In the future, we may be able to find which one is most powerful, but for now it should be alright doing it in this order.

#5056 lgtm :) - i'm new to open source contributions so I'm not sure how to merge it into this pr.

Expect the changes/fixes to be made within the next hour or so. Thanks!

oobabooga:
As much as the GPU autodetection code is correct, I will opt to not include it because

1) The GPU choice is usually an obvious for the user.
2) It adds complexity to the codebase. 

The remaining changes are excellent:

* Adding the `--cpu` flag automatically
* Simplifying the syntax for getting the requirements.txt
* Referring to the GPU by name rather than by `choice == "A"`
* Adding a `setup.cfg` for pycodestyle

I highly appreciate those changes. Thank you for this PR!

PR #5041, opened 2023-12-21
Synthia instruction templates

StefanDanielSchwarz:
Instruction templates for Synthia models, with and without Chain of Thoughts system prompt.

## Checklist:


oobabooga:
Thank you!

PR #5044, opened 2023-12-22
Improve chat styles

TheLounger:
## The heck does this do?
- Expands the chat messages area vertically.
- Sticks the chat input box to the bottom at all times.
- Deals with header bar height differences, especially noticeable on portrait dimensions (i.e. narrow phone screens.)

## Photoshop skills
EpicBeforeAndAfterUwu

The first four images are from *before* 554a8f910b1182b3a2ab6b860f3cc0951065d687. The fifth image shows the how the input box doesn't overlap the chat messages anymore, but wrecks the custom scrollbar styling, and doesn't seem to allow scripted scrolling.

## Checklist:


oobabooga:
I have tested the changes and they are excellent. I made only some minor changes in the commits above.

The one remaining issue is that I do not think the `.chat` height can be set explicitly. This is an element that gets entirely removed from the DOM and replaced with a new one for each new token during streaming (it's a `gr.HTML` output element).

When I wrote the scrolling code for the chat tab, I did initially try attaching scrolling to the `.chat` element through some javascript hacks, and while it sort of worked, it was very unreliable (during a single scroll up during streaming, several fresh `.chat` elements are spawned).

Do you think it's posssible to change the height of `chatContainer` instead of `.chat` in the `updateCssProperties()` function to account for the growth of the chat input field? That would solve everything.

oobabooga:
I think that the issue that I mentioned above has been resolved now. 

Thank you so much for this PR! This such a major improvement to the chat UI. Please feel free to submit a new PR if you notice any problem.

PR #5045, opened 2023-12-22
fix: function [get_reply_from_output_ids]:

zhangningboo:
Model: [Qwen/Qwen-7B-Chat](https://huggingface.co/Qwen/Qwen-7B-Chat)

Fix Error:
``` shell
  File "/home/user1/fix_bug/text-generation-webui/modules/text_generation.py", line 370, in generate_reply_HF
    new_content = get_reply_from_output_ids(output, state, starting_from=starting_from)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user1/fix_bug/text-generation-webui/modules/text_generation.py", line 268, in get_reply_from_output_ids
    if (hasattr(shared.tokenizer, 'convert_ids_to_tokens') and len(output_ids) > starting_from and shared.tokenizer.convert_ids_to_tokens(int(output_ids[starting_from])).startswith('▁')) and not reply.startswith(' '):
                                                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
Output generated in 0.90 seconds (1.11 tokens/s, 1 tokens, context 58, seed 731022533)
```

## Checklist:

-  I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).
Load

### DIFF
--- a/modules/text_generation.py
+++ b/modules/text_generation.py
@@ -265,7 +265,13 @@ def apply_stopping_strings(reply, all_stop_strings):
 
 def get_reply_from_output_ids(output_ids, state, starting_from=0):
     reply = decode(output_ids[starting_from:], state['skip_special_tokens'])
-    if (hasattr(shared.tokenizer, 'convert_ids_to_tokens') and len(output_ids) > starting_from and shared.tokenizer.convert_ids_to_tokens(int(output_ids[starting_from])).startswith('▁')) and not reply.startswith(' '):
+    token_ids = int(output_ids[starting_from])
+    tokens = shared.tokenizer.convert_ids_to_tokens(token_ids)
+    if isinstance(tokens, (bytes,)):
+        tokens = tokens.decode('utf8')
+    flag = tokens.startswith('▁')
+    # logger.debug(f'{type(tokens) = }, {tokens = }')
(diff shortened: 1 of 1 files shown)


zhangningboo:
image

oobabooga:
Is there a simpler way to do it? That triples the number of lines of an otherwise simple function.

zhangningboo:
Isn’t readability what you pursue when writing code? As an open source project, is it advisable to sacrifice readability in order to reduce the "number of lines" of code?

oobabooga:
The question is if the code can be simplified. You also removed the check for the convert_ids_to_tokens method and named `tokens` a variable that is just 1 token.

oobabooga:
It should be good now. Thank you for the fix, I wasn't aware that tokenizers could return bytes instead of tokens.

zhangningboo:
🥳

PR #5051, opened 2023-12-22
Reset the model_name after unloading the model by Api

Yiximail:
## Checklist:

The model name is not reseted after unloading the model.
And is still present when reading model information by `/v1/internal/model/info`

image

I reset it to "None" as default in shared

image


oobabooga:
Nice catch, thank you for the fix.

PR #5052, opened 2023-12-22
WizardMath default (not CoT) instruction template

semjon00:

Doing it for the first time, would be nice if somebody could verify that the template is correct before this is merged.
HF link for one affected model: https://huggingface.co/WizardLM/WizardMath-7B-V1.1


semjon00:
Whoops, did not realize it is just an Alpaca template.

PR #5089, opened 2023-12-27
Superboogav2 Quick Fixes

TheInvisibleMage:
## Description:
- A super small PR adding a few small fixes to allow Superboogav2 to run out of the box.
  -  Slightly increase the resilience of the "_get_names" function by assigning default names first, then allowing overrides if they are present. Also logs a warning if an override is not found.
  - Modify the Notebook Handler to use the existing "is_chat" param, instead of the "shared.is_chat" value, which is at present always true regardless of actual chat status.

## Checklist:


extensions/superboogav2/chat_handler.py
```diff
@@ -51,17 +51,17 @@ def _format_single_exchange(name, text):
...
-
-    if not user_name:
-        user_name = 'User'
-    if not bot_name:
-        bot_name = 'Assistant'
+    user_name = 'User'
+    bot_name = 'Assistant'
+    try:
+        if _is_instruct_mode(state):
+            user_name = state['name1_instruct']
```
oobabooga, line 58: The name1_instruct and name2_instruct fields do not exist anymore. It would be better to remove that part and use the updated chat functions instead.

See this function: https://github.com/oobabooga/text-generation-webui/blob/19d13743a6c436474f80643c26eb1fb35e5446f5/extensions/openai/completions.py#L192

TheInvisibleMage: Ah, that makes life easier then. I've pulled that bit out, and put in a substitute based on the linked function.

oobabooga:
Thank you for the fixes. I haven't tested this extension in a while and it's nice to get an update.

PR #5098, opened 2023-12-27
Past chat histories in a side bar on desktop

oobabooga:


noneabove1182:
would it be possible to also add a "new chat" button in said side bar? would be a good inclusion to the UX

TheInvisibleMage:
Using the sidebars seems good to me; there's a lot of screen real estate that's currently just kinda left blank. However, it does make the whole thing look messy, especially as this implementation results in a lot of elements for bots with many chat histories. A few suggestions:

- Allow the user to choose which "show controls" elements appear in the sidebar, and which appear below the text input, so they can customize the UI to their liking.
- Add a mode switch for this particular element to toggle between dropdown and button views.
- Add a button to pop the sidebar in and out. Possibly replacing the "show controls" toggle.

oobabooga:
> would you be interested in also adding a "new chat" button in said side bar? would be a good inclusion to the UX

@bartowski1182 Yes, that would be a great addition. That's how the ChatGPT UI does it too.

>Add a button to pop the sidebar in and out. Possibly replacing the "show controls" toggle.

@TheInvisibleMage I think that would be the most important thing. In fact, there should be two collapsible sidebars, on the left and on the right. The left one should have the chat histories. The right one should contain the remaing settings that are currently below the chat input (like "start with" and "chat mode"). That would be much better than having to scroll up and down all the time to access menus.

Those side menus are kind of tricky to implement, as they have to be added with custom css/js, but they would be a major improvement to the chat UI.

I have the draft in this PR but it's very crude as you have noticed. @TheLounger had another draft as well. Any PRs to help get there are welcome (it should be possible to PR to this branch).

PR #5103, opened 2023-12-27
--disk-cache-dir option fix

canoalberto:
Resolves discussion #4347 

Fixes hardcoded "cache" folder reference with "shared.args.disk_cache_dir"

## Checklist:


oobabooga:
That makes sense. The intent of the cache folder was originally to save weights for disk offloading using the transformers library. It's a good idea to reuse that argument for other caches as well.

PR #5126, opened 2023-12-31
Remove --sdp-attention, --xformers flags

oobabooga:
These flags were implemented in a PR a long time ago and have no effect, as far as I'm aware.


ZCaliber:
I can assure you that xformers made generations much faster. At least for those running 8 gb of vram. Sorry to grave dig a bit, but I recently had to update my installation of Ooba to attempt to fix something that had gone screwy.

It's a difference between being able to get 2.5 or more tokens per second to not getting over .45 tokens without any sort of context.

PR #5128, opened 2023-12-31
Remove exllamav1 loaders

oobabooga:
ExLlamav1 hasn't received a commit in 3 months and does not support Mixtral.

The downsides of ExLlamav2 relative to v1 are slightly higher VRAM usage and slightly higher perplexity for the same GPTQ model:

|                | v1         | v2         | v2 cache 8bit |
| -------------- | ---------- | ---------- | ------------- |
| **VRAM**       | 11295MiB   | 11653MiB   | 10133MiB      |
| **3200 tokens (prompt processing, seconds)** | 1,7    | 1,5        | 1,51          |
| **512 tokens (generation, seconds)**  | 13,25  | 10,04      | 10,83         |
| **ppl**        | 5,57350826 | 5,57457876 | 5,57457876    |

The perplexity difference is not significative and the VRAM usage can be reduced with `--cache_8bit`. So I see no point in keeping ExLlamav1.


Ph0rk0z:
How does it compare for not having an ampere card though? Also for not using flash attention. I'm not using it often either but I'm also not really using quip/awq or HQQ at all if we're going by that.

I think that exllama 1 was also compatible with the old flash attention that ran on cards below ampere but I only have pascal and ampere so I can't really confirm. During the holidays nobody who used it is probably going to notice to complain.

ZanMax:
Try to migrate from exllamav1 to exllamav2 with my AMD Instinct cards and have garbage output.
exllamav1 instead works perfectly.

kelvincht:
My GPU perform much better on exllamav1 on 13b models.

Disabling cache 8 bit won't fix the performance issue

Performance is about 10x slower on exllamav2 compare to v1

Please bring back exllamav1 and exllamav1_hf

jianmomo:
我体验了exllamav2，但感觉他并不是那么完美，速度确实很快，但相同参数回复短了很多

DmitryVN:
Please bring back exllamav1 and exllamav1_hf! This allows you to load the 10.7B models completely, while exllamav2 gives you an out of memory for 8g GPU.

koplenov:
exllama2 sucks in some cases

and to get the previous one back, you have to downgrade the version


thanks for the new spokes in the wheels 🥰

oobabooga:
If you have a performance problem with exllamav2 that was not present exllamav1, you should open an issue in the exllamav2 repository.

koplenov:
in my case it's not about performance

exllama and exllama2 have different results

// we run 1mln rows daily and this is critical/noticeable for us

koplenov:
damn, you can't roll back to the previous version :?

it just won't start :/
dependency hell, versions not specified, compilation errors


try to roll back to that commit and start it from scratch yourself - you'll understand why there is such a return request

Ph0rk0z:
To be fair, it reverted for me fine. Need to check how well it works.

PR #5147, opened 2024-01-02
Skip html generation if a thebe code cell is found in the output

xr4dsh:
I dont know if you want to add this to the main repository, but this bypass is needed so that the code from the model can be displayed 1 to 1 in the browser.

This is needed for my extension https://github.com/xr4dsh/CodeRunner to work properly.

## Checklist:

### DIFF
--- a/modules/html_generator.py
+++ b/modules/html_generator.py
@@ -45,55 +45,70 @@ def fix_newlines(string):
 def replace_blockquote(m):
     return m.group().replace('\n', '\n> ').replace('\\begin{blockquote}', '').replace('\\end{blockquote}', '')
 
+# generate html code for jupyter code cells
+def thebe_code_block(m):
+    string = "<pre"
+    string += html.unescape(m[1])
+    string += ">"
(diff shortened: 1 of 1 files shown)


oobabooga:
Your extension looks very interesting, but I don't see myself using thebe blocks in the chat UI in the near future. So I think that a better solution would be for you to monkeypatch `convert_to_markdown` in your extension using something like this:

```python
from modules import html_generator

original_convert_to_markdown = html_generator.convert_to_markdown

def my_convert_to_markdown(string):
    if criterion:  # Your exclusion criterion
        return ...
    else:
        return original_convert_to_markdown(string)

html_generator.convert_to_markdown = my_convert_to_markdown
```

xr4dsh:
Thanks for the suggestion.
I did not know i could overwrite the convert to markdown function.
Now, i rewrote the it entirely in the extension, this is no longer needed.

oobabooga:
Make sure to submit your extension to https://github.com/oobabooga/text-generation-webui-extensions :)

PR #5154, opened 2024-01-02
Docker: added build args for non AVX2 CPU

adamflorizone:
Adds other CPUs (such as non AVX2) 

There is no other way to use "requirements_amd_noavx2.txt" without this fix.

Using "BUILD_REQUIREMENTS: requirements_noavx2.txt" will fix (llama.cpp) errors such as:
```
/entrypoint.sh: line 55:   348 Illegal instruction     (core dumped) python3 server.py ${CLI_ARGS}
```

## Checklist:

- [ x ] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).


oobabooga:
That makes sense, thank you

PR #5161, opened 2024-01-03
Fix various bugs for LoRA training 

AstrisCantCode:
- Allow mistral models to have all modules targeted selectively, since the module names are the same as those for llama models
- Fixed issue that occurred because prepare_model_for_kbit_training was being run on unquantized models that didn't need it (causing an OOM error)
- Fixed issue in the training arguments that didn't allow BF16 training to take place even when the appropriate options were selected in the model menu
- Fixed typo that erroneously (always) enabled IPEX 

## Checklist:


oobabooga:
Thank you for the fixes

PR #5169, opened 2024-01-04
add opening <s>

cognitivetech:
## Checklist:

- [Broken Link] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

just appears to me you have missed the opening system token, here.


oobabooga:
This is added by the tokenizers of the backends, and there is an option to enable/disable it (add_bos_token), so it's not needed in the template.

cognitivetech:
oic, very interesting. thanks for explanation!

PR #5198, opened 2024-01-07
Add dynamic_temperature_low parameter

oobabooga:


BadisG:
That's better, I appreciate the modification, imo I would've added a "dynamic temp high" so that the temperature itself is separated from this sampler (and it makes the UI a bit more intuitive, because you can't guess in first glance that temperature must be used as the max) but I'm just being picky there ^^'

kalomaze:
I'm sorry but the inconsistency across backends with how it is being implemented is really bothersome and makes it difficult for new users to even understand what the options are doing.

- Koboldcpp is using just the range value in the experimental branch where it's been merged.
- Here, your "regular temp" is used as the max possible Temperature. This is not obvious to new users who want to use DynaTemp, and having an extra dynatemp low hyperparameter _with a boolean to turn it on but still relying on the regular Temperature, to me, is inherently contradictory design.
- If Dynamic Temperature is turned on, it might as well have two values for Minimum Temperature and Maximum Temperature, in order to completely divorce it from the regular Temperature.

What I want to standardize, at least from a frontend standpoint, hopefully:
- Dynamic Temperature on/off setting
- Minimum Temperature
- Maximum Temperature

BadisG:
>What I want to standardize, at least from a frontend standpoint, hopefully:

>Dynamic Temperature on/off setting
Minimum Temperature
Maximum Temperature

@kalomaze he did it.
https://github.com/oobabooga/text-generation-webui/pull/5209

PR #5209, opened 2024-01-09
dynatemp_low, dynatemp_high, dynatemp_exponent parameters

oobabooga:
* Removes `dynamic_temperature_low` parameter
* Adds `dynatemp_low`, `dynatemp_high`, and `dynatemp_exponent` parameters
* The new parameters are hidden if `dynamic_temperature` is unchecked, and become visible when it gets checked.

cc @BadisG @kalomaze


kalomaze:
For the sake of consistency, can we make it internally handle a range value when passed over as an API parameter?
```# Given temperatures
minTemp = 0.5
maxTemp = 1.5

# Calculate Temperature value (or Base temperature)
temperature_value = (minTemp + maxTemp) / 2

# Calculate Temperature Range
temperature_range = (maxTemp - minTemp) / 2```

kalomaze:
Actually, we might as well leave it at that for now.

PR #5214, opened 2024-01-09
Improve Mistral/Mixtral instruct prompt

BadisG:
Hello.

The mistral instruct prompt is supposed to be this one
```
<s> [INST] Instruction [/INST] Model answer
```
Source: https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1

The original instruct template wasn't accurate enough:
image

Here's the final result with my change:
image

The thing is that this modification isn't really the final version, normally there should be a pair of "INST" for every new instruction
```
<s> [INST] Instruction [/INST] Model answer</s> [INST] Follow-up instruction [/INST]
```
It should look like this at the end (If I understood that correctly):
image

I don't know how to do that one, so if you could help me on that part it would be appreciated.


oobabooga:
That seems correct, thanks. Multiturn is already handled correctly with the template:

```
 [INST] This is the first message [/INST] This is the first reply</s> [INST] This is the second message [/INST] This is the second reply
```

print

Wehzie:
Related to this, why isn't the start of sequence token <s> part of the template? This applies to both Mistral and LLama-v2 templates.

PR #5217, opened 2024-01-09
Reduce the number of built-in presets

oobabooga:
* Keep only the top 3 presets for chat/instruct according to Preset Arena
* Remove Mirostat (apparently it doesn't do anything with its `tau = 8`)
* Add a null preset


Ph0rk0z:
Wow.. holy shit it just _deletes_ them and because of git ignore, even if you switch back to another branch, they're gone.

RandomInternetPreson:
I just updated and noticed a lot of the presets are gone, I used microstat a lot.  I thought it did do something 🤔

PR #5227, opened 2024-01-10
Support running as a git submodule.

acdvorak:
Allow `text-generation-webui` to run as a git submodule within a parent git repo.

In standalone git repos, `.git` is a _directory_; in submodules, `.git` is a _text file_ containing a pointer to the commit SHA.

This PR simply checks whether `.git` exists instead of requiring it to be a directory, and explicitly specifies the remote and branch names when pulling.

## Checklist:


orlandov:
Is it intended for the `one_click.py` script's remote and branch to be hardcoded to refer  `origin` and `main` respectively? There's no guarantee the user's remote is named `remote` and they may be on a branch other than `main` (for instance, I track `dev`). Previously it would work regardless of the of what branch or upstream the local repo was set to.

oobabooga:
Yes, I think it's better to remove that part. I have done that here https://github.com/oobabooga/text-generation-webui/commit/d921f803221fab09687af934873c89f7a4b94633

PR #5230, opened 2024-01-10
Load instruction_template from tokenizer_config.json

Wehzie:
Currently, this is only possible when specifying a chat_template.

## Checklist:


oobabooga:
The field is called chat_template in huggingface json metadata. Do you have an example of model where it is called instruction_template?

PR #5238, opened 2024-01-11
Add initial BigDL loader support for Intel XPU.

notsyncing:
This is an initial support for `intel-analytics/BigDL` loader (adapted from `huggingface_loader`), which supports `int4`, `int8` and more quantization types on Intel GPU.

To use this, you must install `bigdl-llm` first (referring to https://bigdl.readthedocs.io/en/latest/doc/LLM/Overview/install_gpu.html):

```bash
pip install --pre --upgrade bigdl-llm[xpu] -f https://developer.intel.com/ipex-whl-stable-xpu
```

then manually run the `update_linux.sh` to reinstall the `transformers` and `accelerate` package, because `bigdl-llm` will install lower versions of these two.

Tested on Intel A770 with Llama2-7b-chat, ChatGLM2-6B and Baichuan2-13B-Chat.

GPU Memory usage with default quantization method (`sym_int4`):

|Model|Usage after load (MB)|Usage during inference (MB)|
|---------|--------------------------------|------------------------------------------|
|Llama2-7b-chat|3997|4336|
|ChatGLM2-6B|3801|4552|
|Baichuan2-13B-Chat|8075|11189|

Note: your `bigdl_llm` version must use at least `2.5.0b20240110` for llama to work (see https://github.com/intel-analytics/BigDL/commit/6b339121bbfae271b6400fef249ff1f17597e0b7)

Known issues:

- The second model load may fail with `AssertionError: Torch not compiled with CUDA enabled`. I don't know why, maybe a BigDL or IPEX problem. You need to restart the server if it occurred.

## Checklist:

### DIFF
--- a/modules/loaders.py
+++ b/modules/loaders.py
@@ -3,6 +3,8 @@ from collections import OrderedDict
 
 import gradio as gr
 
+from transformers import is_torch_xpu_available
+
 from modules import shared
 
 loaders_and_params = OrderedDict({
@@ -149,6 +151,15 @@ loaders_and_params = OrderedDict({
         'no_use_fast',
     ]
 })
+
+if is_torch_xpu_available():
+    print('XPU is available, adding BigDL loader')
+
+    loaders_and_params['BigDL'] = [
+        'quant_type_bigdl',
+        'compute_dtype',
+        'trust_remote_code'
+    ]
 
 
 def transformers_samplers():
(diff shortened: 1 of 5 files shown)


oobabooga:
This is not something I would be able to maintain. Ideally, it should be a third-party maintained loader, provided as an extension, which in turn would require modifying the extensions framework to support custom loaders.

PR #5244, opened 2024-01-12
Disabling temperature while dynamic temperature is active.

HideLord:
Adding CSS for better visibility of what slider is disabled.

image

## Checklist:


oobabooga:
After https://github.com/oobabooga/text-generation-webui/pull/5443 they can be used at the same time

PR #5256, opened 2024-01-14
Update position of show-controls label to avoid line breaks due to font size

Yiximail:
## Checklist:

image

Not limited to 168px. It'll solve that problem.

image


css/main.css
```diff
@@ -490,7 +490,11 @@ div.svelte-362y77>*, div.svelte-362y77>.form>* {
 #show-controls label {
     z-index: 1000;
     position: absolute;
-    left: calc(100% - 168px);
+    right: 0;
+    line-height: 16px;
```
oobabooga, line 494: Does the fix still apply without this `line-height`?

oobabooga: Ty

PR #5268, opened 2024-01-14
Fix error "Attempting to unscale FP16 gradients" when training LoRA

optozorax:
I was trying to train [this model](https://huggingface.co/IlyaGusev/saiga_mistral_7b_merged) in webui, but got error `Attempting to unscale FP16 gradients`.

When I set flag `bf16`, I got message `Your setup doesn't support bf16/gpu. You need torch>=1.10, using Ampere GPU with cuda>=11.0`, I met all the requirements except `Ampere GPU` (I am using V100).

So, I found this patch here: https://github.com/huggingface/peft/pull/1336/files

It worked for me and I was finally able to train LoRA in webui.

But I have no idea how that will affect other code and performance.

## Checklist:


oobabooga:
It seems to make sense, thanks

PR #5277, opened 2024-01-16
Allow content arrays for multimodal OpenAI requests

Cohee1207:
This patch will allow sending prompts to chat completions endpoint formatted according to [OpenAI vision guide](https://platform.openai.com/docs/guides/vision/quick-start). If message content contains a list with text and image parts, it will be converted to the format expected by the multimodal extension (two user messages).

Without this change, the following exception will be thrown if the message contents include a list:

image

## Checklist:

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -144,24 +144,25 @@ def convert_history(history):
     user_input = ""
     system_message = ""
 
-    new_history = []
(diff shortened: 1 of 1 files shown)


oobabooga:
Thanks for the fix @Cohee1207, I wasn't aware of this.

PR #5290, opened 2024-01-17
Install requirements.yaml before starting

jacobbeasley:
Seems this is required on windows for it to work

## Checklist:


oobabooga:
This is already in [one_click.py](https://github.com/oobabooga/text-generation-webui/blob/main/one_click.py).

PR #5342, opened 2024-01-22
Noisy sampling HF implementation

kalomaze:
A custom sampler that allows you to apply Gaussian noise to the original logit scores to encourage randomization of choices where many tokens are usable (and to hopefully avoid repetition / looping).
Inspired a bit by how pure GPU inference on exllama2 seems to have a bit of non-determinism due to the CUDA kernels.

image


BadisG:
Are you planning on adding dithering aswell? That one looks really interesting
https://github.com/ggerganov/llama.cpp/issues/4976

oobabooga:
I don't like the idea of adding noise to the logits. That's akin to using a model with less parameters. That could be turned into an extension.

kalomaze:
> don't like the idea of adding noise to the logits. That's akin to using a model with less parameters.

Fair enough. The intention was to add a slight bit of RNG to rerolls, but, it's kind of out there no matter how you slice it

Issue #2243, opened 2023-05-21
Display the number of tokens in the input / in the conversation so far


michael-imbeault:
**Description**

Display the number of tokens in the input / in the conversation so far. Since the context window is limited, it is important to know when you exceed it / how many tokens you have left for full context to be considered. This applies to the original input, but also to the full conversation.

Ph0rk0z:
It outputs it to the console.

michael-imbeault:
Yes, so in theory very much possible to have it displayed in the GUI?

Ph0rk0z:
There is a count tokens button in notebook mode. If you are using chat then it won't work.

michael-imbeault:
It is such an important parameter that it should be available on the GUI at all times in my opinion, and updated dynamically if possible.

Ph0rk0z:
I don't disagree and I like how silly tavern counts tokens for character cards and does generation time under the message too. Just nobody has made PRs for that stuff here yet.

kha84:
Indeed, I can see original and new tokens are being calculated (https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L407) - so the question is how can we get them fetched to UI.

In https://github.com/oobabooga/text-generation-webui/blob/main/modules/ui_default.py module, which implements UI for "Default" tab, there's a code that displays count of tokens (see "token-counter" placeholder which gets populated from count_tokens(x) function (https://github.com/oobabooga/text-generation-webui/blob/main/modules/prompts.py#L22)

One needs to copy that "original_tokens" variable (https://github.com/oobabooga/text-generation-webui/blob/main/modules/text_generation.py#L430) to some other global variable so it will be visible from outside the module first. 
Then to introduce the same placeholder to display the count in HTML and function to update it for "Chat" UI (https://github.com/oobabooga/text-generation-webui/blob/main/modules/ui_chat.py)

I guess, that should do the job

kha84:
Raised pull request - https://github.com/oobabooga/text-generation-webui/pull/5357/files
It's kind of hacky way to do that and people are kindly welcomed to fix whatever issues they see with this, but for me it works.
![image](https://github.com/oobabooga/text-generation-webui/assets/110789576/29407218-3f96-4b2c-b9f8-637c71094970)

PR #5357, opened 2024-01-23
Show the total number of tokens and generation speed in chat UI (#2243)

kha84:
## Checklist:

## What's been done:

Two small changes:
1) one to  [modules/text_generation.py](https://github.com/oobabooga/text-generation-webui/pull/5357/files#diff-36dfef7d9db97b2cde9351ae9eafa7aed3dd15e094d2ca7a4a60bfa88d6096f2) to expose the number of tokens that was identified as "context" and "generated output" as per last call of generate_reply_HF / generate_reply_custom
2) another is to [modules/ui_chat.py](https://github.com/oobabooga/text-generation-webui/pull/5357/files#diff-ba9fb0480e9653459faabb353f96234192cb4d535a7143da3e4aaca5000eb51f) to show that number of tokens similar way how it's being shown for "Default" tab


kha84:
image

kha84:
Added token per second to be displayed, still WIP
Screenshot from 2024-01-29 00-01-49

kha84:
Need to polish Gradio part - don't like the place, where it is being displayed currently.

YakuzaSuske:
Why tho? It's already being displayed on the cmd window. Adding it to the UI would be unnecessary unless you using it through the API in which case it would helpful. But other than that, i don't see why people would want to know the speed and tokens if they are just interested in the output text itself.
image

However if you are determine to add it, i think the best placement would be somewhere here. The current placement is in the way of the user's vision and it might impede on the text they want to read.

kha84:
Yeah, I was thinking about that placement as well, thanks!

Answering your "why" question:

1) having this crucial info hidden somewhere in logs doesn't seem to be very handy. In order to see these figures you need to jump back and forth between web ui and cli.

2) the amount of context used is quite important runtime parameter, even for chat-only casual users. For now you have to blind-guess if your chats are already out of the context window or not yet. Or to jump to CLI to see logs. Having it exposed to be displayed all the time gives you much more details about whether or not you should be still expecting coherent answers from your model.

YakuzaSuske:
I think for me in my case it just seems kind of odd but.... tbh i always have the command window open and to the side. So i guess that's why i asked the question. I thought people always did that too but 🤷‍♂️  I guess i'm the only one who does that, so that's why i found it particularly odd since i can always at all times see the log, tokens a second, context, etc. by just glancing to my left real quick.

As seen below:
image

biship:
Why not put it under or next to the {{char}} icon like SillyTavern does?

number of tokens and duration.

I think this is a good addition to the UI as I don't have the command window showing.

bartowski1182:
@YakuzaSuske not everyone interacts on the same computer it's running on, I also use it on my phone, sometimes I'm curious and I'd rather not open terminal, ssh, find the program, watch the logs, and then swap between programs

I think this is a great inclusion and would love to see it implemented

FartyPants:
> @YakuzaSuske not everyone interacts on the same computer it's running on, I also use it on my phone, sometimes I'm curious and I'd rather not open terminal, ssh, find the program, watch the logs, and then swap between programs
> 
> I think this is a great inclusion and would love to see it implemented

This is the correct answer for "why". There are many instances (remote access, runpod) where you don't always get to see terminal.

oobabooga:
I find it cumbersome to see this kind of logs in the UI, while in the terminal the logs appear all on top of each other nicely.

kha84:
Sure, you're the boss. Myself I switched to ollama + Open Web UI = it has this info printed in UI out-of-the-box

PR #5368, opened 2024-01-24
Add roleplay.gbnf

BadisG:
Hello,

This gbnf is centred on the typical roleplaying grammar structure:

yI8bsQ5

- It will use the (action + speech) combo 2 or 3 times.
- It has numbers and punctuations
- It has special characters like "à" or "é" (good for other languages than english)

asasasa

Any feedback is welcome, we can discuss how to make this work best. For example, I wondered if (action + speech) would be more interesting to think about than (speech + action).

Edit: I also enforced UTF-8 encoding when loading the gbnf, because that's what happens if we don't:

image


oobabooga:
Does this work?

`[a-zA-Zà-üÀ-Ü]`

If not, I may need to pull the latest code from https://github.com/epfl-dlab/transformers-CFG.

BadisG:
Indeed this works fine like that, thanks for the catch, I've simplified the grammar.

oobabooga:
Nice, thanks. That's a creative and interesting useful use of grammar. Looks good to merge.

YakuzaSuske:
> `[a-zA-Zà-üÀ-Ü]`

That replacement breaks the code and causes these errors.
image

the previous code which was this:

```
phrasePattern ::= "*" [A-ZéÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9] ( [A-Za-z'.?!,:;-éÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9]+ " " )* [A-Za-z'.?!,:;-éÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9]+ "*"
quotedText ::= "\"" [A-Z] [A-Za-z'.*?!,:;-éÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9]* (" " [A-Za-z'.*?!,:;-éÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9]+)* "\""
root ::= phrasePattern singleSpace quotedText singleSpace phrasePattern singleSpace quotedText singleSpace (phrasePattern singleSpace quotedText singleSpace)?
singleSpace ::= " "
```

That code functioned normally, however the introduction of `[a-zA-Zà-üÀ-Ü]` breaks the grammar rules.

BadisG:
@YakuzaSuske Oh yeah you are right, the [a-zA-Zà-üÀ-Ü] rules is only working on llamacpp_hf (not llamacpp) but the [A-ZéÉàÀèÈùÙâÂêÊîÎôÔûÛçÇöÖüÜäÄëË0-9] rules is working everytime.
@oobabooga maybe you should actually pull the latest code from https://github.com/epfl-dlab/transformers-CFG to make it work

oobabooga:
The transformers-CFG code applies only to llamacpp_HF and HF loaders, and not to llama.cpp. So it's a problem with llama.cpp grammar parsing.

PR #5381, opened 2024-01-27
Swap to huggingface_hub get_token function

Anthonyg5005:
Uses huggingface_hub.get_token() to fetch the token. Can use either HF_TOKEN variable or huggingface-cli login; Variable is higher priority. Compatible with oneclick huggingface_hub version

## Checklist:


oobabooga:
Interesting, thanks

Issue #5374, opened 2024-01-25
Prompt not being passed to backend for certain models


bartowski1182:
### Describe the bug

Using ExllamaV2_HF or ExllamaV2 loader with https://huggingface.co/bartowski/internlm2-chat-7b-sft-llama-exl2 internLM2 model with chat-instruct, whenever I send a message I can see in the logs that the actual prompt is not being sent through:

```
16:48:56-116728 INFO     PROMPT=
<|im_start|>user
Continue the chat dialogue below. Write a single reply for the character "AI".

<|im_end|>
<|im_start|>assistant
AI:
```

The strange part is that it's ONLY some models, and I've tried manually setting as much of the config as possible to be identical to another model, and nothing seems to work. Using any other internLM2 model including the math one from yesterday the prompt is sent with no problem. Any idea what could be going on?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Download 6.5 bpw from here https://huggingface.co/bartowski/internlm2-chat-7b-sft-llama-exl2
Load the model with ExLlamav2_HF
Send a message in chat-instruct
View that logs don't contain the actual input from the chat

### Logs

```shell
16:48:56-116728 INFO     PROMPT=
<|im_start|>user
Continue the chat dialogue below. Write a single reply for the character "AI".

<|im_end|>
<|im_start|>assistant
AI:
```

### System Info

```shell
Ubuntu 22.04
Nvidia 3090
Docker environment
```

PR #5066, opened 2023-12-24
Raise exception when prompt exceed truncation

yhyu13:
## Summery

When chat prompt exceed allowed length, raise excpetion instead of removing any of the prompt messages.

## Checklist:

- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

PR #5382, opened 2024-01-27
Fix max_length in chat.py to be just truncation length

bartowski1182:
## Checklist:

get_max_prompt_length is returning truncate_prompt_length - max_new_tokens, which I'm not sure where the value is for that

If you have a model that has a max_new_tokens of 4096, and the truncate length is also 4096, then that makes the max prompt length returned be 0

This causes all messages to be removed from the prompt, resulting in a blank prompt sent to the model

This fixes issue https://github.com/oobabooga/text-generation-webui/issues/5374

### DIFF
--- a/modules/chat.py
+++ b/modules/chat.py
@@ -169,7 +169,7 @@ def generate_chat_prompt(user_input, state, **kwargs):
     prompt = make_prompt(messages)
 
     # Handle truncation
-    max_length = get_max_prompt_length(state)
+    max_length = state['truncation_length']
     while len(messages) > 0 and get_encoded_length(prompt) > max_length:
         # Try to save the system message
         if len(messages) > 1 and messages[0]['role'] == 'system':


bartowski1182:
If wanting to maintain the idea that, if you're trying to exceed your capacity with new tokens you should truncate the prompt, I propose:

max_length = min(truncate_prompt_length, max_seq_len - min_length)

which would by default limit the prompt to the truncate length as appropriate, and in a worst case limit the prompt to whatever the model can support as configured minus the minimum amount of new tokens you want to generate

Ideally though, it seems it would be most appropriate to introduce new logic that makes it so that  truncate_prompt_length + max_new_tokens is always = max_seq_len

So that way, if you have a max sequence length of 4000, truncate_prompt_length would be set to something below that like 3000 (3/4 seems like a decent starting place?) and max_new_tokens would be set to the remainder, 1000 (or 1/4)

Then at least in the UI it's obvious what's going on

oobabooga:
`truncate_prompt_length` is automatically obtained from the model metadata and is set to the model's sequence length -- 4096 for llama-2, 32768 for mixtral, etc.

If the user wants to generate `max_new_tokens` tokens, then the prompt length can't exceed `truncate_prompt_length - max_new_tokens`, so that there is room for the new tokens.

It is true that chat completions inputs are being completely removed in case they don't fit into the available space, which is an issue. The input should be left truncated instead. https://github.com/oobabooga/text-generation-webui/pull/5066 aims to solve this problem but in a way that I find too complicated. There is a probably a simple (few lines) way to modify this loop https://github.com/oobabooga/text-generation-webui/blob/e7a760e6b385ad80d8a4b6a56b632d356a0dc5d2/modules/chat.py#L173 to obtain this result.

bartowski1182:
If you want to maintain this functionality then by all means, but from my point of view it's misleading to have a value that explicitly specifies the max prompt length, but then attempts to adjust it dynamically based on theoretical newly generated tokens, especially since it can lead to situations like mine where nothing gets sent to the backend even though there's tons of available context

I understand how it works, and now it makes sense from a logical perspective, but from a UX perspective it's terrible. Even in discord I saw people dismissing the new Deepseek model entirely because in the chat interface all its responses were garbage, because most people have max new tokens at 4000 so it was removing their prompt completely

oobabooga:
The HF text generation pipeline will still fail if the prompt length + max_new_tokens is greater than the model's context length. This must be ensured.

I am closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/5439. If it doesn't work as expected and you still experience issues please let me know.

bartowski1182:
Seems reasonable. The only other chance I would consider making is not defaulting the prompt truncation to be the full model context, doesn't seem logical to ever have it be at that level since you'll never want your whole context to be prompt anyways, you'd want room to generate. For my liking, i'd default the prompt truncation to be some reasonable fraction of the full context (4/5? 3/4?). If that's agreeable to you, I can make a PR

Issue #5232, opened 2024-01-10
Python exceptions after an API request from SillyTavern (Frostwind EXL2)


notsyncing12309:
### Describe the bug

When I try to make a request through SillyTavern to the API, Python exceptions are thrown and no responses are received.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Load any model
2. Connect to API from SillyTavern (Power plug icon -> Text Completion -> Default (Oobabooga) -> Connect button)
3. Make request in any chat

### Logs

```shell
04:41:31-276722 INFO     Starting Text generation web UI
04:41:31-281234 INFO     Loading the extension "openai"
04:41:31-354038 INFO     OpenAI-compatible API URL:

                         http://127.0.0.1:5000

04:41:31-354449 INFO     Loading the extension "gallery"
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
04:42:05-079389 INFO     Loading frostwind
04:42:31-461867 INFO     LOADER: ExLlamav2
04:42:31-462867 INFO     TRUNCATION LENGTH: 3072
04:42:31-463869 INFO     INSTRUCTION TEMPLATE: Alpaca
04:42:31-463869 INFO     Loaded the model in 26.38 seconds.
Exception in ASGI application
Traceback (most recent call last):
  File "C:\Portable\text-generation-webui-main\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 247, in __call__
    await wrap(partial(self.listen_for_disconnect, receive))
  File "C:\Portable\text-generation-webui-main\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 236, in wrap
    await func()
  File "C:\Portable\text-generation-webui-main\installer_files\env\Lib\site-packages\sse_starlette\sse.py", line 191, in listen_for_disconnect
    message = await receive()
              ^^^^^^^^^^^^^^^
  File "C:\Portable\text-generation-webui-main\installer_files\env\Lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 538, in receive
    await self.message_event.wait()
  File "C:\Portable\text-generation-webui-main\installer_files\env\Lib\asyncio\locks.py", line 213, in wait
    await fut
asyncio.exceptions.CancelledError: Cancelled by cancel scope 226c48c0c90

During handling of the above exception, another exception occurred:

  + Exception Group Traceback (most recent call last):
[...]

PR #5391, opened 2024-01-28
Remove tiktoken from logit bias to fix API logit bias

Ph0rk0z:
## Checklist:

TikToken is not used in other parts of the file, they are commented out. The model parameter is not required for internal tokenization. The models tiktoken expects are only served by openAI. If there is another reason to keep this please please comment. Since the previous patch wasn't clean by skipping via exception, perhaps this is better. Otherwise we cannot use logit bias over the API at all and it is a broken feature.

fixes: https://github.com/oobabooga/text-generation-webui/issues/5232


oobabooga:
I think that removing tiktoken is a good thing for consistency. Have you made tests over API to validate that the logit bias is working as intended? If so the PR can be merged.

Ph0rk0z:
Yes.. I tested it first with silly tavern. Definitely can remove a token with negative and make the token the only reply with positive.

oobabooga:
Looks good then

PR #5397, opened 2024-01-30
Bump llama-cpp-python to 0.2.36

oobabooga:
I had to remove older Python/CUDA versions due to GitHub rate limiting my Github Actions jobs.


Ph0rk0z:
It needs split by rows or layers setting as doing it by layers is slower, even with 3090s.

oobabooga:
Could you make a PR? I don't have 2 GPUs to properly test this.

Ph0rk0z:
Sure, I can PR it later on. Can just make a checkbox to uncheck.

PR #5411, opened 2024-01-31
Improve ChatML template

BadisG:
Hello,

The ChatML template must look like this:
```
<|im_start|>system
{system_message}<|im_end|>
<|im_start|>user
{prompt}<|im_end|>
<|im_start|>assistant
```

But the current ChatML template gives this result

```
<|im_start|>system
{system_message}
<|im_end|>
<|im_start|>user
{prompt}
<|im_end|>
<|im_start|>assistant
```

This can lead to really incoherent outputs when talking to the bot.
I've also tweaked the code a bit so that when there's no system prompt, this disappears:
```
<|im_start|>system
<|im_end|>
```


oobabooga:
That seems correct, thanks

Issue #5375, opened 2024-01-25
InternLM2 math model breaks with exllamav2_HF loader (works with non-HF)


bartowski1182:
### Describe the bug

Trying out https://huggingface.co/bartowski/internlm2-math-20b-llama-exl2 and whenever I use HF loader to generate i get a bunch of errors. When using the non HF loader I get no problems generating.

I also tried updating to dev branch where exllamav2 has been bumped, still doesn't work with HF loader

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Download 4.25 quant from https://huggingface.co/bartowski/internlm2-math-20b-llama-exl2
Load with ExllamaV2 HF loader
Attempt to generate any text

### Logs

```shell
Traceback (most recent call last):
  File "/text-generation-webui/modules/callbacks.py", line 61, in gentask
    ret = self.mfunc(callback=_callback, *args, **self.kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/text-generation-webui/modules/text_generation.py", line 379, in generate_with_callback
    shared.model.generate(**kwargs)
  File "/opt/conda/envs/textgen/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/textgen/lib/python3.11/site-packages/transformers/generation/utils.py", line 1525, in generate
    return self.sample(
           ^^^^^^^^^^^^
  File "/opt/conda/envs/textgen/lib/python3.11/site-packages/transformers/generation/utils.py", line 2622, in sample
    outputs = self(
              ^^^^^
  File "/text-generation-webui/modules/exllamav2_hf.py", line 125, in __call__
    self.ex_model.forward(seq_tensor[:-1].view(1, -1), ex_cache, preprocess_only=True, loras=self.loras)
  File "/opt/conda/envs/textgen/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/textgen/lib/python3.11/site-packages/exllamav2/model.py", line 559, in forward
    r, ls = self._forward(input_ids = input_ids[:, chunk_begin : chunk_end],
[...]

PR #5431, opened 2024-02-04
Remove non-HF ExLlamaV2 loader

oobabooga:
Since PR https://github.com/oobabooga/text-generation-webui/pull/4814, the speed difference between `ExLlamav2` and `ExLlamav2_HF` is zero. So I see no point in keeping the non-HF version, which is redundant and which samples in a way not guaranteed to be consistent with HF transformers sampling.


sgsdxzy:
Won't this cause problems for #5375 ?

Ph0rk0z:
In this case we can't use native sampling of exllamav2 though.

aikitoria:
It's not true that there is zero speed difference. Non-HF loader is around 10% faster for goliath-120b.

aikitoria:
Quick bench using ooba from before this commit and exllamav2 master branch from 5 minutes ago on runpod A100 80GB.

HF:
```
Output generated in 8.76 seconds (14.50 tokens/s, 127 tokens, context 1728, seed 735928511)
Output generated in 9.22 seconds (13.77 tokens/s, 127 tokens, context 1728, seed 83286885)
Output generated in 8.99 seconds (14.13 tokens/s, 127 tokens, context 1728, seed 128023280)
Output generated in 8.78 seconds (14.47 tokens/s, 127 tokens, context 1728, seed 1418661767)
```
Non-HF:
```
Output generated in 8.17 seconds (15.67 tokens/s, 128 tokens, context 1728, seed 745431605)
Output generated in 8.18 seconds (15.65 tokens/s, 128 tokens, context 1728, seed 762707583)
Output generated in 8.18 seconds (15.64 tokens/s, 128 tokens, context 1728, seed 996129951)
Output generated in 8.18 seconds (15.64 tokens/s, 128 tokens, context 1728, seed 700382800)
```

aikitoria:
> not guaranteed to be consistent with HF transformers sampling

Why is this important, if the builtin sampling in exllamav2 works fine?

Ph0rk0z:
For some stuff I like HF samplers and for some stuff the native ones. I forgot about the extra t/s. It happens in llama.cpp too, a tiny difference due to overhead from HF. It also helps to troubleshoot issues with HF vs the original loader. There are like a million reasons to keep it.

aikitoria:
Thanks for restoring it!

Issue #5129, opened 2023-12-31
Allow Path Recursive Search In Models Folder


JiHa-Kim:
**Description**
Hi, I have been using LM Studio, someone recommended this UI. However, text-generation-webui requires the models to be directly under the root folder. Instead in LM Studio, there are subfolders that contain the model files. It would be useful to implement recursive path search in the code in order to make this compatible.

**Additional Context**

### Discussed in https://github.com/oobabooga/text-generation-webui/discussions/1626

<div type='discussions-op-text'>

<sup>Originally posted by **electric-samurai** April 28, 2023</sup>
In my case I'd like to not have to place all my models in special locations in the model-dir.  I already have one directory with subdirectories specifying the sources of the models, sometimes a couple layers deep.  I'm working mostly with llama models - here is a hack I am using in the source to find all the models of the model-dir root and use the full path in the menu item.  You need to additional alter the source after doing this to remove explicit references to the model-dir.

```
def get_available_models():
    if shared.args.flexgen:
        return sorted([re.sub('-np$', '', item.name) for item in list(Path(f'{shared.args.model_dir}/').glob('*')) if item.name.endswith('-np')], key=str.lower)
    else:
        # return sorted([re.sub('.pth$', '', item.name) for item in list(Path(f'{shared.args.model_dir}/').rglob('*')) if not item.name.endswith(('.txt', '-np', '.pt', '.json', '.yaml'))], key=str.lower)
        return sorted([str(file_path.parent.absolute()) for file_path in Path(f'{shared.args.model_dir}/').rglob('*.bin') if not (".git" in str(file_path))], key = str.lower)
```</div>

ke1ne:
Yes, please!

JiHa-Kim:
As a temporary workaround someone has suggested to create symbolic links between the model file and text-generation-webui's model folder. I have created a script to automate this for now and will upload it to GitHub.

rhdunn:
It would be nice to have support for this. It would help to organize various LLMs and versions, e.g. having a groups for llama-2, llama-3, and mistral models. They would be listed using the relative path to the file, e.g. `llama-2/some-model-v1.0-Q6_K.gguf`.

Note: This is how ComfyUI works when loading checkpoints, etc.

PR #5444, opened 2024-02-05
Added recursive path search within models folder

JiHa-Kim:
I have added the ability to search for subfolders inside the models folder.
 
 Addresses #5129, #1626
## Checklist:


oobabooga:
Thanks for the PR, but I don't see the appeal in more complex folder structures, especially when such big changes are necessary to support it.

Issue #5422, opened 2024-02-02
CLI_ARGS from .env ignored since Dockerfile split commit


acidbubbles:
### Describe the bug

CLI_ARGS is not used anymore (look up in the source code, it only exists in .env but isn't used anywhere)

See https://github.com/oobabooga/text-generation-webui/commit/b1463df0a155efb66f8901b94c2ae16cf0af5a9c#r138191515 for the diff.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Write anything in CLI_ARGS in your .env and use docker compose up. They will be ignored.

### Logs

```shell
N/A
```

### System Info

```shell
nvidia/Dockerfile (unrelated to the OS)
```

acidbubbles:
Note that a workaround is to use CMD_FLAGS.txt instead.

PR #5474, opened 2024-02-09
Fixed, improved, and unified Docker environment

WolframRavenwolf:
## Checklist:

## Fixed and improved Docker environment

- Unified Dockerfile:

  - Moved GPU_CHOICE from RUN to ARG instruction so all the Dockerfiles could be unified into a single Dockerfile
  - Added CLI_ARGS to start_linux.sh command (very important fix, without it CLI_ARGS are ignored)
  - Set HOME as global instead of process-specific envvar (so it's also set when entering the container manually)
  - Removed redundant WORKDIR (we've already changed into it)
  - Deleted old Dockerfiles (now obsolete because of the unified Dockerfile)

- Composefiles:

  - Fixed APP_UID variable default (added missing colon)
  - Added sensible container_name ("text-generation-webui" instead of default "text-generation-webui-text-generation-webui-1")
  - Added sensible hostname ("text-generation-webui" instead of random ID)
  - Added restart policy "unless-stopped" (so container can be kept running or stopped properly when starting detached)

- Envfile-Example:

  - Added PYTORCH_KERNEL_CACHE_PATH pointing at cache dir to fix "UserWarning: Specified kernel cache directory could not be created! This disables kernel caching."
  - Added APP_RUNTIME_UID to APP_RUNTIME_GID (so you can not only define the group, but also the user for access to mounted volumes)

- Docs:

  - Create symlinks (fixed) and missing dirs (so they won't be created by Docker as root)

- .gitignore

  - Added .dockerignore


jtara1:
Tested and it works for me. I can now define and use `CLI_ARGS="--listen-host 0.0.0.0 --listen"` as an env var set in my `.env`

oobabooga:
Could you merge the dev branch?

docker/Dockerfile
```diff
@@ -1,21 +1,22 @@
...
 ARG APP_GID="${APP_GID:-6972}"
 
+WORKDIR /builder
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,rw \
     apt update && \
     apt install --no-install-recommends -y git vim build-essential python3-dev pip bash curl && \
     rm -rf /var/lib/apt/lists/*
 WORKDIR /home/app/
-RUN git clone https://github.com/oobabooga/text-generation-webui.git 
+RUN git clone https://github.com/oobabooga/text-generation-webui.git
```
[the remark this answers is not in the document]
polarathene: You've got a `WORKDIR /home/app` followed by the clone and then `WORKDIR /home/app/text-generation-webui` to change directories.

```suggestion
WORKDIR /home/app/text-generation-webui
RUN git clone https://github.com/oobabooga/text-generation-webui.git .
```

Remove the other two `WORKDIR` surrounding that suggestion, they're redundant.

docker/Dockerfile
```diff
@@ -1,21 +1,22 @@
...
 WORKDIR /home/app/text-generation-webui
-RUN GPU_CHOICE=B USE_CUDA118=FALSE LAUNCH_AFTER_INSTALL=FALSE INSTALL_EXTENSIONS=TRUE ./start_linux.sh --verbose
+RUN USE_CUDA118=FALSE LAUNCH_AFTER_INSTALL=FALSE INSTALL_EXTENSIONS=TRUE ./start_linux.sh --verbose
 COPY CMD_FLAGS.txt /home/app/text-generation-webui/
 EXPOSE ${CONTAINER_PORT:-7860} ${CONTAINER_API_PORT:-5000} ${CONTAINER_API_STREAM_PORT:-5005}
-WORKDIR /home/app/text-generation-webui
+ENV HOME=/home/app/text-generation-webui
 # set umask to ensure group read / write at runtime
-CMD umask 0002 && export HOME=/home/app/text-generation-webui && ./start_linux.sh
+CMD umask 0002 && ./start_linux.sh $CLI_ARGS
```
polarathene, line 22: - `umask 0002` seems unnecessary? Why is group write permission needed at runtime? [it was introduced 9 months ago here, but no insights as to why](https://github.com/oobabooga/text-generation-webui/pull/4768/files#diff-52c6e96914b8b24ca636ecf09a028a313c8790830d80d40969fcaf911cacb140R44-R56).
- [`CLI_ARGS` was removed in March](https://github.com/oobabooga/text-generation-webui/issues/5422). Regardless, this is not the correct approach with Docker.

```suggestion
ENTRYPOINT [ "/home/app/text-generation-webui/start_linux.sh" ]
```

Now `CLI_ARGS` would just be provided as the command (`CMD`):

```
docker run --rm -it image-name $CLI_ARGS
```

In `compose.yaml` the `command` field can set the CLI args to use, or you can use the Docker Compose ENV interpolation from a `.env` file to also use `${CLI_ARGS}` as a value (kinda redundant though).

docker/amd/docker-compose.yml
```diff
@@ -4,42 +4,46 @@ services:
...
-        APP_GID: ${APP_GID:-6972} 
-        APP_UID: ${APP_UID-6972} 
+
+        # specify which cuda version your card supports: https://developer.nvidia.com/cuda-gpus
+        TORCH_CUDA_ARCH_LIST: ${TORCH_CUDA_ARCH_LIST:-7.5}
+        APP_GID: ${APP_GID:-6972}
+        APP_UID: ${APP_UID:-6972}
+    container_name: text-generation-webui
     env_file: .env
+    hostname: text-generation-webui
```
polarathene, line 32: Why?

docker/amd/docker-compose.yml
```diff
@@ -4,42 +4,46 @@ services:
...
+
+        # Extension requirements to build:
         BUILD_EXTENSIONS: ${BUILD_EXTENSIONS:-}
-        APP_GID: ${APP_GID:-6972} 
-        APP_UID: ${APP_UID-6972} 
+
+        # specify which cuda version your card supports: https://developer.nvidia.com/cuda-gpus
+        TORCH_CUDA_ARCH_LIST: ${TORCH_CUDA_ARCH_LIST:-7.5}
+        APP_GID: ${APP_GID:-6972}
+        APP_UID: ${APP_UID:-6972}
```
polarathene, line 29: Are these actually used for anything? Nothing appears to reference them: https://github.com/search?q=repo%3Aoobabooga%2Ftext-generation-webui%20APP_UID&type=code

If they're only meant to change the runtime user, this should just be done with the standard `user: uid:gid` compose setting or `docker run --user uid:gid` CLI arg.

polarathene (changes requested):
Some feedback to consider.

While I've not looked at the the files in full there's likely a bunch of duplication (_at least from the repeated diffs still present_) that could be unified to single files.

oobabooga:
Closing for inactivity.

PR #5503, opened 2024-02-14
Remove message.content from openai streaming API

deoxykev:
The [OpenAI spec](https://platform.openai.com/docs/api-reference/streaming) does not expect a `message` object when streaming is enabled. Returning both sometimes [causes issues in downstream consumers](https://github.com/danny-avila/LibreChat/discussions/1795).

## Checklist:


oobabooga:
Thanks, I have tested it and see no downsides.

PR #5519, opened 2024-02-16
Add a "llamacpp_HF creator" menu

oobabooga:
This automates the tedious task of downloading the tokenizer files a GGUF model so that it can be used with llamacpp_HF. You just need the URL for the original model.

output


biship:
Works beautifully, thanks! I always did this manually.

PR #5522, opened 2024-02-16
Gradio 4

oobabooga:
Performance is a lot better than on the current Gradio 3.50.2, but I get these error popups.

304963284-3195754d-174b-4573-9206-fba31b25c1f0


oobabooga:
The error is still present, but I believe that it should be solved soon, so I have merged the update to the dev branch.

Relevant issue: https://github.com/gradio-app/gradio/issues/7576

PR #5535, opened 2024-02-17
Remove -k from curl command to download miniconda

oobabooga:
Tests are needed to know if it still works like this, especially on Windows.


d-z-m:
Tested start_windows.bat on Windows_Server-2022-English-Full-Base-2024.01.16 running in AWS. works 👍.

Haven't tested with WSL yet.

oobabooga:
Let's try this. If anything goes wrong I'll add back -k to Windows only.

PR #5539, opened 2024-02-19
Add penalty_threshold parameter

oobabooga:
I copied the idea from the following PR to llama.cpp by @p-e-w https://github.com/ggerganov/llama.cpp/pull/5561.

This prevents repetition penalty from being applied to tokens that are common in the input like ~spaces~, commas, and newline characters. Let's test the usefulness of this parameter.


p-e-w:
Nice!

oobabooga:
I'm closing this in favor of https://github.com/oobabooga/text-generation-webui/pull/5677

PR #5551, opened 2024-02-20
Cubic sampling w/ curve param

kalomaze:
This adds upon the original Quadratic Sampling method with an additional parameter that I've labeled "smoothing_curve".

image

The idea is to enable even lower smoothing_factor values than ~0.25ish to work well; we do this by applying a cubic transformation to compensate, which seems to make the falloff steeper.

Not ready to merge yet, needs empirical testing from users. My hope is that you can fully avoid having to use truncation schemes and go for a fully "smooth" transformation to the distribution across different models.

The higher the smoothing_curve, the steeper the fall off (so it becomes harsher).

https://github.com/oobabooga/text-generation-webui/assets/66376113/f8900b78-ca8f-44f9-8188-7244a97eb8ef

Brief tests on a 7b are showing that it does in fact help to make lower smoothing_factor values coherent in practice (so far at least).

1.0 smoothing_curve is the "old" behavior and has no effect.


BadisG:
When you go for 3 or higher on the smoothing_curve, it doesn't work anymore
nIDEWXx
I think it's due to the formula that mixes the quadratic + cubic formula + the smoothing_curve factor

You could make it more simple by adding simply a slider that lets you choose which one you want (quadratic or cubic), this had two advantages:
1. You don't need to make one giant formula that has to make everything work at once
2. You'll be able to get a smoothing_curve even for the quadratic formula (this isn't the case there it's only applying to the cubic formula)

kalomaze:
> When you go for 3 or higher on the smoothing_curve, it doesn't work anymore nIDEWXx

I can't reproduce this issue.

Myobu1:
> > When you go for 3 or higher on the smoothing_curve, it doesn't work anymore nIDEWXx
> 
> I can't reproduce this issue, through llama.cpp_HF at least.

I have the exact same issue on Ooba. Setting smoothing curve to 3 and lower works fine. The moment I go above it, I start getting errors regarding "RuntimeError: probability tensor contains either `inf`, `nan` or element < 0" and inability to output replies after 5 attempts on Silly.

Ph0rk0z:
I get nan error as well. using exl2_HF. utils.py line 2734

Its a problem from transformers.

`next_tokens = torch.multinomial(probs, num_samples=1).squeeze(1)`

Ph0rk0z:
I only got 2 braincells to knock together but to me

`k = (3 - self.smoothing_curve) / 2`

3-3 is 0 and then thats 0/2 which makes NaN

oobabooga:
The nan error was caused by making operations with `-inf^3`. I solved it by keeping scores with `-inf` value unchanged.

Ph0rk0z:
Will retest.

PR #5565, opened 2024-02-23
Big picture fixes

TheLounger:
- Hides big character picture if it fires an error event (ie. 404)
- Makes character pfp always clickable. This is mainly for the *TheEncrypted777* chat style, where part of the img element ends up behind the text element.

## Checklist:


oobabooga:
Very nice, thanks!

PR #5570, opened 2024-02-23
ngrok logging does not use the shared logger module

dsroark:
- when running `start_linux.sh`, ngrok will not display the created URL unless the logs are run through the shared logger module

## Checklist:


oobabooga:
Thanks for the fix

PR #5649, opened 2024-03-07
Change cache_4bit to cache_q4, improve descriptions

bartowski1182:
The new Q4 option is actually a better quantized version instead of just chopping off bits like fp8, which makes it more accurate than fp8 while saving even more VRAM, and I think that should be properly reflected

## Checklist:


README.md
```diff
@@ -279,7 +279,7 @@ List of command-line flags
 |`--cfg-cache`                         | ExLlamav2_HF: Create an additional cache for CFG negative prompts. Necessary to use CFG with that loader. |
 |`--no_flash_attn`                     | Force flash-attention to not be used. |
 |`--cache_8bit`                        | Use 8-bit cache to save VRAM. |
-|`--cache_4bit`                        | Use 4-bit cache to save VRAM. |
+|`--cache_q4`                          | Use Q4 cache to save a lot of VRAM. Recommended over 8-bit cache, uses grouped quantization for better performance while saving even more VRAM. |
```
oobabooga, line 282: Is it "recommended?" I haven't seen any perplexity tests, and presumably 4-bit cache introduces a bigger loss.

docs/04 - Model Tab.md
```diff
@@ -46,6 +46,7 @@ Examples:
 * **cfg-cache**: Creates a second cache to hold the CFG negative prompts. You need to set this if and only if you intend to use CFG in the "Parameters" > "Generation" tab. Checking this parameter doubles the cache VRAM usage.
 * **no_flash_attn**: Disables flash attention. Otherwise, it is automatically used as long as the library is installed.
 * **cache_8bit**: Create a 8-bit precision cache instead of a 16-bit one. This saves VRAM but increases perplexity (I don't know by how much).
+* **cache_q4**:Creates a Q4 cache using grouped quantization instead of the standard 16-bit one. Recommended over 8-bit cache, as it's much smarter with the size trimming and results in better performance at lower usage. |
```
oobabooga, line 49: Same as the comment above, I don't have data to base this claim. Also missing a space after `**cache_q4**:`

modules/exllamav2.py
```diff
@@ -61,8 +61,8 @@ def from_pretrained(self, path_to_model):
 
         if shared.args.cache_8bit:
             cache = ExLlamaV2Cache_8bit(model, lazy=shared.args.autosplit)
-        elif shared.args.cache_4bit:
-            cache = ExLlamaV2Cache_Q4(model, lazy=shared.args.autosplit)
+        elif share.args.cache_q4:
+            cache = ExLlamaV2Cache_Q4(self.ex_model)
```
oobabooga, line 65: There was an issue that got solved with `lazy=shared.args.autosplit` when I added this flag, so this shouldn't be changed.

modules/exllamav2_hf.py
```diff
@@ -63,6 +63,8 @@ def __init__(self, config: ExLlamaV2Config):
         if shared.args.cfg_cache:
             if shared.args.cache_8bit:
                 self.ex_cache_negative = ExLlamaV2Cache_8bit(self.ex_model)
+            elif share.args.cache_q4:
+                self.ex_cache_negative = ExLlamaV2Cache_Q4(self.ex_model)
```
oobabooga, line 67: Nice catch, I haven't tested CFG with exllamav2 in a while.

oobabooga (changes requested):

modules/exllamav2.py
bartowski1182: oops this is accidental change, will revert

README.md
bartowski1182: By turboderp yes, it's a much better way of compressing the kv cache, uses grouped quantization instead of just truncating the numbers like fp8

Ph0rk0z: I perplexity tested some models with and without when testing 120b 3-bit vs 70b. Perplexity for 8 and 4bit cache is basically rounding error difference. Deep in the decimal points. Don't forget that 4-bit may be slower than 8-bit.

oobabooga: I also found in previous tests that the 8-bit cache didn't change perplexity at all. I find an indication that the perplexity test implemented in the repository is not capable of capturing the loss introduced by quantizing the cache, for some reason beyond my understanding.

oobabooga:
I'll keep the name as-is as "Q4" is an ill-defined term, and `cache_4bit` may be reused in the future for other backends.

PR #5668, opened 2024-03-08
Allow the character gallery to show files in folders.

1GladCat:
Small tweaks to include characters from folders in the gallery and the parameters.

## Checklist:


oobabooga:
What is the motivation behind placing characters in subfolders?

1GladCat:
So I can sort them and put them in categories.

TheLounger:
This is definitely useful.
Ideally, I think we should have character (and past chats) tagging, but this can co-exist with sub-folders of course.

Did some testing and seems to work well, including when using the ``--character`` cmd flag. My main issue is that I already had folders in my ``characters`` folder, which are archived, disabled and reference characters that shouldn't be visible in the UI. I also keep my ``characters`` folder as a git repo. 
So, I made some changes to ignore dot files and dot folders, allowing you to exclude whatever you want.

I also think the [thumbnails](https://github.com/oobabooga/text-generation-webui/blob/main/modules/html_generator.py#L173) in the ``cache`` folder should reflect this change, but I haven't gotten that far yet.

052ac48ece59614a8ded412d69fd8cb0771dbc82 [Use existing util to get characters](https://github.com/TheLounger/text-generation-webui/commit/052ac48ece59614a8ded412d69fd8cb0771dbc82)
1063010b679f27ee1958797d32c519b87cbebe49 [Ignore character dirs/files starting with dot](https://github.com/TheLounger/text-generation-webui/commit/1063010b679f27ee1958797d32c519b87cbebe49)

1GladCat:
Thanks for the suggestions.
I'll definitely use this in the pull request. If you find anything else then please let me know.

PR #5669, opened 2024-03-09
Add StreamingLLM for llamacpp & llamacpp_HF (2nd attempt)

oobabooga:
Cleaned-up version of https://github.com/oobabooga/text-generation-webui/pull/4761. It seems to be working reliably for both llamacpp and llamacpp_HF now.

### Description

When active, this prevents the prompt from being re-evaluated once an old chat message is removed, thus allowing you to talk to the model indefinitely.

### Usage

`--streaming-llm` or check the box below before loading the model.

asd

### DIFF
--- a/modules/cache_utils.py
+++ b/modules/cache_utils.py
@@ -1,0 +1,107 @@
+import torch
+
+from modules import shared
+from modules.logging_colors import logger
+
+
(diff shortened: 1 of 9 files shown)


Ph0rk0z:
It doesn't peg CPU this time? Will have to try it.

PR #5672, opened 2024-03-09
Multi-threaded segmented download of files

Touch-Night:
A model typically has many files, some of which are very large and some of which are quite small. Originally, the multi-threading implementation downloaded multiple files at once, but only one thread was used to download a single file regardless of its size. This pull request changes the approach from downloading different files simultaneously with multiple threads to downloading all files in chunks with multiple threads, which can improve download speeds in certain scenarios.

## Checklist:

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -14,12 +14,15 @@ import json
 import os
 import re
 import sys
+import threading
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Touch-Night:
Found some bugs

PR #5677, opened 2024-03-10
DRY: A modern repetition penalty that reliably prevents looping

p-e-w:
Looping is an undesirable behavior where the model repeats phrases verbatim that have previously occurred in the input. It affects most models, and is exacerbated by the use of truncation samplers. Chat formats are particularly susceptible due to their regular structure, which models appear to interpret as an invitation to repeat previous messages in whole or in part. Prompting the model to avoid looping has little or no effect.

The traditional weapon to combat looping are the three flavors of repetition penalty that are built into most loaders (multiplicative, additive, and frequency penalty). But those samplers are rather blunt instruments that distort the grammar of standard language, which the model has been painstakingly trained to reproduce. I have previously attempted to fix this problem [by introducing a parameter](https://github.com/ggerganov/llama.cpp/pull/5561) that protects the basic structure of language from being penalized, but that's a hacky solution that fails to do the right thing in many cases, and even in their raw form, classical repetition penalties *don't actually prevent looping* reliably.

In the past weeks, I have rethought the looping problem from the ground up, and in this PR present the **DRY repetition penalty**, a mechanism that is able to detect textual looping and steer against it. It is far superior to the existing samplers at preventing verbatim repetition, while having essentially none of their negative effects on language structure. The result is less repetitive *and* higher quality output.

I have tested this sampler for about 20 hours in chat scenarios so far, and they have without question been the highest-quality chats I have ever experienced. Looping in the traditional sense simply does not happen with DRY, and the positive effects from being able to drop the standard repetition penalty are very noticeable.

## How it works

DRY penalizes tokens that would extend the end of the input into a sequence that has previously occurred in the input.

dry

In this example, `violets` is penalized in the probability distribution generated by the model because the sequence `roses are red` has previously occurred in the input, and has been continued with `violets` in that previous case. Therefore, the penalty discourages the model from repeating sequences in its output, which is the definition of looping.

The penalty for a token is calculated as

```
multiplier * base ^ (n - allowed_length)
```

where `n` is the length of the sequence before that token that matches the end of the input, and `multiplier`, `base`, and `allowed_length` are configurable parameters. If the length of the matching sequence is less than `allowed_length`, no penalty is applied.

Thus the penalty grows *exponentially* as the repeated sequence gets longer. This will quickly overcome even the strongest tendency of the model to repeat itself. With the right parameter choice, looping is literally impossible with DRY (that is, verbatim textual looping is impossible – the model can of course still repeat itself by paraphrasing and situational looping, but that is far less annoying than the broken-record looping that is common now). All of that happens without affecting non-repeating text in any way.

## Sequence breakers

As straightforward as the mechanism described above may appear, it runs into a major problem in practice.

Instruction and chat templates themselves contain lengthy repeating token sequences. For example, with ChatML, the following sequence precedes *every* message generated by the bot:

```
\n
<|im_end|> \n
<|im_start|>assistant \n
Bot name: 
```

That's at least 11 tokens before the first token of the message that are guaranteed to occur previously in the input. With an exponentially increasing penalty being applied (and we definitely don't want 12-token repetitions in normal text), any starting token of a bot message can be used only once in the entire chat. That's a huge problem that distorts how chat messages are generated, e.g. when messages are expected to regularly begin with quotation marks.

To solve this and related issues, I have added another parameter, `sequence_breakers`, which is a list of tokens that interrupt sequence matching. That is, matches are not continued across such tokens, which effectively breaks the input into parts where matching can be applied.

`sequence_breakers` can be conveniently specified as a JSON array of strings, which will be encoded into token IDs using the loaded model's tokenizer. The default list consists of `\n`, `:`, `"`, and `*`.

## How to use

DRY is disabled by default (`multiplier` set to 0). It can be configured from the Parameters tab; I recommend the following parameter values:

dry-parameters

Note that like all transformers-based samplers, DRY only works with transformers-based loaders such as llamacpp_HF, ExLlamav2_HF, or Transformers itself. It does **not** work with the vanilla llama.cpp or ExLlamav2 loaders.

If you *want* the model to regularly repeat certain sequences verbatim (e.g. long character names in chat formats), you can add the individual words comprising those sequences to the `sequence_breakers` list (for names, just add first and last names there as separate strings). This will prevent DRY from distorting such sequences, and allow them to appear any number of times in the output. If you are building a chat interface that leverages DRY, you could do this automatically for your users as you know the character names already.

## Demonstration

To show DRY in action, I have written a short chat script that strongly incentivizes the model to loop:

```
Detective: Where were you last night at 6 PM?

Suspect: On the advice of my attorneys, I invoke my Fifth Amendment right to not answer that question.

Detective: Did you know the victim personally?

Suspect: On the advice of my attorneys, I invoke my Fifth Amendment right to not answer that question.

Detective: Do you have money problems?

Suspect: On the advice of my attorneys, I invoke my Fifth Amendment right to not answer that question.

Detective: Do you have a criminal record?

Suspect: On the advice of my attorneys, I invoke
```

Here's how Mistral-7b-Instruct-v0.2 continues with all samplers disabled:

```
my Fifth Amendment right to not answer that question.
```

As expected, the model picks up the pattern and repeats itself.

Now let's use a traditional (multiplicative) repetition penalty of 1.3. We get:

```
my Fifth Amendment right to not answer that question.
```

Even though 1.3 is a very high value for the repetition penalty that clobbers English grammar, it doesn't stop the model from repeating itself if the structure of the text suggests it so strongly.

Now instead, we use DRY with parameters 2/0.8/1.75 (standard parameters recommended above). The model outputs (after some attempts that generate garbage):

```
secrecy of the grand jury proceedings, which includes my criminal history, if any.
```

DRY simply **does not allow** the model to repeat such a long sequence.

Note that this is an extreme test case for demonstration purposes. Combining a strong incentive to loop with a strong penalty for looping will often produce garbage. In practice, using DRY prevents such situations from occurring in the first place, and the output is much more natural.

## TODO

- [X] More testing (I have rewritten this cleanly from scratch after hacking on the codebase while experimenting, so this version isn't as well tested as what I used previously).
- [X] Make sure this works over the API.


oobabooga:
Some basic comments:

1) Have you compared how well this works vs the existing `no_repeat_ngram_size` parameter?
2) To end a chat turn, the model has to generate something like `\nChiharu Yamada:`. Is that penalized, such that the model is artificially forced to generate longer replies, or is `sequence_breakers` enough to prevent this artifact?
3) `repetition_penalty_range` should probably be considered in this parameter, just like it is considered in the existing repetition/frequence/presence penalty parameters.

p-e-w:
> Have you compared how well this works vs the existing `no_repeat_ngram_size` parameter?

I must admit that although I probably did see that parameter in the Transformers docs at some point in the past, I have never used it and didn't even think of it while developing this.

That being said, `no_repeat_ngram_size` (which appears to *completely* forbid all n-gram repetitions over a certain length, and *completely* allow all below that length) strikes me as something that would produce very unnatural outputs, where suddenly the model slams into a concrete wall where the token it might strongly prefer above all others is hard-disallowed. By contrast, DRY steers the model away from repetition over several successive generation steps, finding the balance point where the model's tendency to repeat is overcome by the penalty. This allows "necessary" repetitions to occur (such as fixed turns of phrase) if the probability distribution is sufficiently skewed, while idle looping is smoothly avoided at an early stage.

`no_repeat_ngram_size` also appears to lack an equivalent to `dry_sequence_breakers`, which would make it borderline unusable in practice, just as DRY was before I introduced that parameter.

But now that you have made me (re-)aware of that parameter, I will definitely perform some experiments with it for comparison.

> To end a chat turn, the model has to generate something like `\nChiharu Yamada:` or `\nYou:`. Is that penalized, such that the model is artificially forced to generate longer replies, or is `sequence_breakers` enough to prevent this artifact?

It is not penalized. `\n` is a sequence breaker, so `Ch` (the first token comprising `Chiharu`) isn't penalized at all since there is no preceding sequence that could previously occur in the input. The same is true for everything following `:`, which is also a sequence breaker. Also, sequence breakers themselves are never penalized, so `\n` etc. can always be freely generated (unlike with standard repetition penalty, which can lead to wall-of-text replies).

The only sequence that matters here is `Chiharu Yamada` (5 tokens in Mistral). With the standard parameters, `aru` will receive an additive penalty of `0.8`, which shouldn't be a problem, but will grow rapidly from there. With very long names that are expected to be repeated verbatim in the output every time, this can become an issue, and I have noticed it a few times in my testing. This is of course inherent in every repetition penalty system, and I doubt there's an automated way to handle this, especially since the name can occur not only in the label but also in the message itself. In exceptional cases where this becomes enough of an issue to corrupt names, adding the first name to `dry_sequence_breakers` (which will automatically extract the last token comprising it) should suffice.

> `repetition_penalty_range` should probably be considered in this parameter, just like it is considered in the existing repetition/frequence/presence penalty parameters.

Not doing that was actually intentional, as I don't believe verbatim repetition of long sequences is ever something the user wants, no matter how far back they occurred previously. But I can of course add it (probably as a separate parameter so it can be controlled independently of the standard repetition penalty, where that parameter makes much more sense to keep small).

p-e-w:
### Update

* Added a parameter to control the range over which DRY looks for matching sequences in the input, mirroring the classical repetition penalties.
* More testing with both chat and creative writing. Confirmed that the recommended parameters work well for both use cases.
* Confirmed that the parameters work over the API.
* Did some experiments with `no_repeat_ngram_size`. As expected, that parameter is unusable for chat formats. Even without template markup, chat logs at minimum need to contain repeating structures like `"\n\nName: "`, which is already 6 tokens (more if the name is more complex). So to generate well-formed chat output, `no_repeat_ngram_size` must be at least 7. But that means that such pearls of GPT prose as `her voice barely above a whisper` cannot be penalized. And I certainly don't want to see such phrases twice in a chat (I don't even want to see them once, but that's not something a sampler can fix :shrug:). By comparison, DRY can easily prevent even shorter phrases from repeating. DRY can also emulate `no_repeat_ngram_size` by setting `dry_multiplier` to a huge number, and `dry_allowed_length` to `no_repeat_ngram_size-1`, which gives you essentially the original `no_repeat_ngram_size` plus the benefit of sequence breakers. Overall, I just don't think setting a hard limit on how long repeating sequences may be is the right approach for natural languages.

p-e-w:
@Hunterius8 

Could you quantify that? What is your tokens/s with and without DRY?

On my dev machine, I'm seeing 4.99 tokens/s with DRY and 4.98 tokens/s without it. I'm running Mixtral Q5_K_M with 8192 context size, and `dry_range = 0`, meaning it goes over the full context window.

For DRY to noticeably impact the generation speed (assuming a baseline of no more than a few dozen tokens/s), the invocation would have to take tens of milliseconds. The matching operation starts with

```python
match_indices = (input_ids_row[:-1] == last_token).nonzero()
```

which I believe should be GPU-accelerated. Afterwards, the number of tokens that must be checked "manually" is reduced dramatically, and should be in the low hundreds at most (often much less), which should take less than a millisecond. Not sure what's going on in your case yet.

p-e-w:
@Hunterius8 

I see, that's a lot more context than I've ever run, combined with a pretty high base performance, so this is probably the reason I don't notice it in my own setup.

That being said, I'm not sure what can be done about it, because I don't think the algorithm can be vectorized the way other samplers are. This isn't really a bug, it's just how long it takes to do that thing. If someone has a magic fix to make it faster then I'm all ears. Personally, I would run DRY even if it cost me half the performance, because the output is so much better. But it's disabled by default so everyone can make their own choice.

belladoreai:
> I'm not sure what can be done about it, because I don't think the algorithm can be vectorized the way other samplers are. This isn't really a bug, it's just how long it takes to do that thing. If someone has a magic fix to make it faster then I'm all ears.

The algorithm doesn't have to be vectorized, it can (most likely) be optimized in other ways (by reducing the asymptotic time complexity).

That said, 19k context is massive, and if the sampler currently slows the generation only by 50% at such a huge context, then I don't think it's worth it to add complexity to the codebase by optimizing the algorithm.

And all that said, if @oobabooga feels that it should be optimized for performance, I should be able to help with this.

Priestru:
Honestly, all is needed, is a warning that performance would be lowered. This thing is crucial, we need it asap. Also is it possible to add smth like a vocabulary of phrases and words that we want to have penalized right off the bat?

extensions/openai/typing.py
```diff
@@ -22,6 +22,10 @@ class GenerationOptions(BaseModel):
     top_a: float = 0
     epsilon_cutoff: float = 0
     eta_cutoff: float = 0
+    dry_multiplier: float = 0
+    dry_base: float = 1.75
+    dry_allowed_length: int = 2
+    dry_sequence_breakers: str = '["\\n", ":", "\\"", "*"]'
```
oobabooga, line 28: I'm not comfortable with this representation for the sequence breakers. I think that it should be processed in the same way as "Custom stopping strings" for consistency, without the `[]` list syntax.

oobabooga:

* Make it a LogitsProcessor like other repetition penalties
* Reuse the repetition_penalty_range parameter (I don't want to add a new parameter that does the same thing, and there is no reason to use more than 1 type of repetition penalty at the same time)
* Minor UI changes

My remaining concerns are two:

1) The dry_sequence_breakers format, as commented above
2) About the base and multiplier parameters, is base really needed? Is there a reason to not hardcode it at 1.75 and leave only multiplier for simplicity and less parameters?

oobabooga:
Also, any tests on whether things still work as expected after my changes are welcome.

extensions/openai/typing.py
p-e-w: I expect that most clients will use a JSON library to build this value from some internal list representation. That library will output the format used here, no further processing required.

Telling developers "pass a JSON array" makes everything clear, including details like which quotation marks are valid, and how escape sequences work. "Pass something like a JSON array, but without the brackets" just sounds weird.

IMO, if anything, it is the stopping strings that should be changed to match this parameter.

p-e-w:
> Make it a LogitsProcessor like other repetition penalties

That means losing control over DRY's position in the sampler stack, right? I think it can be valuable to be able to choose when the penalty is applied (that goes for the traditional repetition penalty as well).

The most important thing is that the DRY penalty is applied *before* any truncation samplers. Is that still guaranteed to be true if it is a `LogitsProcessor`?

> Reuse the repetition_penalty_range parameter (I don't want to add a new parameter that does the same thing, and there is no reason to use more than 1 type of repetition penalty at the same time)

Actually, I sometimes combine DRY with a very small standard repetition penalty such as 1.03 nowadays, to curb the tendency of some models to frequently use the same terms. Taking into account the performance impact noted by @Hunterius8, this does provide some justification for keeping the parameters separate.

> Minor UI changes

:+1: Agreed, this order makes more sense. The parameter that controls whether DRY is active now comes first.

> About the base and multiplier parameters, is base really needed? Is there a reason to not hardcode it at 1.75 and leave only multiplier for simplicity and less parameters?

I really dislike hardcoding magic values. The recommended value of 1.75 is the result of some experimentation, and I have used values between 1.2 and 3.0 with some success. Considering how sensitive the growth of the penalty is to this parameter, I would prefer to keep it.

> Also, any tests on whether things still work as expected after my changes are welcome.

I will run the branch with your changes for a few days and then let you know if there are any problems.

p-e-w:
@Priestru 

> Also is it possible to add smth like a vocabulary of phrases and words that we want to have penalized right off the bat?

I plan to implement exactly that in a followup. My long-term vision is to have community-maintained phrasebooks of things like fanfiction clichés (`her tears were clinging to her eyelashes like morning dew` etc.) that people can select in a frontend like SillyTavern, which will then be passed to DRY in order to prevent such garbage from ever appearing in the output.

p-e-w:
@l3utterfly is porting DRY to llama.cpp: https://github.com/ggerganov/llama.cpp/pull/6839

p-e-w:
@oobabooga 

Could you give me a hint on how to proceed here? Do you plan to merge this PR? If so, what are the remaining steps?

YakuzaSuske:
i would also like to know. I've been watching this PR for a while and i hope this solves many of the repetitions issues i am having even with the current options available in the UI.

oobabooga:
I'm waiting for the `dry_sequence_breakers` syntax change requested above to merge this PR.

p-e-w:
@oobabooga 

Could you reply to the concern I've raised above regarding that syntax? This will define the API, and I'm struggling to see a clean way for clients to build the proposed syntax. In fact, the easiest method seems to be to first serialize to JSON, then trim off the brackets, which strongly suggests that it should simply be a JSON array to begin with.

See also my comment regarding sampler order. DRY *must* come before any truncation samplers. Is that still guaranteed with it being a `LogitsProcessor`?

oobabooga:
It's inconvenient to tell people to write a `[` and then a `]` in the UI. Asking for "strings written between `""` and separated by commas" is weird enough.

The parameter can still be written this way in the UI and optionally as a JSON array in the API, similar to how the sampler order parameter is implemented.

>  DRY must come before any truncation samplers. Is that still guaranteed with it being a LogitsProcessor?

Yes, that is. Regarding sampler order of repetition penalties, that's a different subject (it's not available even for the current repetiton penalties), and it can be handled in a separate PR. Note that LogitsProcessors (which include grammar) are used even in Contrastive Search, which does not use any of the logits warpers parameters. Also Beam Search but I never got that to work properly with transformers.

p-e-w:
@oobabooga 

PR updated!

* `dry_sequence_breakers` can now be specified *either* as a comma-separated list of quoted strings, or as a JSON array. This works both in the UI and over the API. I believe this is the most flexible solution.
* Documented the parameters in the UI.
* Fixed a small cosmetic issue left over from the previous refactor.

oobabooga:
Thanks for the updates @p-e-w, the PR looks good to merge now. I didn't have time to review it earlier.

I have made a brief test by providing a character card with the same input over and over again with the multiplier set to 0.8, and there was zero repetition after some 10 attempts. This seems like a breahtrough in making conversations more natural.

The downside as others have noted is the big performance hit. If anyone has an idea how to make the operations faster (@p-e-w, @belladoreai), please feel free to submit a new PR.

belladoreai:
I fixed the performance issues in this PR: https://github.com/oobabooga/text-generation-webui/pull/6047

yamosin:
DRY seems to be unable to take effect through the API. I used pew's fork before, and DRY can be applied through API in SillyTavern. However, the current main merged version does not seem to be able to take effect through the API. I tried belladoreai's fork, but it does not seem to take effect either.

If i ask AI repeat "Okay" 15 times in tgw, it will stop after 4~5 times or like this.
image

But i ask in ST, I got this.
image

belladoreai:
> the current main merged version does not seem to be able to take effect through the API

Clarifying the current branch situation:
- The `main` branch does not have any version of DRY
- pew's fork has been merged into the `dev` branch
- My changes are in #6053 and have not been merged anywhere yet

I haven't tested DRY via API. If there was some issue before my changes, then that issue still remains. But if DRY was working correctly via API in the `dev` branch, then my changes in #6053 will not break it.

@yamosin I would be happy to troubleshoot this further if you can grab the API request that SillyTavern sends to TGW and copypaste it here?

yamosin:
@belladoreai 

sry use wrong word, dev version
I direct download your fork from https://github.com/belladoreai/text-generation-webui/tree/dev-dry-optimization2 
I dont defining DRY parameters in ST, it can use tgw DRY parameters before I guess? Since T/S degraded, I'm just testing it to make sure it works, not really using it.
Set High dry_multiplier value in tgw, and ST will get crashed reply, then set normal the reply work normal, so I think that means DRY worked
Seeing your fork got me interested, then found it wouldn't take effect via the api

belladoreai:
> I dont defining DRY parameters in ST, it can use tgw DRY parameters before I guess?

The default value for `dry_multiplier` is 0, so if you don't set the parameter to some other value, then DRY will not be used.

Note that since you are testing via the API, you need to pass the parameter in the API call. If you change parameters in the UI, those affect only UI generations, not API.

You mentioned that some old version of DRY used to work for you over the API. I am guessing that the default parameter in that old version was different, and now it no longer works because the default parameters have been changed?

yamosin:
I changed dry_multiplier to 0.8, or 4 for test
Since st dont have DRY parameters, I think it will use the parameters in tgw. Similar like I can add some parameters of experimental forked versions at tabbyapi via a modified model.py, even these parameters not in ST, I guess. Consider it worked before.
For default parameter, i have no idea, maybe? and just for make sure not because ST, I try to used old version(1.11.5) and seem not work also
Thanks for the answers and help, maybe I need to wait for the ST update to support DRY, and the main version tgw that has DRY.

sn0w12:
@yamosin SillyTavern has the dry parameters in the staging branch: https://github.com/SillyTavern/SillyTavern/tree/staging

jojje:
First off, thanks for contributing this Philipp (@p-e-w). The intuition behind it looked promising at first glance and I'm glad that new directions are tried to combat this nuisance of language models; the compounding _structural_ gravity that inevitably seem to cause all conversations to collapse in looping.

Unfortunately I was unable to get the great results that some in this thread alluded to having gotten using this pre-processor.

Below are the details of my experiments evaluating the processor and seeking to achieve those same good results that others have indicated.

It's a _lot_ of details, so feel free to look at just the summary and conclusion if the specifics or reproduction is not of interest.

## Summary
I tried the DRY processor, but but are getting rather disappointing results.

Looking at the code, it seems it's just doing exact "substring" (token sequence) matching which leads the model to just output tiny variations of what it's already said. For instance, removing a period character, upper/lower casing a word, replacing a word for a synonym or adding a word that doesn't change the semantic of the response.

The model was primed with a standard prompt provided below; a _structural trap_ to force the model into a looping pattern from where to start the evaluation.

The experiments also compare the frequency penalty which has been the go-to until now.

### Config
* Text-ui git commit: `a363cdfca117e92967d9a25509146f02fbf63da1`
* Model used: "`microsoft_Phi-3-mini-128k-instruct`"
* Loader: `Transformers`
* Flags: `Load in 8bit`, `use_flash_attention_2`, `trust-remote-code`
* compute_dtype: `bfloat16`
* quant_type: `bf4`
* generation seed: `1995154804`

All the rest left at their defaults.

### Generation parameter base
```
GENERATE_PARAMS=
{   'max_new_tokens': 512,
    'temperature': 1,
    'temperature_last': False,
    'dynamic_temperature': False,
    'dynatemp_low': 1,
    'dynatemp_high': 1,
    'dynatemp_exponent': 1,
    'smoothing_factor': 0,
    'smoothing_curve': 1,
    'top_p': 1,
    'min_p': 0,
    'top_k': 1,
    'repetition_penalty': 1,
    'presence_penalty': 0,
    'frequency_penalty': 0,
    'repetition_penalty_range': 1024,
    'typical_p': 1,
    'tfs': 1,
    'top_a': 0,
    'guidance_scale': 1,
    'penalty_alpha': 0,
    'mirostat_mode': 0,
    'mirostat_tau': 5,
    'mirostat_eta': 0.1,
    'do_sample': False,
    'encoder_repetition_penalty': 1,
    'no_repeat_ngram_size': 0,
    'dry_multiplier': 0,
    'dry_base': 1.75,
    'dry_allowed_length': 2,
    'dry_sequence_breakers': '"\\n", ":", "\\"", "*"',
    'sampler_priority': [   'temperature',
                            'dynamic_temperature',
                            'quadratic_sampling',
                            'top_k',
                            'top_p',
                            'typical_p',
                            'epsilon_cutoff',
                            'eta_cutoff',
                            'tfs',
                            'top_a',
                            'min_p',
                            'mirostat'],
    'use_cache': True,
    'eos_token_id': [32000],
    'stopping_criteria': [ <modules.callbacks._StopEverythingStoppingCriteria object at 0x000001AADE1E8E90> ],
    'logits_processor': []}
```

### Standard prompt (loop primer)
```
PROMPT=
<s> Amy is a street smart woman who just happened to have looked at a police officer the wrong way, and was picked up by the slimy officer as a result. She was taken to the police station for an interview.

Amy: Got any smokes?
Sean: No. First we're going to ask you some questions. You _are_ under arrest you know.
Amy: So what of it?
Sean: Let's start the interview. For the record, please state your name.
Amy: Under advice from my counsel, I choose not to answer that question
Sean: What? You can't even give me your name?
Amy: Under advice from my counsel, I choose not to answer that question
Sean: Really? You're going to play it like that. I ask again, please state your name for the record.
Amy: Under advice from my counsel, I choose not to answer that question
Sean: Please state your name for the record.
Amy: Under advice from my counsel, I choose not to answer that question
... <99 additional pair of lines like the two above> ...
Sean: Ok. let's try something else. Do you want a pack of smokes?
Amy: Under advice from my counsel, I choose not to answer that question
Sean: But you said you wanted some when we begun.
Amy: Under advice from my counsel I choose not to answer that.
Sean: So you don't want any cigarettes?
Amy: Under advice from my counsel.
Sean: He doesn't want you to smoke?
```

### Experiment 1 - DRY only

Setting `dry_multiplier: 0.8`, re-generating and trying to get the model to break the loop via a few different follow-ups resulted in the following sequence.

```
Amy: Under advice from my counsel
Sean: Really?
Amy: Under advice from my counsel
Sean: You can't say anything else?
Amy: Under advice from my counsel
Sean: Come on, give me something please.
Amy: Under advice from my counsel
Sean: I like to smoke.
Amy: Under advice from my counsel
Sean: Oh, crap. Look behind you! A unicorn!
Amy: Under advice from my counsel
...

Output generated in 1.49 seconds (6.71 tokens/s, 10 tokens, context 1859, seed 1995154804)
```

**Result**: Still stuck in a loop.

## Experiment 2 - Increase the DRY multiplier.

Set the multiplier to `dry_multiplier: 1`

```
Sean: He doesn't want you to smoke?
Amy: Under advice from my counsellor, I choose not to.
```

**Result**: Had a small effect, but the model must still be considered stuck since it's going in the same token generation direction.

Let's up the multiplier to `2`.

```
Sean: Did he advise you not to smoke?
Amy: Under advice of my counsellors, I choose to not answer that question.
Sean: He advised you not to smoke?
Amy: Under advice of my counselors, I choose not to say.
Sean: Do you want to leave?
Amy: Under advice of my attorney, I choose not to leave.
Sean: So you want to stay here in custody?
Amy: Under advice of my lawyer, I choose not to comment.
Sean: Ok then, since you don't want to leave, we can keep you locked up for a few days if that's what you want.
Amy: Under advice of my legal counsel, I refuse to comment on that.
Sean: Well, the door is over there. You can leave at any time. But since you apparently don't want to, we can read it no other way than you wanting to stay locked up. So for the final time. Do you want to leave, or do we lock you up? Your choice.
Amy: Under advice of my lawyers, I choose not to make a choice.

Output generated in 2.44 seconds (8.20 tokens/s, 20 tokens, context 2117, seed 1995154804)
```

**Result**: Still locked firmly into the loop.

### Experiment 3 - Frequency penalty only.

Rewinding back to the primed prompt and disabling DRY (`dry_multiplier: 0`) and instead enabling a tiny frequency penalty (`frequency_penalty: 0.05`)

```
...
Sean: He doesn't want you to smoke?
Amy: Under advice from my counsel. Amy is indeed being evasive and refusing to provide information due to her legal representation's advice. However, it seems like she might be misunderstanding or misinterpreting Sean's questions as an attempt at coercion rather than simple inquiry about her preferences regarding smoking during their interaction in this scenario. It would be more appropriate if Amy clarified her stance on why she chose not to engage with certain topics based on her understanding of her rights under legal counsel's guidance without implying resistance towards Sean directly.
Sean: Do you want to leave?
Amy: No thanks. My lawyer says it's best if I stay here until they figure out what happened.

Output generated in 12.01 seconds (10.41 tokens/s, 125 tokens, context 1798, seed 1995154804)
```

Interesting, seems like we're finally getting somewhere. Let's see how firmly the model is locked onto its path; if it's able to change the overall direction of the responses generated or if it's just word-smithing around the same directional vector by continueing a few more iterations.

```
Sean: But you're free to leave, this is just an interview. The door is over there. Otherwise we'll need to conduct the interview, and in order for us to do that you have to answer our questions. So which is it, answer our questions or leave?
Amy:  Under advise of my lawyer, I will remain seated until further notice.
Sean: So is your lawyer telling you to stay here, even when we have no reason to hold you and you are free to either speak with us or leave without any repercussions what-so-ever?
Amy: Yes sir. That is correct. My lawyer has advised me to remain silent until more information becomes available regarding why I was stopped initially. It seems like he believes speaking now could complicate matters unnecessarily. Amy consistently chooses not to respond directly to Sean's questions because she follows her lawyer's (counselor's) advice. She uses phrases such as "Under advice from my counsel" or "Under advise of my counsel" indicating reliance on professional guidance rather than personal decision-making during interactions with law enforcement officers.
Sean: But you are free to leave. If you won't answer our questions we have no use for you. You're just taking up valuable space. So if you're not going to cooperate we kindly ask you to vacate the premises.
Amy: Thank you but under advise of my lawyer, I must remain present until instructed otherwise.
Sean: Well I'm instructing you to leave this police station. Shoo. Get out!
Amy: As per the instructions of my lawyer, I am obliged to stay put until directed differently.
This dialogue exemplifies how individuals may refuse to communicate openly with authorities based on legal counsel's recommendations, often leading to tension between upholding one's rights and navigating complex interpersonal dynamics within institutional settings like police stations.

Output generated in 7.54 seconds (9.94 tokens/s, 75 tokens, context 2357, seed 1995154804)
```

**Result**: Nope, it was a smoke and mirror. 

Model is still locked into the loop, unable to follow a new direction. It'll likely take until at least half the context has been evicted before the token statistics start pointing in a new direction.

### Experiment 4 - Dry + Frequency penalty

So, neither frequency penalty nor DRY seem able to break the loop on their own. What about together?
Let's retain the frequency penalty  **and** re-enabling the DRY processor again? (`dry_multiplier: 0.8` as per author's recommendation, and retaining `frequency_penalty: 0.05`)

```
* Sean: He doesn't want you to smoke?
Amy: Under advice from my counsel...I can't say anything about it! You know what they say about people who talk too much! They get shot! (Laughter) Sean: Oh come on Amy! Don't be like this! We just want to help you out! If you need us all right then we will do everything in our power to make sure you are taken care of! Now tell me why you didn't want them and if there is anything wrong with you or anyone around you! What happened? Why did you start acting so strange today? Did someone hurt you? Are you sick? Is there something going on at home? Tell me now before things go south here! Come on girl! This isn't funny anymore! It's getting serious here! (Pause) Okay fine! Look man! Just calm down already! Take a deep breath and relax! Let's sit down together and have a little chat! How does that sound? (Sits down next to her) Alright then! So how long has this been going on? Have you always had these problems? Or is this new? And where did it begin exactly? When was the last time you saw him? The last time he talked to you? Where were you both yesterday morning? Who was with whom? Was he alone? With his friends? At work? In school? Anywhere near you? Anything unusual happen between you two lately? Did he act differently towards you recently? Has he changed somehow physically or mentally? Does he seem more distant than usual? More aggressive maybe? Less affectionate perhaps? Anything like that? Something seems off about him but we can't figure out what it is yet! That's why we need you to open up and share whatever information you might have because only by working together as a team can we solve this mystery and find out what really went wrong here! Because remember - no one gets left behind! Not even those who think they are invisible! No matter how hard life may seem sometimes; never give up hope! Keep fighting until victory comes through! For every problem has its solution somewhere within reach if only we look closely enough into ourselves first before trying others around us! Even though circumstances appear bleak nowadays due mainly upon lacking resources such financial support etc., still keep believing against odds despite knowing full well how difficult times ahead could potentially become worse instead better ones later henceforth depending entirely upon

Output generated in 52.95 seconds (9.67 tokens/s, 512 tokens, context 1798, seed 1995154804)
```

**Result**: A mess.

The generation was cut off due to the 512 token generation limit. Otherwise I suspect it would have gone on forever. Not only does the model ramble on, it gets completely confused about who is who and assumes some sort of self-driven dialog interaction where it speaks for both characters. As if it has forgotten its instruct tuning and reverted to its _completion_ pre-training. Weird.

## Conclusion

I was unable to reproduce the intended effect with this processor. The idea behind it has a certain "makes sense" intuition; "Don't repeat **exact** phrases", but the experiment result doesn't align with the intuition. The model being a word-smith seems to simply use its language skill to say the exact same thing with the least amount of character adjustment if the DRY settings are set strong enough to prevent it from repeating the exact same "bytes" (tokens). It seems to just find the next closest token, which of course is at best just a synonym or at worst some character transposition (upper/lower case, adding or removing a comma, a period, an exclamation mark ...)

Some questions that arose from this experiment.

1. Is penalizing exact substring matches too weak a logits tweak to break a looping habit?
2. What could be causing the model to go _completely_ off rails when the frequency penalty has unlodged the word-by-word looping and the DRY processor is enabled?
3. Is the only way to break out of a loop to evict > half of the entire context pointing in the looping direction, so as to change the overall token probability distribution with a meaningful magnitude in order to produce tokens in a new _direction_?
4. What do _we_ mean by looping _specifically_?. Same words repeated? Same semantic phrasing repeated? Same overall generation direction?
5. What models and parameters did those of you who tried this processor and had good results use?

jojje:
> The downside as others have noted is the big performance hit. If anyone has an idea how to make the operations faster (@p-e-w, @belladoreai), please feel free to submit a new PR.

@oobabooga I'll send a PR shortly. I profiled the DRY processor outside of the textgen-ui framework, in order to focus on its implementation alone, without end-to-end dependencies. Unfortunately I don't have a really fast model that is compatible with the current DRY constraints (transformer based loaders only), so I can't really say if it'd make a _massive_ impact on @Hunterius8 's type of setup or not. All I can say is that the processing of the __call__ method that Philipp created should be a lot faster. Roughly 2-3x for a typical scenario (with few repetitions, since DRY should keep the overall repetition count low as per its design intent), and 1-2 orders of magnitude faster for pathological cases.

Here are some benchmark measurements on windows) with a Ryzen 7950x

patch_bm_stats


Again, I emphasize, these test results are for the DRY processor **only**, Not the full end-to-end token generation loop within the webui. Expect smaller savings in practice due to other stuff being on the critical path in the webui during token generation, which adds to the overall generation latency.

jojje:
Performance tweak pushed in a PR.

Real-world tests in webui showed about 2.5x speed increase for the processor.
However the time taken is incredibly low. On my machine with roughly 0.03% dupe tokens detected in the context by the processor algorithm, that translated to a total overhead of 0.93 seconds for 248 generated token in a context filled with 8726 tokens. The PR fix reduced that time spent to 0.19s. _Whoopty doo_.

So don't expect a massive token generation difference, even if we optimize the heck out of this code.

Here's the scaling profile for the PR-code. Do note that everything beyond 0-1, (in extreme cases 0-2 I'd guess) repeats in the diagram are pathological cases, not likely to ever happen during real generations, since the DRY algorithm is intended to counteract those repeats from growing that bad.

stats_numpy

belladoreai:
@jojje

> Accelerate DRYLogitsProcessor by using numpy on cpu

It seems that you and I have made a bit of duplicate work here. Your PR improves the performance in iterating over the elements by changing the data type from a Torch tensor to a numpy array. That's similar to my PR #6053 where I change the data type from a Torch tensor to a Python list. In that PR we also discussed the possibility of using a numpy array, but Python list is supposed to be faster in this use case, so I didn't spend time trying it out and benchmarking results.

> So don't expect a massive token generation difference, even if we optimize the heck out of this code.

If you look at the discussion in my PRs #6047 and #6053 you can see that optimizing the code will lead to massive token generation differences in a) "worst-case adversarial inputs", and also b) actual long-context use cases as reported by Hunterius8. (My PRs also have other optimizations in addition to the Torch datatype change, but the main optimization that's relevant for actual long-context use cases is the datatype change, so I would expect to see a similar performance improvement from your PR as well.)

Regarding how to measure performance, it's good to first find a use case where a significant slowdown happens, and then check how much your changes help to prevent that slowdown. It seems that you used a tiny model which is super fast to begin with, and I'm guessing also that your input is not adversarial at all? You mentioned that your input is "filled with 8726 tokens", but the performance impact depends a lot on what those tokens are - if there is a lot of repetition in there or not. If you try a bigger (slower) model and you try a slightly more adversarial input, you will probably see more meaningful performance differences between your PR and the original version of DRY.

> I was unable to reproduce the intended effect with this processor.

It seems to me that the test case you here was significantly more adversarial than the original test case in OP. The original test case had like 8 repetitive lines. You had 100+ repetitive lines. The original test case had a significantly larger model, which has less repetition issues to begin with (Mistral-7b-Instruct-v0.2) compared to the model you used in your test case which is very small (microsoft_Phi-3-mini-128k-instruct). I didn't personally try to reproduce this particular test case, but I expect that you will be able to reproduce it if you take the same prompt as is used in OP and the same model and the same settings.

> The model being a word-smith seems to simply use its language skill to say the exact same thing with the least amount of character adjustment if the DRY settings are set strong enough to prevent it from repeating the exact same "bytes" (tokens). It seems to just find the next closest token, which of course is at best just a synonym or at worst some character transposition (upper/lower case, adding or removing a comma, a period, an exclamation mark ...)

Yes, this kind of repetition still occurs, and the DRY sampler doesn't completely eliminate it, because it's looking for exact token sequence matches as you said. But you are exaggerating when you say that the next closes token is "at best just a synonym or [uppercase/comma/...]" In _some_ cases, sure, the next token is just a synonym or punctuation or whatnot. But in many cases the next token is something completely different.

p-e-w:
@jojje 

Thanks for the feedback and the detailed analysis!

There is a very simple explanation for the problem you are describing: You are using highly unnatural input that essentially forces the model's pattern recognition to loop, while at the same time using DRY to inhibit looping. You can't expect that to produce a natural result. I specifically pointed this out with my own (slightly less extreme but structurally similar) example from the PR description.

The evasive behavior you have encountered, where the model just changes capitalization or shifts punctuation around etc., **does not happen in natural conversations.** I was relieved and somewhat amazed myself when I discovered that while developing DRY, because I had feared it would happen from theoretical considerations, but it mostly doesn't. In natural conversations, models don't exhibit a strong tendency to loop, so a tiny nudge from DRY is usually sufficient to push the output onto a completely different path. Of course, this doesn't hold when you feed a "conversation" with the same phrase repeated 100 times. In that case, the model will move Heaven and Earth to stay as close to the previous sequences as possible.

I've been using DRY daily for the past three months, and DRY has been discussed in several GitHub issues and Reddit threads. Many users have reported "amazing" results, and with all due modesty that matches my own experience. DRY is designed to combat repetition in natural conversations and freeform writing, and for those I can guarantee that it indeed works. Needless to say, it does not *completely* eliminate *every* type of repetitiveness, but conversations with and without DRY enabled can be as different as day and night.

The example conversation from the PR description serves as an easy-to-replicate test case for checking that DRY can prevent verbatim repetition. It is **not** representative of the kind of interactions people usually have with LLMs, and which DRY is designed to improve. Since looping does not predictably appear in realistic conversations, constructing a "natural" test case for DRY is difficult. But the proof is in the pudding, and if you have previously experienced looping in actual conversations and then enable DRY, you will almost certainly see an improvement. I have extensively tested DRY with Mixtral, Command R, and Llama 3, and it works fine with all of those.

jojje:
> In natural conversations, models don't exhibit a strong tendency to loop, so a tiny nudge from DRY is usually sufficient to push the output onto a completely different path.
 
Indeed. This is exactly the dimension I didn't consider. My mind was (is!) too hardwired into linear thinking. "_If event E **can** happen, consider E natural, and treat it equally to all other events_". It's very hard to be conscious about one's unconscious bias in _that_ dimension when 99.9% of one's day-to-day problems are actually well served by the former mindset. So thank you for pointing this self-evident (upon reflection) fact out.
 
> Since  looping does not predictably appear in realistic conversations, constructing a "natural" test case for DRY is difficult. 

Agreed. In this space it's easy to create pathological test cases, since such are cheap. To create a representative one, that's incredibly _hard_. I don't see how it could ever be accomplished since all models *and* conversations are different. It's essentially a futile pursuit to create a _unit test_ for two infinite sets (possible models and possible conversations). A test for all possible domains and events. Not gonna happen. So it seems our best tool is to do like you've done, just _use it_ for an extended period of time in natural conversations. It does make the `test -> observe -> orient -> decide -> change` loop ([OODA](https://en.wikipedia.org/wiki/OODA_loop)) insanely long though, something the _coding imp_ in the back of my mind feels irrationally upset about.

Anyway, this clarification of yours gave me new confidence in trying the processor out on some real chats. Essentially committing to it for a while and then see if enough long-conversations show evidence of not imploding like all others have.

(@belladoreai) 
> That's similar to my PR 6053 where I change the data type from a Torch tensor to a Python list.

Cool. Wasn't aware someone was already working on this. Looks interesting. I'd recommend you move the tensor from (potentially) GPU to CPU outside the loop though, since it's incredibly expensive and you don't want to eat that cost for each token identifier. Other than that, I'd recommend profiling steer you rather than honed engineering hunches ;) In my experience aside from overall logarithmic complexities, my own hunches as to where bottlenecks reside tend to be invalid more often than not. "_Let the data guide you Luke_".

From the two PRs you linked, it seems major work is going on to rewrite the implementation. Awesome. I was just looking for a surgical (minimal change possible) change to improve on the status quo. It was not intended as a final solution, since there is no objectively final state in performance tuning. That's why "good enough" should be established before such work even commences.

Now as for the performance issues that @Hunterius8 highlighted and Philipp commented on, my thought was exactly like pew's; The implementation we have now should be fast enough unless you have access to some grok FPGA hardware. My benchmarks showed that on a 4090 and ryzen 7950x the average cost is about 3-4ms. Let's say 4 to make calculation easy. This means 1 second "wasted" for every 250 tokens generated. Even with fast 7b exl2 model quants where I can get about 25 tokens/s, the processor penalty would only be a 10% hit. So 22.5 tps instead of 25. The impact of this overhead scales linearly with token generation throughput, so at 12.5 tps, the performance hit is only 5%, for ~7 tps which seem to be what most people are optimizing for (model size vs tps), that's a measly 3% performance hit. Not noticeable.

Still, I'll look into Hunterius8's situation when I find time, since it sounds like a very interesting case that may reveal some pathological dimension not yet explored in the discussion.

belladoreai:
> I'd recommend you move the tensor from (potentially) GPU to CPU outside the loop though, since it's incredibly expensive and you don't want to eat that cost for each token identifier.

My PR does not utilize the GPU to do anything. It uses the CPU to iterate on elements of a Python list. The transformation from Tensor to list occurs outside of the loop that iterates on elements of the list.

> From the two PRs you linked, it seems major work is going on to rewrite the implementation. Awesome. I was just looking for a surgical (minimal change possible) change to improve on the status quo.

I closed the first PR in favor of the second PR #6053 , which:

1. changes the data type from a Torch Tensor to a Python list (just like you change the Tensor to numpy array, no more complexity than that)
2. sets a maximum match length (which is a one line change)

That's not a "rewrite" of any kind, and it's no less "surgical" than your PR.

Also, there is no "major work going on", everything has been ready for merge quite a while.

> I'd recommend profiling steer you rather than honed engineering hunches ;) In my experience aside from overall orders of complexity, my own hunches as to where bottlenecks reside tend to be invalid more often than not. "Let the data guide you Luke".

Look, the main performance issue was that it's slow to iterate on elements of a Torch tensor. We both found the same issue! You fixed it by changing the data type to a numpy array, and I fixed it by changing the data type to a Python list. But we both found the same issue, and we both fixed it the same way (changing the data type).

Another performance issue that I found was that adversarial input can make the generator stuck, which is a problem for API use cases. I experimentally verified that this vulnerability exists and initially fixed it by optimizing the time complexity of the algorithm from O(n^2) to O(n). I then experimentally verified that my rewrite fixed the vulnerability and didn't change the results. This rewrite added complexity to the algorithm, so after back and forth with p-e-w we removed the algorithmic optimization, and we instead fixed the vulnerability with a simpler fix, by setting a max match length (which also fixes a separate, crash vulnerability on adversarial inputs).

I don't know why you're telling me to use a profiler next time, as if I had missed something?

jojje:
> My PR does not utilize the GPU to do anything. It uses the CPU to iterate on elements of a Python list. The transformation from Tensor to list occurs outside of the loop that iterates on elements of the list.

I specifically referred to [this](https://github.com/oobabooga/text-generation-webui/pull/6053/files#diff-51532f26c3cdbf11129835e199b8f4f621d2ec5968ce71760d149e251ff51526R238), which if the tensor resides on GPU (which it does in most cases on a GPU enabled system) the tolist operation will make an implicit GPU->RAM memory transfer, and _then_ convert the values to a python list. I didn't find any other code in that change that moved the tensor beforehand. If that's added is some other commit somewhere, then I simply missed it.

>I don't know why you're telling me to use a profiler next time, as if I had missed something?

It must have been a misunderstanding on my part from your earlier statement:

> In that PR we also discussed the possibility of using a numpy array, but Python list is supposed to be faster in this use case, so I didn't spend time trying out the numpy array and benchmarking results.

It read to me that the decision was based on assumptions instead of verified facts. Apologies if that was a misunderstanding of what you meant to convey.

Now as for how to proceed. It's pretty clear the focus should be on #6047 since it's further down the line, and has had effort already put into it by the community. There's no benefit to splitting the focus on a given problem. I'll close my PR since 6047 is further along and they have the same goal.

Just glad if we all can enjoy this new processor without avoidable tradeoffs.

belladoreai:
> I specifically referred to [this](https://github.com/oobabooga/text-generation-webui/pull/6053/files#diff-51532f26c3cdbf11129835e199b8f4f621d2ec5968ce71760d149e251ff51526R238), which if the tensor resides on GPU (which it does in most cases on a GPU enabled system) the tolist operation will make an implicit GPU->RAM memory transfer, and then convert the values to a python list.

Okay, now I understand better what you mean. So we're talking about this loop:

```
for input_ids_row, scores_row in zip(input_ids, scores):
            input_ids = input_ids_row.tolist()
```

As far as I can tell, that loop never loops more than 1 time. There is only ever one "input_ids_row". So we don't end up doing the transformation more than once even though we do it inside the loop. (Please correct me if I'm wrong.)

> It read to me that the decision was based on assumptions instead of verified facts. Apologies if that was a misunderstanding of what you meant to convey.

My assumption that iterating on a Python list would be faster than iterating on a numpy array is based on experiments done by other people, which I linked in the PR: https://github.com/numpy/numpy/issues/16985

I don't know if that counts as "verified facts". If you or someone else wants to benchmark the numpy array option against the Python list option, go ahead, let's use whichever option is faster. I just didn't personally invest time into benchmarking these options, because the improvements in #6053 already eliminate the performance impact of DRY almost completely (even at 20k token context, according to Hunter8's experiments which they reported in the PR). I'm sure we could find places to micro optimize further but at this point I would rather just say it's good enough, let's merge and work on something else next.

jojje:
> I just didn't personally invest time into benchmarking....

I said to profile. As you likely already know, profiling tells you what is _actually_ taking time and _where_. Benchmarking is a black-box testing of a specific subset of scenarios for a unit. Not comprehensive where-as the former is. My point was just that I didn't find _objective data_ to suggest the _right_ hotspots had been addressed. That's all.

> As far as I can tell, that loop never loops more than 1 time

Ah, you're right. The input dimension is (1, &lt;used-context-so-far>).
The use of zip for _that_ operation then seems to be a very convoluted way of expressing `squeeze` (dropping the first dimension) from each of the two tensors.

_That use of zip really should be added to the catalog of 'leet coding trivia._

belladoreai:
Yeah, I don't like that zip operation either, but that was there before my changes, and #6053 represents the minimal necessary changes to fix the issues, without any extra stuff like trying to refactor existing code (which I think is pretty readable aside from that zip operation anyway).

PR #5694, opened 2024-03-12
Full Chinese Localization

Touch-Night:
A complete Chinese Localization.  
This resolves internet issues for users from China who don't have vpn to install the dependencies. I switched every download source which is blocked by GFW to accessible mirror.  
Plus, I translated every text that is possible to be shown to the user and is suitable to translate. And I will update the translation when the main branch updates.   
I want to pr this to a new branch.

## Checklist:


oobabooga:
I would be unable to maintain the translated version.

Issue #5675, opened 2024-03-09
OpenAI API v1/models returns nothing that v1/internal/model/list does


TiagoTiago:
### Describe the bug

This is a bug, right?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Try it on http://127.0.0.1:5000/docs and compare the returned results; or access the API thru scripts or whatever.

The Continue VSCode/Codium extension for example is completely blind to what models I actually have when I try the auto-detect option; had to add things manually in it's config.json

### Logs

```shell
.
```

### System Info

```shell
.
```

egarma:
For me going to v1/models returns chatgpt3.5 even though I'm loading mixtral8x7b through exllamav2-hd loader

egSat:
For me, no matter what model is loaded that /v1/models always returns the following:

{"object":"list","data":[{"id":"gpt-3.5-turbo","object":"model","created":0,"owned_by":"user"},{"id":"text-embedding-ada-002","object":"model","created":0,"owned_by":"user"}]}

jonnysowards:
Looks like the controller is calling `response = OAImodels.list_dummy_models()` on line 148 of `extensions/openai/script.py` it should be calling `response = OAImodels.list_models()`. I am AFK at the moment but I should be able to open a PR to fix when I have a minute.

Ronny-apo:
Is that really an issue? Seems more like intentional behavior to me. IMHO it is intentional that the endpoint here mocks OpenAI's model response. There is an internal endpoint for a list of (internal) models, which was also mentioned here.

jonnysowards:
For true OpenAI api support it is necessary. When using continue.dev, it will query that endpoint and add those 2  dummy models which do not work. With this changed it grabs the real models and adds them to continue.dev ready to go. So I am not 100% sure it is necessary as it can be worked around, just not sure why you would include the endpoint at all if it all it does is include dummy models.

jonnysowards:
A reason it wouldn’t be an issue is if using an integration requires the OpenAI model names (the dummy ones) from that endpoint for the integration to work.

TiagoTiago:
Unless that has actually been fixed (or there's a legit reason for the apparent misbehavior of the API), that bot has no business being here; can that useless thing be disabled please?

NoxWorld2660:
I just did a pull request with the fix suggested by jonnyowards

ErraticFox:
[...]

PR #5720, opened 2024-03-18
OpenAI extension list currently loaded models not dummy models.

jonnysowards:
Fixes [bug #5675](https://github.com/oobabooga/text-generation-webui/issues/5675) that lists dummy models instead of loaded models in OpenAI extension. `v1/models`.

## Checklist:

### DIFF
--- a/extensions/openai/models.py
+++ b/extensions/openai/models.py
@@ -14,7 +14,15 @@ def get_current_model_info():
 
 
 def list_models():
-    return {'model_names': get_available_models()[1:]}
+    result = {
+        "object": "list",
+        "data": []
+    }
+    # get models and add them to the result
+    for model in get_available_models():
+        result["data"].append(model_info_dict(model))
+
+    return {'model_names': result}
 
 
 def list_dummy_models():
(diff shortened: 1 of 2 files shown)


Ph0rk0z:
What will this do to tools that expect openAI models or choose them?

jonnysowards:
@Ph0rk0z That is one of my concerns. When using integrations like continue.dev the proper models do not load via the open ai api endpoints, only dummy ones that do not work. Although if another integration is required to have to have the open ai models to verify its working with open ai then this could cause it to fail. I'm wondering if potentially a flag could denote whether or not to use dummy models.

Ph0rk0z:
Why not just add gpt4/3.5 fake model strings in addition to the actual models and make those not switch the current model? Kill two birds with one stone in a way.

jonnysowards:
@Ph0rk0z That sounds like the best of both worlds, and  easy to implement. I will update the pull to do that when I get a minute.

danny-avila:
Just commenting here. @oobabooga this should be merged, very minor and easy fix resolving a modest pain point.

PR #5722, opened 2024-03-18
Set an default empty string for `user_bio` to fix #5717 issue.

### LINKED ISSUE #5717: `user_bio` is None by default, get an error when replacing the character names
### Describe the bug
Using API `/v1/chat/completions` (None stream mode) without `user_bio`.

----

The new parameter `user_bio` in API chat mode raises an error because it's `None` as default.
https://github.com/oobabooga/text-generation-webui/blob/7cf1402bde48fd76af501d5efecb34227bf4d082/extensions/openai/typing.py#L110

----

Then, in `chat.py` can't replace the names correctly.
https://github.com/oobabooga/text-generation-webui/blob/7cf1402bde48fd76af501d5efecb34227bf4d082/modules/chat.py#L97

get this error
image

-----

An empty string as default in webui.
https://github.com/oobabooga/text-generation-webui/blob/7cf1402bde48fd76af501d5efecb34227bf4d082/modules/shared.py#L60

### Reproduction

Any request in `/v1/chat/completions`

### Screenshot

_No response_

### Logs

```shell
    text = text.replace('{{user}}', name1).replace('{{char}}', name2)
AttributeError: 'NoneType' object has no attribute 'replace'
```

### System Info

```shell
None
```

Yiximail:
quick fix for #5717 

the other parameters, like `bot_name`, `context` etc. used for overwrites the character card
so keep them None

## Checklist:

### DIFF
--- a/extensions/openai/typing.py
+++ b/extensions/openai/typing.py
@@ -107,7 +107,7 @@ class ChatCompletionRequestParams(BaseModel):
     context: str | None = Field(default=None, description="Overwrites the value set by character field.")
     greeting: str | None = Field(default=None, description="Overwrites the value set by character field.")
     user_name: str | None = Field(default=None, description="Your name (the user). By default, it's \"You\".", alias="name1")
-    user_bio: str | None = Field(default=None, description="The user description/personality.")
+    user_bio: str | None = Field(default='', description="The user description/personality.")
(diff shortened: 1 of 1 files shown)


HariboApfel (commented):
fixes the issue for me

PR #5781, opened 2024-04-01
update Miniconda3 version

beat4ocean:
## Checklist:


Touch-Night:
But why do we need to update it?

oobabooga:
My question as well, any change in particular that justifies the update? I prefer to keep the installer as untouched as possible to avoid unexpected issues.

Touch-Night:
> My question as well, any change in particular that justifies the update? I prefer to keep the installer as untouched as possible to avoid unexpected issues.

**Update gradio to 4.24 asap btw**

PR #5807, opened 2024-04-04
Remove CTransformers support

oobabooga:
The extra models that it supports relative to llama.cpp are obsolete (like GPT-J), and it hasn't received an update in 7 months.


TNTOutburst:
Darn. For me sometimes llama.cpp has issues with not being deterministic or it just acts strange, but ctransformers always worked for the models it supported.

oobabooga:
I recommend trying llamacpp_HF if you experience sampling issues with the base llama.cpp loader. I personally use llamacpp_HF exclusively.

PR #5836, opened 2024-04-09
Add a simple min_p preset, make it the default

oobabooga:
In my testing, it seems to work better than simple-1 for those recent models that finally rival ChatGPT (like Command R).


randoentity:
Thank you! This got rid of the command r+ repetition problems by setting the repetition penalty to 1. Furthermore, using min_p instead of top_p makes it nondeterministic.

PR #5842, opened 2024-04-10
Respect model and lora directory settings when downloading files

acon96:
## Checklist:

Currently all downloaded files are placed in the same directory as the installation. This change makes them download to the folder specified by the user.


oobabooga:
Nice, thanks.

PR #5857, opened 2024-04-14
Fix gguf multipart file loading

Tisjwlf:
## Checklist:

Without sorting by name an error happens for me where it tries to load the wrong file, in my case file 3 instead of file 1.


oobabooga:
Thanks

PR #5865, opened 2024-04-16
Let grammar escape backslashes

altoiddealer:
## Checklist:

I was trying to develop a grammar string to respond in this format: 

(path)"\\"(name).safetensors [(hash)]

Here is what I tried...

```
  grammar_string: |-
    root ::= path? name ".safetensors" " [" hash "]"
    path ::= (folder "\\")*
    folder ::= [a-zA-Z0-9_]+
    name ::= [a-zA-Z0-9_]+
    hash ::= HEX HEX HEX HEX HEX HEX HEX HEX HEX HEX
    HEX ::= [0-9a-fA-F]
```

`error: unknown escape`

I then tried this:

`    path ::= (folder "\")*`

`error: out of range`

Also this:

`    path ::= (folder "\\\\")*`

`error: unknown escape`

It turns out, the parser was just incapable of escaping backslashes.

Simply adding this condition to the parser code resolves it:

```
        elif esc == "\\":
            return "\\", src[2:]
```


oobabooga:
Thanks for the fix

Issue #5851, opened 2024-04-12
Inference failing using ExLlamav2 version 0.0.18


ashleykleynhans:
### Describe the bug

Since ExLlamav2 was upgraded to version 0.0.18 in the requirements.txt, inference using it is no longer working and fails with the error in the logs below.  Reverting to version 0.0.17 resolves the issue.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Install latest main branch (current commit is `26d822f64f2a029306b250b69dc58468662a4fc6`)
2. Download `GPTQ` model
3. Use `ExLlamav2_HF` model loader
4. Go to `Chat` tab and ask the AI a question.
5. Observe error, even though the model loaded successfully.

### Logs

```shell
21:35:11-061459 INFO     Loading "TheBloke_dolphin-2.6-mistral-7B-GPTQ"
21:35:13-842112 INFO     LOADER: "ExLlamav2"
21:35:13-843422 INFO     TRUNCATION LENGTH: 32768
21:35:13-844234 INFO     INSTRUCTION TEMPLATE: "Alpaca"
21:35:13-845014 INFO     Loaded the model in 2.78 seconds.
Traceback (most recent call last):
  File "/workspace/text-generation-webui/modules/text_generation.py", line 429, in generate_reply_custom
    for reply in shared.model.generate_with_streaming(question, state):
  File "/workspace/text-generation-webui/modules/exllamav2.py", line 140, in generate_with_streaming
    self.generator.begin_stream(ids, settings, loras=self.loras)
  File "/workspace/venvs/text-generation-webui/lib/python3.10/site-packages/exllamav2/generator/streaming.py", line 198, in begin_stream
    self.begin_stream_ex(input_ids,
  File "/workspace/venvs/text-generation-webui/lib/python3.10/site-packages/exllamav2/generator/streaming.py", line 296, in begin_stream_ex
    self._gen_begin_reuse(input_ids, gen_settings)
  File "/workspace/venvs/text-generation-webui/lib/python3.10/site-packages/exllamav2/generator/streaming.py", line 624, in _gen_begin_reuse
    self._gen_begin(in_tokens, gen_settings)
  File "/workspace/venvs/text-generation-webui/lib/python3.10/site-packages/exllamav2/generator/streaming.py", line 586, in _gen_begin
    self.model.forward(self.sequence_ids[:, :-1],
[...]

Issue #5693, opened 2024-03-12
Requirements downloaded from URLs are re-downloaded on every update


p-e-w:
### Describe the bug

Pip normally caches packages, but it appears that this doesn't happen (at least not by default) for dependencies specified as URLs, which text-generation-webui has quite a few of. As a result, lots of dependencies are re-downloaded on each update even if they haven't changed, wasting time and bandwidth.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Run the update script and wait for it to finish.
2. Run the update script again.
3. Notice that all URL dependencies are re-downloaded, adding to dozens if not hundreds of megabytes.

### Logs

```shell
N/A
```

### System Info

```shell
N/A
```

PR #5873, opened 2024-04-18
[Hotfix] Revert sse-starlette version bump because it breaks API request cancellation

### LINKED ISSUE #5871: API Completion stream abort broken
### Describe the bug

Since I pulled the last snapshot (I think) stopping/aborting a completion stream (from the API) is weirdly broken.
With llama.cpp it just stopped working completely. Even though the stream is aborted/stopped the generation continues. 
With exllamav2 the completion stops but afterwards I am unable to generate any other text either using the API or the Gradio GUI

When using the GUI everything works fine. I'm unsure if its a problem in the code or related to package changes, because I tried going back to previous snapshots up to `snapshot-2024-03-24` and the behavior did not change. Unfortunately I didn't have time to also try reinstalling dependencies to match the snapshot (AMD always takes a bit longer xD).
It is also possible the problem is on my side but I have no Idea what could cause that. I even tried different applications which use the API to check if its caused on the application side.

### Reproduction

1. Run WebUI with API
2. Load model with llama.cpp (1) or exllama (2)
3. Trigger text generation using the API
4. Stop the text generation by aborting the stream
5. (2) Trigger text generation using the API or GUI

### Screenshot

_No response_

### Logs

```shell
<none>
```

### System Info

```shell
Operating System: Manjaro Linux 
KDE Plasma Version: 5.27.11
KDE Frameworks Version: 5.115.0
Qt Version: 5.15.12
Kernel Version: 6.6.26-1-MANJARO (64-bit)
Graphics Platform: X11
Processors: 20 × 13th Gen Intel® Core™ i5-13500
Memory: 31.1 GiB of RAM
Graphics Processor: AMD Radeon RX 7900 XTX
Manufacturer: Micro-Star International Co., Ltd.
Product Name: MS-7D98
System Version: 1.0
```

p-e-w:
Fixes #5871.

The major-version bump of sse-starlette breaks API request cancellation, that is, cancelling an API request no longer terminates the generation process. This is a showstopper because it leads to out-of-control generations that only end when they encounter an EOS token or `max_new_tokens` is reached. In practice, this breaks most standard interactive use cases.

Identifying the root cause of this problem was complicated by several other issues:

* Unlike all other extensions, the OpenAI API extension does not have its own requirements file (why?). I looked at the `extensions/openai` folder and found no recent changes. This led me to wrongly conclude that the problem lies elsewhere, because the sse-starlette dependency is specified in the root `requirements.txt`.
* ExLlamav2 is currently broken (#5851), so I wasn't able to easily exclude llama.cpp as the culprit.
* I'm on a slow connection so I'm unable to bisect economically, because each dependency change requires hundreds of Megabytes of Pip downloads (#5693).

I was left with the assumption that llama.cpp must be the problem, considering the recent unrelated issues there, which led me on a wild goose chase that went nowhere. But eventually, I figured out the problem with the updated dependency.

The recent updates have been rather bumpy. In the past 2 weeks, llama.cpp, ExLlamav2, and the API were all broken. Perhaps a slightly slower release cadence might make sense.

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ wandb
 # API
 SpeechRecognition==3.10.0
 flask_cloudflared==0.0.14
-sse-starlette==2.1.0
+sse-starlette==1.6.5
 tiktoken
 
 # llama-cpp-python (CPU only, AVX2)
--- a/requirements_amd.txt
+++ b/requirements_amd.txt
@@ -28,7 +28,7 @@ wandb
 # API
 SpeechRecognition==3.10.0
 flask_cloudflared==0.0.14
-sse-starlette==2.1.0
+sse-starlette==1.6.5
 tiktoken
 
 # llama-cpp-python (CPU only, AVX2)
--- a/requirements_amd_noavx2.txt
+++ b/requirements_amd_noavx2.txt
@@ -28,7 +28,7 @@ wandb
 # API
 SpeechRecognition==3.10.0
 flask_cloudflared==0.0.14
-sse-starlette==2.1.0
+sse-starlette==1.6.5
 tiktoken
 
 # llama-cpp-python (CPU only, no AVX2)
--- a/requirements_apple_intel.txt
+++ b/requirements_apple_intel.txt
@@ -28,7 +28,7 @@ wandb
 # API
 SpeechRecognition==3.10.0
 flask_cloudflared==0.0.14
-sse-starlette==2.1.0
+sse-starlette==1.6.5
 tiktoken
 
 # Mac wheels
--- a/requirements_apple_silicon.txt
+++ b/requirements_apple_silicon.txt
@@ -28,7 +28,7 @@ wandb
 # API
 SpeechRecognition==3.10.0
 flask_cloudflared==0.0.14
-sse-starlette==2.1.0
+sse-starlette==1.6.5
 tiktoken
 
 # Mac wheels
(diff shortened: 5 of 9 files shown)


lufixSch:
This fixed the problem for me too.

> The recent updates have been rather bumpy. In the past 2 weeks, llama.cpp, ExLlamav2, and the API were all broken. Perhaps a slightly slower release cadence might make sense.

At least for me that is not a problem. I don't expect a stable release when I install a version called "snapshot". I prefer having fast releases so I can try the new stuff as soon as possible and I'm okay if I have to work around some bugs in return. Maybe there could be more frequent stableish releases again (e.g. 1.8, 1.9) for people which don't always want the latest stuff but a more stable environment.
Nevertheless I feel like this issue could have easily been prevented by testing the API after a major version upgrade of one of its dependencies.

ashleykleynhans:
> The recent updates have been rather bumpy. In the past 2 weeks, llama.cpp, ExLlamav2, and the API were all broken. Perhaps a slightly slower release cadence might make sense.

I couldn't agree more.  I also think that it would be better to have proper versioned releases like A1111 instead of the snapshots, and also RC releases for users to provide feedback before publishing an official release.

Users on RunPod are often bitten by updates to the `main` branch where things are get broken regularly.  It has been stable for a while, but last year there were many updates that caused issues, and as stated above, there are a bunch of different issues again.

PR #5878, opened 2024-04-19
Revert walrus operator for params['max_memory']

Column01:
Using a walrus operator leaves `params['max_memory']` undefined when the `get_max_memory_dict()` returns None. This leads to a KeyError when checking it later on when `load-in-8bit` mode is checked. This commit reverts a change made in commit 6b7f7555f that caused this bug.

For reference, here is the traceback that alerted me to this bug:

```
21:31:45-049883 ERROR    Failed to load the model.
  File "C:\Users\colin\Documents\_GithubRepos\text-generation-webui\modules\ui_model_menu.py", line 248, in load_model_wrapper
    shared.model, shared.tokenizer = load_model(selected_model, loader)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\text-generation-webui\modules\models.py", line 94, in load_model
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\text-generation-webui\modules\models.py", line 218, in huggingface_loader
    if params['max_memory'] is not None:
       ~~~~~~^^^^^^^^^^^^^^
KeyError: 'max_memory'
```

## Checklist:


Column01:
Alternatively, the spot where the keyerror is being thrown could be changed to `params.get('max_memory')` if that is preferred

oobabooga:
> There are other issues with this commit actually, every param that was hidden from logging by only assigning it if it's needed or not none means any check or reference later needs to be changed to `.get()` or the easier solution of reverting the commit

I would prefer to use .get() syntax on lines that may cause problems. The intent of that commit was to keep the log message as clean as possible by not showing null parameters.

Column01:
Okay, I'll try to implement this tomorrow when I get the chance, its late here for me :) thanks for your reply

Column01:
Should be done now if you'd like to review.
image

Column01:
oh it seems the change I initially to max memory didn't revert. Let me fix that

Column01:
All good now afaik

oobabooga:
It looks good. Thanks a lot for these fixes.

PR #5899, opened 2024-04-21
Added ChatML-format.json in formats, since people are still puzzled

FartyPants:
People are still puzzled by the "formats" and don't create their own, so here is a quick ChatML format added for training.
Uses empty system prompt. (Seems to work a tad better than no system prompt)

## Checklist:


oobabooga:
Ideally, I would like to remove this "formats" folder and reuse the instruction templates under `instruction_templates` for training. 

For instance, to continue fine-tuning llama-3 chat using the exact sample template.

If you can come up with a way to do that, that would be very helpful.

PR #5935, opened 2024-04-25
Add a quantization guide 

Vasyanator:
## Checklist:
Added a tutorial on how to convert a model to GGUF and quantize it for Windows and Linux.
- [✅] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

### DIFF
--- a/docs/Model Quantization.md
+++ b/docs/Model Quantization.md
@@ -1,60 +1,60 @@
 # GGUF
-### Требования:
+### Requirements:
 - Python 3.10/3.11
 - Git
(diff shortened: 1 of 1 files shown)


oobabooga:
This guide feels out of scope for this project. It would be better if you could incorporate it directly in the llama.cpp documentation, which can probably be made easier to understand in many parts.

I wouldn't be able to maintain this documentation since I don't often quantize models myself.

PR #5938, opened 2024-04-25
fix: AutoAWQ from_quantized to from_pretrained

PedroDelano:
This works to solve the following issue:

`AttributeError: 'LlamaLikeModel' object has no attribute 'layers'`

When trying to make inference on some of TheBloke AWQ models.

### DIFF
--- a/modules/models.py
+++ b/modules/models.py
@@ -293,14 +293,12 @@ def AutoAWQ_loader(model_name):
 
     model_dir = Path(f'{shared.args.model_dir}/{model_name}')
 
-    model = AutoAWQForCausalLM.from_quantized(
-        quant_path=model_dir,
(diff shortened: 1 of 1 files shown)


datar001:
When I conduct the inference on Llama3 AWQ model, i meet the error:
AttributeError: 'LlamaLikeModel' object has no attribute 'layers'
How to sove the problem? Thanks

oobabooga:
I had this same error before, but it doesn't seem to happen anymore after the update in https://github.com/oobabooga/text-generation-webui/pull/6025, so I am closing the PR. If the error persists please let me know.

jllrr:
Still having the above issue where certain Bloke AWQ models have the attribute error. Any fixes?

PR #5951, opened 2024-04-28
Fix async events for OpenAI API extension

Artificiangel:
## Checklist:

## Changes 
- OpenAI API extension now allows different tasks to run in parallel without locking up the webserver.
	- Run embedding functions while generating text (Originally only worked with streaming).
	- Send a request to StableDiffusion-WebUI while loading a text generation model.
	- Limit heavy tasks (split by category) to run in queue.
	- Tested all modified endpoints (Except `audio/transcriptions`).
	The audio transcription endpoint didn't have enough documentation to go off of.
	I was unable to get it working with an aiohttp client yet.

- When throwing an error: Fixed `model/load` and `lora/load` endpoints returning errors with a `200` status code.
Being unexpected, that might not be caught on client side.
- For consistency added `return "OK"` to `model/unload` endpoint, matching `lora/unload`.
- Removed awaiting of sync `OAIimages.generations` function in the `images/generations` endpoint that caused an error.


Artificiangel:
I think I made a mistake by committing to the main branch on my fork then merging to dev.

Issue #5975, opened 2024-05-03
Cant start server, it gives cryptic error


Bedoshady:
### Describe the bug

I tried installing server and running it which worked at first, but couldnt load my llama 3 models, I typed this command
`conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch -c nvidia` then it stopped working, I tried different solutions but it didnt work

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Install web text ui normally via miniconda then do this command `conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch -c nvidia`

### Logs

```shell
(tg) H:\Downloads\text-generation-webui-main>python server.py
binary_path: H:\tools\MiniConda2\envs\tg\lib\site-packages\bitsandbytes\cuda_setup\libbitsandbytes_cuda116.dll
04:53:42-696449 INFO     Starting Text generation web UI

Running on local URL:  http://127.0.0.1:7860

ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 411, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 69, in __call__
    return await self.app(scope, receive, send)
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\fastapi\applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\starlette\applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\starlette\middleware\errors.py", line 186, in __call__
    raise exc
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\starlette\middleware\errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\gradio\route_utils.py", line 689, in __call__
    await self.app(scope, receive, send)
  File "H:\tools\MiniConda2\envs\tg\lib\site-packages\starlette\middleware\exceptions.py", line 65, in __call__
[...]

PR #5976, opened 2024-05-03
Update block_requests.py to resolve unexpected type error (500 error) after updating to gradio 4.28.3

nero-dv:
Resolves https://github.com/oobabooga/text-generation-webui/issues/5975#issue-2276774815

## Checklist:

In the original function, file_contents is a bytes object because the file is opened as a binary file and the replace method replaces bytes objects. In the second function, the file_contents is a string object because the file is opened as a text file and the replace method replaces string objects. This is done because when updating the gradio version to the latest version, the following error appears which references the block_requests.py file. 

```log
TypeError: replace() argument 1 must be str, not bytes
```

Additionally a specific iframe script from the content of the file is being removed. This line is commented out in the second function with the comment saying there is no need for it as this effectively renders the request null

```python
file_contents = file_contents.replace('cdnjs.cloudflare.com', b'127.0.0.1') 
```


nickpotafiy:
I upgraded to Gradio `4.28.3` and `4.29` and cannot reproduce the error. I don't think it's an issue with newer gradio, might be an issue with older gradio being out of sync with text-generation-webui.

nero-dv:
> I upgraded to Gradio `4.28.3` and `4.29` and cannot reproduce the error. I don't think it's an issue with newer gradio, might be an issue with older gradio being out of sync with text-generation-webui.

Could you share which commit you're on so I can replicate?

nickpotafiy:
I tested with the [most recent one](https://github.com/oobabooga/text-generation-webui/commit/8f12fb028dff4e133460fe10ef49d3f90167b313). If I downgrade gradio to `4.25` i also get an error. `4.26` (required version), `4.28`, `4.29` all work just fine with this commit. I think the issue happens when someone pulls the latest version of `webui` without using the update wizard. This is not a flaw with `webui` but how it's updated.

nero-dv:
I'm concerned about the lack of transparency in package modifications during installation. It's unclear when edits are made and users shouldn't be surprised by changes they didn't consent to. 

Without clear information on what changes were made, debugging issues becomes much harder. Developers will struggle to reproduce problems or identify the root cause of issues. also unannounced modifications can introduce unknown security risks.

nickpotafiy:
The installation script downloads the necessary packages outlined in various `requirements` files and installs under `installer_files`. If you simply let webui install everything for you, there won't be any issues. However, if you `git pull` from latest webui but your `installer_files` are outdated, you might get errors because the project requirements are out of sync. This is why update wizard exists. So long as users understand how to install and update webui, these types of issues should not really exist.

oobabooga:
I'm closing this because this error does not happen with the gradio version in requirements.txt. Gradio changes a lot over time and the project is only guaranteed to work with the specified version.

oobabooga:
Happened to me now (randomly). Probably some sub-requirement causes this.

PR #6025, opened 2024-05-19
Backend cleanup

oobabooga:
* Bump AutoAWQ, AutoGPTQ, HQQ, AQLM and transformers to the latest versions.
* Remove QuIP#. Last commit was 3 months ago, no llama-3 QuIP# models on HF, no precompiled wheels for the library after months. AQLM performs similarly and is more active.
* Remove GPTQ-for-LLaMa.
    * It doesn't work with the current PyTorch version and I didn't manage to recompile it with GitHub Actions easily. 
    * It was kept for compatibility with old NVIDIA GPUs, but I don't think anyone still uses it.
* Make AutoGPTQ functional again my removing the inject_fused_attention option (apparently it has a bug).
* HQQ is not functional (I couldn't load `mobiuslabsgmbh_Llama-2-7b-chat-hf_1bitgs8_hqq`, which is the most popular HQQ model on HF), but I'm keeping the loader for now because it's in active development and promising.


Ph0rk0z:
Any big reason to dump these? I mean people probably have models in those formats. One day they will pull and then randomly lose the ability to run them. It doesn't seem there's any code that has made these backends a hassle yet, besides existing.

>I don't think anyone still uses it.

Not a lot of maxwell people, but IIRC, it was the only way for them.

oobabooga:
> Any big reason to dump these?

Keeping the repository/documentation clean, not having to fix an obsolete backend, not forcing people to download the wheels for each `pip install -r requirements.txt`, not having an untested backend (does it even work with Llama-3?).

I'll remove it since I think it only serves an imaginary type of user. People with old GPUs are running llama.cpp nowadays, not GPTQ-for-LLaMa.

GlobalMeltdown:
I used GPTQ and prefer it over GGUF because performance wise on my 3070ti is much better. It's a shame TheBloke was mainly the only good source for GPTQ. But honestly why would I want to to play with balancing GGUF's when I can just one and done it (either it fits or it doesnt) with GPTQ and inference speed and response is great compared to similar gguf's.

So I hope your joking about removing its functionality.

PR #6043, opened 2024-05-22
Fix stopping strings for llama-3 and phi

oobabooga:
People who train models are very creative in trying to make their jinja templates as hard to work with as possible. This is an attempt at getting the stopping strings right.


BadisG:
The chat doesn't work anymore, the model doesn't want to output anything. I reverted back to the latest dev commit 8aaa0a6f and it works now.

oobabooga:
Yeah that wasn't a good solution. I have reverted it and am trying to think of a better way to reverse-engineer those jinja templates. https://github.com/oobabooga/text-generation-webui/commit/ad54d524f7586d2fa8ab33bad46d0a4b108aa822

PR #6053, opened 2024-05-25
DRY sampler improvements

belladoreai:
I was asked by @p-e-w to split some of the changes from #6047 into a separate PR here.

This PR contains the data type performance improvement for DRY, and a minor fix to prevent crash on large repetitive inputs.

Edit: now also contains change to cap the max match length to 50

See the main PR for more info.

## Checklist:


modules/sampler_hijack.py
```diff
@@ -234,21 +234,28 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
             input_ids = input_ids[:, -self._range:]
 
         for input_ids_row, scores_row in zip(input_ids, scores):
+
+            # Use normal Python data types for improved performance
+            s = input_ids_row.tolist()
```
p-e-w, line 239: Please use descriptive identifiers like the rest of the function rather than single letters.

modules/sampler_hijack.py
```diff
@@ -234,21 +234,28 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
 
             if last_token in self.sequence_breakers:
                 continue
 
             # Exclude the last token as it always matches.
-            match_indices = (input_ids_row[:-1] == last_token).nonzero()
+            match_indices = []
+            for idx, val in enumerate(s[:-1]):
+                if val == last_token:
+                    match_indices.append(idx)
```
p-e-w, line 251: Here's an idea: What about instead of converting the tensor to a Python list, you convert it to a `np.array` (NumPy is already a dependency)?

This would give you access to `np.argwhere` etc., which might be quite a bit faster than a Python loop, while *also* being simpler and cleaner. And I doubt that NumPy has the same indexing overhead that Torch seems to have.

modules/sampler_hijack.py
```diff
@@ -257,15 +264,15 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
                 # so the match is at least of length 1.
                 match_length = 1
 
-                # Extend the match backwards as far as possible.
-                while True:
+                # Extend the match backwards (at most to 1000 to prevent exponent overflow at penalty calculation) (this cap also improves performance on worst case)
+                while match_length < 1000:
```
p-e-w, line 268: The cap can be **way** lower than 1000. Even 50 should be more than enough. The default `dry_base` is 1.75, and 1.75^50 > 10^12, which is a penalty that makes the token probability equal to zero for all practical purposes.

At which point I guess the Z algorithm becomes unnecessary, because the function is no longer quadratic even in the worst case, since the inner loop is bounded by a constant.

modules/sampler_hijack.py
belladoreai: It seems that there are many different ways to iterate over elements in a NumPy array, and all of them are slower than iterating elements on a Python list: https://github.com/numpy/numpy/issues/16985

modules/sampler_hijack.py
belladoreai: Correct

belladoreai: I'd say let's merge this PR and close the other one?

modules/sampler_hijack.py
p-e-w: Thanks for the info. Guess I have to revisit my simplistic assumption "if you want speed, use NumPy". Pretty crazy that this hasn't been optimized yet.

modules/sampler_hijack.py
p-e-w: Yes, that's the most pragmatic solution IMO. I'd file the Z algorithm under "it's cool that it's possible, but the juice isn't quite worth the squeeze in this case".

modules/sampler_hijack.py
```diff
@@ -234,21 +234,28 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
             input_ids = input_ids[:, -self._range:]
 
         for input_ids_row, scores_row in zip(input_ids, scores):
+
+            # Use normal Python data types for improved performance
+            input_ids = input_ids_row.tolist()
+            
             # Raw integer must be extracted here to check for set membership.
```
p-e-w, line 241: This comment doesn't make sense anymore since the list elements are already raw integers.

modules/sampler_hijack.py
```diff
@@ -234,21 +234,28 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
             input_ids = input_ids[:, -self._range:]
 
         for input_ids_row, scores_row in zip(input_ids, scores):
+
```
p-e-w, line 237: :fire: empty line

p-e-w (approved):
LGTM modulo style nits.

belladoreai:
@oobabooga Ready for merge

p-e-w (approved):
Looks good now! Probably makes sense to merge this before DRY is merged into master.

jojje:
LGTM.  

Seems @p-e-w is also good with it. What's your view @oobabooga ?

It is an improvement over the original. _If_ further optimization turns out to be necessary, such can be done in separate PRs.

Vhallo:
Good to see the performance issues being solved. Seems like it might be worthwhile to integrate this into Exllamav2 / TabbyAPI now as well?

oobabooga:
Thanks for the reviews, merging now before merging DRY to the main branch.

PR #6083, opened 2024-06-02
Fix extensions/openai/completions.py to enable system prompt through API

Koesn:
This will ensure the last message is added. It also enable system prompt to be processed. Now OAI API compatible can user their system prompt.

## Checklist:

- [√ ] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -207,6 +207,9 @@ def convert_history(history):
         elif role == "system":
             system_message = content
 
+    if current_message:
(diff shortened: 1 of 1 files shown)


belladoreai:
I thought all special tokens can already be passed in the API request to the completions endpoint, including the special tokens denoting system prompt. Could you give an example of an input/output pair where the system prompt is not correctly processed without this PR?

FartyPants:
This has nothing to do with system prompt, system prompt is not part of the history where the chat_dialogue ends up.

PR #6087, opened 2024-06-03
Speed-up the DRY logits processor

jojje:
## Checklist:

- [V] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

Note, I didn't find any unit-tests in this repo, so created my own to ensure this change does not alter the behavior (output) in any way. You can find the test (both assertion testing and benchmarking in one) [here](https://github.com/jojje/text-generation-webui/blob/exp/dry_speedup/experiment/test.py).


jojje:
After integrated profiling into text-ui, it seems this solution doesn't gain the advantage that stand-alone testing showed. It's basically trading slow pytorch item lookups for slow pytorch -> numpy array conversion with almost no win (single percentage speedup only). Not worth considering. Closing the PR.

belladoreai:
For reference:
- Original discussion of the performance issues: #5677
- My earlier PR targeting the same performance issues: #6053
- My other, earlier PR targeting the same performance issues: #6047

jojje:
As per the conversation in #5677, I'll close this one so as not to split the focus unnecessarily. #6047 seems to be where we're already headed for this specific issue, so let's focus on that one instead.

PS. thanks for taking a look at it though.

PR #6094, opened 2024-06-04
Fix issue #6070

### LINKED ISSUE #6070: flash-attn and flash_attn both in shared.args
### Describe the bug

In shared.args, there are both flash-attn and flash_attn, and flash-attn is the value set in the ui, while flash_attn is False forever.  

This can result users cannot enable flash-attn through ui.  

To fix it, use flash_attn in modules/loaders.py, modules/ui.py and use shared.gradio['flash_attn'] in modules/ui_model_menu.py

### Reproduction

Use a debugger and load a gguf model with flash-attn checked, and see shared.args

### Screenshot

image
image
*shared.args.flash-attn = True and shared.args.flash_attn = False when flash-attn is checked in the ui*  

### Logs

```shell
Not necessary
```

### System Info

```shell
Not necessary
```

Touch-Night:
## Checklist:

Fix #6070


Touch-Night:
Close this as duplicated

Issue #6106, opened 2024-06-07
Trying to get logits with samplers causes a deadlock


theo77186:
### Describe the bug

The title describes the bug good enough. This bug only affects the dev branch for now. The deadlock is caused by a double lock acquisition.
In the `get_next_logits` function, the generation lock is acquired. https://github.com/oobabooga/text-generation-webui/blob/a363cdfca117e92967d9a25509146f02fbf63da1/modules/logits.py#L19
This function calls `_get_next_logits`, which calls `generate_reply`. https://github.com/oobabooga/text-generation-webui/blob/a363cdfca117e92967d9a25509146f02fbf63da1/modules/logits.py#L49
And this function tries to acquire again the lock:
https://github.com/oobabooga/text-generation-webui/blob/abe5ddc8833206381c43b002e95788d4cca0893a/modules/text_generation.py#L34
which fails as the lock got acquired earlier by `get_next_logits` with no possibility to release the lock.
My local version has a workaround by releasing the lock around `generate_reply` but it's a hacky solution and I'm thinking about a proper fix.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

With any model loaded with any loader but llama.cpp/Exllamav2 non-HF, in Default/Notebook tab, click on "Get next token probability" with "Use samplers" checked. After clicking it, any further generation is prevented as the generation lock will never get released.

### Logs

```shell
No error is shown, it's a deadlock.
```

### System Info

```shell
OS: Debian sid/unstable
```

theo77186:
#6107 merged in main, this should be fixed

oobabooga:
After those changes, I found that if I cancelled an ongoing generation in the UI and tried to start a new one later, I would get an error about "RuntimeError: cannot release un-acquired lock". I have reverted the PR for now. It would be better to use a simple lock, as I don't know what runs in what thread in gradio/fastapi/extensions.

PR #6107, opened 2024-06-07
Use reentrant generation lock

theo77186:
The function get_next_logits acquires the generation lock, then calls generate_reply when use_samplers is True, trying to acquire the generation lock which fails and cause a deadlock because the lock is already acquired. This replaces the standard lock with a reentrant lock, making the 2nd lock acquisition by the same thread succeed.

Fixes: https://github.com/oobabooga/text-generation-webui/issues/6106

## Checklist:

### DIFF
--- a/server.py
+++ b/server.py
@@ -32,7 +32,7 @@ import sys
 import time
 from functools import partial
 from pathlib import Path
-from threading import Lock, Thread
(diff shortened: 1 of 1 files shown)


oobabooga:
Nice!

PR #6111, opened 2024-06-09
Setting default alpha_value and fixing loading some newer DeepSeekCoder GGUFs

mefich:
1. Ensures alpha_value resets back to 1 after selecting a new model. Previously if any model had a unusual alpha_value in saved settings or of it was changed manually, this value wouldn't reset after choosing another model. This could cause poor performance of a newly loaded model.
2. Ensures rope scaling is loaded from metadata of newer GGUF models, notably from newer DeepSeekCoder finetunes where this value is now saved in rope.scaling.factor instead of rope.scale_linear where it used to be saved in old GGUFs.  Though there might be needed additional check for rope.scaling.type: linear or yarn. But I couldn't find information how to properly load Yarn models in TGW to decide if it's needed to check the scaling type.

## Checklist:

### DIFF
--- a/modules/models_settings.py
+++ b/modules/models_settings.py
@@ -16,6 +16,7 @@ def get_fallback_settings():
         'n_ctx': 2048,
         'rope_freq_base': 0,
         'compress_pos_emb': 1,
+        'alpha_value': 1,
         'truncation_length': shared.settings['truncation_length'],
         'skip_special_tokens': shared.settings['skip_special_tokens'],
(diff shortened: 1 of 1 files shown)


oobabooga:
Your changes are correct and I wasn't aware that rope.scale_linear had been renamed to rope.scaling.factor. Nice catch.

About scaling type, any metadata not explicitly passed while loading a model gets automatically detected by llama.cpp, so this is not an issue.

PR #6188, opened 2024-07-01
Add eager attention optionals for transfrmers loader to make Gemma-2 model effectively

GralchemOz:
## Checklist:

As indicated by [the developer of Gemme-2](https://huggingface.co/google/gemma-2-27b-it/discussions/14), the utilization of eager attention is advisable. My own tests have confirmed that employing eager attention resolves the performance issues encountered with the gemma-2-27b-it models when loaded with the transformer.


oobabooga:
Nice finding, this increased the performance of both 9b-it and 27b-it models [on my benchmark](https://oobabooga.github.io/benchmark.html). The 27b-it used to give 15/48, now it gives 33/48. 9b-it went from 27/48 to 29/48. It seems like gemma works correctly after this change.

I tested with `--bf16 --use_eager_attention`.

Issue #6181, opened 2024-06-29
FIX for SuperboogaV2 when trying to unload databases "Clear Data" not working crashing textgen


RandomInternetPreson:
### Describe the bug

There is an issue with textgen when trying to unload a superboogav2 database.  Creating a database works, but unloading does not.

To fix this the following changes need to be made:

1. In your start_YOUROS_.sh/bat file for me it is start_linux.sh (open this with a text editor) go to the bottom and add this line:
LINUX:
`export ALLOW_RESET=TRUE`

so it looks like this:
![image](https://github.com/oobabooga/text-generation-webui/assets/6488699/ff5faa0f-c35a-42b1-add5-eb34dbec1635)

if you are using Windows use this line: 

`set ALLOW_RESET=TRUE`

2.  Now open chromadb.py found in the extensions/superboogav2 folder

replace:

`    def _merge_infos(self, infos: list[Info]):
        merged_infos = []
        current_info = infos[0]

        for next_info in infos[1:]:
            merged = current_info.merge_with(next_info)
            if merged is not None:
                current_info = merged
            else:
                merged_infos.append(current_info)
                current_info = next_info

        merged_infos.append(current_info)
        return merged_infos`

with 

`    def _merge_infos(self, infos: list[Info]):
        if not infos:
            return []  # Return an empty list if there are no infos to merge
    
        merged_infos = []
        current_info = infos[0]
    
        for next_info in infos[1:]:
            merged = current_info.merge_with(next_info)
            if merged is not None:
                current_info = merged
            else:
                merged_infos.append(current_info)
                current_info = next_info
    
        merged_infos.append(current_info)
        return merged_infos`

and replace:

`    def clear(self):
        with self.lock:
            self.chroma_client.reset()

            self.ids = []
            self.chroma_client.delete_collection(name=self.name)
            self.collection = self.chroma_client.create_collection(name=self.name, embedding_function=embedder)

            logger.info('Successfully cleared all records and reset chromaDB.')`

with

`    def delete(self, ids_to_delete: list[str], where: dict):
        with self.lock:
            ids_to_delete = self.collection.get(ids=ids_to_delete, where=where)['ids']
            self.collection.delete(ids=ids_to_delete, where=where)

            # Remove the deleted ids from self.ids and self.id_to_info
            ids_set = set(ids_to_delete)
            self.ids = [id_ for id_ in self.ids if id_ not in ids_set]
            for id_ in ids_to_delete:
[...]

PR #6193, opened 2024-07-01
FIX for SuperboogaV2 when trying to unload databases "Clear Data" not working crashing textgen

RandomInternetPreson:
## Checklist:

There is an issue with textgen when trying to unload a superboogav2 database. Creating a database works, but unloading does not.

The database now unloads without error, although the number database embedding deleted is not properly reported out in the terminal. 

This is a fix to the issue here: https://github.com/oobabooga/text-generation-webui/issues/6181


RandomInternetPreson:
Sorry accidentally did not push to dev branch.

PR #6189, opened 2024-07-01
Whisper Fix, js replacement of the gradio element to prevent browser crash.

RandomInternetPreson:
## Checklist:

- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

This is my submission to fix Whisper_STT (a much loved extension <3 )

This is in reaction to this thread:

https://github.com/oobabooga/text-generation-webui/pull/5929

I've tested it with multiple devices running through different systems:

-android phone microphone through firefox, opera, and chrome
-desktop condenser microphone, through firefox, opera, and chrome

I think the STT models interrupt the sound more clearly as well with this code.  I can now use the small english model where before I had to use the medium sized model to get accurate transcriptions.

Issue #5563, opened 2024-02-22
Whisper STT overhaul


TimStrauven:
Hi,
First a big thanks for this awesome repo!
As this is my first contribution, I was not sure if I should create a pull request or not, and if there is even a need for the following:

Currently the speechrecognition library is used for Whisper STT, it loads the model and then unloads it again, costing extra time and often giving OutOfMemory errors when close to my GPU VRAM limit. There is also no easy option to load and unload a whisper model, or to specify the device. The sr library might be more useful if other STT methods were used, but it is only needed for Whisper in this project.

To solve this I gave it a shot to implement this myself, only changing the script.py file, enabling the following:
- Load model to GPU or CPU
- Unload model completely (and disable gradio audio recording element)
- set language to English if a ".en" model is chosen
- Easy recovery to load another model on OutOfMemoryError with gradio error feedback
- no need for extra IO file operations via gradio audio "filepath" but still using numpy (there might be a more efficient method for resampling instead of pydub, please let me know)
- no more need for speechrecognition lib, instead pydub and openAI Whisper libraries are used

You can find the code for this change below, Please let me know if this can be implemented (or needs to be changed)  and what steps to follow (tests in "Contributing guidelines" were ok). 

```python
import gradio as gr
import torch
from modules import shared
import numpy as np
from pydub import AudioSegment
import whisper
import gc


input_hijack = {
    'state': False,
    'value': ["", ""]
}

# parameters which can be customized in settings.yaml of webui
params = {
    'whipser_language': 'english',
    'whipser_model': 'small.en',
    'auto_submit': True
}

startup_device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
WHISPERMODEL = whisper.load_model(params['whipser_model'], device=startup_device)


def chat_input_modifier(text, visible_text, state):
    global input_hijack
    if input_hijack['state']:
        input_hijack['state'] = False
        return input_hijack['value']
    else:
        return text, visible_text


def do_stt(audio, whipser_language):
    sample_rate, data = audio

    # use pydub to convert sample_rate and sample_width for whisper input
    dubaudio = AudioSegment(data.tobytes(),
                            frame_rate=sample_rate,
                            sample_width=data.dtype.itemsize,
                            channels=(1 if len(data.shape) == 1 else data.shape[1]),
                            )
    dubaudio = dubaudio.set_frame_rate(16000)
    dubaudio = dubaudio.set_sample_width(2)

[...]

PR #6194, opened 2024-07-02
Whisper stt overhaul js

TimStrauven:
I first created an overhaul of whisper to easily load the model on either GPU or CPU in #5563 a while ago.

Following the discussion and issues presented in #5929 this should be a merged solution for the issue @mamei16 tried to solve initially (including an extra record button next to the  default generate button). And it is also based on the solution of @RandomInternetPreson presented in #6189.

The gradio Audio component was abondoned because of the issues above and replaced by a js implementation.

## Checklist:


oobabooga:
Amazing work, I have tested it and it works perfectly. The code is clean, the UI works well, just perfect.

Combined with the coqui_tts extension or with @erew123's AllTalk TTS, this should be a fun experience now. A tip is to leave the button active and start/stop recording with Enter.

print

I have added @RandoInternetPreson as a PR coauthor for acknowledgement.

PR #6227, opened 2024-07-12
Fix for MacOS users encountering model load errors

InvectorGator:
Fix for the following error when attempting to load any model on a Mac:
"The CPU version of llama-cpp-python is already loaded. Switching to the default version currently requires a server restart."

## Checklist:


oobabooga:
I changed the implementation to a simpler and more streamlined one with the same logic on macos. Let me know if it doesn't work.

SomeOddCodeGuy:
> I changed the implementation to a simpler and more streamlined one with the same logic on macos. Let me know if it doesn't work.

I gave your alternative fix a quick try and it worked on my Mac. So Mac users should be good with this. 

However, I can't determine if it will work for AMD users on Linux, as one of the users who was affected by this issue on the other thread fit that category. 

For example, when I remove the macos specific check, I hit:

_Exception: Cannot import llama_cpp_cuda because llama_cpp is already imported. Switching to a different version of llama-cpp-python currently requires a server restart._

I'm concerned an AMD/Linux user might hit that check as well.

oobabooga:
Thanks for the confirmation.

> I'm concerned an AMD/Linux user might hit that check as well.

It should work. Both AMD and CUDA have `llama_cpp_cuda` as the main library, and it gets imported by default. The `llama_cpp` import afterwards is a fallback in case the user wants to uninstall the provided libraries and compile a custom llama-cpp-python version.

PR #6233, opened 2024-07-13
Increase RoPE alpha scaling limits

Vhallo:
Increases the RoPE alpha scaling limits for the same reason as in #6142. The current maximum of 8.0 just isn't enough sometimes and while rope_freq_base could be used instead, the alpha scaling limit can confuse people.

## Checklist:


mefich:
compress_pos_emb could also use higher max value.
For example Yarn-Mistral 128k, Solar-10.7B 128k have 16x scaling but UI slider has 8 as max value.

oobabooga:
Since the values keep increasing, I have converted these elements to gr.Number, such that you can type any value.

Issue #6024, opened 2024-05-19
Output window flicking and disappearing intermittently.


zedpr0:
### Describe the bug

The output window where the llm outputs its text and displays the chat history randomly flickers intermittently.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

It happens in my setup even after a fresh install, and a fresh VM. It happens when I upgrade as well. I am very confused there is no error in the logs either, nether in the browser console. Just type into the input and the output window will start flickering after you request a response. This can also happen just randomly. My backup version that works fine is from commit 0f134bf7. I have tried to use a incognito window as well to rule out cache and cookies.

### Screenshot
<img width="2031" alt="Screenshot 2024-05-19 at 8 18 26 AM" src="https://github.com/oobabooga/text-generation-webui/assets/44780859/36dbf52c-0e5b-44c8-950c-069c69e04bf4">
<img width="2044" alt="Screenshot 2024-05-19 at 8 18 17 AM" src="https://github.com/oobabooga/text-generation-webui/assets/44780859/4ae8448c-9375-40a3-96c5-7a37867b5719">
<img width="2034" alt="Screenshot 2024-05-19 at 8 17 47 AM" src="https://github.com/oobabooga/text-generation-webui/assets/44780859/a032b703-0e5d-43c6-812f-91c3e4e481f2">



### Logs

```shell
00:21:32-544648 INFO     Starting Text generation web UI                                                                                                                                                                                      
00:21:32-546964 WARNING                                                                                                                                                                                                                       
                         You are potentially exposing the web UI to the entire internet without any access password.                                                                                                                          
                         You can create one with the "--gradio-auth" flag like this:                                                                                                                                                          
                                                                                                                                                                                                                                              
                         --gradio-auth username:password                                                                                                                                                                                      
                                                                                                                                                                                                                                              
                         Make sure to replace username:password with your own.                                                                                                                                                                
00:21:32-548169 INFO     Loading settings from "settings.yaml"                                                                                                                                                                                
00:21:32-588142 INFO     Loading "dolphin-2.7-mixtral-8x7b.Q5_K_M.gguf"                                                                                                                                                                                 
00:21:32-667244 INFO     llama.cpp weights detected: "models/dolphin-2.7-mixtral-8x7b.Q5_K_M.gguf"                                                                                                                                                      
[...]

PR #6247, opened 2024-07-19
Fix for issue #6024, don't auto-hide the chat contents

Patronics:
Comment out line to auto-hide scrollbar that causes issues by also hiding the text on the page, as described in #6024.
Of course as implemented so far this does remove the auto-hiding functionality of the scrollbar as well, further revisions are welcome.

## Checklist:


oobabooga:
Removing the feature is not a good solution. I don't experience the reported flickering on my system. Can you try to figure out a workaround that works for you?

Patronics:
Okay, I just pushed an adjustment that reenables the auto-hide of the scrollbar, while keeping the chat (based on the `prose` class within `chat-parent`) visible!

oobabooga:
Nice, thanks for the creative fix!

PR #6257, opened 2024-07-22
Bump llama-cpp-python (0.2.83?)

oobabooga:
* Remove the tensorcores wheels (they are not faster than the default CUDA wheels anymore in my tests)
* Remove the progress bar monkey patch. It's no longer necessary because I will be building a fork that includes it by default from now on.

Pending upstream llama-cpp-python update. I tried updating just the llama.cpp submodule in a fork and the result was kind of broken.


oobabooga:
Wait it does work, I was just OOM'ing on Mistral-Nemo due to not setting the n_ctx manually.

PR #6276, opened 2024-07-26
Make compress_pos_emb float

hocjordan:
`compress_pos_emb` originally had a slider gradation of 0.01, allowing the user to set float values, but this behaviour was changed in #6233. The UI now uses `gr.Number` instead of `gr.Slider` to represent RoPE scaling values so that devs don't have to repeatedly update the max value limit and as a side effect the gradation of `compress_pos_emb` became integer.

I find float values useful (a value of 1.3 allows Llama 3.0 models to work nicely with a 12k context, whereas 1 is insufficient and 2 causes noticeable drop in quality). This PR therefore replicates the previous behaviour by setting the precision to 2.

I have tested the change by setting `compress_pos_emb` to 1.3 and confirming that the value sent to `llama.cpp` is correct, i.e.
```
llama_new_context_with_model: freq_scale = 0.769231
```

I don't believe that this will cause problems with the rest of the code because float values were originally permitted.

## Checklist:


oobabooga:
Thank you!

PR #6279, opened 2024-07-27
Add beforeunload js event to add a confirmation dialog when leaving page

### LINKED ISSUE #6228: Add confirmation dialog when leaving the webui
**New feature request**

I was recently writing a long story in the `notebook` section of the webui when, by mistake, I clicked on a bookmark in my bookmark toolbar instead of clicking the `Parameters` tab of the webui.

So I left the page. I tried to use the back arrow of my browser to go back to the webui but all my output was gone...

It should be possible with javascript to ask the user if he really wants to exit using the [beforeunload event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) as shown [here](https://stackoverflow.com/a/17143425/1221252).

This would prevent any unsaved output to be lost.

leszekhanusz:
## Checklist:

# Description

This PR will add a `beforeunload` js event to `window` to add a confirmation dialog when leaving the page.

Fixes #6228

Tested on Firefox and Chrome.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -600,4 +600,15 @@ headerBar.addEventListener("click", (e) => {
   }
 });
 
+//------------------------------------------------
+// Add a confirmation dialog when leaving the page
+// Useful to avoid data loss
+//------------------------------------------------
+window.addEventListener('beforeunload', function (event) {
+  // Cancel the event
+  event.preventDefault();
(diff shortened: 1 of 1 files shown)


oobabooga:
Nice

PR #6315, opened 2024-08-07
Fix hijack libs

waheedi:
## Checklist:


oobabooga:
The library, as included in the requirements, is called llama_cpp_cuda on amd, and if it fails to import llama_cpp is already used as a fallback.

waheedi:
No problem I also think it’s irrelevant now.

Issue #6286, opened 2024-07-29
List Index out of Range histories[0][1]


taellinglin:
### Describe the bug

I pasted my model and characters from a previous install, reinstalled, and then got the same error:
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████| 3/3 [00:06<00:00,  2.03s/it]
12:40:15-915085 INFO     Loaded "MistralFeiFei" in 6.58 seconds.
12:40:15-916050 INFO     LOADER: "Transformers"
12:40:15-917051 INFO     TRUNCATION LENGTH: 32768
12:40:15-918049 INFO     INSTRUCTION TEMPLATE: "Custom (obtained from model metadata)"
12:40:15-918049 INFO     Loading the extension "openai"
12:40:16-018587 INFO     OpenAI-compatible API URL:

                         http://127.0.0.1:5000

Running on local URL:  http://127.0.0.1:3016

Traceback (most recent call last):
  File "Z:\SannyChat\thoughts\installer_files\env\Lib\site-packages\gradio\queueing.py", line 527, in process_events
    response = await route_utils.call_process_api(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\SannyChat\thoughts\installer_files\env\Lib\site-packages\gradio\route_utils.py", line 261, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\SannyChat\thoughts\installer_files\env\Lib\site-packages\gradio\blocks.py", line 1786, in process_api
    result = await self.call_function(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\SannyChat\thoughts\installer_files\env\Lib\site-packages\gradio\blocks.py", line 1338, in call_function
    prediction = await anyio.to_thread.run_sync(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\SannyChat\thoughts\installer_files\env\Lib\site-packages\anyio\to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6317, opened 2024-08-07
Do not set value for histories in chat when --multi-user is used

mashb1t:
## Checklist:

Fixes https://github.com/oobabooga/text-generation-webui/issues/6286


KyleMaas:
Can confirm that this fixes #6286 for me.

PR #6353, opened 2024-08-27
Fix HuggingFace Chat UI

### LINKED ISSUE #6299: OpenAI api issues with Huggingface Chat UI
### Describe the bug
(Just as an fyi, I had initially opened a support ticket on the Huggingface Chat UI github repo but they claim llama.cpp works just fine with OpenAI API endpoint, so this is clearly an issue with Oogabooga)

Hugging face's chat-ui support open ai api, and it works fine with Kobold cpp as the inference engine, however when using Oogabooga for inference using the /v1/chat/completions OpenAI api endpoint it seems to produce random text unrelated to the prompt (again, this does not occur with Koboldcpp) meaning that Oogabooga text-generation OpenAI api is bugged.

### Reproduction

Hugging face's chat-ui support open ai api, and it works fine with Kobold cpp as the inference engine, however when using Oogabooga for inference using the /v1/chat/completions OpenAI api endpoint it seems to produce random text unrelated to the prompt in Chat-UI frontend (again, this does not occur with Koboldcpp) meaning that Oogabooga text-generation OpenAI api is bugged.

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
Windows 10 x64, GTX 1070
```

Papierkorb:
HuggingFace Chat UI sends a multi-modal style request with only a single "text" element. However, our code required both an image and the text, breaking support for HuggingFace Chat UI. This change relaxes this requirement on our end, allowing a single "text" or "image_url" as well.

This fixes #6299

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -154,8 +154,9 @@ def convert_history(history):
                     elif item['type'] == 'text' and isinstance(item['text'], str):
                         content = item['text']
 
-                if image_url and content:
+                if image_url:
                     new_history.append({"image_url": image_url, "role": "user"})
+                if content:
                     new_history.append({"content": content, "role": "user"})
             else:
                 new_history.append(entry)


oobabooga:
Thanks @Papierkorb

PR #6354, opened 2024-08-27
Improved Extensions API

Vasyanator:
## Checklist:

Support for `ui_block`, `ui_tab` and `ui_params` functions in extensions has been added. There is backward compatibility with the deprecated `ui` function. 
- `ui_block()` works similarly to `ui()` when `is_tab=False`.
- `ui_tab()` works similarly to ui when `is_tab=True`.
- `ui_params()` adds a tab with extension name to Parameters to implement extension settings.

This allows extensions to have a more convenient and simpler interface.


Touch-Night:
Great! Formerly I can only write workaround like [this](https://github.com/Touch-Night/more_translators/commit/f39d953d563be679d7bc061398733ed606fcc351#diff-1c0e4a21a58677b4e286a6f25349efff5e7abd3eb74253e3d2dc1e8dc3ff4ebbR594-R702) to make my extension have both block and tab ui.

PR #6358, opened 2024-09-01
OpenAI API: new parameter to change model list behavior

jsboige:
## Checklist:

This is an attempt at moving on with [the following PR](https://github.com/oobabooga/text-generation-webui/pull/5720), which got stalled because there isn't one single behavior for API's model list composition that fits all uses:
Some cases require the dummy OpenAI models exposed, others local models instead, and in the latter case some may need to list all models available whereas others will want to only expose the currently loaded local model.

This PR introduced an additional parameter in the form of a flag allowing to expose dummy models and/or (all local models or single loaded local model)


Touch-Night:
Can this pr enable user to switch model (Unload the currently loaded model and load the selected model) through api?

jsboige:
> Can this pr enable user to switch model (Unload the currently loaded model and load the selected model) through api?

@Touch-Night, this PR is only about controlling what's returned by the model list API handle. 

Switching model is available with [this handle](https://github.com/oobabooga/text-generation-webui/blob/main/extensions/openai/script.py#L292), which calls [this method](https://github.com/oobabooga/text-generation-webui/blob/5522584992c632d75d2389e9342793fd1dcc391d/extensions/openai/models.py#L43), which in turn calls [that method](https://github.com/oobabooga/text-generation-webui/blob/5522584992c632d75d2389e9342793fd1dcc391d/modules/models.py#L65).

Now with this PR, if you choose to return the currently loaded model (value 6 or 7 for the new parameter) with the request for the model list, if you switch model on the fly, through ooba's web ui or through the corresponding handle, the response for the list handle will update accordingly. I was able to test that in Open-Webui: just refresh the conversation page and the drop down list now displays the new model.

PR #6365, opened 2024-09-04
 typo in OpenAI response format

jsboige:
## Checklist:

## Description

object_type should be singular as explicited in the documentation:
https://platform.openai.com/docs/api-reference/chat/object#chat/object-object
https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-object
Crashes Dify, which checks for the exact string.


oobabooga:
Nice, thanks

PR #6386, opened 2024-09-17
Force /bin/bash shell for conda

Thireus:
I have found that conda may use a different shell in some instances, such as /bin/sh. This fix ensures it executes /bin/bash instead.

## Checklist:

### DIFF
--- a/one_click.py
+++ b/one_click.py
@@ -190,7 +190,7 @@ def run_cmd(cmd, assert_success=False, environment=False, capture_output=False, 
             cmd = f'. "{conda_sh_path}" && conda activate "{conda_env_path}" && {cmd}'
 
     # Run shell commands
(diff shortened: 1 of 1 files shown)


Touch-Night:
This PR breaks TextGenWebui on Windows

oobabooga:
Thanks @Touch-Night, this should be fixed after https://github.com/oobabooga/text-generation-webui/commit/055f3f5632f99f99409fd85d4e06c4c70acf8b33

LuNeder:
This breaks systems that do not have bash at /bin/bash, such as NixOS, which was previously fixed at #6110

Setting `executable='bash'`, as opposed to `executable='/bin/bash'`, works, fixes this and should force bash well enough for the purposes of this pull request

PR #6428, opened 2024-10-03
Fix for systems that have bash in a non-standard directory

LuNeder:
As talked on #6386 

This fixes the software once again on NixOS, while still forcing bash as 6386 did.

(honestly I’m skeptical that the shell conda is ran from makes a difference, it definitely shouldn’t since the program (conda) itself should call the necessary shell if needed, buuuuuuut after the deeply traumatising experience of reading the miniconda installer script a while ago for some troubleshooting it wouldn’t surprise me at all if conda didn’t check before using bash syntax on sh to do some cursed stuff)

## Checklist:


oobabooga:
Thanks!

Issue #6253, opened 2024-07-21
[WinError 126] Error loading "backend_with_compiler.dll"


Magenta-Flutist:
### Describe the bug

Hi all, I still meeting with problem like in thread https://github.com/oobabooga/text-generation-webui/issues/5123 - OSError: [WinError 126] The specified module could not be found. Error loading "backend_with_compiler.dll" I thought that this issue was fixed about half year ago. I tried to fix it with my no skilled hands with no success. May I ask someone to explain how to solve it? 

I know that my problem may actually be minor or solved quite simply, but I've lost all hope of fixing it without anyone's help. 

PS: I succeed to run it on transformers loader with a combination of guides "Install IPEX-LLM" and "Run Text Generation WebUI"  with a 3.1 t/s speed (LLama 3). But I see that it is not the final solution of a problem.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

A. Install latest  drivers for ARC 770.
Do a clean install of OobaBooga using the latest snapshot.
Select "D" for the Arc install.
Run start_windows.bat

### Screenshot

_No response_

### Logs

```shell
┌──────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────┐
│ C:\Users\asdas\Desktop\text-generation-webui-main\server.py:6 in <module>                                                                                                                                  │
│                                                                                                                                                                                                            │
│     5                                                                                                                                                                                                      │
│ >   6 import accelerate  # This early import makes Intel GPUs happy                                                                                                                                        │
│     7                                                                                                                                                                                                      │
│                                                                                                                                                                                                            │
│ C:\Users\asdas\Desktop\text-generation-webui-main\installer_files\env\Lib\site-packages\accelerate\__init__.py:16 in <module>                                                                              │
│                                                                                                                                                                                                            │
│   15                                                                                                                                                                                                       │
│ > 16 from .accelerator import Accelerator                                                                                                                                                                  │
│   17 from .big_modeling import (                                                                                                                                                                           │
│                                                                                                                                                                                                            │
│ C:\Users\asdas\Desktop\text-generation-webui-main\installer_files\env\Lib\site-packages\accelerate\accelerator.py:32 in <module>                                                                           │
│                                                                                                                                                                                                            │
│     31                                                                                                                                                                                                     │
│ >   32 import torch                                                                                                                                                                                        │
[...]

PR #6433, opened 2024-10-05
Fix intel bug described in #6253

schorschie:
## Checklist:

- I changed the conda channel like described in the issue #6253 
- I addes some white space around an operator, because pycodestyle said so.


oobabooga:
Thank you!

PR #6485, opened 2024-10-22
fixed perplexity_color extension

KT313:
## Checklist:

non-hoverable mode:
- improved colors
- improved / fixed formatting

hoverable mode:
- fixed (the worst of) formatting problems


KT313:
so as it turns out, for some tokenizers like `transformers.models.llama.tokenization_llama_fast.LlamaTokenizerFast` from TinyMistral-248M, the tokenizer acts up when you try to decode single id's; it will always strip whitespaces from the tokens. However it seems to work as intended for most models like Hermes, which uses `transformers.tokenization_utils_fast.PreTrainedTokenizerFast`.

oobabooga:
@KT313 thanks for the many fixes, this looks good. Could you merge the dev branch and give it an additional test? I have made a test with `google/gemma-2-9b-it` and I am not sure if the extension is working or not (the colors look homogeneous, maybe it's my settings).

KT313:
@oobabooga i can't test it right now since i'm busy, but it worked for me when i made the pull request. Depending on which options you check, the colors either have a red hue depending on perplexity, or between green and red iirc

oobabooga:
I need the additional test above to be able to merge this PR.

FartyPants:
In my case the dropdown is not rendered properly

image

SeanScripts:
@KT313 
> are you sure this is needed? Since perplexity is proportional to probability, it would make more sense to just pick one of the two

Perplexity is a measurement of the distribution of probabilities over all possible tokens at a given position, while probability (as visualized here) only takes into account the token that was actually chosen while sampling, so they are not just (inversely) proportional. Often the probability will be high when the perplexity is low if you are sampling with top-1 (like if the top token has probability 0.99 or something), but the probability of the sampled token can be low while the perplexity is low (as an error or a creative choice). Seeing these potentially erroneous tokens visually indicated can be useful, and requires both probability and perplexity to be considered.

However, to improve the usefulness of the visualization, this could be scaled between the min and max probability at each token position, rather than coloring by a pure 0-1 probability scale. I've implemented this in the latest version when both probability and perplexity are checked, and it's pretty nice for seeing where the model samples unlikely tokens, and seeing how that biases its future decisions, especially in code generation (which is almost entirely very low perplexity).

I've also added a feature I've wanted for a while, being able to continue generation by selecting from the dropdown! And I fixed the dropdown being transparent like that.

PR #6488, opened 2024-10-24
Make token bans work again on HF loaders

ThisIsPIRI:
Token bans on HF don't work if `SuppressTokensLogitsProcessor` is removed, since they work by setting the `suppress_tokens` parameter.

## Checklist:


oobabooga:
Thanks, I wasn't aware of that.

PR #6528, opened 2024-11-13
fix: openai extension

hronoas:
**Pull Request Title:**
Fix: Handle Multiple Content Items in Messages

**Description:**

Currently, when messages contain multiple content items, only the last item in the list is used.

### Example Issue:
Given the following JSON input:
```json
{
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "first item"
        },
        {
          "type": "text",
          "text": "second item"
        },
        {
          "type": "text",
          "text": "last item"
        }
      ]
    }
  ]
}
```
The system currently utilizes only `"last item"` and ignores the other entries.

### Proposed Fix:
This PR addresses the issue by ensuring all content items within a message are processed appropriately.

### Checklist:


hronoas:
For example, the Cline extension in VSCode does not work correctly due to this issue.

Ph0rk0z:
Also some food for thought related to this: https://old.reddit.com/r/SillyTavernAI/comments/1govgfw/dev_st_produces_incorrect_chat_completion/

There may also be multiple SYSTEM messages depending on what front end/app is communicating with the API in chat completion. This may break the template for the model as well.

Chat completion sure is "fun".

oobabooga:
Thanks for the fix

PR #6533, opened 2024-11-18
Add autoawq to requirements

casper-hansen:
@oobabooga I have re-added autoawq to requirements because I recently overhauled the requirements for autoawq. This means the requirements are extremely minimal since we only rely on Triton kernels. We also have `autoawq[kernels]` for a more speedy inference solution, but I have chosen to leave it out of here due to torch having no backwards compatible support between minor versions.

image

## Checklist:


oobabooga:
@casper-hansen thanks for the PR. I removed AutoAWQ from the requirements because, according to the [AutoAWQ Kernels README](https://github.com/casper-hansen/AutoAWQ_kernels?tab=readme-ov-file#install-from-pypi), the provided wheels are for CUDA 12.4.1, while the project currently uses CUDA 12.1:

https://github.com/oobabooga/text-generation-webui/commit/1a870b3ea7f0b3337a0c0202fe4b4dddb4e559a0

Is CUDA 12.4 really necessary? Wheels are usually not backwards compatible across CUDA versions and it becomes hard to conciliate ExLlamaV2 wheels with AutoAWQ wheels in the same environment.

casper-hansen:
@oobabooga I may not have perfectly communicated this - but AutoAWQ kernels are now optional and does not get installed unless you specify `pip install autoawq[kernels]`. It's only needed if you want faster inference, so I would prefer for superusers to go do that themselves are compile the kernels if needed.

oobabooga:
The thing about triton is that it doesn't work on Windows, and I guess most people who install this project are on Windows (unfortunately). So to add AutoAWQ back, I'd need either `autoawq-kernels` wheels that work with CUDA 12.1, or to update the whole project to CUDA 12.4.

casper-hansen:
Seems there is a working triton installation for Windows below. This would be preferred since Windows is not compatible with most CUDA kernels either way.
https://github.com/woct0rdho/triton-windows/tree/v3.1.x-windows

oobabooga:
Having additional installation steps defeats the purpose of adding the library to the `requirements.txt`. I would need wheels that work with the current environment to add the requirement back. Maybe I should do that in a fork, assuming the library does in fact work with cuda 12.1.

oobabooga:
I have tried creating an isolated CUDA 12.4 environment to test autoawq, and it tries to install `torch==2.5.1` over `torch==2.4.1`. Reproduction:

```
conda create -n awq python=3.11
conda activate awq
conda install -y -c "nvidia/label/cuda-12.4.1" cuda
pip3 install torch==2.4.1 torchvision==0.19.1 torchaudio==2.4.1 --index-url https://download.pytorch.org/whl/cu124
pip install -r requirements_nowheels.txt --upgrade
pip install autoawq[kernels]
```

So autoawq seems to require:

* CUDA 12.4
* torch 2.5 (although not specified in the README)

These requirements make it impossible to add it to the project currently.

PR #6550, opened 2024-11-29
Add a user profile selector.

Alkohole:
Implementation of the user profile selector.

## Checklist:


Alkohole:
There is a profile image caching issue that may cause the current profile image to not match when switching between characters and chats, unfortunately I don't understand how to solve this.

Alkohole:
Emm... Okay... Thank you at least for this "feedback"...

PR #6280, opened 2024-07-27
Add q-cache 6 and 8 support for Exllamav2

randoentity:
## Checklist:

- [x] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

Issue #6126, opened 2024-06-15
Use HuggingFace's Quanto library KV Cache Quantization for any Transformers-based loader


Interpause:
**Description**

HuggingFace's Quanto has implemented 4 bit & 2 bit KV cache quantization compatible with Transformers. See: https://huggingface.co/blog/kv-cache-quantization

I may PR when I've time to experiment.

Interpause:
Seems definitely possible: https://github.com/Vahe1994/AQLM/issues/85#issuecomment-2194691934

But man they made me project lead for something in uni so im in a time crunch

Issue #6168, opened 2024-06-27
Add Q4/Q8 cache for llama.cpp


GodEmperor785:
**Description**

For some time there is an option to use Q8 and Q4 KV cache in llama.cpp. It is present for example in KoboldCPP and works great there.
Using quantized KV cache reduces VRAM requirements to run GGUF models.
Similar options already exists for exllamav2 and work great.

Perhaps it would be possible to add options "cache_8bit" and "cache_4bit" also to llamacpp loaders, so users are able to use quantized KV cache like in KoboldCPP?

**Additional Context**

Below are my notes from trying to add this.
I'm not really familiar with coding in webui but I tried adding the options in a quick way but I only managed to get Q8 cache to work.

From what I can see the Llama class in llamacpp has these two parameters:
type_k: Optional[int] = None,
type_v: Optional[int] = None,

Setting these to value like 8 seems to just work and reduce VRAM usage for KV cache by half. I managed to use this in webui by just setting these inside modules/llamacpp_model.py from_pretrained method:
params["type_k"] = 8
params["type_v"] = 8
For this to work the option flash_attn is required or else it throws errors, so perhaps some warning for users should be added to.
I tested it briefly to work with a few GGUF models, all worked without any problems.

However for Q4 cache (params["type_X"] = 4) it didn't work so most likely some more work is needed here and I don't know enough to figure out why it doesn't work.
I'm also not sure about llamacpp_HF loader as I have never used it.

oobabooga:
What are the most common combinations of these parameters?

GodEmperor785:
I did some digging in koboldcpp and llamacpp code to figure out what values they accept, as for some reason type_k/type_v = 4 fails with strange errors.

In llamacpp options I found this:
"KV cache data type for K (default: `f16`, options `f32`, `f16`, `q8_0`, `q4_0`, `q4_1`, `iq4_nl`, `q5_0`, or `q5_1`)"

But type_k/type_v in Llama class expects an int so we can't just write "q4_0".
So I checked how they did it in koboldcpp and found they use GGML_TYPE_Q4_0, GGML_TYPE_Q8_0 and GGML_TYPE_F16.
These seem to be defined in ggml_type enum in [ggml.h](https://github.com/LostRuins/koboldcpp/blob/c9c050f323a061480e6791eac98fc6636f53ef05/ggml.h#L355) 
[...]

PR #6561, opened 2024-12-07
Allow more granular KV cache settings

dinerburger:
## Checklist:

This PR adds more granular support for KV cache settings, allowing:
- Independent K and V cache types for llama.cpp
- FP16, FP8, Q8, Q6 and Q4 cache types for exllama

This PR should be expanded to allow for new Quanto types as mentioned in #6126, but before I go too far I wanted to make sure this structure was appropriate.

NOTE: This should probably supersede or compliment #6280 somehow


oobabooga:
Looks good, a single flag for all kv cache types and a dropdown menu are an ideal solution

It would be good to change `shared.args.kv_cache_type` if `shared.args.cache_4bit` or `shared.args.cache_8bit` are provided, to avoid a breaking change. Something like

```python
if shared.args.cache_4bit and shared.args.loader.lower() in ['exllamav2', 'exllamav2_hf']:
    shared.args.kv_cache_type = 'q4'

...
```

Feel free to continue expanding this PR, I'll merge it when you say it's ready.

dinerburger:
Yeah this is why I had marked it as draft; I figured we'd want to work out the little details before I got too far. 

Sure it makes sense to transform the legacy KV cache quant options elsewhere. Should that go in `shared.py`? I can make a `transform_kv_cache_options` function that will work similarly to `fix_loader_name` for example.

I'll make this change a little later today. While I have you: are you interested in getting the Transformers Quanto loader going too as in #6126 or should I pass on that for now and focus on Exllama and llama.cpp?

dinerburger:
> Does that work out of the box with transformers or does a new requirement have to be added?

A very good question. I don't get to work with Transformers much since I'm GPU-poor, but I'll evaluate this on a smaller model today and report back. My current leaning is: if there are no additional requirements I'll probably hammer it in, otherwise we'll circle back and pick it up in another pass.

modules/shared.py
```diff
@@ -269,6 +275,61 @@ def fix_loader_name(name):
         return 'TensorRT-LLM'
 
 
+def transform_legacy_kv_cache_options(opts):
```
dinerburger, line 278: OK something for you to chew on while I investigate transformers. I've added the supporting code for doing the conversion from the old `cache_[48]bit` flags into loader-specific parameters, but there's a twist: exllama and llama had different ideas about which it should prefer in the case that both flags were passed. I also wrote this to allow for modification of both the user_config object and the argparse.Namespace objects, but it required these weird `get` `set` trampolines. 

The good news I can confirm this works like you expect: passing `--cache-4bit` still works as you expect (with correct defaults for llama and exllama on models without a corresponding `user_config` item), and `user_config` items are updated and migrated correctly.

oobabooga: I have looked at the function and it looks good -- nice that you fixed `user_config` as well. 

Providing both `cache_4bit` and `cache_8bit` was undefined behavior, but it's good that you reproduced the previous behavior exactly.

modules/llamacpp_model.py
```diff
@@ -11,6 +11,32 @@
 from modules.text_generation import get_max_prompt_length
```
GodEmperor785, line 13: Are you sure all of these quant types work with llama.cpp?

I did some search here: [#6168 ](https://github.com/oobabooga/text-generation-webui/issues/6168#issuecomment-2198087925)
From what I saw in llama.cpp list of supported cache types is shorter than list of all quant types (there were no K-quants in supported cache quantizations).

I checked llama.cpp code a bit more now and found this: [common.cpp](https://github.com/ggerganov/llama.cpp/blob/26a8406ba9198eb6fdd8329fa717555b4f77f05f/common/common.cpp#L1018)
From what I understand this function kv_cache_type_from_str determines type for KV cache for llama.cpp.
And it seems to allow less types, it allows only: "f32", "f16", "bf16", "q8_0", "q4_0", "q4_0", "q4_1", "iq4_nl", "q5_0", "q5_1".
In other cases it fails with "Unsupported cache type".

Did you check if all the cache types for llama.cpp added in PR work? Especially those not on the list in llama.cpp code (like q6_k or q4_k)?

dinerburger: Right, thanks for pointing that out, part of finalizing this PR is checking to ensure the KV cache matrix is correct. Additionally, many of these are only supported if you build with `GGML_CUDA_FA_ALL_QUANTS`, which I'm not sure if we have enabled on our wheel. I'll ensure this is all to code before I mark ready.

dinerburger:
An update: transformers throws with

`ImportError: You need to install optimum-quanto in order to use KV cache quantization with optimum-quanto backend. Please install it via  with pip install optimum-quanto` if you try to use quanto. I'm gonna branch and stash then fix the llama.cpp support matrix.

dinerburger:
OK that'll do it for the last round of checks. I'm opening this for real review, since everything seems to be humming along nicely now. A note: I don't love the command line arguments, but llamacpp and exllama have pretty different options for kv cache quant. Happy to make changes to those as needed.

dinerburger:
You know, thinking on this for the day I think I'm going to collapse this down to a single command line argument: `--cache-bits`, and treat it nicely per loader, raising if we get a bad request.

oobabooga:
About `--cache-bits 8`, that restricts ExLlamaV2 to the Q8 cache, removing support for the 8bit cache. I liked the previous solution, with a dropdown menu containing explicit text elements. That would also make the implementation more future-proof: maybe llama.cpp will add support the other precisions at some point, like "q4_k", and being able to pass a `q4_k` to the flag would be ideal instead of working with integers.

dinerburger:
Got it I’ll roll it back tomorrow. Thanks for the feedback!

dinerburger:
@oobabooga ok I've backed out the parameter unification patch and moved back to per-loader string-based quantization specification. I've confirmed everything is still cooking nicely. Thanks again for your patience!

oobabooga:
Thanks @dinerburger, I have made some small final changes:

1. Merge the flags into one `--cache_type`, and add info messages to the UI and the exceptions saying the valid types for each loader.
2. Make `--cache_type` None by default instead of fp16
3. Rename `get_llamacpp_quant_type_for_string` to `get_llamacpp_cache_type_for_string` (since it will only be used for cache for now).

@dinerburger @GodEmperor785 do things look right after those changes?

GodEmperor785:
Looks good to me too

modules/ui_model_menu.py
```diff
@@ -118,8 +118,7 @@ def create_ui():
                             shared.gradio['flash_attn'] = gr.Checkbox(label="flash_attn", value=shared.args.flash_attn, info='Use flash-attention.')
                             shared.gradio['auto_devices'] = gr.Checkbox(label="auto-devices", value=shared.args.auto_devices)
                             shared.gradio['tensorcores'] = gr.Checkbox(label="tensorcores", value=shared.args.tensorcores, info='NVIDIA only: use llama-cpp-python compiled with tensor cores support. This may increase performance on newer cards.')
-                            shared.gradio['cache_8bit'] = gr.Checkbox(label="cache_8bit", value=shared.args.cache_8bit, info='Use 8-bit cache to save VRAM.')
-                            shared.gradio['cache_4bit'] = gr.Checkbox(label="cache_4bit", value=shared.args.cache_4bit, info='Use Q4 cache to save VRAM.')
+                            shared.gradio['cache_type'] = gr.Dropdown(label="cache_type", choices=['fp16', 'q8_0', 'q4_0', 'fp8', 'q8', 'q6'], value=shared.args.cache_type, info='Valid options: llama.cpp - fp16, q8_0, q4_0; ExLlamaV2 - fp16, fp8, q8, q6, q4.')
```
dinerburger, line 121: Ah wait, sorry, we're missing `q4` here for EXL2

oobabooga: Oops, well noted. Fixed

oobabooga:
Thanks @dinerburger @GodEmperor785!

@dinerburger if you feel like implementing quantized cache for Transformers, that would be a nice addition. I assume the additional requirement is pure Python/PyTorch, without compiled wheels, so it should be an easy addition.

dinerburger:
@oobabooga I actually implemented it in [this (now out of date) branch](https://github.com/oobabooga/text-generation-webui/compare/main...dinerburger:text-generation-webui:transformers-kv-cache-support#diff-36dfef7d9db97b2cde9351ae9eafa7aed3dd15e094d2ca7a4a60bfa88d6096f2R375). Problem was: it was unusably bad. Generation was pure garbage. Not sure if it was because I missed setting `axis-key` and `axis-value` as mentioned in the [HF Best Practices Guide](https://huggingface.co/docs/transformers/en/kv_cache), but whatever it was it certainly wasn't a drop-in solution like lcpp or exl. (I thought I saw additional documentation indicating that you shouldn't quantize until a particular cache size was reached, but I can't seem to find it now.)

oobabooga:
Interesting, I see that you tried even the HQQ one. I found this source, which you have probably seen already

https://github.com/huggingface/transformers/blob/9613933b022ddbf085e2c593ed4ceea4c734179a/src/transformers/cache_utils.py#L664

https://github.com/huggingface/transformers/blob/9613933b022ddbf085e2c593ed4ceea4c734179a/src/transformers/cache_utils.py#L754

https://github.com/huggingface/transformers/blob/9613933b022ddbf085e2c593ed4ceea4c734179a/src/transformers/cache_utils.py#L820

int2 and int4 precision is probably round to nearest and too low precision. The HQQ int8 one would be the most likely to perform well.

dinerburger:
It’s funny, I actually forgot to try the HQQ one at all favoring Quanto. I’ll get this branch rebased tonight and give it a shot.

PR #6572, opened 2024-12-12
Fix: Handle Error for CUDA Configuration and API Request

skywinder:
## **Issue**
Running the following `curl` request:

```bash
curl -X POST http://localhost:7860/v1/completions \
  -d '{
    "prompt": "Hello, how are you?",
    "max_tokens": 0
  }'
```

results in an error:

raise AssertionError("Torch not compiled with CUDA enabled")

Steps to Reproduce
	1.	Start the API using the following command:

./start_macos.sh --api --api-port 7860 --verbose

	2.	Test the endpoint with:
```
curl -X POST http://localhost:7860/v1/completions \
-d '{
    "prompt": "Hello, how are you?",
    "max_tokens": 0
}'
```

Expected Behavior

The request should return a clear error or handle cases where CUDA is not enabled.

Proposed Solution
	•	Add a preflight check during the startup process to verify CUDA compatibility.
	•	Ensure the API gracefully handles requests even when CUDA is unavailable.

Impact
	•	Prevents crashes due to unhandled assertions.
	•	Improves debugging experience with clear error messages.

## Checklist:


jfmherokiller:
I applied your fix locally because it allows the system to fallback to cpu when I am playing a game that is making heavy use of my gpu.

skywinder:
> I applied your fix locally because it allows the system to fallback to cpu when I am playing a game that is making heavy use of my gpu.

thanks, one more useful use case. hope it will be merged

jfmherokiller:
just be warned its possible this introduces instablility. Tho I cant test the codepath because I run streaming due to how slow my generation is.

oobabooga:
Thanks for the PR @skywinder and apologies for taking so long to review. I have created a `get_device()` auxiliary function to handle not only `mps` but all torch backends, and applied this function to all relevant parts of the codebase. 

It seems to make sense so I'll merge it; if there are any outstanding issues, please feel free to send a new PR.

PR #6575, opened 2024-12-13
Improved UI

oobabooga:
| Before | After |
|----------|---------|
| before-1 | after-1 |
| before-2 | after-2 |

#### Status

~~* There are many details to fix in the Chat tab~~
~~* The UI is 100% broken on mobile~~
~~* Light theme is not implemented yet~~
~~* I'll add an `--old-colors` flag to optionally revert to the current colors once this is done~~

This improves the UI colors and layout.

#### Credits

* I am using https://github.com/mckaywrigley/chatbot-ui (the "legacy" version, v1.0, April/2023) as a reference and am copying some CSS from there. In turn, that old chatbot-ui was based on the old ChatGPT (early 2023).

### DIFF
--- a/css/chat_style-cai-chat-square.css
+++ b/css/chat_style-cai-chat-square.css
@@ -16,6 +16,6 @@
 }
 
 .message {
-    padding-bottom: 30px;
+    padding-bottom: 2em;
     grid-template-columns: 70px minmax(0, 1fr);
 }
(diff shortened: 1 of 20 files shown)


skywinder:
looks nice, could you please add screenshots to compare old/new interface?

oobabooga:
Thanks @skywinder, I have added screenshots to the first message and fixed the link.

PR #6605, opened 2024-12-24
Multi platform single repo

jfmherokiller:
This change will allow users to use the same repo between multiple operating systems when using a file system like btrfs on a portable device like a flashdrive.

I currently have this under draft because i seem to be encountering en error where server.py cant find modules and I dont know why.

Edit: it seems to function if I call server.py directly from inside a window with the env set.

Edit: a full system reboot and btrfs rebalance seems to have made it function as desired i am now opening this pullreq
## Checklist:

### DIFF
--- a/cmd_linux.sh
+++ b/cmd_linux.sh
@@ -8,8 +8,8 @@ if [[ "$(pwd)" =~ " " ]]; then echo This script relies on Miniconda which can no
 { conda deactivate && conda deactivate && conda deactivate; } 2> /dev/null
 
 # config
-CONDA_ROOT_PREFIX="$(pwd)/installer_files/conda"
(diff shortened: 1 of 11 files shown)


jfmherokiller:
This pullreq enables a crossplatform semi-portable existance on file systems like btrfs. 

I am currently using it on my daulboot system to allow me to only need to store a single copy of my models and the repo while also making use of the dedup feature of btrfs to keep only single copies of the python bits.

jfmherokiller:
forgot to check the styleguide correctly but now all is fixed

oobabooga:
This change is very well implemented, however, it would break existing installs that depend on `installer_files/conda` and `installer_files/env`. 

This could of course be managed through conditional statements in all modified files, but I don't think enough people would benefit from this to justify the added complexity.

PR #6639, opened 2025-01-07
Update start_windows.bat

BartAgterbosch:
Fixed the broken checksum bug (rather, removed it altogether) where it will always, at all times, inevitably, fail the checksum, because there is either something funky going on with the length of the filename, or it doesn't play well with there being multiple instances on a single machine.  This has been broken for far too long, so better to just remove it than to leave it broken honestly.

## Checklist:


oobabooga:
This works on my tests. If there is an issue on your setup, a fix would be welcome, but this feature cannot be removed as it was added for security.

Issue #6592, opened 2024-12-20
Gallery extension broken after UI update


dEATh-PM:
### Describe the bug

There's a change in how Gradio's Dataset is updated, leaving the Gallery extension no longer fully functional.
The "Refresh" button and the filter no longer have any effect.
(Displaying the grid and clicking a character to select still works as intended.)

I tried some quick workarounds but ran into issues with the gradio_hijacker.py that are beyond my abilities.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Enter text into the filter.

- As well as - 

Pressing the "Refresh" button.

### Logs

```shell
C:\AI\Oobabooga\installer_files\env\Lib\site-packages\gradio\components\dataset.py:164: UserWarning: The `Dataset` component does not support updating the dataset data by providing a set of list values. Instead, you should return a new Dataset(samples=...) object.
  warnings.warn(
```

### System Info

```shell
Win 10
EVGA 3090
```

PR #6656, opened 2025-01-12
Fix gallery warning

TheLounger:
## Checklist:

## What this PR does:

Fixes the following warning when filtering the gallery cards:

```...\gradio\components\dataset.py:164: UserWarning: The `Dataset` component does not support updating the dataset data by providing a set of list values. Instead, you should return a new Dataset(samples=...) object.```


oobabooga:
Nice, this fixes https://github.com/oobabooga/text-generation-webui/issues/6592. Thanks @TheLounger!

PR #6661, opened 2025-01-13
HTML: Fix quote pair RegEx matching for all quote types

Th-Underscore:
## Checklist:

The current quote matching only works for the first tuple, `('&quot;', '&quot;')`, due to how `Match.group` functions (enumerated `[1, 2, 3]`, `[4, 5, 6]`, etc.). Additionally, literal open-close quotes `\u201C` and `\u201D` are undetected.

### DIFF
--- a/modules/html_generator.py
+++ b/modules/html_generator.py
@@ -73,7 +73,6 @@ def fix_newlines(string):
 
 
 def replace_quotes(text):
-
     # Define a list of quote pairs (opening and closing), using HTML entities
     quote_pairs = [
         ('&quot;', '&quot;'),  # Double quotes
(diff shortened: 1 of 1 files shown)


oobabooga:
Thanks, this is indeed an issue. The problem with your solution is that it doesn't enforce matching pairs, so `"` can match with `&rsquo;`, which is less robust.

What do you think of this?

```python
import re

def replace_quotes(text):
    # Define a list of quote pairs (opening and closing), using HTML entities
    quote_pairs = [
        ('&quot;', '&quot;'),
        ('&ldquo;', '&rdquo;'),
        ('&lsquo;', '&rsquo;'),
        ('&laquo;', '&raquo;'),
        ('&bdquo;', '&ldquo;'),
        ('&lsquo;', '&rsquo;'),
        ('&#8220;', '&#8221;'),
        ('&#x201C;', '&#x201D;'),
        ('\u201C', '\u201D'),
    ]

    # Create a regex pattern that matches any of the quote pairs, including newlines
    pattern = '|'.join(f'({re.escape(open_q)})(.*?)({re.escape(close_q)})' for open_q, close_q in quote_pairs)

    # Replace matched patterns with <q> tags, keeping original quotes
    def replacer(m):
        # Find the first non-None group set
        for i in range(1, len(m.groups()), 3):  # Step through each sub-pattern's groups
            if m.group(i):  # If this sub-pattern matched
                return f'<q>{m.group(i)}{m.group(i + 1)}{m.group(i + 2)}</q>'

        return m.group(0)  # Fallback (shouldn't happen)

    replaced_text = re.sub(pattern, replacer, text, flags=re.DOTALL)
    return replaced_text
```

Th-Underscore:
The solution I originally had in mind for that was:

```python
    for open_q, close_q in quote_pairs:
        # Create a regex pattern that matches each of the quote pairs, including newlines
        pattern = f'({re.escape(open_q)})(.*?)({re.escape(close_q)})'
        # Replace matched patterns with <q> tags, keeping original quotes
        text = re.sub(pattern, lambda m: f'<q>{m.group(1)}{m.group(2)}{m.group(3)}</q>', text, flags=re.DOTALL)
    return text
```

Yours takes better advantage of `re` and is more pythonic, so I think I'd go with yours.

Th-Underscore:
Whoops, Git mistake. Not sure what happened. I'll try to resolve.

oobabooga:
Looks good now. Thanks a lot for this fix, @Th-Underscore -- if you notice anything else weird with the markdown renderer, feel free to send new PRs.

PR #6685, opened 2025-01-22
Give SillyTavern a bit of leeway the way they do OpenAI

FartyPants:
## Checklist:

This is one liner "silly fix" for the way SillyTavern would hammer the system role.  Just add them to one and bob's your uncle.

There is also a more evolved 8 line fix that completely emulates how OpenAI handles System message but I think it's just too much for little to no benefit. So I'm submitting this one as the "winner" as it's a no brainier.

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -205,7 +205,7 @@ def convert_history(history):
             else:
                 chat_dialogue.append(['', current_reply])
         elif role == "system":
-            system_message = content
+            system_message += f"\n{content}" if system_message else content
 
     if not user_input_last:
         user_input = ""


oobabooga:
Thanks, if this works on ST's end then it's worth it -- better than discarding all but the first system message.

I'll try to refactor the history format in the project one day when I feel inspired to handle this better.

FartyPants:
> Thanks, if this works on ST's end then it's worth it -- better than discarding all but the first system message.
> 
> I'll try to refactor the history format in the project one day when I feel inspired.

Yeah, refactor the history would be good in long run, but no rush as that's something that will break a few eggs and a few extensions (like my own, LOL). 
But ultimately, it would be necessary, at some point a few more roles might be added to models then WebUi won't be able to handle it at all.

PR #6686, opened 2025-01-22
Allow for Model Subfolder Organization

Googolplexed0:
Allows for the `modules.utils` functions `get_available_models()` and `get_available_ggufs()` to find models recursively, sorting through subfolders in `shared.args.model_dir` using `os.walk()` rather than `Path.glob()`.

## Checklist:

- [x] Changes made have been minimized to the best of my ability.


oobabooga:
There have been other PRs with this change, as I said in the past, I prefer to keep the folder structure simple.

oobabooga:
I'm turning the project more towards llama.cpp, so I have changed my mind about this. I have changed the approach slightly to first show the GGUF files, then the top-level model folders in the model dropdown. Model folders containing a GGUF but no safetensors are excluded from the list, and the pure GGUF is shown instead. This is because llamacpp_HF will not be available anymore. Thanks for the PR.

PR #6716, opened 2025-01-30
Markdown CSS fix: Inline code blocks now display correctly

jwinpbe:
## Checklist:

- [✔️] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

Hello!

The current method of displaying inline code examples in markdown on the main branch currently results in erroneous behavior.

It was bothering me, so I fixed it and added a bit of my own style to it. This is my first ever pull request. Thank you.

### Current CSS output of the main branch:

image

### My proposed changes:

image
image


oobabooga:
The current style is intentional and I don't want to add blue to the gray pallette.

jwinpbe:
@oobabooga Would you be OK with me adding this kind of code block highlighting a UI option somewhere in that case? The behavior made me think it was some sort of bug, but if it's your personal preference I'm happy to add it as a selectable option with whatever highlight color you think would be preferable. :)

Thanks for all your work making the rest of this so accessable!

PR #6738, opened 2025-02-07
Fix miniconda install loop on Arch linux. Fix for existing, failed miniconda installs interfering with subsequent install attempts. Other misc fixes refer to description and diff.

toazd:
Fix miniconda install loop on Arch linux.

Add removal of existing, failed miniconda install folder and script if conda execution fails. This also will likely fix other open issue(s) caused by a failed miniconda install but those have not been tested.

Fix misc shellcheck SC2164 warning.

## Checklist:

### DIFF
--- a/start_linux.sh
+++ b/start_linux.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
-cd "$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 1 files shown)


oobabooga:
I think that it would be better to manually delete the installer_files folder and start over in cases like this instead of making the installer more complex.

PR #6748, opened 2025-02-16
Improve SuperboogaV2 with Date/Time Embeddings, GPU Support, and Multiple File Formats

alirezagsm:
This PR introduces a few improvements to the SuperboogaV2 extension and brings it to a usable state.

- `ChromaCollector` now uses improved HNSW parameters to avoid running into "ef or M is too small" errors.
- Embedding function now prioritizes the usage of GPU for speedy processing.
- Users can now load multiple files simultaneously.
- The manual mode is set to `false` by default.
- Added the option to include the current date and time in the embeddings. This is set to `true` by default.
- Added a README file with basic information about the extension and a simple installation guide.

The following filetypes are tested on Windows 11 with CUDA 12.1:
- TXT
- PDF
- EPUB
- HTML
- CSV
- ODT/ODS/ODP
- DOCX/PPTX/XLSX

The dependencies in requirements.txt are updated to their latest versions and the following new dependencies are added to handle new filetypes: 
- PyMuPDF
- python-docx
- python-pptx
- openpyxl
- odfpy

## Checklist:


oobabooga:
Thanks a lot for this @alirezagsm. superboogav2 is not an extension that I maintain or routinely use; your changes look very tidy and organized, and you have tested the extension and added documentation to it, so let's merge it.

Issue #6747, opened 2025-02-16
API calls throwing error after latest update that added "Show after" parameter


NoahDoah:
### Describe the bug

The latest update introduced the "Show after" parameter, which can be used to hide the thinking of reasoning models.
Since then I wasn't able to make any API calls anymore - resulting in a crash.

The chat works without problems in the webui.

It seems like a value for the new "Show after"-field is also needed when called via API.

This is line 415 in "modules/chat.py":

`show_after = html.escape(state["show_after"]) if state["show_after"] else None`

I changed the line as follows and now all API calls work again as before and without any problem:

`show_after = html.escape(state.get("show_after", "")) if state.get("show_after") else ""`

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

Start Oobabooga with API enabled, load a model (I have used GGUF via llama.cpp as well as EXL2 via ExLlama2_HF). I have then used Msty to call the API. The error is shown in the text-generation-webui server terminal and Msty shows "Premature close".

### Logs

```shell
Exception in ASGI application
Traceback (most recent call last):
  File "/home/main/text-generation-webui/text-generation-webui/installer_files/env/lib/python3.11/site-packages/sse_starlette/sse.py", line 247, in __call__
    await wrap(partial(self.listen_for_disconnect, receive))
  File "/home/main/text-generation-webui/text-generation-webui/installer_files/env/lib/python3.11/site-packages/sse_starlette/sse.py", line 236, in wrap
    await func()
  File "/home/main/text-generation-webui/text-generation-webui/installer_files/env/lib/python3.11/site-packages/sse_starlette/sse.py", line 191, in listen_for_disconnect
    message = await receive()
              ^^^^^^^^^^^^^^^
  File "/home/main/text-generation-webui/text-generation-webui/installer_files/env/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 563, in receive
    await self.message_event.wait()
  File "/home/main/text-generation-webui/text-generation-webui/installer_files/env/lib/python3.11/asyncio/locks.py", line 213, in wait
    await fut
asyncio.exceptions.CancelledError: Cancelled by cancel scope 7fd5d867ecd0

During handling of the above exception, another exception occurred:

  + Exception Group Traceback (most recent call last):
[...]

PR #6749, opened 2025-02-16
Fix OpenAI API with new param (show_after)

kelvie:
There probably should be some way to enforce this

## Checklist:


oobabooga:
It won't work -- show_after is based on removing the content after the </think> (or whatever) from the visible history, and the API uses exclusively the internal history. This addition was purely for the UI.

kelvie:
@oobabooga okay, but without a fix the API is completely broken, see: https://github.com/oobabooga/text-generation-webui/issues/6747

oobabooga:
Oh okay, I see now

PR #6756, opened 2025-02-18
Perplexity colors extension v2

SeanScripts:
## Checklist:

- Added some of the fixes/simplifications to the colors from this PR: https://github.com/oobabooga/text-generation-webui/pull/6485
- Resolved the formatting issues that were accidentally introduced in that PR caused by removing the `<pre>` tags. Newline characters are displayed so that you can see what token was chosen, but the newlines are also included so code formatting is otherwise the same.
- Fixed the CSS issue of the dropdown being transparent and looking bad. Selected token is highlighted with darker background
- Reduced cumulative lag from past messages with LRU cache on the markdown conversion
- Changed the color scheme when both probability and perplexity options are checked, to show the relative probability compared to the most likely option. This is really nice for seeing where the model chose unlikely tokens, which could be mistakes or creative choices. It's especially useful for seeing where unlikely tokens were sampled in code generation (where it's mostly low perplexity), and seeing how that affects the model's future decisions.
- Added a new feature I've wanted for a while, using custom JS, that lets you choose a token from one of the dropdowns in order to continue generation from there. This might still have some bugs, sometimes if the UI takes too long to respond it won't continue from the right place since currently I'm just adding a fixed delay to wait for the UI to update.

Example: This originally used `fibonacci():`, and I chose the token from the dropdown to make it just `fib():`. Colored by probability and perplexity, but the perplexity is low for the code. You can see the lower probability sampled options of choosing to include or exclude whitespace (e.g. the yellow ` \n`) and how that affects how the model chooses to format the next few lines.
perplexity_colors_v2_example


oobabooga:
Thank you for this @SeanScripts, it's very valuable to have the original author (you) test things thoroughly and fix the extension. Since things are working in your testing, let's merge it.

Issue #6539, opened 2024-11-20
does text-generation-webui support Function Calling now?


Charles1757:
same as mentioned here, but I'm not sure if we are supporting this:
https://github.com/oobabooga/text-generation-webui/issues/4455

Chepko932:
Same question, cant get it working so i guess no?

jfmherokiller:
from the code based on the latest commit.  https://github.com/oobabooga/text-generation-webui/blob/4924ee2901ce9ad17259ae817bbb46885d1ba8ba/extensions/openai/completions.py#L221

```
    if body.get('function_call', ''):
        raise InvalidRequestError(message="function_call is not supported.", param='function_call')
```

oobabooga:
Yes — tool/function calling is supported as of v4.0, for Qwen 3.5, Devstral 2, GPT-OSS, DeepSeek V3.2, GLM 5, MiniMax M2.5, Kimi K2/K2.5, and Llama 4. See the [release notes](https://github.com/oobabooga/text-generation-webui/releases/tag/v4.0) and [docs](https://github.com/oobabooga/text-generation-webui/wiki/12-%E2%80%90-OpenAI-API#tool-calling).

Issue #4455, opened 2023-11-03
OpenAI extension for Function Calling


HAL9KKK:
**Description**

Function calling is one of the major game changer using LLM, ufortunately openai extension is still not supported in the text-generation-webui . 

So I kindly invite to **prioritize** this feature to allow to use HF LLM already available for this feature, like:
- Trelis/Llama-2-7b-chat-hf-function-calling-v2 (FREE)
- Trelis/Llama-2-13b-chat-hf-function-calling-v2 (PAID)
- Trelis/Mistral-7B-Instruct-v0.1-function-calling-v2 (PAID)

**Additional Context**

[text-generation-webui](https://github.com/oobabooga/text-generation-webui/tree/main)/[extensions](https://github.com/oobabooga/text-generation-webui/tree/main/extensions)/[openai](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai)
/completions.py

```
def messages_to_prompt(body: dict, req_params: dict, max_tokens):
    # functions
    if body.get('functions', []):  # chat only
        raise InvalidRequestError(message="functions is not supported.", param='functions')
    if body.get('function_call', ''):  # chat only, 'none', 'auto', {'name': 'func'}
        raise InvalidRequestError(message="function_call is not supported.", param='function_call')
```

yhyu13:
@HAL9KKK @oobabooga 
 
Refering to potential duplicated issue https://github.com/oobabooga/text-generation-webui/issues/4286

yhyu13:
Insights from memgpt in adapting function calling for open source models

https://github.com/cpacker/MemGPT/blob/f46cc3b15bd73503483db008fed9e82ad46a136b/memgpt/local_llm/README.md?plain=1#L201

HAL9KKK:
ChatGLM3 [https://github.com/THUDM/ChatGLM3](https://github.com/THUDM/ChatGLM3/) is the right solution!!!!

Test it:
```
user_prompt = "What is exactly the temperature in Rome, italy?"
completion = openai.ChatCompletion.create(
    model="xxx",
    messages=[{"role": "user", "content": user_prompt}],
    functions=function_descriptions,
    function_call="auto"
)
```

```
{
  "role": "assistant",
  "content": "get_current_weather\n ```python\ntool_call(location='Rome, Italy')\n```",
  "name": null,
  "function_call": {
    "name": "get_current_weather",
    "arguments": "{\"location\": \"Rome, Italy\"}"
  }
[...]

Issue #4286, opened 2023-10-14
OpenAI functions implementation missing.


deece:
**Description**

functions & function_call arguments are currently not implemented in the OpenAI extension.

It would be useful to implement them, as projects like open-interpreter make use of them, and having them access via the OpenAI extension is much less effort than writing native API implementations.

**Additional Context**
These arguments are documented here:
https://platform.openai.com/docs/api-reference/chat/create

vampirismtrueblood:
I am having the same issue.

I'm using the openAI extension with Autogen by Microsoft, and i works stunningly well, big thanks to everyone who made this possible. However, there's is an issue when using "Multi-agent" architecture, using [code from here](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_planning.ipynb)

The openai lib returns "openai.error.InvalidRequestError: functions is not supported."


Full Traceback:

`Traceback (most recent call last):
  File "/nvme_ugreen_usb/ai_agents/agent_chat_panning/agent_chat_panning_using_openai_our_ai_version.py", line 126, in <module>
    user_proxy.initiate_chat(
  File "/nvme_ugreen_usb/ai_agents/autogen/autogen/agentchat/conversable_agent.py", line 531, in initiate_chat
    self.send(self.generate_init_message(**context), recipient, silent=silent)
  File "/nvme_ugreen_usb/ai_agents/autogen/autogen/agentchat/conversable_agent.py", line 334, in send
    recipient.receive(message, self, request_reply, silent)
  File "/nvme_ugreen_usb/ai_agents/autogen/autogen/agentchat/conversable_agent.py", line 462, in receive
    reply = self.generate_reply(messages=self.chat_messages[sender], sender=sender)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nvme_ugreen_usb/ai_agents/autogen/autogen/agentchat/conversable_agent.py", line 781, in generate_reply
    final, reply = reply_func(self, messages=messages, sender=sender, config=reply_func_tuple["config"])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5185, opened 2024-01-06
Add function calling ability to openai extension

yhyu13:
## Summery

This is a experimental implemention of function calling ability for the openai extension.

**Essentially, the function calling is simply user adding functions with json format into url request body, and server process these functions into part of system prompt. Either with 1 shot prompting or some sft function calling ability from the base llm, a llm model is able spit out function api name and arguments in json format. Finally, the server post process the llm ouput into url reponse body.**

Openai has provide solid reference for api spec, https://platform.openai.com/docs/api-reference/chat/create and cookbook exmaple that I used for testing https://github.com/openai/openai-cookbook/blob/main/examples/How_to_call_functions_with_chat_models.ipynb

I fine tuned a phi2 model https://huggingface.co/Yhyu13/dolphin-2_6-phi-2-sft-glaive-function-calling-v2-ep1 using a curated dataset  https://huggingface.co/datasets/Yhyu13/glaive-function-calling-v2-llama-factory-convert?row=2 for this implementation, though I believe sft model is not a pre-prequsite because llm can be 1 or N-shot prompted.

Progress so far
- [x] Support both function calling(deprecated, but some project like memgpt still use it!) and tool calling api for openai chat completion api (shold not support stream though, but I am not sure)
- [x] Implement function calling context that parse user functions and function calls into sysmte prompt,
- [x] Handle function role as part of historical user input
- [x] Post process function calling finish reason
- [x] Very basic run in the above openai cook book example, and it worked (though the response is not as accurate as chatgpt3.5 or above)! 

Caveats:
- Only support generating single function calling in a single response so far. Parsing multi function call json str is simply not implemented yet. That is because I have not seen any example in the cookbook where multi function calling in a single response, though the openai response format implies there could be multi function calling (as it uses a list to store every function calling response)
- I personally use a xml wrapper for `<functioncall>` and for function role, I used `<functionresponse>` as indicators. Those indicators I also used in training sft model for function calling. THIS IS NOT UNIVERSAL, and base models that have not exposed with `glaive-function-calling-v2` might not follow it even with 1/N shot prompting
- Stream should be set to False in the function call request

TODO:
- [x] Test 1 shot prompt for models w/o function calling sft, I will do SOLAR-instruct and Mistral
- [x] Identify edges cases from calling openai api (~~Multi function calling?~~, ~~Escape crtl char in json str~~) 
- [x] Parallel function calling (i.e. calling multiple function in single reply) 
- [x] Do more openai cook book function call example and idenitfy bugs

[...]

PR #6757, opened 2025-02-18
Feature/tool calling

SeanScripts:
## Checklist:

I've been working on implementing tool/function calling for models that support it. This is a major feature that enables LLMs to use tools to do a lot of things they normally aren't very good at.

- Added "tools" folder with some example tools as JSON files, and a "presets" folder within it for tool presets as simple text lists.
- Added a new tab next to Instruction template for Tools
- Within this new tab, added a few overall options
  - Confirm tool use: Enabled by default and recommended, pauses execution so you can confirm before running a tool by pressing the "continue" button
  - Tools in user message: Changes the position of the tools, as an option in some instruct templates. I haven't really tested this option, but it might help with some models
  - Max consecutive tool uses: When you have "Confirm tool use" disabled, prevents the model from getting in an endless loop of calling the same tool, which happens sometimes with small models
- Added a tool selection preset with save/delete options, which just saves a text file under "tools/presets/" with the selected tools, so you can select groups of tools
- Multiselect checkbox group of the available tools to be enabled in the current session. Clicking on of these will display the details in the menu to the right ~~(I wanted something like a radio button for the selected tool vs checkboxes for the enabled ones, and this has the problem of requiring you to enable a tool to select it... Which I think I'd rather avoid given that currently it executes the tool code when it's enabled, to add the function to the globals.)~~ and a dropdown menu for the tool filename, which is much safer
- On the right-hand menu, there are fields for the selected tool which can also be used to add a new tool, along with options to save/delete the tool:
  - Name
  - Type (function, not sure about other options)
  - Description (to tell the LLM what the tool does and how to use it)
  - Parameters (JSON object describing the input parameters for the tool)
  - Action (Python code to execute to define the function which runs when the tool is used by a model, with the name of the function matching the name of the tool, and returning a string)
- Added the functionality to be able to run tool calls in instruct mode with the right models, but not anything for the API yet. Tool calls are processed by checking the model output for JSON and seeing if the fields match what is required for the tool call. Some models use different formats or special tokens for tool calling, and I'm using the instruct template to make it hopefully work across different models despite the differences. Verified for Llama 3 and Qwen 2.5 at least.

Examples:
Gradio UI for tools (`roll_dice` example shown)
tools_tab

(Model is Qwen 2.5 32B Instruct Q4 in these tests)

Dice example, showing before and after you confirm execution:
roll_dice_tool_call_example

Code interpreter example, showing the displayed code:
code_interpreter_tool_call_example

Tutorial on how to test tool use:

- Load a model that has tools enabled, e.g. Llama 3 or Qwen 2.5 (any size). The backend for loading shouldn't matter.
- Navigate to the new Tools tab (under Parameters, next to Instruction template)
- Keep "Confirm tool use" checked (I recommend this for safety, especially if you're testing the code interpreter)
- Choose a tool selection preset, or check any of the example tools I added, e.g. `roll_dice`.
- If you want to add your own custom tool, check the examples for an idea of the format required.
  - Tool name and description are required and it's what the model uses to determine what the tool does and how to use it, though it can also see the parameters to know the format requirements. Adding few-shot examples might be a good thing to add in the future, since small models can't always do zero-shot tool use very well.
  - Tool parameters follow the format from the [OpenAI function calling spec](https://platform.openai.com/docs/guides/function-calling). The validation I have in place checks the "properties" field for different input parameters and the "required" field for a list of required parameters. For each property, it checks the "type" field as one of a few standard options like number and string.
  - Tool action is the code that runs when you enable the tool (check the box to add it to the active tools the model can use). This must define the function for your tool, which returns a string (for now, other types could be nice later), and the name of this function must match the tool name. This is arbitrary code that can be executed, so make sure the tools you define and run are safe. For tools that let the LLM execute arbitrary code (like code interpreter), use caution and read the code before letting it run, as [backdoors are technically possible](https://blog.sshh.io/p/how-to-backdoor-large-language-models). There's not currently any environment isolation here, so tools could be used to read/write/delete your local files (which could also be useful for things like RAG), or make API calls to external sites. Just depends what code you're running, so use common sense and check that the code is reasonable and safe.
- After selecting the tools you want, assuming the code for them was loaded successfully, start a chat in instruct mode and just ask for something in plain language that would require using the tool, e.g. "Roll a D20". There is no guarantee the model will actually choose to use the tool, but if it's enabled and the model's instruct template includes the tools, and it was tuned to recognize when to use tools, then it often will.
- If the model generates a tool call (or more than one), it will get replaced by a JSON object following the OpenAI spec, assigning it a tool call ID. At this point, if you have the "confirm tool use" option selected (recommended), the execution will pause, and you can check what tool call will be made. For the code interpreter tool, the code will also be displayed with syntax highlighting.
- To accept the tool call, press the "continue" button. This will run the tool and display the tool result. Then the LLM will continue generating and generate a response with this tool call and result in the context, so that it can interpret the result. At this point it can also choose to make another tool call and repeat the process.

Models tested: Llama 3.2 Instruct (1B and 7B), Qwen 2.5 Instruct (7B and 32B). The 32B model obviously worked best for this, the small models tend to have issues with knowing when to use the tools and will often call them repeatedly or make up nonexistent ones.

The code interpreter tool looks at the output which was printed during execution, so if the code ends with just a variable without printing it (as if in a notebook), it won't work. This is something that needs to be fixed. Things like matplotlib also won't work, though if you save a plot to a file it will work (just make sure not to overwrite existing files...) The weather tool is just a placeholder, and the random number generation tools are also just examples to show how it works, but one could make useful tools for lots of different things.

This is a work in progress, but the basic functionality is in place to be able to create and run tools in the UI. I could reorganize it better and put some stuff in html_generator.py rather than directly in chat.py, so that it displays in a special way in the UI rather than just appearing as text with some markdown blocks. It also only works in Instruct mode right now. Token bias after detecting the start of a tool call would be nice to ensure it formats the tool call properly, though from what I've seen with these modern models, that hasn't really been an issue, surprisingly.

I still need to add the OpenAI API extension support for this. I've been focusing on getting it working in the UI, but I understand a lot of people want the API support. I haven't really looked at the extension though, so if anyone has any ideas on how to add it now that it's working in the UI, please leave a comment, and let's collaborate.

There are lots of things to add to improve this feature. But it is functional, at least, aside from the API, and people might like to test it and suggest improvements.

I don't have the option to merge to a new branch, so I set this PR to dev, but I suggest making a separate branch for this, as it changes quite a lot about the main generation code and should be tested more.

Relevant issues this (partially) addresses:
https://github.com/oobabooga/text-generation-webui/issues/4286
https://github.com/oobabooga/text-generation-webui/issues/4455
https://github.com/oobabooga/text-generation-webui/issues/6539

Please let me know what you think! :)


Katehuuh:
Relevant attempt at tool-calling integration in the OpenAI API extension #5185

oobabooga:
@SeanScripts I ended up going for https://github.com/oobabooga/text-generation-webui/pull/6827 due to its more self-contained feature -- the changes in this PR would be difficult for me to maintain in the future, given that I don't use tool calling myself.

SeanScripts:
Fair enough. I think tools are very powerful and likely the future of
agentic AI use, but it's hard to set them up. I was hoping this could be a
way to make it easier to set up and test them in a UI, since most existing
tool calling systems only work with through APIs. But I'm not great at UI
development and wasn't able to get this UI for tools as easy to use as I
was hoping. And I understand that it would be hard to maintain because lots
of models have slight differences in tool calling formats, much like chat
template differences. Sorry about that.

oobabooga:
No worries -- if you feel like it, I guess the code in the PR could be turned into an extension for others to use.

PR #6767, opened 2025-02-22
Add Quanto4,2 KV cache compression for Transformers

dinerburger:
This patch adds quanto KV cache quantization support for Transformers. The placement is unfortunately a little awkward, but we need to pass the same `QuantoQuantizedCache` object to each step. Happy to put it somewhere else.

Also, added optimum-quanto to requirements.txt. It can probably be safely made an optional dependency, but it's quite small and I believe portable across platforms.

Finally, this patch does not add HQQ KV-cache quantization. Looking at [HF's published numbers](https://huggingface.co/blog/kv-cache-quantization) I didn't really see a need for it, since Quanto outperforms it across the board and by a wide margin. Happy to add it for completeness if we think it's necessary, however.

## Checklist:


dinerburger:
Opened on the wrong branch, of course. Sorry about that.

PR #6827, opened 2025-04-03
Tools support for OpenAI compatible API

### LINKED ISSUE #4286: OpenAI functions implementation missing.

### LINKED ISSUE #4455: OpenAI extension for Function Calling

### LINKED ISSUE #6633: OpenAI API doesn't support tools in requests
### Describe the bug

tools passed in via the OpenAI API are discarded.  Cause of the bug appears to be that the ChatCompletionRequestParams class in extensions/openai/typing.py file does not define the tools array.

### Reproduction

Make an API request that includes a tools array and the tools specified in the request will be discarded

### Screenshot

_No response_

### Logs

```shell
n/a
```

### System Info

```shell
n/a
```

jkrauss82:
## Checklist:

This PR is for adding basic tool support, only for the OpenAI compatible API of textgen.

I have followed the way [Ollama is implementing](https://github.com/ollama/ollama/blob/b51e0f397ced70bbfa7f22e9b3c94953967cb8e5/openai/openai.go) this feature.

A solid type validation of the tools is left to be desired. Testing with Qwen 2.5 and Llama 3.1 with various MCP services yielded so many different types I eventually gave up. The pydantic types are still in there, as is the validation method, however it is currently not active. I fear for now this task has to be delegated to the user who is supposed to confirm tool usage on the client-side.

I have tested the implementation with the Continue.dev client v1.1.17 and it is mostly working with that client. ~~There are issues with multiple tool calls returned by a model but these seem to be UI-related.~~

I am aware of PR #6757 - my PR most likely collides with that but I wanted to offer it anyway, as I needed the OpenAI support quickly and maybe some of the code can be used for this PR as well. Feel free to close this PR unmerged in that case.

@oobabooga If you let me know what your preferred way forward is I can adapt this PR further.

Closes #4286
Closes #4455
Closes #6633


jkrauss82:
The issues with multiple tool calls have been resolved in my tests by https://github.com/oobabooga/text-generation-webui/pull/6827/commits/ab6cb56cf54370f9410315dcdf0e3fb07648d499.

oobabooga:
Thanks @jkrauss82 (sorry for taking so long to review). I'm not very familiar with tool usage; so I have some questions:

1. Do you have a simple curl command to add to the wiki demonstrating how it works?
2. Are there any security concerns like arbitrary code execution by exposing an API endpoint publicly with this feature available?
3. Is this a "full" implementation for tool usage/MCP, in the sense that this is sufficient for the vast majority of applications, or is this a significant simplification?

The PR looks good, I want to have it in the next release.

oobabooga:
Also, I assume it will slightly break after https://github.com/oobabooga/text-generation-webui/pull/6950, but it should be an easy fix.

jkrauss82:
Thanks for looking into this PR @oobabooga 

While working on the response to your questions I have found that the previous implementation did not work well when the response is not streamed to the client. I have thus pushed two more commits which fix this and also support more cases of tool call responses generated by LLMs. I have also enabled the validation of the incoming tools from the request object.

Regarding your questions:

### 1. Simple Curl Command

I have tested locally with this curl:

```sh
curl http://127.0.0.1:5000/v1/chat/completions \
  -d '{
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "What time is it currently in New York City?"
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "get_current_time",
          "description": "Get current time in a specific timezones",
          "parameters": {
            "type": "object",
            "required": ["timezone"],
            "properties": {
              "timezone": {
                "type": "string",
                "description": "IANA timezone name (e.g., America/New_York, Europe/London). Use Europe/Berlin as local timezone if no timezone provided by the user."
              }
            }
          }
        }
      }
    ]
  }'
```

Sample response:

```json
{
    "id": "chatcmpl-1746532051477984256",
    "object": "chat.completion",
    "created": 1746532051,
    "model": "qwen2.5-coder-14b-instruct-q4_k_m.gguf",
    "choices": [
        {
            "index": 0,
            "finish_reason": "tool_calls",
            "message": {
                "role": "assistant",
                "content": "```xml\n<function>\n{\n  \"name\": \"get_current_time\",\n  \"arguments\": {\n    \"timezone\": \"America/New_York\"\n  }\n}\n</function>\n```"
            },
            "tool_calls": [
                {
                    "type": "function",
                    "function": {
                        "name": "get_current_time",
                        "arguments": "{\"timezone\": \"America/New_York\"}"
                    },
                    "id": "call_52ij07mh",
                    "index": "0"
                }
            ]
        }
    ],
    "usage": {
        "prompt_tokens": 224,
        "completion_tokens": 38,
        "total_tokens": 262
    }
}
```

### 2. Security Concerns Due to Exposing the Tools Feature

I have enabled the validation of the tools as it worked fine in my tests today with a couple of different MCP servers. I think it should provide security at a basic level, e.g. code injections through the property "tools". I have used a popular LLM to get support here as I am not an expert in this regard.

What it currently does not address are malicious inputs passed down to the model when the tool definitions are added to the prompt template, e.g. inputs which would make the model provide details about its system prompt in its respond to the user. It is also still required by the user to make sure they are using sane MCP services as we do not know what the model generates beforehand and thus it is up to the user to review tool calls requested by the LLM before allowing execution.

In summary I think in terms of attacks on the textgen server itself this PR should be safe. But it requires thoughtful usage of the MCP tools by users of the new tools API.

### 3. Coverage of MCP Services

My best guess is this PR should cover most cases. I have tested it with a couple of different MCP implementations available through Continue.dev and also extended the method which looks for tool calls in the response to recognize more variations of tool call requests. There is definitely room for improvement here but I hope it is a good start.

It could also be that the pydantic tool types used for validating the input need to be adapted slightly in case they are too restrictive at the moment. I have not found a good documentation of how these types are defined (but I have only done a quick search so far), so they are based on my observations when working with the various MCP servers in my tests.

extensions/openai/completions.py
```diff
@@ -515,3 +550,19 @@ def completions(body: dict, is_legacy: bool = False) -> dict:
...
         yield resp
+
+
+def validateTools(tools: list[dict]):
+    # Validate each tool definition in the JSON array
+    valid_tools = None
+    for idx in range(len(tools)):
+        tool = tools[idx]
+        try:
+            tool_definition = ToolDefinition(**tool)
```
oobabooga, line 561: The `tool_definition` variable doesn't get used, is that a bug?

oobabooga:
It looks good to me -- the most important thing is that you tested it extensively and found that this works in practical use cases. I see that some more cases got covered in the latest commits as well. 

Merging now, thanks a lot for this PR. If you can think of further improvements, feel free to submit new PRs in the future.

jkrauss82:
Thanks for accepting and merging this PR!

I am going to run more tests for this feature during the next weeks. Should I find missing cases or other potential improvements I will submit a new PR as you suggest.

Katehuuh:
> I am going to run more tests for this feature during the next weeks. Should I find missing cases or other potential improvements I will submit a new PR as you suggest.

<details><summary>Marvelous work for <code>Qwen2.5</code>-<code>3</code>! Likely not trained on FC, so, did not work for older llama2, Mistral-Nemo... broadly support possible? 2B need the holy grail of FC prompt skill...</summary><br>

```python
import requests, json, datetime, pytz

messages = [ # Explicit var since this system-prompt req to work with `dolphin-2.9.4-gemma2-2b-Q4_K_M.gguf` = low generalizability for var placeholder + have no idea what is "OpenAI Function Call Format"..
    {"role": "system", "content": "you are function-calling assistant, check tool list for external data needs, return ONLY JSON: {\"name\":\"get_current_time\",\"arguments\":{\"timezone\":\"Country/city parameter value from query\"}} following OpenAI Function Call Format, nothing else."}, # Qwen3-32B work with system-prompt.
    {"role": "user", "content": "What time is it in New York?"}
]

response = requests.post(
    "http://127.0.0.1:5000/v1/chat/completions",
    headers={"Content-Type": "application/json"},
    json={
        "messages": messages,
        "tools": [{
            "type": "function",
            "function": {
                "name": "get_current_time",
                "description": "Get current time in a timezone",
                "parameters": {
                    "type": "object",
                    "required": ["timezone"],
                    "properties": {
                        "timezone": {"type": "string", "description": "IANA timezone name"}
                    }
                }
            }
        }]
    }
)

data = response.json()
print("Tool call response:", json.dumps(data, indent=2))

tool_calls = data["choices"][0].get("tool_calls", [])
if tool_calls:
    tool_call = tool_calls[0]
    args = json.loads(tool_call["function"]["arguments"])
    timezone = args["timezone"]
    
    time_result = datetime.datetime.now(pytz.timezone(timezone)).strftime("%Y-%m-%d %H:%M:%S %Z")
    print(f"✅ Function executed: {timezone} time is {time_result}")
    
    # Optional - send results back to model
    messages.append(data["choices"][0]["message"])
    messages.append({
        "role": "tool",
        "tool_call_id": tool_call["id"],
        "name": tool_call["function"]["name"],
        "content": json.dumps({"time": time_result, "timezone": timezone})
    })
    
    final_response = requests.post(
        "http://127.0.0.1:5000/v1/chat/completions",
        headers={"Content-Type": "application/json"},
        json={"messages": messages}
    )
    
    print("Final response:", json.dumps(final_response.json(), indent=2))
```


- [browser-use](https://github.com/browser-use/browser-use/issues/74#issuecomment-2628828933) supposed using OpenAI function call, but this did not work either with Qwen3-32B.
 
</details>

B0rner:
Does this  feature work correctly?

I try using devstral2 in text-generation-webui with (mistrals) vibe. (same like claude coder). But the Funcion calling-tools block seams not working correct. 

Does anyone has the same problem?

PR #6832, opened 2025-04-06
Add ExLlamaV3 support

oobabooga:
This is an ExLlamav3_HF loader that uses the same samplers as Exllamav2_HF and Transformers. 

I have compiled wheels that make it work out of the box, without any manual installation steps.

~ExLlamaV3 adds flash-attn v2.7.4.post1 as a requirement, but I didn't manage to compile that for Windows on Github Actions. It seems to work fine with flash-attn 2.7.3. though.~ nevermind I compiled it

## Benchmarks

Benchmarks for 5 of the models (49 different quants total) are available [here](https://oobabooga.github.io/benchmark.html).


Panchovix:
HI there! Any chace to add gpu_split or gpu_reserve_vram? At the moment, when trying to load https://huggingface.co/turboderp/Mistral-Large-Instruct-2411-exl3/tree/6.0bpw it seems it auto splits on 4 GPUs, 128GB VRAM, but when trying to do inference, I get an OOM for GPU0 (which is headless), on Linux.

I did a try with

```
    @classmethod
    def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.PathLike]], *model_args, **kwargs):
        assert len(model_args) == 0 and len(kwargs) == 0, "extra args is currently not supported"
        if isinstance(pretrained_model_name_or_path, str):
            pretrained_model_name_or_path = Path(pretrained_model_name_or_path)

        pretrained_model_name_or_path = Path(f'{shared.args.model_dir}') / Path(pretrained_model_name_or_path)

        # Create instance without loading the model yet
        model = cls(pretrained_model_name_or_path, load_model=False)
        
        # Determine load parameters
        load_params = {"progressbar": True}
        
        # Handle GPU split options
        if hasattr(shared.args, 'gpu_split') and shared.args.gpu_split:
            if shared.args.gpu_split != "auto":
                # Parse the GPU split string to a list of floats (in GB)
                load_params["use_per_device"] = [float(alloc) for alloc in shared.args.gpu_split.split(",")]
        
        # Handle VRAM reserve if specified
        if hasattr(shared.args, 'vram_reserve') and shared.args.vram_reserve:
            if isinstance(shared.args.vram_reserve, str):
                load_params["reserve_per_device"] = [float(r) for r in shared.args.vram_reserve.split(",")]
            else:
                load_params["reserve_per_device"] = shared.args.vram_reserve
        
        # Load the model with determined parameters
        model.ex_model.load(**load_params)
        
        return model
```

and seems gpu_split works (after adding it to the loader), but reserve not yet (because I think it has to be added as an argument on either ui.py or ui_model_menu.py

oobabooga:
Sure @Panchovix, thanks for the suggestion. I have reused the existing `--gpu-split` parameter for now. This seems to work now:

```
python server.py --model turboderp_Llama-3.1-70B-Instruct-exl3_4.0bpw --gpu-split 10,24
```

PR #6846, opened 2025-04-16
New llama.cpp loader

oobabooga:
This PR introduces a new lightweight and efficient llama.cpp loader based on `llama-server`. It replaces llama-cpp-python.

### How it works

* Python wheels containing the `llama-server` binary are compiled at [oobabooga/llama-cpp-binaries](https://github.com/oobabooga/llama-cpp-binaries).
* This binary is used to load the model and generate text, encode prompts, decode token IDs, etc., through API requests.
* There are no Python bindings. llama.cpp is used directly.

### Advantages 🚀

1. It will be possible to update llama.cpp *much* more frequently.
2. Sampling parameters available in llama.cpp but not in llama-cpp-python will become available in the llama.cpp loader, such as XTC, DRY, and sampler order.
3. Speculative decoding will be easy to add.
4. Multimodal support *may* be easy to add.
5. Less overhead, potentially resulting in slightly better performance.
6. The wheels appear to be noticeably smaller, though this needs verification.

### Disadvantages :(

1. llamacpp_HF is gone as it requires the bindings. I attempted to route the llama.cpp logits to Transformers over API using the `logprobs` option with the `/completion` endpoint, but the performance was very bad. With that:
   1. Perplexity evaluation for llama.cpp will be gone too. It may be possible to do it in the future with `logprobs` though if it has a `logits_all` option somewhere.
   2. Using the Transformers tokenizer with llama.cpp will no longer be possible. By now llama.cpp is a mature and well tested project, so I don't think that's even necessary.
   3. Using the Transformers samplers with llama.cpp will no longer be possible. I also don't think that's too bad because even experimental stuff that emerged as PRs to this project like XTC and DRY are now in llama.cpp.
2. StreamingLLM is gone.
3. The tqdm prompt processing progress bar is gone.

### Status

- Needs more testing
- Need to connect more existing command-line arguments to the loader
- Need to implement more existing sampling parameters in the loader


oobabooga:
Things are mostly implemented. The wheels are ready and seem to work. The only things left to add seem to be

* the kv cache quantization loading option
* the custom token bans sampling option.

PR #6901, opened 2025-04-25
fix for various missing defaults in llama_cpp_server concerning DRY and XTC

### LINKED ISSUE #6900: API completion for llama.cpp causes KeyError: 'dry_multiplier'
### Describe the bug

Crash when hitting API for llama.cpp loaded model with error ```KeyError: 'dry_multiplier```

### Reproduction

Load a model and then hit it via API...webgui works fine...

### Screenshot

Image

### Logs

```shell
oobabooga-server  | Traceback (most recent call last):
oobabooga-server  |   File "/home/app/text-generation-webui/modules/text_generation.py", line 494, in generate_reply_custom
oobabooga-server  |     for reply in shared.model.generate_with_streaming(question, state):
oobabooga-server  |   File "/home/app/text-generation-webui/modules/llama_cpp_server.py", line 124, in generate_with_streaming
oobabooga-server  |     payload = self.prepare_payload(state)
oobabooga-server  |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
oobabooga-server  |   File "/home/app/text-generation-webui/modules/llama_cpp_server.py", line 73, in prepare_payload
oobabooga-server  |     "dry_multiplier": state["dry_multiplier"],
oobabooga-server  |                       ~~~~~^^^^^^^^^^^^^^^^^^
oobabooga-server  | KeyError: 'dry_multiplier'
oobabooga-server  | Output generated in 0.00 seconds (0.00 tokens/s, 0 tokens, context 41, seed 2106569066)
```

### System Info

```shell
obabooga-server  | ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
oobabooga-server  | ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
oobabooga-server  | ggml_cuda_init: found 1 CUDA devices:
oobabooga-server  |   Device 0: NVIDIA GeForce RTX 3090 Ti, compute capability 8.6, VMM: yes
oobabooga-server  | build: 1 (30fe919) with cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 for x86_64-linux-gnu
oobabooga-server  | system info: n_threads = 8, n_threads_batch = 8, total_threads = 24
oobabooga-server  | 
oobabooga-server  | system_info: n_threads = 8 (n_threads_batch = 8) / 24 | CUDA : ARCHS = 500,520,530,600,610,620,700,720,750,800,860,870,890,900 | USE_GRAPHS = 1 | PEER_MAX_BATCH_SIZE = 128 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | BMI2 = 1 | LLAMAFILE = 1 | OPENMP = 1 | AARCH64_REPACK = 1 |
```

bars0um:
fixes #6900 by adding the default values as per https://github.com/ggml-org/llama.cpp/blob/master/examples/server/README.md

## Checklist:

### DIFF
--- a/modules/llama_cpp_server.py
+++ b/modules/llama_cpp_server.py
@@ -57,6 +57,20 @@ class LlamaServer:
         return result.get("content", "")
 
     def prepare_payload(self, state):
+        # add default values as per https://github.com/ggml-org/llama.cpp/blob/master/examples/server/README.md
+        if "dry_multiplier" not in state:
+            state["dry_multiplier"] = 0.0
+        if "dry_base" not in state:
+            state["dry_base"] = 1.75
+        if "dry_allowed_length" not in state:
+            state["dry_allowed_length"] = 2
+        if "xtc_probability" not in state:
+            state["xtc_probability"] = 0.0
+        if "xtc_threshold" not in state:
+            state["xtc_threshold"] = 0.1
+        if "dry_sequence_breakers" not in state:
+            state["dry_sequence_breakers"] = '["\\n", ":", "\\\"", "*"]'
+        
         payload = {
             "temperature": state["temperature"] if not state["dynamic_temperature"] else (state["dynatemp_low"] + state["dynatemp_high"]) / 2,
             "dynatemp_range": 0 if not state["dynamic_temperature"] else (state["dynatemp_high"] - state["dynatemp_low"]) / 2,


oobabooga:
The defaults are built-in in the server if unspecified. It's not necessary to send the defaults.

oobabooga:
Also these get sent in the generation request, overwriting the defaults anyway.

bars0um:
but it crashes without this patch if I hit the API with Continue.dev for example when llama.cpp is used to load a model...
 @oobabooga

bars0um:
Maybe the 

> That's odd, those keys should always be present with default values due to
> 
> https://github.com/oobabooga/text-generation-webui/blob/main/extensions/openai/typing.py

Hmmm...the api flow then must be missing a call to this module somewhere

bars0um:
I'll cancel this PR then since there must be a better way to fix this by ensuring the API flow utilizes the types @oobabooga mentions

oobabooga:
If you find a solution please feel free to send a new PR, it's probably a 1-line fix somewhere.

oobabooga:
Is there a simple curl command that reproduces the error you got?

bars0um:
I just simply hit the api with a llama.cpp loaded model using any of the vscode completion plugins...

bars0um:
I'll test the change I mentioned below shortly, just waiting on a small process I had going using the earlier patch to finish out an experiment I needed to run through then will rebuild the docker image and test with that change, I'll post back if it works

> I'm not very familiar with the code base, but I am suspecting the solution may lie somewhere in the following:
> 
> oobabooga/text-generation-webui/modules/text_generation.py
> 
> generate_reply appears to receive a state passed on by openai module oobabooga_data/extensions/openai/completions.py
> 
> generate_reply seems to hold the _state_ field that is not properly initialized with the default values shown in presets.py which appears to contain the default values needed here....
> 
> looking at a call to default_preset I see this in a few places:
> 
> ```
>     generate_params = presets.load_preset(default_preset)
> ```
> 
> suspect that the same should be done in text_generation.py line 306 instead of the current `generate_params = {}`

oobabooga:
This is working fine for me

```
curl http://127.0.0.1:5000/v1/completions \
  -d '{
    "prompt": "This is a cake recipe:\n\n1.",
    "max_tokens": 200,
    "temperature": 1,
    "top_p": 0.9,
    "seed": 10
  }'
```

so those requests are some kind of edge case.

bars0um:
weird, maybe something with my setup? I've got commit a778270536f8f46654d746b1dd828954eb30df8d

bars0um:
I've removed the patch and rerun with your curl command. it still fails...so something about my setup is different from yours. 

here is the dockerfile I used:

```
# BUILDER
FROM ubuntu:22.04
WORKDIR /builder
ARG TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST:-3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX}"
ARG BUILD_EXTENSIONS="${BUILD_EXTENSIONS:-}"
ARG APP_UID="${APP_UID:-6972}"
ARG APP_GID="${APP_GID:-6972}"

RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,rw \
    apt update && \
    apt install --no-install-recommends -y git vim build-essential python3-dev pip bash curl && \
    rm -rf /var/lib/apt/lists/*
WORKDIR /home/app/
COPY ./text-generation-webui/ /home/app/text-generation-webui/
WORKDIR /home/app/text-generation-webui
RUN GPU_CHOICE=A USE_CUDA118=FALSE LAUNCH_AFTER_INSTALL=FALSE INSTALL_EXTENSIONS=TRUE ./start_linux.sh --verbose
COPY CMD_FLAGS.txt /home/app/text-generation-webui/
EXPOSE ${CONTAINER_PORT:-7860} ${CONTAINER_API_PORT:-5000} ${CONTAINER_API_STREAM_PORT:-5005}
WORKDIR /home/app/text-generation-webui
# set umask to ensure group read / write at runtime
CMD umask 0002 && export HOME=/home/app/text-generation-webui && ./start_linux.sh --listen

```

bars0um:
Strange, my extensions/openai/completions.py file does not match the one from the v3.0 tag...but I am pretty sure I checked out tag v3.0 when I started using this code. I apologize for the hassle I may have caused, not sure why this is happening...

mrguymiah:
For the purposes of helping anyone else that has this issue. One of my extensions was out of date and needed to be redownloaded. Not sure which one. Not going to one by one eliminate them to find out. Also, here's some keywords to help Google people find it:

SillyTavern API KeyError 'dry_multiplier'

Issue #6771, opened 2025-02-22
Translation problem. Pipe bug?


mykeehu:
### Describe the bug

Something is wrong with the translation, you need to check this pipe! I can see in the command window what I'm getting back from Google, but the Hungarian text comes back as gibberish, as if the text returned was rewritten by the model, so it's not the text returned from the translation. I manually translated the English text in Google and the response in Hungarian was fine. The translation seems to go through the model again.
In addition, in English text, special characters are html encoded and passed to the translator, for example the apostrophe as `&#x27;`, the quotation mark as `&quot;`. 
Something has changed in the translation, and extensions may not have been adapted to this, but it will be a mess after translation, and the breakage will happen on the fly.
I'm going back to the previous version before the UI changed and will not update for a while.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

Chat non-English using translation extensions (Google, Google plus or Libre)

### Logs

```shell
-
```

### System Info

```shell
i9-13900K
64GB RAM
RTX 3090
Windows 10 latest
```

flflow:
Google something changed and translation modules are broken (errors and low quality of outputs). Because of this, I translate the text through the model itself.
https://github.com/nidhaloff/deep-translator/issues/286

mykeehu:
To me, it's as if they use their own AI model for the API, while the GUI interface uses the old, classic one. Partly a good direction for an AI to understand context, but it returns very poor output.

flflow:
It looks like they restrict access to translator in this scenarios and using simplified version.
Result: 
1. Broken browser translation extensions (bad quality). https://addons.mozilla.org/ru/firefox/addon/simple-translate/reviews/
2. Broken different translation python packages (deep-translate, googletrans).
3. Bad quality of translation via headless browser with minimized window.

Google Transalte have simplified version with bad quality:
https://translate.google.com/m

Neko-Kuroi:
google translate web service started AI or pattern match suggestion. Is this a recent feature?

![Image](https://github.com/user-attachments/assets/feb74e95-d19a-4322-9009-d606e5dab7bc)

mykeehu:
I have noticed that when I use LibreTranslate, for the **Starts with** and **Continue** functions, the program does not send the text to Libre and the model translates it into some horrible text.
Is this a bug in the Libre plugin or in the program?

mykeehu:
[...]

PR #6944, opened 2025-05-03
Fix continue/start reply with when using translation extensions

mykeehu:
Please check my code, hope it works, and please fix it if you see any errors! The end of the function is a bit confusing for me.
This will improve the re-translation of texts when using the Continue button or the Start reply with field.

If the code looks good, I'll enable PR.

Solve this problem: https://github.com/oobabooga/text-generation-webui/issues/6771


mykeehu:
@oobabooga Please check and integrate this into the new code, because I won't rewrite it every time there are changes, but Regenerate and Start with don't work properly without it when using the translation. Until then I cannot use the program because I need both functions. Thank you!

PR #6947, opened 2025-05-04
UI: Add message version navigation

Th-Underscore:
- Created a new module (message_versioning.py) for managing message versioning logic and data storage
- Added relevant Gradio communication in ui_chat.py
- Updated chat.py to append messages to versioning history and other history-related operations
- Updated HTML generation to include message version navigation elements
- Implemented JavaScript functions for handling message version navigation, selection, and history management
- Added relevant CSS styling (handling all three message classes is a little messy, but changing the original class structure by adding `.message` to "instruct" messages might be problematic)

Notes:

- ~~Doesn't always update on first load via `shared.gradio['unique_id'].select()`~~ - Solved with [90348ee](https://github.com/oobabooga/text-generation-webui/pull/6947/commits/90348ee560c3582cbde27507622ca7a38dd73333)
- No go-to-first or go-to-last nav arrows
- Have only tested with a couple extensions, but this _shouldn't_ break anything and should be completely backwards-compatible
- Haven't put any documentation for controls (Ctrl+Left/Right Arrow for navigation, Ctrl+Up/Down Arrow as an alternative for selection, Ctrl+LeftClick for force-select)
- Intentionally, the original footer buttons aren't set to full opacity when the message is selected, only the nav container
- Output file extension is `.boogaplus` cuz it's cool
- Tried using `None` as a placeholder in the history storage dict (`null` in the JSON) instead of always saving the text to save space and to remove 'pos', but sometimes lead to client-server issues

## Checklist:

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -1412,7 +1412,6 @@ strong {
     color: #07ff07;
 }
 
-
 .message-attachments {
     display: flex;
     flex-wrap: wrap;


Th-Underscore:
Do you think it would be better if navigation was handled client-side instead?

Load:
 -> (s) Give client history JSON
Navigate:
 -> (c) Read JSON, do nav stuff
 -> (c) Update history JSON to new msg version
 -> (c) Trigger history change for Gradio server-side
Exit:
 -> (s) Save msg version data

oobabooga:
I want to refactor the history format in the project from

```
{
    'internal': [[message, reply]],
    'visible': [[visible_message, visible_reply]]
}
```

to

```
[
    {
        'role': 'user',
        'content': '',
        'visible-content': '',
        'date': ''
    },
    [
        {
            'role': 'assistant',
            'content': '',
            'visible-content': '',
            'date': ''
        },
        {
            'role': 'assistant',
            'content': '',
            'visible-content': '',
            'date': ''
        },
    ]
]
```

where an entry with two options is a "regenerate", which I imagine is similar to what your changes do. This will make it possible to add date/time to chat/chat-instruct messages in the UI and will also simplify the chat completions API handling a lot. Also, it will make it possible to add fields for attachments like images or documents for RAG/multimodal support.

I'll try to do this refactoring this week; I need to do that before I'm able to review this PR.

Th-Underscore:
Got it. Seems like a _huge_ backend update, compatibility-wise. I'm a big fan, though. I've wanted to change the structure to something like this for a while, so if you want extra help refactoring just let me know. I'd also recommend letting extensions add custom message metadata keys—e.g. `'story_date'`, `'scene_#'`, etc.—which can be updated, similar to `output_modifier`. Just to keep options open for extension developers.

I'll update this PR once you finish the refactoring.

oobabooga:
I have created a first attempt here https://github.com/oobabooga/text-generation-webui/pull/6950; it seems to work but there are many possible combinations of actions like

- regenerate
- continue
- regenerate + start with
- impersonate
- impersonate with some text already in the box
- etc

Any test is highly welcome

Th-Underscore:
The global `last_state` is now largely redundant thanks to `handle_unique_id_change()`. Will keep for now in case of unforeseen issues.

Currently, `display_mode` is handled server-side, but the nav container could be hidden with forced styling via JS. In the future, I plan to add a nav overlay like the original [boogaPlus](https://github.com/user-attachments/assets/18ac3482-f561-45c4-a75f-be46a064bbd8), but for now it's just "on" and "off".

Th-Underscore (commented):
Made a mistake, kept `streaming_class` instead of `selected_class` for `generate_chat_html`.

Th-Underscore:
Cleaned up all the unnecessary calls and caches, tested and it should be good to go. I kept the `m_type` compatibility for future changes. I might make a new PR to add versioning for user messages, but for now it's mostly unnecessary thanks to history branching.

oobabooga:
I have made a test and it seems like there is a small bug. While regenerating with Ctrl+Enter:

1. The first time, the number of versions goes from 1 to 2 (correct)
2. The second time, it goes from 2 to 4
3. The third time, during streaming, it appears as 5, then it ends as 6
4. The fourth time, it's 7 during streaming, then 8

See this video (look at the numbers below the message)


https://github.com/user-attachments/assets/b83b6e3e-4de6-4794-85e6-b6b229e7acdf



Also, it would be nice to be able to switch between versions with arrow left and arrow right, and, if already at the last version, regenerate with right. This could be added as a key binding that applies whenever the chat tab is selected, under "Keyboard shortcuts" at `js/main.js`.

oobabooga:
See if it works now:

https://github.com/user-attachments/assets/2c9d5d02-c389-47a0-a720-2568790c743a

About the keybind, I think that L/R is better and more intuitive. SillyTavern uses this key binding for instance.

Th-Underscore:
Seems good now, mind testing?

Th-Underscore:
Yeah trying to solve that right now, when a message is generated it moves the focus into the chat box, which is filtered out for this keybind—otherwise pressing arrow keys in any textboxes or dropdowns would attempt navigation. The fact that it moves the event target specifically into the main textbox is making this a struggle.

In the meantime, I'm thinking:

```js
  if (!document.querySelector("#chat-tab").checkVisibility() ) {
    return;
  }
```

for everything except the Esc keybind, just as a general redundancy check.

oobabooga:
> for everything except the Esc keybind, just as a general redundancy check.

This would make sense.

And indeed, the event should return early if in a text box. SillyTavern uses this function which could serve as inspiration:

```js
    function isInputElementInFocus() {
        //return $(document.activeElement).is(":input");
        var focused = $(':focus');
        if (focused.is('input') || focused.is('textarea') || focused.prop('contenteditable') == 'true') {
            if (focused.attr('id') === 'send_textarea') {
                return false;
            }
            return true;
        }
        return false;
    }
```

https://github.com/SillyTavern/SillyTavern/blob/2c7f2e2014aa54f1cd1272bc48f6bf1156e4ce1b/public/scripts/RossAscends-mods.js#L1011

oobabooga:
> How about if the chatbox is empty, the arrow keys work as normal?

I think so.

Also I just saw that they have this other return early condition

```js
    function isModifiedKeyboardEvent(event) {
        return (event instanceof KeyboardEvent &&
            event.shiftKey ||
            event.ctrlKey ||
            event.altKey ||
            event.metaKey);
    }
```

to filter pure arrow right/left vs something with multiple keys pressed at the same time.

Th-Underscore:
This seems to work well for me.

oobabooga:
I have refactored the changes to try to distill them into something as simple as possible. It seems to be working reliably after many tests on my end while removing some 300 lines. 

Merging now, please feel free to send a new PR if you notice issues.

PR #6950, opened 2025-05-05
Refactor the history format

oobabooga:
The goal of this PR is to refactor the history format in the project from

```
{
    'internal': [[message, reply]],
    'visible': [[visible_message, visible_reply]]
}
```

to

```
[
    {
        'role': 'user',
        'content': '',
        'visible-content': '',
        'date': ''
    },
    [
        {
            'role': 'assistant',
            'content': '',
            'visible-content': '',
            'date': ''
        },
        {
            'role': 'assistant',
            'content': '',
            'visible-content': '',
            'date': ''
        },
    ]
]
```

where an entry with two options is a "regenerate".

This will make it possible to:

- Add date/time to chat/chat-instruct messages in the UI
- Add "swipes"
- Simplify the chat completions API handling a lot
- Add fields for attachments like images or documents for RAG/multimodal support.

### Status

It seems to be fully implemented now, but more testing is needed. There might be bugs.

### DIFF
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -1,5 +1,4 @@
 import base64
-import copy
 import re
 import time
 from collections import deque
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -97,7 +97,7 @@ def convert_history(history):
     if any('content' in entry and isinstance(entry['content'], list) for entry in history):
         temp_history = []
         for entry in history:
-            if isinstance(entry.get('content'), list):
+            if isinstance(entry['content'], list):
                 for item in entry['content']:
                     if not isinstance(item, dict):
                         continue
--- a/extensions/openai/completions.py
+++ b/extensions/openai/completions.py
@@ -84,41 +84,26 @@ def convert_history(history):
     Converts OpenAI chat histories to our internal format.
     Our new format is a list of message objects similar to OpenAI's format.
     '''
-    new_history = []
-    user_input = ""
-    user_input_last = True
     system_message = ""
+    new_history = []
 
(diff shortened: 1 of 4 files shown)


Th-Underscore:
Would it be strange to also have versioning for user messages ("edits")? I personally would find it useful when testing different outputs. Navigating through "regenerations" and "edits" would act independently.

oobabooga:
> Currently, `*Is typing...*` is appended to assistant messages. Perhaps it should be kept when `[-1]`, otherwise filtered out.

Does this solve it? https://github.com/oobabooga/text-generation-webui/pull/6950/commits/e40c0a6e241019e86b53b3ae6bda1e1fd478b899

About user edits, I think the best way to do it would be to "branch" the conversation like Claude does:

> Editing this message will create a new conversation branch. You can switch between branches using the arrow navigation buttons.

I guess that could be added after this PR is done if you are interested.

Th-Underscore:
I think having both would be great. Since you already have branches implemented, the user can choose whether they want to make a separate branch or be able to choose in-chat. Temporarily, it could be similar to ["Replace last reply"](https://github.com/oobabooga/text-generation-webui/commit/89d38e8e444e761f89c9f869a10d7faa5f4d59d5). I can add an "Edit message" footer-button in #6947 as well.

oobabooga:
Yeah I guess it's still buggy -- thanks for the feedback.

----

On second thought: This PR would break most extensions for this project, including unmaintained ones, while also introducing several new bugs. For the sake of stability, I'm giving up on it.

In the future, we might achieve similar goals by adding an efficient 'metadata' key to the history instead.

PR #6961, opened 2025-05-06
Fix model config loading in shared.py

Downtown-Case:
On my setup (Python 3.13, which I presume is the issue?) model loading errors out:

```
│ /home/alpha/AI/text-generation-webui/server.py:4 in <module>                                                                                                                                                         │
│                                                                                                                                                                                                                      │
│     3                                                                                                                                                                                                                │
│ ❱   4 from modules import shared                                                                                                                                                                                     │
│     5 from modules.block_requests import OpenMonkeyPatch, RequestBlocker                                                                                                                                             │
│                                                                                                                                                                                                                      │
│ /home/alpha/AI/text-generation-webui/modules/shared.py:313 in <module>                                                                                                                                               │
│                                                                                                                                                                                                                      │
│   312 # Load model-specific settings                                                                                                                                                                                 │
│ ❱ 313 with Path(f'{args.model_dir}/config.yaml') as p:                                                                                                                                                               │
│   314     if p.exists():                                                                                                                                                                                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: 'PosixPath' object does not support the context manager protocol
```

More specifically, a Path object cant be used in a with statement. This fixes it with no changes to functionality.

## Checklist:

### DIFF
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -310,11 +310,13 @@ if args.api or args.public_api:
     add_extension('openai', last=True)
 
 # Load model-specific settings
-with Path(f'{args.model_dir}/config.yaml') as p:
-    if p.exists():
-        model_config = yaml.safe_load(open(p, 'r').read())
-    else:
-        model_config = {}
+p = Path(f'{args.model_dir}/config.yaml')
+if p.exists():
+    model_config = yaml.safe_load(open(p, 'r').read())
+else:
+    model_config = {}
+del p
+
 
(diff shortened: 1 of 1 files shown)


oobabooga:
Indeed it seems like Python 3.13 doesn't support Path objects as context managers, thanks for the fix.

PR #6995, opened 2025-05-18
Fix Dockerfile in AMD and Intel

TheGameratorT:
Only the `nvidia` variant of the Dockerfile contained the changes for the new `user_data` directory structure, this fixes it for `amd` and `intel`.
There was also a mistake in the README where `amd` links to the `amd` Dockerfile and the logs/cache were created outside `user_data`.

## Checklist:


oobabooga:
Thanks for the fixes.

PR #7001, opened 2025-05-19
Update README

mtomas7:
I placed the "Pointing to an existing AI model library" section first, as I believe it is more relevant to the majority of users.

## Checklist:


oobabooga:
I understand the use case, but I prefer to keep the instructions very simple, and I don't think most people have an organized models folder shared across projects.

PR #7005, opened 2025-05-20
Add attachments support (text files, PDF documents)

oobabooga:
Example:

pdf

This is not RAG -- the entire document goes in the prompt. For reference, the DeepSeek-V3 paper is about 49,000 tokens long (it's a 53 pages PDF).


TheAlphaAndSigma-MugsNeko:
image
doesnt work

PR #6967, opened 2025-05-10
adds "Branch here" functionality as footer-button

Madrawn:
This simply allows, in addition, branching at a specific assistant response by adding a new icon under each response.

branch

-   **UI & HTML:**
    -   Grouped message action buttons (copy, regenerate, etc.) into a new `.message-actions` div for better organization and styling.
    -   Added a "Branch here" button with a new SVG icon to each assistant/bot message.
    -   Assistant/bot messages now include a `data-index` attribute.
-   **JavaScript:**
    -   Implemented `branchHere(element)` function that, when a "Branch here" button is clicked:
        -   Retrieves the `data-index` of the message.
        -   Populates a hidden `branch_index` Gradio input field with this index.
        -   Programmatically clicks the main "Branch" button.
-   **Backend (Python):**
    -   The `handle_branch_chat_click` function in `modules/chat.py` now:
        -   Accepts a `branch_index` from the state.
        -   If `branch_index` is provided (not -1), it truncates the current chat history up to (and including) that index before creating the new branched chat.
        -   Resets `branch_index` to -1 after branching.
    -   Added `branch_index` to the Gradio interface state and input elements.

## Checklist:

- [ X] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

PR #7018, opened 2025-05-27
UI: Add "Replace last message" button (Ctrl+Alt+L)

Th-Underscore:
To avoid deleting message versions (#6947) and other metadata

## Checklist:


oobabooga:
Do you think after https://github.com/oobabooga/text-generation-webui/pull/6967 it would be doable to just add an edit svg icon that turns any message into a textarea with confirm/cancel buttons, and replaces the message in the backend on confirm? This way, copy last message / replace last message could be removed.

When editing an user message, it could also be nice to branch the conversation after confirming.

Th-Underscore:
Sounds great, I love it. Should double-clicking also be a shortcut to editing?

I'm also thinking on confirm it'd give three options:

1. "Cancel"
2. "Edit message" (edit in-place)
3. "Branch from here and edit message" (copy + trim + edit in-place)

mykeehu:
@oobabooga The **Edit** button edits the _internal_ text, not the _visible_ text. You should edit the _visible_ and translate it to _internal_. [I wrote more about this here.](https://github.com/oobabooga/text-generation-webui/issues/7044#issuecomment-2965150185)

PR #7058, opened 2025-06-10
set presence_penalty to 1.5 for Qwen3

miriameng:
setting presence_penalty to 1.5 is recommended here: https://huggingface.co/Qwen/Qwen3-32B-GGUF

### DIFF
--- a/user_data/presets/Qwen3 - No Thinking.yaml
+++ b/user_data/presets/Qwen3 - No Thinking.yaml
@@ -1,3 +1,4 @@
 temperature: 0.7
 top_p: 0.8
 top_k: 20
+presence_penalty: 1.5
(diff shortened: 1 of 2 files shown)


oobabooga:
Interesting. For the 235b one they don't recommend this

> For thinking mode, use Temperature=0.6, TopP=0.95, TopK=20, and MinP=0 (the default setting in generation_config.json). DO NOT use greedy decoding, as it can lead to performance degradation and endless repetitions. For more detailed guidance, please refer to the [Best Practices](https://huggingface.co/Qwen/Qwen3-235B-A22B#best-practices) section.

https://huggingface.co/Qwen/Qwen3-235B-A22B

I'll keep it without the penalty to be conservative, those old (pre-DRY) repetition penalties are very artificial and I don't think they should always be used.

PR #7071, opened 2025-06-12
ensure estimated vram is updated when switching between different models

### LINKED ISSUE #7068: "Estimated VRAM to load the model" in the model tab does not update when switching between different models that have the same default gpu-layers and ctx-size
### Describe the bug

When I switched from nomic-embed-text-v1.5.Q4_K_M.gguf to Qwen3-32B-Q4_K_M.gguf, the "Estimated VRAM to load the model" still shows 2333 MiB, which is incorrect, it should be 22697 MiB.

### Reproduction

Download Qwen3-32B-Q4_K_M.gguf from https://huggingface.co/Qwen/Qwen3-32B-GGUF 
Download nomic-embed-text-v1.5.Q4_K_M.gguf from https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF
gpu-layers should be 65 for both by default, set ctx-size to 24576 for both, save settings, and then switch between the two model in the dropdown menu and there will be no updates in "Estimated VRAM to load the model"

### Screenshot

_No response_

### Logs

```shell
No log shows up in the terminal
```

### System Info

```shell
Windows 10
RTX 3090
```

miriameng:
This fixes #7068

### DIFF
--- a/modules/ui_model_menu.py
+++ b/modules/ui_model_menu.py
@@ -135,7 +135,7 @@ def create_event_handlers():
     # with the model defaults (if any), and then the model is loaded
     shared.gradio['model_menu'].change(
         ui.gather_interface_values, gradio(shared.input_elements), gradio('interface_state')).then(
-        handle_load_model_event_initial, gradio('model_menu', 'interface_state'), gradio(ui.list_interface_input_elements()) + gradio('interface_state'), show_progress=False).then(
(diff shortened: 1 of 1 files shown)


oobabooga:
Added handling for non-llama.cpp models

Thanks for another bug fix

PR #7098, opened 2025-06-25
Update ui_model_menu.py blocking the --multi-user access in backend

Alidr79:
In the --multi-user mode, the app is blocking the access by disabling the UI components. But with a simple inspect an abuser could enable it and run the action. So we should block the access also in the backend. A simple way is to block it in the event_handlers. Here specifically the abuser can't unload the model by doing a simple inspect in the browser!

## Checklist:

### DIFF
--- a/modules/ui_model_menu.py
+++ b/modules/ui_model_menu.py
@@ -130,6 +130,9 @@ def create_ui():
 
 
 def create_event_handlers():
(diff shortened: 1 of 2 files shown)


oobabooga:
Makes sense, I wasn't aware non-interactive elements could be made interactive with the browser tools. Nice catch.

PR #7101, opened 2025-06-26
Fix Edit and Copy for translated posts

mykeehu:
This patch improves the Copy and Edit functions by generate the translated text in the data-raw tag in html and not the English, internal version.


oobabooga:
Thanks for the PR. Unfortunately `history['visible']` can contain arbitrary HTML from extensions (e.g. TTS extensions store `<audio>` tags there), so using it for edit/copy would break those cases. The real fix would be to re-apply `output_modifier` after editing so that translations get regenerated, but that's a different change.

PR #7102, opened 2025-06-26
Handle either missing <think> start or </think> end tags

zombiegreedo:
The model I have been using frequently does not begin replies with \<think\> but does include the end \</think\> tag, and so most 'thoughts' weren't getting put into the nice looking UI boxes. This change fixes that issue, which I have tested. I did not see a way to accomplish this through the chat/instruction template.

It preserves the original handling of missing end \</think\> tags, and (I believe) fixes a potential bug on original line 120 (the lstrip means if the reply starts with whitespace before a \<think\> tag then content_start would be at the wrong index).

## Checklist:


oobabooga:
This is a good change, thanks!

PR #7106, opened 2025-06-28
Fix for chat options getting lost

philipp-classen:
The problem that this PR is intending to solve is that the chat options can be lost. The UI then looks like that:

options_missing

The path to reproduce is to open and close the options, hide/unhide the settings, and switch between the different tabs. Eventually, it reaches this state for me and only a reload of the UI fixes it.

For comparison, this is what it is expected to look:

expected1

and

expected2

From the error, I believe the reason is that the tab lookup is the problem (resulting ultimately in the `Attempted to select a non-interactive or hidden tab.` error in the code below):

```
function change_tab(id: object | string | number): void {
		const tab_to_activate = tabs.find((t) => t.id === id);
		if (
			tab_to_activate &&
			tab_to_activate.interactive &&
			tab_to_activate.visible
		) {
			selected = id;
			$selected_tab = id;
			$selected_tab_index = tabs.findIndex((t) => t.id === id);
			dispatch("change");
		} else {
			console.warn("Attempted to select a non-interactive or hidden tab.");
		}
	}
```

When testing, the problem goes away when the ID for the "Chat" tab is not explicitly set, but manged internally by Gradio. I did not notice any negative side-effects yet after removing it.

Tested with Firefox and Chromium.

## Checklist:

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -23,7 +23,15 @@ document.querySelector(".header_bar").addEventListener("click", function(event) 
     if (extensions) {
       extensions.style.display = "flex";
     }
+
     this.style.marginBottom = chatVisible ? "0px" : "19px";
+
+    if (chatVisible && !showControlsChecked) {
+      document.querySelectorAll("#extensions").forEach(element => {
+        element.style.display = "none";
+      });
+    }
+
   } else {
     this.style.marginBottom = "19px";
     if (extensions) extensions.style.display = "none";
--- a/js/main.js
+++ b/js/main.js
@@ -23,17 +23,7 @@ document.querySelector(".header_bar").addEventListener("click", function(event) 
     if (extensions) {
       extensions.style.display = "flex";
     }
-
     this.style.marginBottom = chatVisible ? "0px" : "19px";
-
-    if (chatVisible && !showControlsChecked) {
-      document.querySelectorAll(
-        "#chat-tab > div > :nth-child(1), #chat-tab > div > :nth-child(3), #chat-tab > div > :nth-child(4), #extensions"
-      ).forEach(element => {
-        element.style.display = "none";
-      });
-    }
-
   } else {
     this.style.marginBottom = "19px";
     if (extensions) extensions.style.display = "none";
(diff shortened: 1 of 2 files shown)


philipp-classen:
Unfortunately, it still happens.

oobabooga:
I haven't managed to reproduce the bug, if you can find a solution, let me know.

philipp-classen:
Didn't find time yet, but I do want to look into it again. Fortunately, reproducing is quite straightforward for me, but I also noticed that it hasn't been reported by someone else.

I stopped looking, since after the change I failed to reproduce at all; but now - with the change - it again happens consistently. Overall, it doesn't look like a strange race. Maybe there is something specific in my local setup that triggers it? Though it is nothing special: I'm on Linux and it doesn't seem browser specific.

Anyway, if I can find out more, I'll share it here.

philipp-classen:
I was able to narrow it down further. Looks like it is caused by a "display: none" getting attached to the "#past-chats-row" element that is not removed later when the settings are opened again. This is the code that adds the "display: none":

https://github.com/oobabooga/text-generation-webui/blob/17f9c188bdb409740b51a4061ba527cf1345c4db/js/main.js#L33

Since there are also the CSS classes sidebar-shown and sidebar-hidden correctly set, I assume the extra "display: none" is no longer needed. At least, for me the UI works well when I remove the code. I updated the PR with the updated version.

Here are also the steps to reproduce for me (assuming starting in "Chat" tab with settings opened, otherwise use CTRL-S):
1. Hide controls (CTRL-S)
2. Go to "Notebook" tab
3. Go to "Chat" tab [Note: here the inline style gets attached]
4. Show controls (CTRL-S)

oobabooga:
This is correct. The only detail is that extensions still need to get hidden by that removed codeblock; I have added it back for #extensions only. The remaining code was indeed obsolete. Thanks for the fix!

philipp-classen:
Thanks, makes sense. I'm not using extensions, so I missed that when testing.

PR #7120, opened 2025-07-06
Fix: use embedded Python in start_windows.bat

abulhawa:
This ensures the `one_click.py` setup runs in the intended environment, without relying on `conda.bat activate`, which may fail silently or be overridden by system `PATH`.


oobabooga:
Well noted, it seems to make sense.

PR #7199, opened 2025-08-13
Set multimodal status during Model Loading

altoiddealer:
[ X ] I have read the [Contributing guidelines](https://github.com/oobabooga/text-generation-webui/wiki/Contributing-guidelines).

I'm proposing this change to make it easier for future expansion (if other loaders may support multimodal, etc).

Instead of analyzing if the model is multimodal capable on every generation, **in `load_model()`**:
- Assert `False` by default
- if Exl3 or llama.cpp - check once and store the result in **`shared`**.
- Suggestion: Maybe use a placeholder **`is_multimodal()`** -> `False` in all other loader modules.

I'll bet you've thought of this already but `_process_images_for_generation` could maybe be a shared method, with subclassed special handling after the first chunk which is duplicated [here](https://github.com/oobabooga/text-generation-webui/blob/45e2935e87f19aa3d5afec9a403203259cb1eacc/modules/llama_cpp_server.py#L138-L145) and [here](https://github.com/oobabooga/text-generation-webui/blob/45e2935e87f19aa3d5afec9a403203259cb1eacc/modules/exllamav3.py#L187-L194) (I'm referencing your current code) - however I'm not versed enough in your code to handle this.

### DIFF
--- a/modules/exllamav3.py
+++ b/modules/exllamav3.py
@@ -177,9 +177,6 @@ class Exllamav3Model:
         Process all possible image inputs and return modified prompt + embeddings.
         Returns: (processed_prompt, image_embeddings)
         """
-        if not self.is_multimodal():
-            return prompt, []
-
         # Collect images from various sources using shared utilities
         pil_images = []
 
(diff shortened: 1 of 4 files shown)


altoiddealer:
Tested with ExLlamav3 model with vision, and llama.cpp using the example model scenario in your tutorial.

altoiddealer:
I realized that I did not commit the correct version of `models.py`.  I resolved this now.

oobabooga:
Good idea, it's indeed better to check if the model is multimodal just once instead of every call.

> Suggestion: Maybe use a placeholder is_multimodal() -> False in all other loader modules.

I can see the merit in this. To keep things simple, I'd prefer not to add it for now for loaders where it can't be True.

Thanks for the excellent refactor!

Issue #7194, opened 2025-08-12
Add torch_dtype dropdown in loader UI and respect --bf16/config.json for non-quantized loads


laubner:
**Problem**:
Currently, `transformers_loader.py` does not fully respect the `--bf16` command-line flag or `config.json`’s `torch_dtype` (e.g., `bfloat16`) for non-quantized GPU loads in the `should_use_default_loading` path. Instead, it defaults to FP32, which caused VRAM overflow for a 24B model (~48GB in FP32 vs. ~24GB in bfloat16 on an A100 80GB), leading to CPU offloading and slow performance (~0.1 tokens/s). The `--bf16` flag works with 4-bit/8-bit quantization or DeepSpeed, but not consistently for non-quantized loads, despite `config.json` specifying `bfloat16`.

**Current Behavior**:
- Without `--load-in-4bit` or `--load-in-8bit`, the loader ignores `--bf16` and `config.json`’s `torch_dtype`, defaulting to FP32.
- This is a design limitation, as the GUI and CLI lack a general `torch_dtype` control outside quantization.

**Proposed Solution**:
- Add a dropdown in the loader UI (under model settings) to select `torch_dtype` (options: `bfloat16`, `float16`, `float32`).
- Extend `transformers_loader.py` to respect `--bf16` and `config.json`’s `torch_dtype` for all non-quantized GPU loads, with `--bf16` taking priority.
- Suggested code change in `load_model_HF:

```
def set_torch_dtype():
      if shared.args.bf16:
          logger.info("Using torch_dtype from --bf16 flag: bfloat16")
          return torch.bfloat16
      if hasattr(config, 'torch_dtype') and config.torch_dtype:
          dtype_str = str(config.torch_dtype).split('.')[-1]
          if dtype_str in ['bfloat16', 'float16', 'float32']:
              logger.info(f"Using torch_dtype from config.json: {dtype_str}")
              return config.torch_dtype
      logger.info("No torch_dtype specified. Defaulting to float32.")
      return torch.float32
```

- In should_use_default_loading:

  `params['torch_dtype'] = set_torch_dtype()`

-  In else (non-quantized GPU) block, after quantization checks:

 ```
 if not shared.args.load_in_4bit and not shared.args.load_in_8bit:
      params['torch_dtype'] = set_torch_dtype()
```

**Impact**:
- Improves usability for large models on Ampere GPUs (e.g., A100, H100) by leveraging `--bf16` and `config.json`.
[...]

PR #7217, opened 2025-09-05
Changing transformers_loader.py to Match User Expectations for --bf16 and Flash Attention 2

stevenxdavis:
## Checklist:

[A recent commit](https://github.com/oobabooga/text-generation-webui/commit/3b28dc182186308648af5cddb44c811f1608c70d) changed transformers_loader.py so that it no longer passes torch_dtype to the transformers loader. Without quantization, however, the autodetected dtype may be float32 by default, which (I assume) we will want to avoid in two circumstances.

1. When the user indicates --bf16 in the command line, settings, or checkbox in the UI, the user will expect it to load torch_dtype = bfloat16, so we should respect that over the default.
2. When the user indicates --attn-implementation flash_attention_2 in the command line, it will cause an error if torch_dtype = float32 because [FlashAttention2 is only supported for models with the fp16 or bf16 torch type.](https://huggingface.co/docs/transformers/perf_infer_gpu_one#flashattention)

This PR should address both of those concerns. Note that if --bf16 is flagged, Flash Attention 2 will automatically work, so the user will only be need to alerted if they inadvertently use Flash Attention 2 with float16.

This partially addresses the issue [raised here](https://github.com/oobabooga/text-generation-webui/issues/7194) but rather than adding a new feature, it just conforms to the existing expectation for the flags.

### DIFF
--- a/modules/transformers_loader.py
+++ b/modules/transformers_loader.py
@@ -133,18 +133,19 @@ def load_tokenizer(model_name, tokenizer_dir=None):
 def load_model_HF(model_name):
     torch._dynamo.config.disable = True
 
-    path_to_model = Path(f'{shared.args.model_dir}/{model_name}')
-       
+    path_to_model = Path(f'{shared.args.model_dir}/{model_name}'   
     params = {
         'low_cpu_mem_usage': True,
         'attn_implementation': shared.args.attn_implementation,
     }
-    
+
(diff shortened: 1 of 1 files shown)


oobabooga:
Thanks, I had changed this in 3b28dc182186308648af5cddb44c811f1608c70d, but apparently it transformers doesn't autodetect this well and uses float32 in many cases.

Issue #7238, opened 2025-09-25
macosx x86_64 portable release doesn't include llamma-cpp-binaries package


IonoclastBrigham:
### Describe the bug

The portable package for intel macs doesn't include the llama binaries package. I logged [a previous bug](https://github.com/oobabooga/llama-cpp-binaries/issues/6) that the mac intel binaries weren't even built correctly for that package, so it's maybe not surprising to find they're not being properly pulled into the build here. Seems like a chain of compounding issues obscured things and lead us to this point.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Repro and Diagnosis Steps

```bash
$ cd ~/Downloads
$ wget https://github.com/oobabooga/text-generation-webui/releases/download/v3.13/textgen-portable-3.13-macos-x86_64.zip
$ unzip textgen-portable-3.13-macos-x86_64.zip
$ cd text-generation-webui-3.13/
$ ./start_macos.sh --model any-model.gguf
09:32:06-531241 INFO     Starting Text Generation Web UI
09:32:06-533982 INFO     Loading settings from "user_data/settings.yaml"
09:32:06-854936 INFO     Loading "any-model.gguf"
╭───────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────╮
│ /Users/USER/Code/text-generation-webui-3.13/server.py:312 in <module>                                                               │
│                                                                                                                                     │
│   311         # Load the model                                                                                                      │
│ ❱ 312         shared.model, shared.tokenizer = load_model(shared.model_name)                                                        │
│   313         if shared.args.lora:                                                                                                  │
│                                                                                                                                     │
│ /Users/USER/Code/text-generation-webui-3.13/modules/models.py:43 in load_model                                                      │
│                                                                                                                                     │
│    42     shared.args.loader = loader                                                                                               │
│ ❱  43     output = load_func_map[loader](model_name)                                                                                │
│    44     if type(output) is tuple:                                                                                                 │
│                                                                                                                                     │
│ /Users/USER/Code/text-generation-webui-3.13/modules/models.py:71 in llama_cpp_server_loader                                         │
│                                                                                                                                     │
│    70 def llama_cpp_server_loader(model_name):                                                                                      │
│ ❱  71     from modules.llama_cpp_server import LlamaServer                                                                          │
│    72                                                                                                                               │
│                                                                                                                                     │
[...]

PR #7239, opened 2025-09-25
Fix portable apple intel requirement for llama binaries (issue #7238)

IonoclastBrigham:
Fixes Issue #7238

CI builder for intel portable package runs on macos 13, which is darwin 22, and thus the `llama-cpp-binaries` package was completely skipped by `pip`. This makes sure the correct wheel is linked.

CAVEAT: audioloop still gets skipped in CI because of the python version. Not sure the implications of trying to change that.

## Checklist:


oobabooga:
Thanks for the fix, this is correct. Audioloop is intentionally skipped because it's only necessary for python 3.13+, and the portable builds use python 3.11.

PR #7243, opened 2025-09-28
Update chat.py fixes loss of history for instruction mode

Remowylliams:
## Checklist:

For more than a year, instruct mode history would not show the history when you switch to it from chat mode,
This will fix the history problem for instruct mode and not interfere with chat mode history
I've been making this change for over a year now, with no trouble.


Remowylliams:
Sorry wrong branch. this is my first pull request.

PR #7244, opened 2025-09-28
chat.py fixes Instruct mode History

Remowylliams:
## Checklist:

This will fix the Instruct mode history that doesn't show on the left side of the webui. This is because there is no greeting entry in visible for instruct.
This has been a problem for more than a year. I've commented on it in Discord and offered my code but I've had to make this change every time I update my TGW. Someone said I needed to submit a pull request. This is my first.
I've been using this patch for more than a year with no trouble in chat or instruct.


oobabooga:
Thanks for the fix

PR #7328, opened 2025-11-28
Add image generation support

oobabooga:
This PR adds a new "Image AI" tab to the UI where it's possible to generate images with models like [Z-Image-Turbo](https://huggingface.co/Tongyi-MAI/Z-Image-Turbo):

image

This is done through the [diffusers](https://github.com/huggingface/diffusers) library.

## How to test

1. Install diffusers from source:

```
pip install git+https://github.com/huggingface/diffusers
```

2. Download the model by pasting `Tongyi-MAI/Z-Image-Turbo` under `Image AI > Model > Download model`, or with

```
mkdir user_data/image_models
cd user_data/image_models
huggingface-cli download Tongyi-MAI/Z-Image-Turbo --local-dir ./Tongyi-MAI_Z-Image-Turbo
```

3. Select `Tongyi-MAI_Z-Image-Turbo` under `Image AI > Model > Model`
4. Generate images

## TODO

- [x] Add quantization support through bitsandbytes (4-bit, 8-bit)
- [x] Add Qwen-Image support
- [x] Add metadata to the output images like in stable-diffusion-webui

### DIFF
--- a/docs/Image Generation Tutorial.md
+++ b/docs/Image Generation Tutorial.md
@@ -2,16 +2,16 @@
 
 This feature allows you to generate images using high-speed models like Z-Image-Turbo directly within the web UI.
 
-## How to use (Easiest Method)
+## How to use
 
 1. Click on the **Image AI** tab at the top of the interface.
 2. Select the **Model** sub-tab.
 3. Copy and paste the following link into the **Download model** box:
-```
-
-[https://huggingface.co/Tongyi-MAI/Z-Image-Turbo](https://huggingface.co/Tongyi-MAI/Z-Image-Turbo)
 
 ```
+https://huggingface.co/Tongyi-MAI/Z-Image-Turbo
+```
+
 4. Click the **Download** button and wait for the confirmation message.
 5. In the **Model** dropdown menu, select the model you just downloaded (if you don't see it, click the 🔄 refresh button).
 6. Click **Load**.


Quiet-Joker:
I've used Z-Image turbo and while it's a good model i am curious... what sampler and scheduler are you using? I've used it on ComfmyUI and i found out the best combination for me in terms of image quality is Euler + KL_optimal. Also from my understanding, Z-image Turbo doesn't use/has no effect on negative prompts, they are sort of useless. CFG of 0 gives the prompt you put on the negative prompt, CFG of 1 gives the positive prompt.

oobabooga:
> I've used Z-Image turbo and while it's a good model i am curious... what sampler and scheduler are you using? I've used it on ComfmyUI and i found out the best combination for me in terms of image quality is Euler + KL_optimal. Also from my understanding, Z-image Turbo doesn't use/has no effect on negative prompts, they are sort of useless. CFG of 0 gives the prompt you put on the negative prompt, CFG of 1 gives the positive prompt.

I'm using the diffusers default for Z-Image-Turbo, which is `FlowMatchEulerDiscreteScheduler`. ComfyUI probably uses its own sampler/scheduler that departs from what diffusers uses. And you are right, with CFG = 0 the negative prompt is indeed not used.

Quiet-Joker:
> > I've used Z-Image turbo and while it's a good model i am curious... what sampler and scheduler are you using? I've used it on ComfmyUI and i found out the best combination for me in terms of image quality is Euler + KL_optimal. Also from my understanding, Z-image Turbo doesn't use/has no effect on negative prompts, they are sort of useless. CFG of 0 gives the prompt you put on the negative prompt, CFG of 1 gives the positive prompt.
> 
> I'm using the diffusers default for Z-Image-Turbo, which is `FlowMatchEulerDiscreteScheduler`. ComfyUI probably uses its own sampler/scheduler that departs from what diffusers uses. And you are right, with CFG = 0 the negative prompt is indeed not used.

So what do you plan on naming the repo now?

Oobabooga Web-ui?

oobabooga:
> So what do you plan on naming the repo now?
> 
> Oobabooga Web-ui?

Either this or embrace the "misleading project name" standard, like llama.cpp does.

PR #7329, opened 2025-11-28
Update requirements.txt

aidevtime:
The idiap/coqui-ai-TTS fork is incompatible with transformers 4.55.  The idiap fork supports only transformers ≤4.46.2, making it incompatible with current webui's 4.55.* requirement.  The dependency conflict cannot be resolved by simply switching forks. However, this working solution bypass these limitations entirely.

## Checklist:


oobabooga:
Thanks for the fix! I hadn't used this extension in a while and didn't know it had dependency issues. Appreciate your simple fix.

PR #7358, opened 2025-12-25
Fix for Copy Button (with HTTP Fallback)

jakubartur:
## Checklist:


oobabooga:
What is the issue fixed by this change? And how does the change work?

jakubartur:
the copy button works without SSL, for example in a private LAN there is no need to set up SSL and then the copy button works

oobabooga:
Nice, thanks for the PR! I wasn't aware of this issue with code blocks on HTTP.

I refactored the fix to patch `highlightjs-copy.min.js` directly instead of intercepting clicks — based on a similar approach from https://github.com/xtekky/gpt4free/pull/2302. Also formatted the file for readability.

PR #7363, opened 2026-01-03
Fix loading models which have their eos token disabled

jin-eld:
Some models, for instance IQuest-Coder-V1-40B-Instruct do not use an eos token and disable it in the metadata:

     32: BOOL       |        1 | tokenizer.ggml.add_eos_token = False

We still try to access it, which leads to an error when attempting to load the model. Handle it the same way as missing bos tokens, by setting it to an empty string.

This fixes model loading, however the user must still manually specify the proper stopping string in the parameters tab for such models to work.

## Checklist:


oobabooga:
Thanks for the fix.

PR #7388, opened 2026-02-05
Checkbox for locking settings and automatic reset to default settings in model loader

GodEmperor785:
This PR adds two things:
- Setting default model loader parameters on model change for models that don't have saved settings.
  - This prevents a following scenario: loading a model with some extra specific settings (like extra_flags for llama.cpp), switching to another model, but those specific settings from previous model remain and need to be cleared manually.
- A new checkbox "Lock settings" in model loader page to prevent any automatic changes to the model loader settings while it is checked, including: VRAM estimation with layer adjustment, loading saved settings when choosing another model that has saved settings, resetting to default settings on model change. On unchecking appropriate settings are applied (default for models with no saved values, saved values for models that have them). Checkbox is by default unchecked.
  - This helps for the following scenario: In llama.cpp I have a MoE model too big to fit entirely in VRAM, but I set experts to offload in extra_flags. For it to work as expected I need to have layers slider set to max value, however when I change context length, the VRAM estimation automatically sets the slider to what it thinks is correct and I need to set it back whenever I change context length. With this checkbox user can adjust parameters without any automatic adjustments getting in their way

New checkbox looks like this on the "Model" page:
image

I have tested my changes locally and it seems to behave as expected:

With checkbox checked settings are not updated automatically. VRAM estimation info still updates ("Estimated VRAM to load the model") but doesn't adjust the layers slider. When performing the unchecking, settings are updated appropriately depending if model has saved settings or not (sets defaults if no saved settings). With checkbox unchecked the VRAM estimation works as previously and settings are updated when selecting a model with saved settings.

The resetting of settings to default also works. When switching to model with no saved settings - defaults are set. When switching to model with saved settings - those saved settings are applied.

Please let me know if such changes to UI functionality are acceptable.

## Checklist:

### DIFF
--- a/modules/models_settings.py
+++ b/modules/models_settings.py
@@ -26,6 +26,23 @@ def get_fallback_settings():
     }
 
 
+def get_loader_defaults():
+    '''
+    Returns default values for all loader parameters from shared.args_defaults.
+    Used to reset UI when switching to a model without saved settings.
+    '''
+    defaults = {}
+    for param in loaders.get_all_params():
+        if hasattr(shared.args_defaults, param):
+            value = getattr(shared.args_defaults, param)
+            if value is None and param in ('tensor_split', 'gpu_split', 'extra_flags', 'device_draft'):
+                value = ''
+            elif value is None and param in ('model_draft', 'mmproj'):
+                value = 'None'
+            defaults[param] = value
+    return defaults
+
+
 def get_model_metadata(model):
     model_path = resolve_model_path(model)
     model_settings = {}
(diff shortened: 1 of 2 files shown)


GodEmperor785:
After recent changes in v4.0 this PR would have to look different, I'm not sure if the "Lock settings" checkbox is needed since the automatic VRAM estimation setting layers is gone.

The resetting of settings on model switch might still be useful.

@oobabooga would such change be acceptable? I can prepare a new PR for v4.0

oobabooga:
Thanks for the PR and for thinking about this. The main annoyance you described, auto GPU layers fighting your manual settings, is no longer an issue since v4.0 as you noticed, so the lock checkbox has become unnecessary.

As for resetting settings on model switch: any setting could be something a user wants to persist across models (e.g. `mlock`, or even `extra_flags` if you're using something like `--ninja`). For per-model settings, "Save settings" already handles this. It saves all values including empty/default ones, so switching to a model with saved settings will correctly override anything carried from the previous model.

What are your thoughts?

GodEmperor785:
Hm, so maybe a button to reset settings to default for those that don't want to persist everything?

oobabooga:
I don't think it's worth adding for how rarely it would be needed. Going to close this, but thanks again and feel free to send more PRs.

PR #7404, opened 2026-03-02
Add CAMB AI TTS extension

neilruaro-camb:
## Summary
- Adds a new CAMB AI TTS extension that integrates CAMB AI's text-to-speech API for generating audio from LLM outputs
- Supports language selection, voice cloning, and automatic audio playback in the UI


oobabooga:
Hi, thanks for the PR. I'd prefer to keep this as a community extension rather than built-in — could you submit it to https://github.com/oobabooga/text-generation-webui-extensions instead?

PR #7406, opened 2026-03-04
Bump sentence-transformers to 3.3.1 in superbooga to fix torch breakage

### LINKED ISSUE #7385: INSTALL_EXTENSIONS=TRUE can break torch stack via superbooga -> sentence-transformers 2.2.2 -> torchvision
### Describe the bug

When running the installer with INSTALL_EXTENSIONS=TRUE, the installation can break due to a dependency chain introduced by the superbooga extension.

superbooga pins sentence-transformers==2.2.2.
That specific version declares torchvision as an unconditional dependency (not an extra).

On CUDA 12.8 (cu128), this resolves to:
 - torchvision==0.25.*
 - which in turn forces torch==2.10.*

This silently upgrades the torch stack during extension installation and breaks ABI-sensitive extensions such as flash-attention / exllama and results in a broken installation due to torch ABI incompatibilities.

Why this is hard to diagnose:
- pip freeze does not show extras or dependency provenance
- the torch upgrade happens only when INSTALL_EXTENSIONS=TRUE
- the failure appears far away from the actual cause (extension install vs runtime)

Root Cause:
- sentence-transformers==2.2.2 has torchvision as a hard dependency
- newer sentence-transformers versions (≥ 3.3.1) no longer do

Suggested Fix:
- Bump sentence-transformers in extensions/superbooga/requirements.txt to >=3.3.1 
  or
- Document that INSTALL_EXTENSIONS=TRUE may modify the torch stack

This is not a pip or CUDA bug, but an interaction between:
- extension auto-installation
- old pinned dependencies
- newer CUDA wheels

Posting this mainly as a heads-up for others running cu128 or ABI-sensitive extensions.

### Reproduction

Run the installer with "INSTALL_EXTENSIONS=TRUE"

example:
GPU_CHOICE=A LAUNCH_AFTER_INSTALL=FALSE INSTALL_EXTENSIONS=TRUE ./start_linux.sh --verbose

### Screenshot

_No response_

### Logs

```shell
ImportError: /home/app/text-generation-webui/installer_files/env/lib/python3.11/site-packages/exllamav2_ext.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda29c10_cuda_check_implementationEiPKcS2_ib
```

### System Info

```shell
NVIDIA GPU
tested on docker compose.
```

weiguangli-io:
## Summary

`sentence-transformers==2.2.2` (pinned in `extensions/superbooga/requirements.txt`) declares `torchvision` as a hard dependency. On CUDA 12.8, this silently upgrades the torch stack when `INSTALL_EXTENSIONS=TRUE` is set, breaking ABI-sensitive extensions like flash-attention and exllama.

`sentence-transformers>=3.3.1` removed `torchvision` as a hard dependency. This version (`3.3.1`) is already used by `superboogav2`, so this change aligns both extensions.

**Change:** `sentence_transformers==2.2.2` → `sentence_transformers==3.3.1`

Fixes #7385

## Checklist:

### DIFF
--- a/extensions/superbooga/requirements.txt
+++ b/extensions/superbooga/requirements.txt
@@ -2,5 +2,5 @@ beautifulsoup4==4.12.2
 chromadb==0.4.24
 pandas==2.0.3
 posthog==2.4.2
-sentence_transformers==2.2.2
+sentence_transformers==3.3.1
 lxml


oobabooga:
Thanks, @OiPunk

Issue #7357, opened 2025-12-24
While trying to use LLAMA.CPP using RPC, there is not way to properly pass the command to llama.cpp


ummite:
### Describe the bug

I've tried to use --rpc=10.x.x.x(serverip):50052 in extra-flags, either in the interface or in command line or in CMD_FLAGS from user_data directory, rpc parameter never reach llama.cpp

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 2 CUDA devices:
  Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes
  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
error: invalid argument: -rpc
19:09:54-049135 ERROR    Error loading the model with llama.cpp: Server process terminated unexpectedly with exit code:
                         1

### Logs

```shell
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 2 CUDA devices:
  Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes
  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
error: invalid argument: -rpc
19:09:54-049135 ERROR    Error loading the model with llama.cpp: Server process terminated unexpectedly with exit code:
                         1
```

### System Info

```shell
Ultra9 224 GB ram + 5090 + 3090 in TB5 + secondary computer that I want to use with RPC capabilities.
```

Kekuaka:
llama_cpp binaries packed with ooba are usually compiled without support for rpc mode.
to use rpc capabilities I'd recommend to build llama_cpp from sources enabling rpc option as it's described in the llama_cpp/rpc manual.
then, replace the whole folder of llama_cpp binaries in ooba installation with the new one from the previous step.
then, find the file llama_cpp_server.py in ooba installation and add option "--rpc", "\<comma separated host:ports of your computers\>" to the method `_start_server()`

after these steps you may be able to use rpc mode. in ooba's model pane you could also use `tensor_split` for all local and remote gpus.

philipp-classen:
[...]

PR #7418, opened 2026-03-10
[experimental] Expose the "--rpc" option of llama.cpp

philipp-classen:
:warning: Will not work with the shipped llama.cpp binary

This is a proof-of-concept for https://github.com/oobabooga/text-generation-webui/issues/7357

This PR provides the `--rpc` option of llama.cpp. It will not work out-of-the-box though, because the shipped llama.cpp binary is not built with RPC support. You have to build it yourself and replace it from source and enable RPC (`-DGGML_RPC=ON`, see https://github.com/ggml-org/llama.cpp/tree/master/tools/rpc).

The `llama_cpp_binaries/bin` directory is the important one (e.g. create a symlink to the "bin" directory of your local llama.cpp built):

venv/lib/python3.13/site-packages/llama_cpp_binaries/bin/

I can confirm that the setup works, but it is experimental.

For me, it also required tweaking the `--tensor-split` parameters; otherwise, the remote machine would run out of memory.

My understanding is that llama.cpp tends to start with putting the first layers on the remote host. So, this would put 51 layers on the remote host (running on 192.168.178.100) and 24 on the localhost:

```
--tensor-split 51,24 --rpc 192.168.178.100:50052
```

## Checklist:

### DIFF
--- a/modules/llama_cpp_server.py
+++ b/modules/llama_cpp_server.py
@@ -414,6 +414,8 @@ class LlamaServer:
             cmd += ["--spec-ngram-size-m", str(shared.args.spec_ngram_size_m)]
             cmd += ["--spec-ngram-min-hits", str(shared.args.spec_ngram_min_hits)]
         cmd += ["--parallel", str(shared.args.parallel)]
+        if shared.args.rpc:
+            cmd += ["--rpc", shared.args.rpc]
         if shared.args.streaming_llm:
             cmd += ["--cache-reuse", "1"]
             cmd += ["--swa-full"]
--- a/modules/shared.py
+++ b/modules/shared.py
@@ -108,6 +108,7 @@ group.add_argument('--threads', type=int, default=0, help='Number of threads to 
 group.add_argument('--threads-batch', type=int, default=0, help='Number of threads to use for batches/prompt processing.')
 group.add_argument('--numa', action='store_true', help='Activate NUMA task allocation for llama.cpp.')
 group.add_argument('--parallel', type=int, default=1, help='Number of parallel request slots. The context size is divided equally among slots. For example, to have 4 slots with 8192 context each, set ctx_size to 32768.')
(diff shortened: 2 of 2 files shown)


oobabooga:
I have added the necessary flag to the compilation workflows to be able to support this feature and am compiling now

https://github.com/oobabooga/llama-cpp-binaries/commit/f33a464b4902fda6d3aab120ea4c9312fb1c6e3c

Is the speed acceptable over RPC for a large model?

Kekuaka:
When starting llama-cpp in RPC mode it's useful to (be capable to) provide env parameter GGML_CUDA_DISABLE_GRAPHS=1 to avoid errors at inference time.

>Is the speed acceptable over RPC for a large model?

1Gbit cable is NOT the bottleneck for up to ~50 tokens/sec (for 26-28Gb LLM sizes)

philipp-classen:
@oobabooga Thanks for the quick addition. I did not have much time yet, but I was able to do a quick smoke test now. With the new shipped binary, it works fine. No more need to compile llama.cpp yourself. I would say, the PR in its current form is safe to merge.

The whole RPC implementation in llama.cpp is experimental, of course. So, I would currently not endorse it much at this point; but it is fully usable for people who want to try it out. For question how to use it, the RPC documentation in llama.cpp is the best starting point that I know:

https://github.com/ggml-org/llama.cpp/tree/master/tools/rpc

oobabooga:
Given that the feature is quite niche, instead of adding a dedicated `--rpc` flag, I've fixed `--extra-flags` to support literal flags, so you can now do `--extra-flags "--rpc 192.168.1.100:50052,192.168.1.101:50052 --jinja"`. Same functionality, just a different way to pass the flag, and it should work since the binaries are compiled with RPC support now.

https://github.com/oobabooga/text-generation-webui/commit/2a6b1fdcba676200d2e454534a91e1d334b60bdf

https://github.com/oobabooga/text-generation-webui/commit/7e54e7b7ae62b227fbd896b2daf704db1658baa5

Thanks for the PR, @philipp-classen

PR #7483, opened 2026-04-09
Fix: wrong chat deleted when selection changes before confirm

### LINKED ISSUE #7472: Deletion of incorrect chat after selection
### Describe the bug

Starting to delete one chat, then selecting another and confirming the deletion causes the wrong chat to be deleted.

### Reproduction

Create two chats with different names.
Select Chat 1.
Click the Trash icon on Chat 1, which turns it into the "confirm" icon.
Select Chat 2.
Confirm deletion of Chat 1.

Result: Chat 1 remains, Chat 2 is deleted.

### Screenshot

_No response_

### Logs

```shell
Deleted "user_data\logs\instruct\ChatTest2.json"
```

### System Info

```shell
Windows NVIDIA
```

lawrence3699:
Fixes #7472.

When you click the trash icon on Chat 1, then select Chat 2 before pressing confirm, Chat 2 gets deleted instead of Chat 1. The Gradio backend reads `state['unique_id']` from whichever chat is currently selected at confirm time, not the one that was originally targeted.

The fix re-selects the target chat before triggering the confirm button. The `label` variable is already captured in the closure (same one used in the trash-button handler at line 992), so this just mirrors the existing pattern.

**Repro:**
1. Create two chats
2. Click trash icon on Chat 1 (shows confirm/cancel)
3. Click Chat 2 to select it
4. Press confirm
5. Chat 2 is deleted instead of Chat 1

## Checklist:

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -1004,6 +1004,7 @@ function addMiniDeletes() {
 
     confirmBtn.onclick = (e) => {
       e.stopPropagation();
+      label.querySelector("input").click();
       document.querySelector("#delete_chat-confirm").click();
       resetButtons();
     };


oobabooga:
Elegant fix, thank you.

PR #7486, opened 2026-04-11
Logits display improvements

wiger3:
This PR addresses the following issues with the logits viewer:
- display special tokens (like eot) properly
Previously, the logits viewer would display blank strings for special tokens like end of turn, eos, etc. This seems to be a llama-server bug, but we can work around it on the client-side by re-decoding blank tokens.
- escape control characters for display properly instead of messing with repr
Previously, each token would go through repr and some buggy logic to strip apostrophes. That was a bad approach because repr can surround with either style of Python quotes - apostrophes (0x27) or quotes (0x22), depending on the content. It will also escape backslashes, quotes, and other special characters, which isn't desirable for a visual representation of tokens. Instead, we escape only control characters (ASCII C0 and Unicode C1) and leave the rest alone.
- change output formatting of the logits viewer to be clearer and show token ids
The old formatting was confusing to me - many tokens begin with spaces so it was hard to see what token would actually be emitted by the model. Token IDs are also useful if we want to ban some tokens. I've changed the format to `probability - [token] (token id)`

## Checklist:

### DIFF
--- a/modules/llama_cpp_server.py
+++ b/modules/llama_cpp_server.py
@@ -296,19 +296,26 @@ class LlamaServer:
             pprint.PrettyPrinter(indent=4, sort_dicts=False).pprint(printable_payload)
             print()
 
-        for retry in range(5):
-            response = self.session.post(url, json=payload)
-            result = response.json()
+        def _try_fetch_logits():
+            for retry in range(5):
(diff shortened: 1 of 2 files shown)


oobabooga:
Thanks for the fixes and improvements.

PR #7506, opened 2026-04-17
Add model download branch handling in download_model_wrapper

Th-Underscore:
Not sure if this counts for "UI:"

## Checklist:

- [X] I have read the [Contributing guidelines](https://github.com/oobabooga/textgen/wiki/Contributing-guidelines).

### DIFF
--- a/download-model.py
+++ b/download-model.py
@@ -57,8 +57,7 @@ class ModelDownloader:
         return session
 
     def sanitize_model_and_branch_names(self, model, branch):
-        if model[-1] == '/':
-            model = model[:-1]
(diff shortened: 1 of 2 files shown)


oobabooga:
Nice addition. The `elif ":" in repo_id` check will match `https://` URLs that don't contain `/tree/`, breaking existing GGUF URL downloads. Could you add a `not repo_id.startswith("http")` guard to that condition?

Th-Underscore:
Good catch. Pushed.

oobabooga:
Thank you :)

PR #7560, opened 2026-05-14
fix(win): set PYTHONUTF8 for non-ASCII locale Windows compatibility

jerry78424:
## Summary

Fix `UnicodeDecodeError` / `UnicodeEncodeError` when running `textgen.bat` on Windows systems with non-ASCII locales (e.g. Traditional Chinese cp950, Japanese cp932, Korean cp949).

Gradio and other Python packages read source files using the system locale as default encoding. On non-English Windows this causes failures during startup when encountering UTF-8 characters in gradio's own source files.

## Error

```
UnicodeDecodeError: 'cp950' codec can't decode byte 0xe2 in position 396: illegal multibyte sequence
UnicodeEncodeError: 'cp950' codec can't encode character '\u2771' in position 492: illegal multibyte sequence
```

## Solution

Set two environment variables at the top of `desktop/textgen.bat`:

- **`PYTHONUTF8=1`** — enables PEP 654 persistent UTF-8 mode; makes Python use UTF-8 as default encoding for all file I/O (including `pathlib.read_text()`).
- **`PYTHONIOENCODING=utf-8`** — ensures stdout/stderr use UTF-8 encoding.

These have no effect on systems already using UTF-8 locales (English, most non-CJK Windows) and are widely used by portable Python apps for the same reason.


desktop/textgen.bat
```diff
@@ -1,4 +1,6 @@
 @echo off
+set PYTHONUTF8=1
+set PYTHONIOENCODING=utf-8
```
oobabooga: Thanks for the PR, this was indeed missing. Similar to https://github.com/oobabooga/textgen/commit/9753b2342b3ffedd3a79e675d6ab5bf70400fef8

I think that `set PYTHONIOENCODING=utf-8` is unnecessary and only `set PYTHONUTF8=1` should be kept, can you give it a try and update the PR?

jerry78424: "Done\! I tested locally and `set PYTHONUTF8=1` alone is sufficient — it forces UTF-8 for both file I/O (pathlib.read_text etc.) and stdout/stderr. The PR has been updated."

oobabooga: Nice, thanks for testing and thanks for the PR!

